PR #16, opened 2024-10-15
Added OpenAI compatible options and examples

hunkim:
Added base_url, model, and api_key options to make it compatible with all Open AI-like chat/embeddings APIs, including Mistral, Llama, Together.ai, Fireworks, Croq, etc.

It will significantly improve the compatibility and usability of LightRAG.

See `examples/lightrag_openai_compatible_demo.py`.


LarFii:
Thank you for improving LightRAG! This significantly increases its usability and flexibility!

PR #39, opened 2024-10-17
update Step_3.py and openai compatible script

russellkim:
update reproduce/Step_3.py slightly
add open ai compatible script for reproduce/Step_1 and Step_2


LarFii:
Thanks for your contribution!

PR #62, opened 2024-10-20
Graph Visualization

BinNong:
Add a graph visualization util to examples directory. Once `lightrag_*_demo.py` have been executed and produced a `graphml` file, we can run `graph_visual.py` file to generate a html file which illustrates the graph structure produced by `lightrag`.


BinNong:
Graph visualization like this:

LarFii:
Thank you! That’s indeed a very convenient visualization solution.

PR #70, opened 2024-10-20
make graph visualization become colorful

BinNong:
Add colors to nodes to make graph visualization look like this:


LarFii:
Thanks!

Epsilon-Lee:
Hi, thanks for the contribution. I have a question. What does different color mean? Does each color represents a type of entity in the ontology?

PR #78, opened 2024-10-21
Fix example in README

sebastianschramm:
Fix the example in the README to make it work without any changes.

- add os import
- rm duplicate working dir definition


LarFii:
Thanks!

PR #79, opened 2024-10-21
Multifile processing, prompts config

ivs:
Currently WIP!

Prompts moved to toml config files, so later prompts can be easily tuned for concrete task. Source code prompt config added to prompts/code.tom

There're some doubts regarding entities capitalization, currently I switched it of, cos for code it harmful I think, mb it's nice to control by parameter, cos it looks like it beneficial for natural language.

Every chunk embeded with metadata now, inserted text has !none in that fields
```bash
####
## FILENAME: README.md
## FILEPATH: ./README.md
## CHUNK_NUM: 0
###
```

insert accepts either text or Path object, or list of one of them:
```python
def insert(self, input_data: Union[str, os.PathLike, List[Union[str, os.PathLike]]])
```

File name, path and chunk number added in entity and relationship extraction.


Jaykumaran:
You should have synced fork before creating PR, as there are merge conflicts

PR #98, opened 2024-10-22
Fix encoding error when running demo in some occasions

JavieHush:
When running demos under `examples` directory, may encounter `UnicodeDecodeError` at the code below
```python
with open("./book.txt") as f:
    rag.insert(f.read())
```
this frustrating error frequently happens on windows, but never happens on Linux.
I've added `encoding` param to avoid it and apply such change to all demos.
```python
with open("./book.txt", "r", encoding="utf-8") as f:
    rag.insert(f.read())
```


LarFii:
Thanks!

PR #110, opened 2024-10-23
[FIX] fix infinite loading hf model bug that cause oom

tackhwa:
- according to current hf code in llm.py, it will initailize a new hf_model object everytime the function is invoke, which causing duplicate loading model in gpu and cause oom.

- this pr cache the hf_model with lru_cache, so that it will not load a new model everytime when the function is invoke.

- before:
image

- after:
image
b6b95b20a8a01ec1e06f16231a9ac8b

### DIFF
--- a/examples/lightrag_siliconcloud_demo.py
+++ b/examples/lightrag_siliconcloud_demo.py
@@ -19,7 +19,7 @@ async def llm_model_func(
         prompt,
         system_prompt=system_prompt,
         history_messages=history_messages,
-        api_key=os.getenv("UPSTAGE_API_KEY"),
+        api_key=os.getenv("SILICONFLOW_API_KEY"),
         base_url="https://api.siliconflow.cn/v1/",
         **kwargs,
     )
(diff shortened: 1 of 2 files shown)


LarFii:
Should the follow code be put into the initialize func?
```python
 if hf_tokenizer.pad_token is None:
        # print("use eos token")
        hf_tokenizer.pad_token = hf_tokenizer.eos_token
```

tackhwa:
> Should the follow code be put into the initialize func?
> 
> ```python
>  if hf_tokenizer.pad_token is None:
>         # print("use eos token")
>         hf_tokenizer.pad_token = hf_tokenizer.eos_token
> ```

i think so, i alr move the code, please check

tackhwa:
> Should the follow code be put into the initialize func?
> 
> ```python
>  if hf_tokenizer.pad_token is None:
>         # print("use eos token")
>         hf_tokenizer.pad_token = hf_tokenizer.eos_token
> ```

i already move the code and tested the code, please feel free to check

PR #138, opened 2024-10-25
[FIX] fix hf output bug (current output contain user prompt which cause logical error in entity extraction phase)

tackhwa:
- the current `response_text` in `hf_model_if_cache` will contain user prompt as shown as follows:
image

-  even though the response_text is being stripped in `operate.py` which make the final response look like normally, but in the process of creating the knowledge graph and vector databse (entity extraction phase), the user prompt is not strip, which will result the prompt `PROMPTS["entity_extraction"]` in `prompt.py` will also consider as assitant response as well and causes the entity in example prompt is being extracted, but these entity is not actually in the text file (pure text database), instead, it just an example entity.

- what does thir pr do: remove user prompt in response as shown as follows:
8933b3a26fef0bed0bab6ae8ad2c5ef

### DIFF
--- a/lightrag/llm.py
+++ b/lightrag/llm.py
@@ -266,10 +266,11 @@ async def hf_model_if_cache(
     input_ids = hf_tokenizer(
         input_prompt, return_tensors="pt", padding=True, truncation=True
     ).to("cuda")
+    inputs = {k: v.to(hf_model.device) for k, v in input_ids.items()}
     output = hf_model.generate(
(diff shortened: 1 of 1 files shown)


LarFii:
This might only solve the issue for certain models, as each model’s output logic can vary quite a bit. : (

tackhwa:
> This might only solve the issue for certain models, as each model’s output logic can vary quite a bit. : (

ya, but at least i tested with few popular model, `qwen`, `llama-3.2`, `internlm`. this code does solve the problem.

PR #139, opened 2024-10-25
Streamlit demo and template for LightRAG

aiproductguy:
Great work on LightRAG! I made a streamlit demo and template that I would like to contribute back at https://lightrag.streamlit.app.

CleanShot 2024-10-25 at 16 10 18@2x

Thoughts?


Dormiveglia-elf:
Good job. But there are several problems I want to address:  First, I think it is a little heavy, which means it is not neccessary to use streamlit to build since it contains many redundant elements and hard to customize styles and functions; Second, I think the primary goal should be to emphasize the Q&A process rather than Knowledge Graph (KG) visualization. Currently, the system may give the impression that KG visualization is the main feature, which could be confusing. Maybe it is better to adjust the layout and workflow to ensure the Q&A component is the focal point, with KG visualization serving as an optional, complementary feature. Third, the visualization currently displays all nodes at the same size, which doesn’t effectively convey the relationships and importance of various nodes. I think it is better to adjust this so that nodes with a higher degree (more connections) are larger. Fourth, you give an option to select mode, what if I want to query with different mode simutaneously to do a comparison? Last but not least, I don't think it is a wise choice to input the api key explicitly to the website which is very dangerous.
Anyway, it is a good try and good demo!

aiproductguy:
Great feedback. Thank you.

Streamlit provides simple cloud deployments for prototyping. What do you suggest?

I want to make another version that focuses from on the Verification. Any input is welcome. 

I like the idea of giving weight to the nodes. I am still learning how to query the base storage.

Let me think about parallel modes feature further.

Trying to balance simplicity with security.

Issue #163, opened 2024-10-29
RuntimeWarning: coroutine 'LightRAG.ainsert' was never awaited   print(f"An error occurred: {e}") 


small122:
运行lightrag_openai_compatible_demo.py报错

Dormiveglia-elf:
Please refer my PR #168 , I think you can run without error now.

small122:
> Please refer my PR #168 , I think you can run the program smoothly without error now.

thank you，i get it

PR #168, opened 2024-10-29
[hotfix-#163] Fix asynchronous problem

Dormiveglia-elf:
I fixed the asynchronous problem. The issue address is [no.163]
#163


LarFii:
Thanks!

PR #177, opened 2024-10-30
fix:Step_3.py context

WinstonCHEN1:
Fix the same bug of Step_3_openai_compatible.py in Step_3.py in reproduce steps


LarFii:
Thanks!

PR #217, opened 2024-11-06
Bug fix & Code Lint: fix RuntimeError: no running event loop

JavieHush:
When trying to get a loop, we'll always get an error said `RuntimeError: no running event loop`. The code could run correctly before several commits, however, it somehow got a bug now...

The problematic code is as follows
```python
def always_get_an_event_loop() -> asyncio.AbstractEventLoop:
    try:
        loop = asyncio.get_running_loop()
    except RuntimeError:
        logger.info("Creating a new event loop in main thread.")
        # loop = asyncio.new_event_loop()
        # asyncio.set_event_loop(loop)
        loop = asyncio.get_event_loop()
    return loop
```
to fix this bug, the code can be modified as follows
```python
def always_get_an_event_loop() -> asyncio.AbstractEventLoop:
    try:
        loop = asyncio.get_running_loop()
    except RuntimeError:
        logger.info("Creating a new event loop in main thread.")
        loop = asyncio.new_event_loop()
        asyncio.set_event_loop(loop)
    return loop
```
This fix could solve [issue #214](https://github.com/HKUDS/LightRAG/issues/214)
Plus, I format the code in the project using ruff-format


wiltshirek:
The Neo4J Async lib has an issue leveraging the existing loop but it can use get_event_loop().   I tested both networKx and Neo4J with the following:
def always_get_an_event_loop() -> asyncio.AbstractEventLoop:
    try:
        **loop = asyncio.get_event_loop()**. 
    except RuntimeError:
        logger.info("Creating a new event loop in main thread.")
        loop = asyncio.new_event_loop()
        asyncio.set_event_loop(loop)
    return loop
    
    This seems to work for both cases.    I can do a quick pull for this.

wiltshirek:
#221

PR #241, opened 2024-11-10
Added GUI (linked), updated README.md

aiproductguy:
I am not sure streamlit is the interface I want to contribute back to the LightRAG project, but willing to share it all under the same MIT license. It seemed you did not like streamlit as well. I am working on taipy or flask next and can contribute that how you suggest.

I figured I would at least share the link and source code in the README.

I would like to understand how you benchmark and want to visualize the investigation/verification workflows for future version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <center><h2>🚀 LightRAG: Simple and Fast Retrieval-Augmented Generation</h2></center>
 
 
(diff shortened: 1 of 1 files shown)


ASAD-BE18:
If you could, please add Ollama support in the streamlit application using API endpoints

LarFii:
Thanks for your GUI! I'm not very familiar with these visualization methods, so I don't have any specific suggestions. For the evaluation of LightRAG, we use a one-to-one approach similar to GraphRAG. We generate users and tasks for a specified dataset using GPT, then have different RAG systems answer the same query.

PR #263, opened 2024-11-12
Only update storage if there was something to insert

detaos:
Before, the `finally` block would always call `_insert_done()`, which writes out the `vdb_*` and `kv_store_*` files ... even if there was nothing to insert (because all docs had already been inserted).  This was causing the speed of skippable inserts to become very slow as the graph grew.


LarFii:
Thanks, you're right.

PR #266, opened 2024-11-13
fix hf embedding to support loading to different device

davidleon:
as title. currently local hf model embed func doesn't take care of loading model to cuda.


LarFii:
Thanks!

PR #295, opened 2024-11-18
change the type of binding parameters in Oracle23AI

tmuife:
I have update the parameter binding type in oracle_impl, so that it can handle some key words appearing in column's content

change:
lightrag/kg/oracle_impl.py


LarFii:
Thanks!

PR #325, opened 2024-11-25
Enhance Query Logic and Add Configurable Features

jin38324:
### 1. Unified Query Functions
The previously separate query functions (`local_query`, `global_query`, `hybrid_query`) shared significant duplicated code. I merged them into a single function, `kg_query`, to improve maintainability and clarity.

To achive this:
- The query mode is now determined by the `query_param.mode` parameter.
- Enhanced keyword handling: when no valid keyword is returned, the query mode adapts dynamically (e.g., switching from hybrid to global or returning an error).
- Unified context generation using `_build_query_context`, streamlining the process into a single return point.

### 2. Configurable Example Count in Prompts
In the original prompt logic, examples occupied a large number of tokens, increasing cost and response time. This update introduces a configurable parameter, `addon_params = {"example_number": 1}`, to limit the number of examples used in entity extraction and keyword generation.

To achive this:
- Updated prompt.py to split prompts into a list of examples.
- If example_number is not specified or exceeds the available examples, all (default: 3) are used.

### 3. Added Language Support for Entity Extraction
Entity extraction now supports language customization via `addon_params = {"language": "Simplified Chinese"}`.

To achive this:
- Added a `language` variable in the prompt's string template.
- Introduced `default_language` in prompt.py (default: English). If no language is specified in addon_params, the default value is used.


lightrag/operate.py
```diff
@@ -399,42 +412,63 @@ async def local_query(
...
+    logger.info("kw_prompt result:")
+    print(result)
     try:
+        json_text = locate_json_string_body_from_string(result)
         keywords_data = json.loads(json_text)
-        keywords = keywords_data.get("low_level_keywords", [])
-        keywords = ", ".join(keywords)
-    except json.JSONDecodeError:
-        try:
-            result = (
```
LarFii, line 440: This section is an adaptation for responses that don't fully adhere to the JSON format. It needs to be kept as is. For example, some responses are enclosed with:
\```json
xxxxxxxxx
\```

lightrag/llm.py
```diff
@@ -69,12 +69,15 @@ async def openai_complete_if_cache(
     response = await openai_async_client.chat.completions.create(
         model=model, messages=messages, **kwargs
     )
-
+    content = response.choices[0].message.content
+    if r"\u" in content:
+        content = content.encode("utf-8").decode("unicode_escape")
+    print(content)
```
LarFii: Should the `print` statement here be removed?

lightrag/operate.py
```diff
@@ -461,22 +497,114 @@ async def local_query(
...
+
+
+async def _get_node_data(
     query,
     knowledge_graph_inst: BaseGraphStorage,
     entities_vdb: BaseVectorStorage,
     text_chunks_db: BaseKVStorage[TextChunkSchema],
     query_param: QueryParam,
 ):
+    # 获取相似的实体
```
LarFii: The comments may need to remain in English.

examples/lightrag_oracle_demo.py
```diff
@@ -97,6 +97,7 @@ async def main():
             graph_storage="OracleGraphStorage",
             kv_storage="OracleKVStorage",
             vector_storage="OracleVectorDBStorage",
+            addon_params={"example_number": 1, "language": "Simplfied Chinese"},
```
LarFii: In the demo, let's keep the default language as English. You can modify the README to add usage instructions.

examples/lightrag_api_oracle_demo..py
```diff
@@ -81,7 +81,7 @@ async def get_embedding_dim():
 
 async def init():
     # Detect embedding dimension
-    embedding_dimension = await get_embedding_dim()
+    embedding_dimension = 1024  # await get_embedding_dim()
```
LarFii: Should we keep `await get_embedding_dim()` here?

lightrag/operate.py
```diff
@@ -461,22 +497,114 @@ async def local_query(
...
+            ll_text_units_context,
+        )
+    elif query_param.mode == "global":
+        entities_context, relations_context, text_units_context = (
+            hl_entities_context,
+            hl_relations_context,
+            hl_text_units_context,
+        )
+    return f"""
+# -----Entities-----
```
LarFii: Should we delet the `#` here?

lightrag/operate.py
jin38324: The function `locate_json_string_body_from_string` use `re` to parse json text, all formate like`{......}` can be  matched. So I think no need to prcess the LLM output text use `replace`.

lightrag/llm.py
jin38324: It's for debug, should be removed. I have fixed follow your advice.

lightrag/operate.py
jin38324: I have fixed follow your advice.

examples/lightrag_oracle_demo.py
jin38324: I have fixed follow your advice.
Information about all parameters is added into the readme.

examples/lightrag_api_oracle_demo..py
jin38324: It's for debug, should be removed. I have fixed follow your advice.

lightrag/operate.py
jin38324: It's typo. I have fixed follow your advice.

jin38324:
I forgot one more feature.
The `only_need_prompt parameter` is added to` QueryParam`. 
The purpose of this is to enable lightrag to be efficiently integrated into other applications. By setting `only_need_prompt=true`, you can get the prompt constructed after the graph query, and send this prompt directly to LLM to get the answer.

Use case:
1. Send the prompt word to the LLM API to get streaming output;
2. By enabling the lightrag API server, other applications (such as `Dify`) can use lightrag as an external knowledge base, get the prompt from lightrag, and then call LLM by the application.

PR #343, opened 2024-11-28
fix templating of language in prompts

### LINKED ISSUE #342: BUG: use chinese language is now hardcoded in prompts
I believe the intention was to template the language and make it configurable in the prompts but some instances have been missed and are now hard-coded to "chinese".

- https://github.com/HKUDS/LightRAG/blame/main/lightrag/prompt.py#L181
- https://github.com/HKUDS/LightRAG/blob/main/lightrag/prompt.py#L134

sebastianschramm:
resolves #342 

- Replace hardcoded chinese language in prompts with templated language
- Template missed instance of English language
- Adjust usage of fixed prompts to supply language as context


LarFii:
Thanks for your contribution!

PR #345, opened 2024-11-28
add jina embedding support

davidleon:


LarFii:
Thank you so much for your contributions! It would be great if you could also add a demo for Jina embedding. Before submitting, please make sure to run pre-commit run --all-files to ensure everything is up to standard.

Issue #209, opened 2024-11-05
Lots of RuntimeError: Event loop is closed


XenonAesir:
Running lightrag_ollama_demo.py with gemma:2b orlightrag_openai_compatible_demo.py with qwen2.5-3b-instruct wil cast a lot of this:
```
Exception ignored in: <function _ProactorBasePipeTransport.__del__ at 0x0000024507A5AB90>
Traceback (most recent call last):
  File "C:\Environment\anaconda3\envs\lightrag\lib\asyncio\proactor_events.py", line 116, in __del__
    self.close()
  File "C:\Environment\anaconda3\envs\lightrag\lib\asyncio\proactor_events.py", line 108, in close
    self._loop.call_soon(self._call_connection_lost, None)
  File "C:\Environment\anaconda3\envs\lightrag\lib\asyncio\base_events.py", line 745, in call_soon
    self._check_closed()
  File "C:\Environment\anaconda3\envs\lightrag\lib\asyncio\base_events.py", line 510, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```
And some type of outputs will disappear, such as "global", "hybrid"

TashaSkyUp:
Yeah I think I found the issue. Ill make PR and update this issue.

TashaSkyUp:
fix is here : https://github.com/HKUDS/LightRAG/pull/354

buaali1025:
have you solved your issue?  What should i do? Thanks

PR #354, opened 2024-11-29
fix for #209

TashaSkyUp:
function was returning a closed event loop.


LarFii:
Thanks for your contribution!

PR #359, opened 2024-11-30
Update NanoVectorDBStorage upsert to use hash values for IDs

FatRicePaddyyyy:
- In the delete_entity method, a hash value is calculated when deleting an entity.
- On the other hand, in the upsert method, the index was used directly as the ID when adding an entity.
- Therefore, I modified the upsert method to also calculate a hash value.

### DIFF
--- a/lightrag/storage.py
+++ b/lightrag/storage.py
@@ -86,7 +86,7 @@ class NanoVectorDBStorage(BaseVectorStorage):
             return []
         list_data = [
             {
-                "__id__": k,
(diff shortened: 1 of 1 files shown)


LarFii:
Thanks for your suggestion! But the hash value is already calculated in `operate.py` before calling the `upsert` method.

FatRicePaddyyyy:
I see. Sorry for jumping to conclusions. I'll cancel the pull request.

PR #360, opened 2024-11-30
Azure OpenAI Embedding

ahmadhatahet:
+ The "azure_openai_embedding" function was missing the "api_version" argument.
Previously, it was only read from an environment variable not passed from the user.

+ The "max_token_size" was incorrect. Now it corrected to 8191, according to OpenAI documentation.
https://platform.openai.com/docs/guides/embeddings/#embedding-models


ahmadhatahet:
+ same with "azure_openai_embedding". The function "azure_openai_complete_if_cache" lacked "api_version"

LarFii:
Thanks for your contribution!

PR #383, opened 2024-12-03
added nvidia text-embedding API and example of using nvidia API llm a…

MRX760:
Changes:
- Adding nvidia text-embedding API (**llm.py**)
- Adding simple usage example of nvidia API for LLM and text-embedding model (**lightrag_nvidia_demo.py**)

Additional note:
- Text-embedding docs: https://docs.nvidia.com/nim/nemo-retriever/text-embedding/latest/reference.html#:~:text=NV-Embed-QA%20and%20E5%20models%20operate%20in%20passage%20or,is%20very%20important%20to%20use%20the%20correct%20input_type.
- Supported nvidia NIM model type: https://build.nvidia.com/nim

### DIFF
--- a/examples/lightrag_nvidia_demo.py
+++ b/examples/lightrag_nvidia_demo.py
@@ -1,0 +1,159 @@
+import os
+import asyncio
+from lightrag import LightRAG, QueryParam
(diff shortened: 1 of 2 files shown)


LarFii:
Thanks for your contribution!

PR #384, opened 2024-12-03
fix examples' prompt bug

Ecocytus:
1. examples in the latest main branch was changed to be cutomizable, but a bug was introduced, that the string literal format is missing for the `PROMPTS["entity_extraction_examples"` in `lightrag/prompt.py` . It is directly used in the line 265 of `lightrag/operate.py` without formating.
2. fixed a typo (missing right bracket) in the system prompt for entity extraction.


Ecocytus:
Just saw this problem has already been issued here #377

LarFii:
非常感谢！

PR #396, opened 2024-12-04
fix: example llm_model_func unexpected keyword argument error

linbuxiao:
- If the keyword_extraction argument is not passed explicitly, it is passed as kwargs to the `openai_async_client.chat.completions.create` function. Raised an error `TypeError: AsyncCompletions.create() got an unexpected keyword argument 'keyword_extraction'`.
- how to fix: add the `keyword_extraction` parameter just like implementations such as `gpt_4o_mini_complete`.


LarFii:
Thanks! Regarding this issue, I have made the necessary updates.

Issue #77, opened 2024-10-21
Does it support streaming output?


redpintings:
If supported, how should I proceed? Please help me.
If not supported, will it be supported later?

LarFii:
We do not support streaming output at the moment. It might be added in the future, but it will take some time as we have many other tasks to work on right now.

PR #407, opened 2024-12-06
Add support for Ollama streaming output and integrate Open-WebUI as the chat UI demo

partoneplay:
#77 and #93

`lightrag_api_open_webui_demo.py` is a demo API server that mocks the Ollama API, enabling  integration with the [Open-WebUI](https://github.com/open-webui/open-webui). The default RAG implementation in Open-WebUI lacks sufficient quality.


LarFii:
Thank you very much for sharing this! However, since we've recently updated a caching mechanism, there might be some conflicts that need to be resolved.

PR #415, opened 2024-12-06
Fix: Ensure entity_or_relation_name is a string in _handle_entity_relation_summary

SaujanyaV:
# Description

I encountered an issue where the function `_handle_entity_relation_summary` raised the following error:

```python
Argument 'entity_or_relation_name' has incorrect type (expected str, got tuple)
```

This error occurred because the function was being called with a tuple `(src_id, tgt_id)` instead of a string. To resolve this issue, I made the following changes:

### Converted the Tuple to a Formatted String

Replaced the call to `_handle_entity_relation_summary` with:

```python
description = await _handle_entity_relation_summary(
    f"({src_id}, {tgt_id})", description, global_config
)
```

This ensures that the `entity_or_relation_name` argument is always passed as a string.

### Maintained Consistency

Updated the code to use formatted strings for better readability and alignment with the function's expectations.

---

## Steps to Reproduce the Issue

1. Use the `_merge_edges_then_upsert` function with a graph edge where the tuple `(src_id, tgt_id)` is passed to `_handle_entity_relation_summary`.
2. Observe the type mismatch error when the function is executed.

---

## Solution

Ensure `entity_or_relation_name` is a string by using a formatted string `f"({src_id}, {tgt_id})"` instead of the tuple.


LarFii:
Thanks!

PR #429, opened 2024-12-09
fix extra kwargs error: keyword_extraction.

davidleon:
add lazy_external_load to reduce external lib deps whenever it's not necessary for user.


LarFii:
Thanks!

PR #444, opened 2024-12-10
Fix/lazy import

davidleon:


LarFii:
Thank you for your contribution! Next time, you can consider combining multiple PRs into one for better organization.

PR #447, opened 2024-12-10
feat: Add ChromaDB integration for vector storage

spo0nman:
- Implemented `ChromaVectorDBStorage` class in `lightrag/kg/chroma_impl.py` to support ChromaDB as a vector storage backend.
- Updated `lightrag.py` to include `ChromaVectorDBStorage` in the storage class mapping.
- Added a test script `test_chromadb.py` to demonstrate the usage of ChromaDB with LightRAG, including configuration for embedding functions and ChromaDB connection settings.
- fix lazy import function to support package context for dynamic class loading. https://github.com/HKUDS/LightRAG/pull/444/commits/288d4b8355eb3fd1fdfaf165f1380501b58aaf67


LarFii:
Thanks for sharing!

PR #464, opened 2024-12-13
Maybe very important!!! Fix embedding error

billvsme:
The asyncio.as_completed() function does not guarantee that the results are ordered

When using this project, I found that the entities and relationships of the recall were somewhat related to the low-level keywords and the high-level keyword, but it was not very related, and I felt that the ability of the vector model was not fully utilized, it maybe a strange problem. 

I looked at the code of inserting vectors and found problem. ```asyncio.as_completed```  is used incorrectly, ```asyncio.as_completed``` results are not returned in the order in input tasks, but in the order in which the tasks were completed. I changed to use ```asyncio.gather``` to make sure the results were orderly. And ensure that ```tqdm``` works properly.

This code need ```asyncio.as_completed``` results were orderly 

https://github.com/HKUDS/LightRAG/blob/b7552f35aa6a7e61de585c2a6beaed59a7e7b937/lightrag/storage.py#L110-L112

### DIFF
--- a/lightrag/kg/milvus_impl.py
+++ b/lightrag/kg/milvus_impl.py
@@ -59,7 +59,7 @@ class MilvusVectorDBStorge(BaseVectorStorage):
         embedding_tasks = [self.embedding_func(batch) for batch in batches]
         embeddings_list = []
         for f in tqdm_async(
-            asyncio.as_completed(embedding_tasks),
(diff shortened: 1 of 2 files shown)


LarFii:
Thank you so much for your valuable contribution. This is indeed an issue that can occur.

PR #494, opened 2024-12-19
Add Gremlin graph storage

alllexx88:
Add Gremlin graph storage.

Note that at least my setup suffered from an `anyio` issue, that has been fixed a couple of days ago, with commit https://github.com/agronholm/anyio/commit/c518300593f2411196fd492d917269dcb6f7682b so I had to install it from git before the next version is released.

### DIFF
--- a/examples/lightrag_ollama_gremlin_demo.py
+++ b/examples/lightrag_ollama_gremlin_demo.py
@@ -1,0 +1,85 @@
+import asyncio
+import inspect
+import logging
+import os
+
+logging.basicConfig(format="%(levelname)s:%(message)s", level=logging.WARN)
+
+from lightrag import LightRAG, QueryParam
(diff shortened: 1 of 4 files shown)


alllexx88:
I see the linting error report, due to `logging.basicConfig(format="%(levelname)s:%(message)s", level=logging.WARN)` line before the lightrag imports. I can fix it, but setting logging level with this line doesn't change the loglevel when issued *after* lightrag imports.

LarFii:
Thank you for your contribution! To avoid the linting error during , you can temporarily comment out this line. Additionally, you can add a comment to describe its functionality.

alllexx88:
Thanks @LarFii ! Done together with one little fix (using `await asyncio.gather` instead of consecutive `await`s in `get_node_edges()`)

LarFii:
Thanks!

PR #505, opened 2024-12-23
GremlinStorage: support more Gremlin compatible DBs

alllexx88:
Tested on ArcadeDB with Gremlin plugin. The main change is using "entity_name" vertex property instead of the label as a node_id since different implementations have different restrictions on label names.


LarFii:
Thanks!

PR #515, opened 2024-12-27
fix keywords_extraction_examples format error

zhangch-ss:
hi, thanks for your great work. 
when I ran lightrag_azure_openai_demo.py, 
I found that there were some problems with the format of ‘keywords_extraction_examples’. 
```
{{
"high_level_keywords": ["International trade", "Global economic stability", "Economic impact"],
"low_level_keywords": ["Trade agreements", "Tariffs", "Currency exchange", "Imports", "Exports"]
}}
```
 it  caused unexpected output.
image

I fixed the format of keywords_extraction_examples and it seems to work well now.
image

### DIFF
--- a/lightrag/prompt.py
+++ b/lightrag/prompt.py
@@ -211,30 +211,30 @@ PROMPTS["keywords_extraction_examples"] = [
 Query: "How does international trade influence global economic stability?"
 ################
 Output:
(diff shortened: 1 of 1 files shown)


LarFii:
Thanks!

PR #516, opened 2024-12-27
Fix the async issue while running on Windows

ShanGor:
When running with Apache AGE impl on Windows (with psycopg), it will got async loop error, the fix is to register asyncio with WindowsSelectorEventLoopPolicy. Only effect on Windows. Tested okay.


LarFii:
Thanks!

PR #545, opened 2025-01-06
Enhance the llm_cache_kv_store, enable the llm_cache for entity extraction and revise readme

ShanGor:
To solve the paint point that when the beginners trying to run the first test, it might failed somewhere, and the LLM response for entity extractions are all wasted, gone with money and time.
- Provide an option to enable the llm cache for entity extraction (optional and by default not enabled). Which can significantly save the beginner's money and time.
- Revised the readme about postgres implementation.
- Added tutorial for contributors to do pre-commit.

Tested the codes with below scenarios:
- Test with enabling the llm cache, for default storage: passed
- Test with disabling the llm cache, for default storage: passed
- Test with enabling the llm cache, for postgres storage: passed
- Test with disabling the llm cache, for postgres storage: passed


LarFii:
Thanks for your contributions!

PR #553, opened 2025-01-07
Implement custom chunking feature

GurjotSinghShorthillsAI:
I implemented a custom insert function by the name 'insert_custom_chunks' where user can create chunks themselves and pass the list of chunks as an input to this function (instead of usual text only)

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -458,6 +458,72 @@ class LightRAG:
                     # Ensure all indexes are updated after each document
                     await self._insert_done()
 
+    def insert_custom_chunks(self, full_text: str, text_chunks: list[str]):
+        loop = always_get_an_event_loop()
+        return loop.run_until_complete(self.ainsert_custom_chunks(full_text, text_chunks))
+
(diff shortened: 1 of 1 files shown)


mariowolframm:
So does this imply I should rather use one of the more sophisticated chunking methods (easily available as part of Langchain, LLamaindex etc.) and may expect better results? I love it

LarFii:
That's cool! Thanks for your contribution. But there are some linting errors. Please make sure to run pre-commit run --all-files before submitting to ensure all linting checks pass.

GurjotSinghShorthillsAI:
> So does this imply I should rather use one of the more sophisticated chunking methods (easily available as part of Langchain, LLamaindex etc.) and may expect better results? I love it

Yes! You can make chunks as you want using different text splitters from langchain or you can just make chunks yourself based on certain delimitters and then give those chunks to lightrag to get inserted!

GurjotSinghShorthillsAI:
> That's cool! Thanks for your contribution. But there are some linting errors. Please make sure to run pre-commit run --all-files before submitting to ensure all linting checks pass.

Screenshot from 2025-01-09 00-40-49
I've made the changes! Thanks for the review!

PR #572, opened 2025-01-11
fix: Resolve 500 error caused by missing `len()` for `LightRAG`

iridium-soda:
This PR fixes a critical issue where the `/documents/text` endpoint returned a 500 error due to a missing `len()` implementation for the LightRAG object. The issue was caused by passing `len(rag)` to the `document_count` field in the InsertResponse but the class `LightRAG` has no `len()` method to use. This bug affects the /documents/text endpoint across multiple servers, including Ollama, OpenAI, and LOLMS. A mock test has been implemented to simulate the issue, and all tests have passed successfully, confirming the fix.


LarFii:
@ParisNeo LightRAG indeed does not have a `len()` method. Since the API Servers have already been consolidated, should we modify `document_count=1` at the indicated location in the picture?
image

ParisNeo:
Hi,

I have fixed this issue in last pull request.

Thanks for the information. That was an oversight.

PR #585, opened 2025-01-14
Add custom function with separate keyword extraction for user's query and a separate prompt

gurjot-05:
This PR introduces a significant enhancement to our RAG module's keyword extraction capabilities by implementing the `query_with_separate_keyword_extraction` function. Previously, our approach involved extracting keywords from a combined string of both the user's prompt and the query. This often led to the extraction of irrelevant keywords, as the prompt might contain additional language that does not necessarily reflect the core intent of the query.

## Changes Made
The new function, query_with_separate_keyword_extraction, addresses this issue by explicitly separating the user's query from the prompt. This allows the system to perform keyword extraction solely on the query, ensuring that only relevant keywords are considered when forming the response. The prompt is then utilized to format the response appropriately, without influencing the keyword extraction process.

## Example usage
```
rag.query_with_separate_keyword_extraction(
    "What are the top themes in this story?",
    "You need to answer me this question in as simple language as you can. Just know that a 5 year old should understand it.",
    param=QueryParam(mode="hybrid")
)
```
This improvement is expected to enhance user interaction by making the responses not only more relevant but also tailored to the context provided by the prompt. It prevents the dilution of keyword significance when the prompt includes supplementary instructions or context not directly related to the content of the query.


LarFii:
Thank you so much for your contribution! Could you please add a detailed explanation in the `README.md` and include specific usage examples in the `examples/`?

gurjot-05:
> Thank you so much for your contribution! Could you please add a detailed explanation in the `README.md` and include specific usage examples in the `examples/`?

On it!

gurjot-05:
> Thank you so much for your contribution! Could you please add a detailed explanation in the `README.md` and include specific usage examples in the `examples/`?

I have added the detailed explanation in `README.md` and also included specific usage example in `examples/`

PR #590, opened 2025-01-16
Enhance Robustness of insert Method with Pipeline Processing and Caching Mechanisms

jin38324:
The current `insert` method in LightRAG writes results to persistent storage only upon successful completion of all operations. This design means that any error—be it network issues, database failures, or bugs—forces users to restart the entire process, even if 99% of the work was completed. This all-or-nothing approach can lead to significant inefficiencies and user frustration.

**Enhancements Introduced:**

1. **Data Storage Optimization with LLM Cache:**
   - Implemented `enable_llm_cache_for_entity_extract = True` to ensure that previously sent prompts are retrieved directly from the cache, reducing redundant API calls and enhancing efficiency.

2. **Pipeline-Based Knowledge Graph Construction:**
   - Refactored the `insert` process into three distinct asynchronous pipeline stages, each capable of resuming from interruptions without reprocessing completed tasks:

     1. **`apipeline_process_documents(self, string_or_strings)`**:
        - Removes duplicate input texts.
        - Generates document IDs.
        - Filters out documents already stored.
        - Stores new documents.
        - Sets document status to pending.

     2. **`apipeline_process_chunks(self)`**:
        - Retrieves pending and failed documents.
        - Splits documents into chunks.
        - Inserts chunks, invoking the embedding model.
        - Sets chunk status to pending.
        - Updates document status to processed.

     3. **`apipeline_process_extract_graph(self)`**:
        - Retrieves pending or failed chunks.
        - Extracts entities and relationships from each chunk using the LLM model.
        - Inserts entities and relationships.
        - Updates chunk status to processed.

   - This structured approach ensures that each LLM call is utilized effectively, with results either stored in the cache or directly in entities and relationships, eliminating wasted computations.

**Note:**

The pipeline methods have been adapted for **Oracle database**, necessitating the addition of a status field to track the state of documents and chunks. The recently committed `JsonDocStatusStorage` was not utilized due to its complexity.

The original `insert` is still available, so non-Oracle storage can still use the old code

These enhancements collectively improve the robustness and efficiency of the `insert` method, ensuring that users can resume interrupted processes without redundant operations, thereby enhancing the overall user experience with LightRAG.

**Additional Optimizations:**

- **Test File Organization:**
  - Moved certain test files to the `example` folder for better project structure.

- **Selective Storage Loading:**
  - Modified `lightrag.py` to load only the necessary storage classes instead of all 20, improving resource utilization.

- **Global Configuration Handling:**
  - Embedded `global_config` as a fixed parameter in `lightrag.py`, removing the need to pass it during each class instantiation.

- **Non-ASCII Character Support:**
  - In `operate.py` at line 356, added `ensure_ascii=False` to `json.dumps(history_messages)` to prevent non-English characters from being converted to ASCII.

- **LLM Call Statistics:**
  - Introduced `statistic_data` in `utils.py` to track the number of LLM calls during knowledge graph construction.

- **Logging Level Adjustment:**
  - Set `logging.getLogger("httpx").setLevel(logging.WARNING)` in `utils.py` to suppress excessive log outputs.

- **Bug Fixes:**
  - In `utils.py` at line 462, changed `if mode == "naive"` to `if mode == "default"`, resolving a bug that previously hindered execution.
  - At line 484 in `utils.py`, updated `embedding_model_func = hashing_kv.global_config["embedding_func"]["func"]` to append `.func`, correcting a bug that prevented proper function execution.


LarFii:
Thanks for your contributions!

PR #608, opened 2025-01-20
Temporary fix to introduced problem in ollama and lollm

ParisNeo:
Hi,

Sorry. I just found out that passing the headers to the ollama and lollms directly from llm has a conflict with another header generated inside this so we need to rewrite it.

To fix the problem I just removed the header asignment temporarily and I'll rewrite it this evening.

I tested it last night but it seems that I wasn't testing the right version (my bad)

Sorry for the error. This should fix it.
Have a nice day

### DIFF
--- a/lightrag/llm.py
+++ b/lightrag/llm.py
@@ -354,7 +354,7 @@ async def ollama_model_if_cache(
                 "Content-Type": "application/json",
                 "Authorization": f"Bearer {api_key}"
                 } if api_key else {"Content-Type": "application/json"}
-    ollama_client = ollama.AsyncClient(host=host, timeout=timeout, headers=headers)
+    ollama_client = ollama.AsyncClient(host=host, timeout=timeout)# temporary fix: (TODO: rewrite this with better compatibility), headers=headers)
     messages = []
     if system_prompt:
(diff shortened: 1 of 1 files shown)


LarFii:
No worries. But there are some linting errors that need to be fixed.
Thanks, and have a great day!

ParisNeo:
Ok, I suspect that the real problem comes fom another update beceause yesterday it worked on my PC.
I'll analyze the problem more then make a pull request when nready. I'll also do linting

ParisNeo:
OK, found the problem and fixed it. Now everything is working fine. We just need to update the linting and we're back on track.

ParisNeo:
Would you please accept and merge the update then do a linting with pre-commit. I can't do that immediately. I can do it this evening when I'm back home, but It would be wize not to keep the repo with a serious error like this for long.

Thanks

ParisNeo:
TO sum up the problem:
I have added the possibility to access lollms or ollama behind a secure access with BEARING Key.

In the code I specified that the header will be None if no key is specified.
But it shouldn't be none, it should at least be an empty dict or a dict with content type.

So I have updated this so that if no key is pecified, we use a default header.

LarFii:
Sure, no problem.

PR #633, opened 2025-01-23
Optimize Dependencies and Minimal Setup for fewer dependencies

ceberhardt:
# Optimize Dependencies and Add Minimal Setup

## Changes
- Restructured package dependencies:
  - Set minimal_requirements for core functionality as default:
    - numpy
    - pydantic
    - tiktoken
    - aiohttp
    - nano-vectordb
    - networkx
    - openai
    - ollama
    - tenacity
    - tqdm
    - xxhash
    - setuptools
    - python-dotenv
  - Moved heavy dependencies (`torch`, `transformers`, `aioboto3`) to be imported only when needed
  - Added full extra for all dependencies from requirements.txt `pip install -e .[full]`
- Code improvements:
  - Moved imports inside functions to reduce startup time and memory footprint

This PR makes LightRAG more lightweight and modular, allowing users to install only the dependencies they need for their specific use case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,16 +51,35 @@ This repository hosts the code of LightRAG. The structure of this code is based 
 
 ## Install
 
-* Install from source (Recommend)
-
+* Install from source (Recommended)
 ```bash
 cd LightRAG
-pip install -e .
+pip install -e .  # Minimal install with core dependencies
+pip install -e ".[full]"  # Full install with all dependencies
 ```
+
 * Install from PyPI
 ```bash
-pip install lightrag-hku
+pip install lightrag-hku  # Minimal install with core dependencies
+pip install "lightrag-hku[full]"  # Full install with all dependencies
 ```
+
+The minimal install includes only core dependencies:
+- numpy
+- pydantic
+- tiktoken
+- aiohttp
+- nano-vectordb
+- networkx
+- openai
+- ollama
+- tenacity
+- tqdm
+- xxhash
+- setuptools
+- python-dotenv
+
+For additional features like using local models (torch, transformers) or specific storage backends, use the full install.
 
 ## Quick Start
 * [Video demo](https://www.youtube.com/watch?v=g21royNJ4fw) of running LightRAG locally.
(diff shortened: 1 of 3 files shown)


LarFii:
Thanks for your contributions! There are some linting errors. Please make sure to run `pre-commit run --all-files` before submitting to ensure all linting checks pass.

ParisNeo:
Hi there, I have completely reworked the llm system so that it only installs things when needed and did split all llms into an llm folder.
Make sure to pull latest updates and take a look at what you can enhance from there.

YanSte:
Hi,
Will be merge soon ?

Thanks by advance.

ParisNeo:
Hi there.

I think we don't need this update anymore.
The repo has evolved too much and the structure has changed in order to minimize the number of dependancies to the maximum (more than the ones you propose in your PR)

Take a look at the new structure, it became more flexible and modular.

Sorry for the inconvenance, but I think you can pull the last version and see for yourself.

Best regards

YanSte:
Why aren’t you using a pyproject.toml like @ParisNeo for dependencies? Many projects are based on that rather than relying on pip’s master.

YanSte:
@LarFii 

Currently, the project does not use a pyproject.toml file for managing dependencies. Many modern projects follow this approach.

[pipmaster](https://github.com/ParisNeo/pipmaster) is a small library maintained by a single person, making it difficult to maintain.
Many people have complained about the dependencies.

Please consider using pyproject.toml for dependency or using this PR.

ParisNeo:
Pip master is in its final stage. It is a simple layer on top of pip which is the most ancient and stable package managing system so there is no big risk here.

You can add pyproject.toml on top of what was done. The new version has everything split into modules. pipmaster allows you to have modular system. you don't need to install everything if you only use 1% of the things. Install them as they go.

Just take a look at the structure. I think there is a space for both.

To be honest I prefer using uv but using toml projectfile is also fine. Just use it for the main dependencies if you want.

YanSte:
Ok I will make a pull request.

YanSte:
@LarFii we can close this PR

PR #642, opened 2025-01-24
asyncio optimizations

dimatill:
- Makes the hybrid mode 2x faster by executing local and global pipelines in parallel
- Optimizes other pieces of code in the same way


LarFii:
Thanks for your contribution!

PR #647, opened 2025-01-25
feat:  Added webui management, including file upload, text upload, Q&…

18277486571HYB:
Added webui management, including file upload, text upload, Q&A query, graph database management (can view tags, view knowledge graph based on tags), system status (whether it is good, data storage status, model status, path),request /webui/index.html

### DIFF
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -62,7 +62,7 @@ VECTOR_STORAGE = "NanoVectorDBStorage"
 
 # read config.ini
 config = configparser.ConfigParser()
-config.read("config.ini")
+config.read("config.ini", "utf-8")
 # Redis config
 redis_uri = config.get("redis", "uri", fallback=None)
 if redis_uri:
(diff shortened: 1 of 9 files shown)


LarFii:
Thanks for your contributions! It would be helpful to change all the comments to English so that other developers can easily work on the project in the future. Thank you again!

ParisNeo:
Hi, I can translate them if you want.
And I think we can merge my webui with this one if you don't mind.
I can do that after the pull request is merged.

18277486571HYB:
Whatever，you can do

18277486571HYB:
Do you need any assistance from me?

ParisNeo:
I just need to wait for it to be accepted before I can fuse it so we can start from a clean slate.

I took a look at your code, I think I can handle it without help :) .
If I really don't understand something I'll just ask you.

Thanks alot.

ParisNeo:
Well I just checked it. In fact, it is not much more advanced than my webui. The things you add (the graph management) are not implemented so they don't work. Do you plan on adding implementation of that?

Also, you shouldn't use BASE_URL in the lightrag.js script as it is already served from a specific url and you don't need to use any BASI_URLK, just use /the end point and it will work even if you change the server address.

I'm keeping my webui, but I'll add some ideas from yours and i'll also add a placeholder so that you can implement graph visualization if you like that.

PR #658, opened 2025-01-27
Query with your custom prompts

MdNazishArmanShorthillsAI:
Feature:
- Query by adding your own prompt default prompt type will be of light rag
Bring your own prompt and generate response in the format you want


MdNazishArmanShorthillsAI:
@LarFii Please approve

PR #684, opened 2025-01-31
Fixes for concurrent document indexing and querying

danielaskdd:
### Problems Addressed

1. When `enable_llm_cache_for_entity_extract` is True and `enable_llm_cache` is False, there is race condition when doing document indexing and user query at the same time.  Func `extract_entities` temporary modify global config `enable_llm_cache` to True, causing user query also using cache temporary.
2. Performance bottleneck during document indexing due to LLM function instance exhaustion, causing user query latency. The `limit_async_func_call` decorator  implementation using sleep to wait for another thread to release resource is so bad, that user query will not have a chance to run until document index is over.
3. Thread safety issues in NanoVectorDB's disk persistence operations.

### Solutions

1. Entity Extraction Improvements:

- Intruduce `force_llm_cache` parm to `handle_cache`, forcing the function handle cache no mater what global config settings are. 
-  Modify `extract_entities` to use this new parm to control the behavior of handle_cache. 

2. Concurrent Operation Optimization:

- Replaced function call limiting mechanism of limit_async_func_call with asyncio.Semaphore, ensuring FIFO order to prevent query starvation during indexing.
- Removed redundant semaphore logic from EmbeddingFunc

3. Introduce asyncio.Lock to protect file save operations in NanoVectorDBStorage

Above changes improve concurrent document indexing and querying reliability while maintaining system performance and improve thread safety.

### Bug fixes

1. Fix `quantize_embedding` for not support embedding vector of list type
2. Fix `cache_type` missing in llm_response_cache, and add new cache_type `extract` for entities extraction llm respond cache.
3. Fix prompt respond cache fail when `is_embedding_cache_enabled` is true.
4. Fix `llm_model_func` retrieval error in `handle_cache`,  this function is for LLM similarity verification.
5. Update similarity_check prompt to avoid generating two scores sometimes.


danielaskdd:
@ParisNeo This update involves some core functionalities in LightRAG, and I hope you could help with a preliminary review of the PR.

PR #696, opened 2025-02-02
Add the ability to specify a path for saving lightrag.log

ultrageopro:
When initializing LightRAG, you can specify `log_dir`, where `lightrag.log` will be saved.
The default value is the project’s root directory.

This can be useful if the project contains other log files that should be stored in a dedicated directory.


varyakchau:
Please, review this pull request, I badly need this feature!

PR #709, opened 2025-02-04
Reapply "Integrated the GraphML Visualizer as an optional component of LightRAG"

ArnoChenFx:


LarFii:
Thank you so much!

LarFii:
What I was unsure about is whether `extra/VisualizationTool/graph_visualizer.py` was handled correctly, because I noticed that there isn’t this file under `lightrag/tools/lightrag_visualizer`.

LarFii:
Oh, I found it. It's in `init.py`. Do we need to change the file name?

LarFii:
I looked through the doc, and I think I understand it now. Thank you so much for your help!

PR #715, opened 2025-02-05
Fix: AttributeError in NanoVectorDB Initialization

Vasundhhara:
Issue
The following error occurred during the initialization of NanoVectorDB in nano_vector_db_impl.py:

File ".../LightRAG/LightRAG/lightrag/kg/nano_vector_db_impl.py", line 92, in __post_init__
    self.embedding_func.embedding_dim, storage_file=self._client_file_name
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'function' object has no attribute 'embedding_dim'

Root Cause
The embedding_func variable was declared locally instead of being assigned to self.embedding_func.
As a result, self.embedding_func was referring to a function instead of an instance of EmbeddingFunc, causing the AttributeError.

Changes Made
1. Explicitly Initialized self.embedding_func in __post_init__ and query using:

self.embedding_func = EmbeddingFunc(
    embedding_dim=4096, max_token_size=8192, func=gpt_4o_mini_complete
)

2. Imported EmbeddingFunc and gpt_4o_mini_complete to avoid reference errors.

### DIFF
--- a/lightrag/kg/nano_vector_db_impl.py
+++ b/lightrag/kg/nano_vector_db_impl.py
@@ -54,6 +54,10 @@ from tqdm.asyncio import tqdm as tqdm_async
 from dataclasses import dataclass
 import numpy as np
 import pipmaster as pm
+from lightrag.base import EmbeddingFunc
+from lightrag.llm.openai import gpt_4o_mini_complete
+
+
 
 if not pm.is_installed("nano-vectordb"):
     pm.install("nano-vectordb")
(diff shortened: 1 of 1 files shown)


LarFii:
Thanks for your contribution!  But there are some linting errors. Please make sure to run `pre-commit run --all-files` before submitting to ensure all linting checks pass.

danielaskdd:
@LarFii the embedding function is setup by post_init of LightRAG according to user preference. It can not be set to a fix method in the DB implementation.

ArnoChenFx:
It's not a good idea to create the embedding_func in NanoVectorDBStorage. The embedding_func of VectorDB is automatically passed in by LightRAG. You only need to pass the embedding_func as a parameter when creating LightRAG.

```python
from lightrag.llm.openai import gpt_4o_complete, openai_embed

rag = LightRAG(
    llm_model_func=gpt_4o_complete,
    embedding_func=EmbeddingFunc(
        embedding_dim=1024, 
        max_token_size=8192, 
        func=lambda texts: openai_embed(
            texts,
            model="text-embedding-3-small"
        )
    )
)
```

LarFii:
> @LarFii the embedding function is setup by post_init of LightRAG according to user preference. It can not be set to a fix method in the DB implementation.

Oh, I got it. Thanks!

Vasundhhara:
@ArnoChenFx This solved the error without having to initialize the embedding_func in NanoVectorDBStorage. Thanks.

PR #724, opened 2025-02-06
feat: trimming the model’s reasoning

ultrageopro:
#### Using "Thinking" Models (e.g., DeepSeek)

To return only the model's response, you can pass `reasoning_tag` in `llm_model_kwargs`.

For example, for DeepSeek models, `reasoning_tag` should be set to `think`.

This may be useful to handle responses such as 

```
<think>

</think>

Hello! How can I assist you today? 😊
```


wronkiew:
Why was this removed with https://github.com/HKUDS/LightRAG/commit/24ae083284d5979f71942a7c363555e5a18abb97?

PR #734, opened 2025-02-08
Fixed Pipe Insert & Code Cleanup

YanSte:
This pull request addresses the issue with the Pipe insert and ensures proper functioning of the following pipeline processes:

apipeline_process_documents

apipeline_process_chunks

apipeline_process_extract_graph

Additionally, the code has been cleaned up to improve clarity and maintainability, with type annotations added for better readability and robustness.

And process more document in Async way.
----

**Remarks**

@LarFii 

I have a concern regarding the project's focus.  

Initially, the primary goal was RAG Graph, but I notice an increasing number of unnecessary tool integrations, such as openwebui_tool and others etc. These additions may lead to a loss of focus and introduce extra maintenance overhead.

Would it be beneficial to reassess and streamline the scope to ensure the project remains aligned with its core objective?

And also if writing test is in the scope to not have regression.

PR #735, opened 2025-02-08
Updated documentation

YanSte:
Some documentations update to make more clear some point.


LarFii:
After merging [#734](https://github.com/HKUDS/LightRAG/pull/734), there may be some conflicts that will need to be resolved.

PR #741, opened 2025-02-10
fix(postgres): update document status with partial update instead of …

Brenon28:
### Description
This PR fixes two key issues in PostgreSQL storage:

- Document content was not being properly stored due to incorrect data handling.
- Status updates were overwriting all fields instead of modifying only status, chunks_count, and updated_at.

### Changes

- Fixed content insertion issue in PGDocStatusStorage.
- Updated status handling to prevent unnecessary field overwrites.
- Ensured proper datetime handling in PostgreSQL queries.

### Testing

- Verified that document content is now correctly stored.
- Checked that status updates only modify relevant fields.


LarFii:
Thanks for your contribution!

PR #750, opened 2025-02-11
Integrate gemini client into Lightrag

gurjot-05:
I integrated Gemini client into LightRAG. So users can now use Gemini too to query using LightRAG.


LarFii:
Thanks!

PR #802, opened 2025-02-17
Improved variable assignment to use your own azure open ai embedding …

MdNazishArmanShorthillsAI:
…model

- Improved the variable assignment for the embedding model
- By default would be "text-embedding-3-small"
- If you want to use your own embedding model define a varibale EMBEDDING_MODEL and use the model you want to use


MdNazishArmanShorthillsAI:
@LarFii  Please check and approve

PR #806, opened 2025-02-17
Added system prompt support in all modes

MdNazishArmanShorthillsAI:
- Bring your own custom system prompt and do query
- Supported in all forms of search type
- If you don't pass your own system prompt, answer will be generated using lightRAG by default system prompt defined in each mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ rag = LightRAG(working_dir=WORKING_DIR)
 
 # Create query parameters
 query_param = QueryParam(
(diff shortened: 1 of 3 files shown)


MdNazishArmanShorthillsAI:
@LarFii Please review let me know if there is any changes

YanSte:
Looks good to me. Thanks!

MdNazishArmanShorthillsAI:
> Looks good to me. Thanks for sharing.

Welcome

PR #824, opened 2025-02-17
Update README.md - Imports

VeiledTee:
Edited examples to use ```from lightrag.lightrag import LightRAG, QueryParam``` instead of ```from lightrag import LightRAG, QueryParam``` (latter causes Unresolved reference error upon pip install of LightRAG v1.1.6)


YanSte:
Thanks !

PR #842, opened 2025-02-18
Lightrag Server Launch fix.

YanSte:
This pull request addresses and hide unnecessary temporary lib using PM usage.

It also includes testing upon launching lightrag-server to verify stability.”

- Removing tqdm library
    - Why is mostly using in some Notebook, in production code we don't need.


lightrag/llm/ollama.py
```diff
@@ -79,15 +40,15 @@
...
-async def ollama_model_if_cache(
+async def _ollama_model_if_cache(
     model,
     prompt,
     system_prompt=None,
     history_messages=[],
     **kwargs,
 ) -> Union[str, AsyncIterator[str]]:
     stream = True if kwargs.get("stream") else False
-    reasoning_tag = kwargs.pop("reasoning_tag", None)
```
YanSte, line 90: never use, in all project.

PR #844, opened 2025-02-18
Fix office file indexing problem

danielaskdd:
- Fix docx pptx indexing error
- Add xlsx support

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -653,7 +653,7 @@ class LightRAG:
             if update_storage:
                 await self._insert_done()
 
-    async def apipeline_enqueue_documents(self, input: str | list[str]):
+    async def apipeline_enqueue_documents(self, input: str | list[str]) -> bool:
         """
         Pipeline for Processing Documents
 
@@ -691,11 +691,12 @@ class LightRAG:
 
         if not new_docs:
             logger.info("No new unique documents were found.")
-            return
+            return False
 
         # 4. Store status document
         await self.doc_status.upsert(new_docs)
         logger.info(f"Stored {len(new_docs)} new unique documents")
+        return True
 
     async def apipeline_process_enqueue_documents(
         self,
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -653,7 +653,7 @@ class LightRAG:
             if update_storage:
                 await self._insert_done()
 
-    async def apipeline_enqueue_documents(self, input: str | list[str]) -> bool:
+    async def apipeline_enqueue_documents(self, input: str | list[str]):
         """
         Pipeline for Processing Documents
 
@@ -691,12 +691,11 @@ class LightRAG:
 
         if not new_docs:
             logger.info("No new unique documents were found.")
-            return False
+            return
 
         # 4. Store status document
         await self.doc_status.upsert(new_docs)
         logger.info(f"Stored {len(new_docs)} new unique documents")
-        return True
 
     async def apipeline_process_enqueue_documents(
         self,
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -1343,8 +1343,7 @@ def create_app(args):
                     from docx import Document
                     from io import BytesIO
 
-                    docx_content = await file.read()
-                    docx_file = BytesIO(docx_content)
+                    docx_file = BytesIO(file)
                     doc = Document(docx_file)
                     content = "\n".join(
                         [paragraph.text for paragraph in doc.paragraphs]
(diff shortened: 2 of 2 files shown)


lightrag/lightrag.py
```diff
@@ -653,7 +653,7 @@ async def ainsert_custom_chunks(self, full_text: str, text_chunks: list[str]):
             if update_storage:
                 await self._insert_done()
 
-    async def apipeline_enqueue_documents(self, input: str | list[str]):
+    async def apipeline_enqueue_documents(self, input: str | list[str]) -> bool:
```
YanSte: Hello,

Thanks for sharing.

We can't modify apipeline_enqueue_documents to returns a boolean.
Lightserver should not modify the library code.

Could you remove ?

danielaskdd: If Lightserver could recognize the status of file additions, it would be able to provide more user-friendly notifications. Currently, when a user uploads a duplicate file, the logs shows "Successfully processed and enqueued file", which may cause confusion for user.

I have reviewed all instances where apipeline_enqueue_documents is used and found that the function returning a Boolean value does not cause any compatibility issues. Therefore, I believe it would be a better solution to have apipeline_enqueue_documents explicitly return the status of file additions.

YanSte: Thanks for your feedback.

However, this change is not useful in the current context, as the value is only used for logging. Adding a boolean for logging purposes doesn’t add much value.

Additionally, pipe needs to return None, as that aligns with its intended behavior.

danielaskdd: iShot_2025-02-18_20 02 54

danielaskdd: WebUI also need to know this status to make a sensible respond .

YanSte: Thanks for your input and the feature suggestion.

However, as I mentioned earlier, this adds unnecessary complexity for developers to understand. 

Since you only used for logging in the Lightrag server, it doesn’t seem to provide enough value and changing lightrag for that.

If you really need this functionality, a separate method to check whether it’s a new document could be a better approach, but it shouldn’t be added to lightrag.

danielaskdd: How can we use a separate method to check if a previously added document is a duplicate? Is this approach significantly more complicated? When adding hundreds of files to LightRag, there must be a way to inform the user which files have failed or are duplicates.

YanSte: As you can see, you using only to show a logging print.

```
                has_new_docs = await rag.apipeline_enqueue_documents(content)
                if has_new_docs:
                    logging.info(
                        f"Successfully processed and enqueued file: {file_path.name}"
                    )
                else:
                    logging.info(
                        f"File content already exists, skipping: {file_path.name}"
                    )
```     

I understand the need to track duplicates when processing large volumes of files.

However, using apipeline_enqueue_documents solely to determine whether a document is new—just for logging—does not align with its intended functionality. (ainsert doent give this functionality for exemple)

Would it make sense to introduce a dedicated method for checking duplicates before insertion? That way, we keep the pipeline clean while still providing the logging functionality you need.

Let me know your thoughts.

danielaskdd: Separate checking and enqueuing into two step is not a good idea. Think about multiple users adding same file to LightServer at the same time, how to make checking reliable in this situation.

danielaskdd: What your concern about return a status for enqueuing a file?

YanSte: I see your concern about handling concurrent file additions, but separating checking from enqueuing is actually a cleaner and more scalable approach.

Returning a status from apipeline_enqueue_documents blurs its responsibility. Its job is to enqueue, not to validate whether a document is new. If we mix concerns here, we introduce unnecessary complexity and make the function behave inconsistently across different projects.

If the goal is to reliably check for duplicates in a multi-user scenario, the right place to handle this logic in a new method like.

```
def check_new_file(..)
new_doc = compute_mdhash_id(content, prefix="doc-")
all_new_doc_ids = set(new_doc)
unique_new_doc_ids = await self.doc_status.filter_keys(all_new_doc_ids)
new_docs = {doc_id: new_docs[doc_id] for doc_id in unique_new_doc_ids}
return new_docs
```

Would a dedicated is_new_document method in LightRag serve your needs better? (here it's add value)
This would provide the validation you need without overloading apipeline_enqueue_documents with extra responsibility.

YanSte: @LarFii What do you think ?

danielaskdd: I get  your point, you are right.

danielaskdd: I am currently contemplating how to program LightRAG into a universal knowledge base backend. Therefore, I am considering many aspects. I hope that LightRAG can conveniently create and delete Namespaces, easily add files in batches to a Namespace, and have each Namespace simulate an Ollama model to provide services externally. I wonder if everyone supports my idea, or have any suggestion? @ParisNeo

danielaskdd: > Hello,
> 
> Thanks for sharing.
> 
> We can't modify apipeline_enqueue_documents to returns a boolean. Lightserver should not modify the library code.
> 
> Could you remove ?

The codes is removed.

YanSte:
Thanks for sharing 🙏🏻

PR #846, opened 2025-02-18
Refactor Database Connection Management and Improve Storage Lifecycle Handling

ArnoChenFx:
This PR refactors the database connection management for PostgreSQL, Oracle, and TiDB by introducing a `ClientManager` class. It also enhances the lifecycle management of storage components in `LightRAG`, ensuring consistency and better maintainability.

#### Key Changes:
1. **Unified Database Initialization for All Backends:**
   - Previously, PostgreSQL, Oracle, and TiDB could not be automatically created and configured like other backends when used outside the `lightrag_server` environment. This update ensures that all backends now exhibit consistent behavior.

2. **Decoupled Backend Implementations from `lightrag_server`:**
   - Previously, `lightrag_server` explicitly imported and initialized PostgreSQL, Oracle, and TiDB, which increased backend maintenance complexity and violated the principle of backend isolation.
   - With this refactor, these backends are now fully self-contained within their respective implementation files, restoring modularity and reducing unnecessary dependencies in the server code.

3. **Improved Storage Lifecycle Management:**
   - Added `initialize()` and `finalize()` methods to `StorageNameSpace` to handle storage setup and cleanup.
   - `LightRAG` now explicitly manages storage lifecycle with `initialize_storages()` and `finalize_storages()`, ensuring proper resource management.

#### Benefits:
- **Consistent Behavior Across Backends:** PostgreSQL, Oracle, and TiDB now behave like other backends, automatically initializing when needed.
- **Reduced Maintenance Complexity:** Backend implementations are now self-contained, making them easier to maintain and extend.
- **Improved Modularity:** Decouples backend logic from the server, making the system more flexible and scalable.

#### Testing:
- Verified that PostgreSQL, Oracle, and TiDB are now automatically initialized and configured correctly.
- Confirmed that storage lifecycle methods are called appropriately during server startup and shutdown.

### DIFF
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -31,17 +31,13 @@ import configparser
 import traceback
 from datetime import datetime
 
-root_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "../.."))
-print(root_path)
-sys.path.append(root_path)
-
 from lightrag import LightRAG, QueryParam
 from lightrag.base import DocProcessingStatus, DocStatus
 from lightrag.types import GPTKeywordExtractionFormat
 from lightrag.api import __api_version__
 from lightrag.utils import EmbeddingFunc
 from lightrag.utils import logger
-from ollama_api import OllamaAPI, ollama_server_infos
+from .ollama_api import OllamaAPI, ollama_server_infos
 
 
 # Load environment variables
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -15,11 +15,6 @@ import logging
 import argparse
 from typing import List, Any, Literal, Optional, Dict
 from pydantic import BaseModel, Field, field_validator
-from lightrag import LightRAG, QueryParam
-from lightrag.base import DocProcessingStatus, DocStatus
-from lightrag.types import GPTKeywordExtractionFormat
-from lightrag.api import __api_version__
-from lightrag.utils import EmbeddingFunc
 from pathlib import Path
 import shutil
 import aiofiles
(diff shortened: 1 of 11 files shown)


lightrag/kg/mongo_impl.py
```diff
@@ -40,31 +39,59 @@
...
-                config.get("mongodb", "database", fallback="LightRAG"),
-            )
-        )
-
         self._collection_name = self.namespace
 
-        self._data = database.get_collection(self._collection_name)
-        logger.debug(f"Use MongoDB as KV {self._collection_name}")
+    async def initialize(self):
+        if not hasattr(self, "db") or self.db is None:
```
YanSte: To solve the issue of `self hasattr(self, "db")`

You can add:

`db: TypeOfTheClient = field(init=false)`

This solve your issue, and leave the type.

YanSte:
Thanks a lot, I will have a look. 🙏🏻

lightrag/kg/mongo_impl.py
ArnoChenFx: Indeed

lightrag/api/lightrag_server.py
```diff
@@ -36,39 +31,17 @@
...
 from datetime import datetime
 
-from lightrag.utils import logger
-from .ollama_api import (
-    OllamaAPI,
-)
-from .ollama_api import ollama_server_infos
-
+root_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "../.."))
+print(root_path)
```
YanSte: does the print is need ?

lightrag/lightrag.py
```diff
@@ -348,6 +349,9 @@ class LightRAG:
     # Extensions
     addon_params: dict[str, Any] = field(default_factory=dict)
 
+    # Ownership
+    is_managed_by_server: bool = False
```
YanSte: I think what you did is generally good!

However, I believe is_managed_by_server should be removed. Instead, we should modify the LightRaG setup to follow a more structured approach with docs, like this:

```
rag = LightRaG()
await rag.initialize()
```

And later:

```
await rag.finalize()
```

Does this align with what you’re thinking? Let me know if you need clarification.

YanSte:
Really nice feature 🙏🏻

lightrag/api/lightrag_server.py
ArnoChenFx: reverted

lightrag/lightrag.py
ArnoChenFx: By default, the LightRAG class automatically calls `initialize_storages` and `finalize_storages` at the appropriate times. However, in lightrag_server, these two functions are managed by FastAPI's lifespan. Maybe we can try using automatic management entirely to see if it still works properly.

lightrag/kg/mongo_impl.py
```diff
@@ -83,8 +83,8 @@ async def release_client(cls, db: AsyncIOMotorDatabase):
 @final
 @dataclass
 class MongoKVStorage(BaseKVStorage):
-    db: AsyncIOMotorDatabase = field(init=False)
```
YanSte: It was great before; the standard approach is to set init=False and initialize in __post_init__. Why did you change it?

lightrag/lightrag.py
YanSte: Ok got it. after review

lightrag/kg/mongo_impl.py
ArnoChenFx: init=False raised no attribute error

YanSte:
Let me know when is good I will test it directly.

ArnoChenFx:
> Let me know when is good I will test it directly.

It‘s ready for test!

YanSte:
Let's go

YanSte:
Got some error see my issue: https://github.com/HKUDS/LightRAG/issues/849

PR #858, opened 2025-02-19
Introduced sorting of nodes

MdNazishArmanShorthillsAI:
- We are sorting edges/relationship on the basis of rank and weight
- It will be usefull when we can sort the nodes on the basis of rank/degree as we know highest degree node will be more connected and possibly containing more informations

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -1095,6 +1095,9 @@ async def _get_node_data(
         if n is not None
     ]  # what is this text_chunks_db doing.  dont remember it in airvx.  check the diagram.
     # get entitytext chunk
+
+    node_datas = sorted(node_datas, key=lambda x: x["rank"], reverse=True)
+
     use_text_units, use_relations = await asyncio.gather(
         _find_most_related_text_unit_from_entities(
             node_datas, query_param, text_chunks_db, knowledge_graph_inst
@@ -1452,6 +1455,8 @@ async def _find_most_related_entities_from_relationships(
         {**n, "entity_name": k, "rank": d}
         for k, n, d in zip(entity_names, node_datas, node_degrees)
     ]
+
+    node_datas = sorted(node_datas, key=lambda x: x["rank"], reverse=True)
 
     len_node_datas = len(node_datas)
     node_datas = truncate_list_by_token_size(


MdNazishArmanShorthillsAI:
@LarFii @YanSte  Can you guys have a look into this please?

MdNazishArmanShorthillsAI:
> Hi, Thanks for sharing.
> 
> Could please have more context on this feature? What that provides.
> 
> Thanks.

The purpose of re-ranking nodes is to prioritize more relevant and important entities in the graph. This ensures that the most significant nodes (i.e., entities) appear first in any downstream processing, improving the accuracy and efficiency of queries.

Initial Ranking (Node Degree Assignment) which is already currently implemeneted

Each node is initially assigned a rank based on its degree (i.e., the number of connections it has).
A higher degree suggests that the node is more central and important in the graph.


Re-ranking After Retrieval -> Now raised

Once all relevant nodes are retrieved, they are sorted based on rank to ensure the most connected and significant entities appear first.
This prevents less relevant nodes from overshadowing highly relevant ones.

This is already implemented for _get_edges function for your refernce

MdNazishArmanShorthillsAI:
In the future, to handle larger datasets efficiently while maintaining accuracy, we can enhance the node ranking process by first retrieving a broader set, such as the top 100 nodes instead of 60, ensuring wider coverage of relevant entities. Then, an additional filtering step can be introduced to refine this selection to only the top 40 most relevant nodes based on a secondary ranking mechanism. This extra layer of filtering significantly improves performance by reducing unnecessary computations while maintaining high precision. Once the top 40 nodes are identified, we fetch only their related edges and text units, avoiding the overhead of processing all 100 nodes. This approach enhances scalability, speeds up queries, and ensures that the most significant entities are prioritized, making the system more adaptable to large-scale datasets while preserving retrieval quality.

YanSte:
@LarFii Could you have a look. Thanks

MdNazishArmanShorthillsAI:
> I personally feel this might not be entirely reasonable, because for the retrieved nodes, the matching degree with the query is more important than the degree itself. Similarly, the retrieved edges are also sorted based on similarity, so what’s really needed are the nodes corresponding to these edges, rather than reordering them based on degree.

Thanks I am closing it

Issue #871, opened 2025-02-19
After updating the branch, I have networkx error


ParisNeo:
Hi, I just did a git pull and now I can't run the application with the default confioguration:
I get this
```
  File "/opt/lollms/lightrag/LightRAG/lightrag/kg/networkx_impl.py", line 27, in <module>
    raise ImportError(
ImportError: `networkx` library is not installed. Please install it via pip: `pip install networkx`.
```

This is not correct. We have a pipmaster check before networkx import.

So it is actually installed. When I do pip install networkx it says that it is already installed. So I removed the try except to actually see what is the real error here and it looks like it comes from this :

hyppo/kgof/fssd.py", line 4, in <module>
    from past.utils import old_div
ModuleNotFoundError: No module named 'past'

After some investigation it seems that I have to install future to get past working. LOOOOOL

```bash
pip install future
```

So I have removed the try except thing, it was hiding real errors. If it fails let it fail and we can see what's happening. That's better than hiding the errror begding another except.

In all case, if it wasn't installed pipmaster would have detected that.

I'll do a PR

PR #872, opened 2025-02-19
Fix ImportError by Removing Redundant Try-Except Block

ParisNeo:
This PR addresses an issue where an ImportError for the networkx library was incorrectly raised due to a try-except block that masked the real problem. The actual issue was a ModuleNotFoundError for the past module, which is part of the future package.

Since pipmaster is used to verify that the module is already installed, no need to use that extra try except especially that it hides the real error.

This fixes issue: #871


YanSte:
Thanks for sharing could please remove all try catch ?

ParisNeo:
Ok, I'll do that

YanSte:
Thanks a lot

PR #873, opened 2025-02-19
Removed all useless try except around imports as pipmaster already handles that

ParisNeo:
The try catch hides the real errors when importing the libraries. pipmaster ensures that the library is installed so if the import fails, it is a problem related to something else and we need to make it visible to allow detection of potential problems.


YanSte:
Merci 🙏🏻

ParisNeo:
I just linted everything correctly

PR #885, opened 2025-02-19
Cleanup of code

YanSte:
## Description

- Remove unsused method
- Cleaned get_all_labels never implemented everywhere
- Cleaned import

## Changes Made

- Info print
- Remove get_all_labels

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

### DIFF
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -1682,11 +1682,6 @@ def create_app(args):
             trace_exception(e)
             raise HTTPException(status_code=500, detail=str(e))
 
-    # query all graph labels
-    @app.get("/graph/label/list")
-    async def get_graph_labels():
-        return await rag.get_graph_labels()
-
     # query all graph
     @app.get("/graphs")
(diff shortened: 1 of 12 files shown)


lightrag/api/lightrag_server.py
```diff
@@ -1682,11 +1682,6 @@ async def stream_generator():
             trace_exception(e)
             raise HTTPException(status_code=500, detail=str(e))
 
-    # query all graph labels
-    @app.get("/graph/label/list")
```
ArnoChenFx, line 1686: This API is also used by the WebUI. We can't remove it.
@YanSte

ArnoChenFx:
All the APIs used by the WebUI are here: https://github.com/HKUDS/LightRAG/blob/main/lightrag_webui/src/api/lightrag.ts

lightrag/api/lightrag_server.py
YanSte: Okay.

PR #897, opened 2025-02-20
Added more file types support to the webui

ParisNeo:
## Description

There are many other filetypes that can be imported to the webui (like json files from databases, csv files etc ..)
I added the support for all these files in both backend and frontend :
      '.txt', '.md', '.html', '.htm', '.tex', '.json', '.xml', '.yaml', '.yml',
      '.rtf', '.odt', '.epub', '.csv', '.log', '.conf', '.ini', '.properties',
      '.sql', '.bat', '.sh', '.c', '.cpp', '.py', '.java', '.js', '.ts',
      '.swift', '.go', '.rb', '.php', '.css', '.scss', '.less'

## Related Issues

No issue is related to this
## Changes Made

Changed the server to support these files
Changed the components of the front end as well as the different ui elements to support this

Added  an extra configuration to allow compiling the front end on windows as I am using windows. This helps ensuring 

On windows there is no Bun support so I created a custom command build-no-bun that can be called by windows users to compile the frontend easily.

## Checklist

- [X] Changes tested locally
- [X] Code reviewed


ParisNeo:
@ArnoChenFx FYI I added some options to package.json to allow me to build the webui on windows. In fact Bun doesn't run on windows so I just used vite. This changes nothing for you, you still have your classic build, dev etc methods. I just added no-bun versions of them.

I hope this is ok for you.

With this configuration we both can collaborate on the webui

ArnoChenFx:
> @ArnoChenFx FYI I added some options to package.json to allow me to build the webui on windows. In fact Bun doesn't run on windows so I just used vite. This changes nothing for you, you still have your classic build, dev etc methods. I just added no-bun versions of them.
> 
> I hope this is ok for you.
> 
> With this configuration we both can collaborate on the webui

But i am developing using Bun on Windows.

ParisNeo:
> > @ArnoChenFx FYI I added some options to package.json to allow me to build the webui on windows. In fact Bun doesn't run on windows so I just used vite. This changes nothing for you, you still have your classic build, dev etc methods. I just added no-bun versions of them.
> > 
> > I hope this is ok for you.
> > 
> > With this configuration we both can collaborate on the webui
> 
> But I am indeed developing with Bun on Windows, so it should support Windows. If remove the `bunx --bun`, vite will use the nodejs to build the project.

Hi. We have both build options. I think there is no harm having both. No?

ArnoChenFx:
> > > @ArnoChenFx FYI I added some options to package.json to allow me to build the webui on windows. In fact Bun doesn't run on windows so I just used vite. This changes nothing for you, you still have your classic build, dev etc methods. I just added no-bun versions of them.
> > > I hope this is ok for you.
> > > With this configuration we both can collaborate on the webui
> > 
> > 
> > But I am indeed developing with Bun on Windows, so it should support Windows. If remove the `bunx --bun`, vite will use the nodejs to build the project.
> 
> I already have node.js working. The new setup allows both to be used. I did not remove the option to use bun to build (it is still the default build way). But if someone has no Bun and can't install it beceause he is not admin or something, node.js can be an options.
> 
> You can rebuild it with bun when you need. But this allows me to easily fix and update things without waiting to be on another PC.
> 
> We have both build options. I think there is no harm having both. No?

Okay, you've convinced me.
I just think different builders would produce different results. If something goes wrong, it could lead to additional debugging work.

YanSte:
@ArnoChenFx Again sorry for my mistake.

PR #898, opened 2025-02-20
Database Cleanup

YanSte:
## Description

Database Cleanup

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -82,8 +82,8 @@ class LightRAG:
     log_level: int = field(default=logger.level)
     """Logging level for the system (e.g., 'DEBUG', 'INFO', 'WARNING')."""
 
-    log_dir: str = field(default=os.getcwd())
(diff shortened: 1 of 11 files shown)


lightrag/lightrag.py
```diff
@@ -43,210 +47,22 @@
...
-
-    except RuntimeError:
-        # If no event loop exists or it is closed, create a new one
-        logger.info("Creating a new event loop in main thread.")
-        new_loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(new_loop)
-        return new_loop
-
 
+@final
```
yash98, line 51: @YanSte I am trying to extend LightRAG, and want to keep my implementation very loosely coupled with LightRAG class (not changing the LightRAG class itself). 

I wanted to understand whether `@final` here is primarily to ensure implementation stability (preventing subclasses from breaking some internal assumptions) or is it more of a design pattern preference encouraging composition over inheritance?

PR #925, opened 2025-02-23
Simplify cli arguments

danielaskdd:
## Description

Optimize command-line interface (CLI) by streamlining argument complexity. Eliminate parameters that are infrequently modified by users, as these typically remain consistent across executions. This simplification enhances usability and reduces cognitive load for end-users.

## Changes Made

- Remove storage config from CLI
- Remove  LLM and embedding settings from CLI
- Remove chunk config from CLI
- Change `--verbose` to store_true action
- Change default timeout of LLM to 150s
- Update README.md for API Server
- Improve entity extraction logging and metrics

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed


YanSte:
Thanks a lot 🙏🏻

danielaskdd:
> Thanks a lot 🙏🏻

You are welcome :)

ParisNeo:
Hi there. This did break my setup :(

Well, I see the purpose of simplifying the arguments, but then, I think we need to update the ini file to support the rest of configurations.

I fixed it by using environment variables, but It would be a good thing if we have a full config.ini file centralizing all the configurations.

And let's try to mak the workflow simple and stable. I have some experience in this. People don't like it when we modify the code to the point where we break the compatibility with previous versions. We need to keep retrocompatibility because some people may have created a setup for themselves and they do a pull and everything breaks. They get pissed off and may leave the project.

I am OK with this update, just let's try not to have these kinds of updates break things in the future.

Thanks for the great work.

danielaskdd:
@ParisNeo Completely agree with your perspective! While this change was necessary given the confusing and potentially misleading nature of the previous configuration approach, for any API or configuration system changes, we need to plan carefully. Sorry for any inconvenience cause by this PR.

ParisNeo:
Would you like to make the config.ini capable of handling the other parameters? It would be nice if we can add an argument for example --config and in that one put the link to the file to use and get all the parameters from the config.ini file then we can give an example so people can just change what they want and use that? Also we can have a default one in /etc for example for linux users and if not, just find a config.ini in the current folder.

This can be a good way to put all these. Ini files have sections so we can add sections for llm configuration, section for embedding configuration, for storages etc.

PR #994, opened 2025-03-04
[EVO] Add language configuration to environment and argument parsing

Brocowlee:
The language was not considered when using lightRAG_server. Even though the variable SUMMARY_LANGUAGE was defined in the env.example, it was not used anywhere.
I fixed that.


YanSte:
Thanks a lot !

danielaskdd:
This pull request conflicts with #989, which focuses on refining the summary language for entities and relations. @YanSte @LarFii

PR #995, opened 2025-03-04
Made .env file loaded from the current folder instead of the script folder

ParisNeo:
This pull request does the following:
1 - removed relative imports to simplify debugging on vscode
2 - .env file is now loaded from the current folder instead of the script folder allowing for launching multiple instances of lightrag in different folders and with different configurations.


YanSte:
Thanks a lot !

PR #1011, opened 2025-03-06
Added docling option to load files

ParisNeo:
Docling is a superior tool to load data from files and format them.

I added a new option to select document loading engine.

I did put docling as the default one as my tests confirml its superiority compared to PyPDF.

I updated the documentation too.

I made the arguments visible to any class that needs them so that we can use them when needed.


danielaskdd:
That's great. Do you check if Docling is  blocking frontend any more?

ParisNeo:
Docling seems to work fine forme.
But since you don't like that I made it optional and the default value does not use docling anymore.
Now you have a new variable named: DOCUMENT_LOADING_ENGINE that can be either DEFAULT or DOCLING.

With this, I think everyone is happy.
To me, docling has a far superior extraction capabilities than the classic pypdf2.

danielaskdd (approved):

PR #1022, opened 2025-03-08
feat: adding optional id into text insert

KennyFukuya:
## Description

Added param for id into `/documents/text` and `/documents/texts`

## Checklist

- [ X ] Changes tested locally
- [ X ] Code reviewed

## Additional Notes


danielaskdd:
Thank you for your contribution. Could you please clarify the purpose of requiring an ID for document insertion? Additionally, it would be helpful if you could provide several use cases for us to evaluate the implementation.

KennyFukuya:
> Thank you for your contribution. Could you please clarify the purpose of requiring an ID for document insertion? Additionally, it would be helpful if you could provide several use cases for us to evaluate the implementation.

The goal is to correctly map a document from another database by its doc uid. Currently the lightrag-server wasn't allowing that, but it is implemented into the package. My modification is just to allow the same method into the server.

Use case can be the insertion of any new text with its given id, if correctly its id on lightrag_doc_full must be the same as the one inserted.

PR #1026, opened 2025-03-08
infer_new_relationships

choizhang:
## Description

Using a large model to discover hidden relationships based on explicit relationships between characters in text

## Changes Made

1. lightrag\lightrag\lightrag.py
Added infer_new_relationships function

2. lightrag\lightrag\prompt.py
Added PROMPTS["relation_inference"]

[List the specific changes made in this pull request.]

## Checklist

- [ ] Changes tested locally
- [ ] Code reviewed


LarFii:
I appreciate the effort in exploring LLMs for inferring hidden relationships based on explicit ones in a knowledge graph, but I have a few concerns about the current approach:

1. *Scalability of Node Processing*: The algorithm doesn't seem to consider the number of retrieved nodes. Sending all nodes at once to an LLM is unrealistic, and the allocation strategy needs to be carefully designed. Even if we batch the nodes, if the majority of them are retrieved, it essentially turns into a reindexing operation rather than an efficient relationship inference method. I'm not convinced this will lead to better quality results.

2. *LLM Hallucination Risk*: The prompt doesn’t strictly enforce the LLM to infer relationships solely based on the provided context. This raises the concern that the model might generate hallucinated relationships that don’t actually exist. I believe this needs proper validation before it can be considered reliable.

Overall, I think extracting relationships through a knowledge graph is a promising direction, but the current method feels quite rudimentary. Instead of relying solely on LLMs, I believe we should leverage KG-based reasoning, considering different types of relationships. If that's the case, incorporating graph reasoning at the retrieval stage might be a more reasonable and effective solution.

PR #1070, opened 2025-03-12
fix stream

lybtt:
```
async def initialize_rag():
    rag = LightRAG(
        working_dir="yanbao",
        embedding_func=EmbeddingFunc(
            embedding_dim=1536,
            max_token_size=8192,
            func=embedding_func
        ),
        llm_model_kwargs={"stream": True}
    )

    return rag
```

If I add the configuration llm_model_kwargs={"stream": True}, the following error occurs.

```
ERROR: Error in get_kg_context: expected string or bytes-like object
```

Perhaps I should not use the stream mode?

my code might not be correct, and I would greatly appreciate any guidance you can provide."

### DIFF
--- a/lightrag/llm/openai.py
+++ b/lightrag/llm/openai.py
@@ -123,18 +123,21 @@ async def openai_complete_if_cache(
 
         async def inner():
             try:
+                _content = ''
                 async for chunk in response:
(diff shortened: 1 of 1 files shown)


LarFii:
Thanks for sharing! This modification seems to be correct.

PR #1085, opened 2025-03-13
Postgresql - refactor the store entity/relation data

OxidBurn:
## Description

This PR refactors the PostgreSQL implementation to improve how chunk IDs are stored for entities and relationships in LightRAG. It addresses issues related to chunk ID handling by:

  1. Modifying the storage of chunk IDs from VARCHAR(255) to arrays of VARCHAR(255) in the database schema
  2. Adding logic to properly parse source_id strings containing multiple chunk IDs (separated by "< SEP >")

##  Related Issues

  Resolves issues with chunk ID storage in the PostgreSQL database, improving entity extraction and relationship storage.

##  Changes Made

  - Added logic to parse source_ids containing multiple chunks
  - Modified code to store chunk IDs as arrays in PostgreSQL rather than a string field
  - Changed SQL templates to handle arrays properly in the upsert operations for entities and relationships
  - Improved the entity and relationship storage process by properly handling multiple chunk references

 ## Checklist

  - [x] Changes tested locally
  - [ ] Code reviewed

##  Additional Notes

  This change enhances data integrity by properly storing all chunk references without truncation, which is essential for maintaining entity-document relationships in the knowledge graph.


lightrag/kg/postgres_impl.py
```diff
@@ -950,10 +962,14 @@ def _record_to_dict(record: asyncpg.Record) -> dict[str, Any]:
                             vertices.get(edge["end_id"], {}),
                         )
             else:
-                if v is None or (v.count("{") < 1 and v.count("[") < 1):
+                if v is None:
```
JoramMillenaar, line 965: I believe we can simplify this whole clause to a single line:
```py
d[k] = json.loads(v) if isinstance(v, str) and ("{" in v or "[" in v) else v
```

JoramMillenaar:
Quick related question; is the webui's graph working for you when using postgres? I seem to be getting parse errors inside `get_knowledge_graph`. Asking here because it might be a related bug/fix that can be included in this PR.

```python
if result["r"]:
    edge = result["r"]
    src_id = self._decode_graph_label(edge["start_id"])  # <---- start_id does not exist
    tgt_id = self._decode_graph_label(edge["end_id"])
    edges.append((src_id, tgt_id))
```

lightrag/kg/postgres_impl.py
OxidBurn: @JoramMillenaar, you're absolutely right! 
However making the solution like I did really helps with debugging and reading the code. I also find the ternary conditional operator a bit tricky to read in that particular case.

OxidBurn:
> Quick related question; is the webui's graph working for you when using postgres? I seem to be getting parse errors inside `get_knowledge_graph`. Asking here because it might be a related bug/fix that can be included in this PR.
> 
> ```python
> if result["r"]:
>     edge = result["r"]
>     src_id = self._decode_graph_label(edge["start_id"])  # <---- start_id does not exist
>     tgt_id = self._decode_graph_label(edge["end_id"])
>     edges.append((src_id, tgt_id))
> ```

Honestly, these days, I only use LightRAG as an RAG tool and have never tried the web UI project.

Issue #1099, opened 2025-03-17
[Bug]: When using the postgres@15 version, there will be an error


pengjunfeng11:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

Here is bug detail
```
PostgreSQL database, error:subquery in FROM must have an alias
HINT:  For example, FROM (SELECT ...) [AS] foo.
```
It's should be the SQL Template problem

### LightRAG Config Used

# Paste your config here

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

pengjunfeng11:
And i have commit pr in #1100

PR #1100, opened 2025-03-17
Update postgres_impl.py

pengjunfeng11:
There is aBUG. 

## Description

When lighttrag uses the postgres@15 version, there will be an error.The BUG detail as follow: 
```
PostgreSQL database, error:subquery in FROM must have an alias 
HINT:  For example, FROM (SELECT ...) [AS] foo.
```
It roughly means that there is an error in SQL template, and you need to use the alias alias to declare the content of the subquery.

## Related Issues

#1099 

## Changes Made

[List the specific changes made in this pull request.]

## Checklist

- [x] Changes tested locally
- [x] Code reviewed


LarFii:
Thanks for sharing!

PR #1105, opened 2025-03-17
Improve Logging Configuration

spo0nman:
Problem: The logging configuration currently lacks flexibility incase you want to run it in a container and want to only log to STDOUT.

The change adds an `enable_file_logging` parameter to `setup_logger` which:

- When `True` (default): Sets up both console and file logging
- When `False`: Sets up console logging only, skipping all file-related operations

The change also adds better error handling for file permission issues, allowing the logger to fall back to console-only logging if it can't create log files, rather than crashing.

This makes the logging system more flexible and resilient, especially in restricted environments where file writing might not be possible or desired.

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -794,10 +794,12 @@ class LightRAG:
                         "cur_batch": 0,
                         "request_pending": False,  # Clear any previous request
                         "latest_message": "",
+                        "history_messages": [],  # Initialize empty history_messages list
                     }
                 )
                 # Cleaning history_messages without breaking it as a shared list object
-                del pipeline_status["history_messages"][:]
+                if "history_messages" in pipeline_status:
+                    del pipeline_status["history_messages"][:]
             else:
                 # Another process is busy, just set request flag and return
                 pipeline_status["request_pending"] = True
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -794,12 +794,9 @@ class LightRAG:
                         "cur_batch": 0,
                         "request_pending": False,  # Clear any previous request
                         "latest_message": "",
-                        "history_messages": [],  # Initialize empty history_messages list
                     }
                 )
-                # Cleaning history_messages without breaking it as a shared list object
-                if "history_messages" in pipeline_status:
-                    del pipeline_status["history_messages"][:]
+                del pipeline_status["history_messages"][:]
             else:
                 # Another process is busy, just set request flag and return
                 pipeline_status["request_pending"] = True
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -109,33 +109,23 @@ def setup_logger(
     logger_name: str,
     level: str = "INFO",
     add_filter: bool = False,
-    log_file_path: str = None,
+    log_file_path: str | None = None,
+    enable_file_logging: bool = True,
 ):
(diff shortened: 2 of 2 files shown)


lightrag/lightrag.py
```diff
@@ -794,10 +794,12 @@ async def apipeline_process_enqueue_documents(
...
                         "request_pending": False,  # Clear any previous request
                         "latest_message": "",
+                        "history_messages": [],  # Initialize empty history_messages list
                     }
                 )
                 # Cleaning history_messages without breaking it as a shared list object
-                del pipeline_status["history_messages"][:]
+                if "history_messages" in pipeline_status:
+                    del pipeline_status["history_messages"][:]
             else:
```
danielaskdd, line 800: o ensure proper initialization, the initialize_pipeline_status() function must be invoked prior to the creation of any LightRAG instance. Consequently, the history_messages parameter should never be null under any circumstances. Could you please provide details regarding the specific scenario in which you encountered the reported KeyError?

lightrag/lightrag.py
```diff
@@ -794,10 +794,12 @@ async def apipeline_process_enqueue_documents(
                         "cur_batch": 0,
                         "request_pending": False,  # Clear any previous request
                         "latest_message": "",
+                        "history_messages": [],  # Initialize empty history_messages list
```
danielaskdd: It will break history_messages in multi worker environments of Gunicorn, cause history_messages is proxy of shared object.

danielaskdd (changes requested):

lightrag/lightrag.py
spo0nman:
```
        # Initialize LightRAG instance
        logger.info("Initializing LightRAG instance")
        rag = LightRAG(
            working_dir=settings.WORKING_DIR,
            embedding_func=embedding_func_instance,
            graph_storage="Neo4JStorage",
            vector_storage="ChromaVectorDBStorage",
            kv_storage="MongoKVStorage",
            embedding_batch_num=32,
            auto_manage_storages_states=False,  # Prevent automatic initialization
            **CHROMADB_CONFIG,
        )
       # And then  later on I do 
       
            if doc_id:
            # Store the full document text directly
            await rag.full_docs.upsert(
                {doc_id: {"content": document.text, "created_at": time.time()}}
            )

            # Use the two-stage pipeline for document processing
            await rag.apipeline_enqueue_documents([document.text])
            await rag.apipeline_process_enqueue_documents()
        else:
            # No custom ID - use standard insertion
            await rag.ainsert(document.text)

        
```
When in this process should I be doing a `initialize_pipeline_status()` ?

danielaskdd: Before you create LightRAG. Pls refer to lightrag_api_openai_compatible_demo.py :

```
```

lightrag/lightrag.py
spo0nman: Great! initialize_pipeline_status() resolves the issue with history_messages.

lightrag/lightrag.py
spo0nman: Removed the changes to lightrag.py to make this change only about the ability to have STDOUT only logs.

spo0nman:
I Closed this in favour of https://github.com/HKUDS/LightRAG/pull/1111

PR #1111, opened 2025-03-18
Add ability to conditionally disable file logging

spo0nman:
Problem: The logging configuration currently lacks flexibility incase you want to run it in a container and want to only log to STDOUT.

The change adds an `enable_file_logging` parameter to `setup_logger` which:

- When `True` (default): Sets up both console and file logging
- When `False`: Sets up console logging only, skipping all file-related operations

The change also adds better error handling for file permission issues, allowing the logger to fall back to console-only logging if it can't create log files, rather than crashing.

This makes the logging system more flexible and resilient, especially in restricted environments where file writing might not be possible or desired.

### DIFF
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -109,33 +109,23 @@ def setup_logger(
     logger_name: str,
     level: str = "INFO",
     add_filter: bool = False,
-    log_file_path: str = None,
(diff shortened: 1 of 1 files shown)


LarFii:
Thanks for sharing!

Issue #1088, opened 2025-03-14
[Bug]: <title>value too long for type character varying(255)


jasperchen01:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

I'm using the main branch, demo code as below:
```python
import asyncio
import logging
import os
import time
from dotenv import load_dotenv

from lightrag import LightRAG, QueryParam
from lightrag.llm.zhipu import zhipu_complete
from lightrag.llm.ollama import ollama_embedding
from lightrag.llm.openai import openai_embed,openai_complete_if_cache
from lightrag.utils import EmbeddingFunc
from lightrag.kg.shared_storage import initialize_pipeline_status
import numpy as np


CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
WORKING_DIR = f"{CURRENT_DIR}/lightrag_data"

FILE_PATH = f"{CURRENT_DIR}/../../data_dir/marker_output/2305_15323v1.md"
FILE_NAME = os.path.basename(FILE_PATH)

logging.basicConfig(format="%(levelname)s:%(message)s", level=logging.DEBUG)

if not os.path.exists(WORKING_DIR):
    os.mkdir(WORKING_DIR)

# PG
os.environ["AGE_GRAPH_NAME"] = "dickens"
os.environ["POSTGRES_HOST"] = "localhost"
os.environ["POSTGRES_PORT"] = "5432"
os.environ["POSTGRES_USER"] = "postgres"
os.environ["POSTGRES_PASSWORD"] = ""
os.environ["POSTGRES_DATABASE"] = "lightrag"

# neo4j
os.environ["NEO4J_URI"] = "neo4j://localhost:7687"
os.environ["NEO4J_USERNAME"] = "neo4j"
os.environ["NEO4J_PASSWORD"] = "admin123"

async def _llm_model_func(
    prompt, system_prompt=None, history_messages=[], keyword_extraction=False, **kwargs
) -> str:
    return await openai_complete_if_cache(
        model="qwen-plus-latest",
        prompt=prompt,
        system_prompt=system_prompt,
        history_messages=history_messages,
        api_key="***",
        base_url="https://dashscope.aliyuncs.com/compatible-mode/v1",
        **kwargs
    )

async def _embedding_func(texts: list[str]) -> np.ndarray:
    return await openai_embed(
        texts,
        model="text-embedding-v3",
[...]

PR #1127, opened 2025-03-19
Resolved issue with upsert of PG entity and relation

OxidBurn:
## Description
This PR complements [PR #1120 "fix: correct chunk_ids as array type and remove incorrect filepath type conversion"](https://github.com/HKUDS/LightRAG/pull/1120) with a few additional improvements

## Related Issues
Addresses the "entity and relation upsert error" as well as missing "metadata" key errors encountered during the insertion of entities and relationships into the PostgreSQL database.

## Changes Made
- Fixed SQL template for entity upsert by adding CURRENT_TIMESTAMP for update_time column
- Added conditional check for metadata and file_path keys in graph node extraction
- Fixed parameter type casting for content_vector in relationship insertion query
- Ensured proper alignment between SQL parameter placeholders and data dictionary values

## Checklist
- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes
The fix ensures data integrity during entity and relationship extraction operations by resolving SQL syntax errors that were preventing proper document processing.

### DIFF
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -8,6 +8,7 @@ import numpy as np
 import configparser
 
 from lightrag.types import KnowledgeGraph, KnowledgeGraphNode, KnowledgeGraphEdge
+from prompt import GRAPH_FIELD_SEP
 
 import sys
 from tenacity import (
@@ -434,8 +435,8 @@ class PGVectorStorage(BaseVectorStorage):
     def _upsert_entities(self, item: dict[str, Any]) -> tuple[str, dict[str, Any]]:
         upsert_sql = SQL_TEMPLATES["upsert_entity"]
         source_id = item["source_id"]
-        if isinstance(source_id, str) and "<SEP>" in source_id:
-            chunk_ids = source_id.split("<SEP>")
+        if isinstance(source_id, str) and GRAPH_FIELD_SEP in source_id:
+            chunk_ids = source_id.split(GRAPH_FIELD_SEP)
         else:
             chunk_ids = [source_id]
 
@@ -454,8 +455,8 @@ class PGVectorStorage(BaseVectorStorage):
     def _upsert_relationships(self, item: dict[str, Any]) -> tuple[str, dict[str, Any]]:
         upsert_sql = SQL_TEMPLATES["upsert_relationship"]
         source_id = item["source_id"]
-        if isinstance(source_id, str) and "<SEP>" in source_id:
-            chunk_ids = source_id.split("<SEP>")
+        if isinstance(source_id, str) and GRAPH_FIELD_SEP in source_id:
+            chunk_ids = source_id.split(GRAPH_FIELD_SEP)
         else:
             chunk_ids = [source_id]
 
@@ -1570,7 +1571,7 @@ TABLES = {
                     content_vector VECTOR,
                     create_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
                     update_time TIMESTAMP,
-                    chunk_id TEXT NULL,
+                    chunk_ids VARCHAR(255)[] NULL,
                     file_path TEXT NULL,
 	                CONSTRAINT LIGHTRAG_VDB_ENTITY_PK PRIMARY KEY (workspace, id)
                     )"""
@@ -1585,7 +1586,7 @@ TABLES = {
                     content_vector VECTOR,
                     create_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
                     update_time TIMESTAMP,
-                    chunk_id TEXT NULL,
+                    chunk_ids VARCHAR(255)[] NULL,
                     file_path TEXT NULL,
 	                CONSTRAINT LIGHTRAG_VDB_RELATION_PK PRIMARY KEY (workspace, id)
                     )"""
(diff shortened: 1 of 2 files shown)


lightrag/kg/postgres_impl.py
```diff
@@ -1570,7 +1571,7 @@ def namespace_to_table_name(namespace: str) -> str:
                     content_vector VECTOR,
                     create_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
                     update_time TIMESTAMP,
-                    chunk_id TEXT NULL,
+                    chunk_ids VARCHAR(255)[] NULL,
```
LarFii, line 1574: This fix introduces another issue: when there are too many associated chunks, the `chunk_id` exceeds the limit. Do you have any good solutions for this?

canesin: This is a VARCHAR so 255 unique characters. The code uses a md5 hash which is only 32 VARCHAR wide. It is unlikely that there will ever be such a large ID.

Howe829: In PostgreSQL, using a join table is an efficient way to store large amounts of related data. By creating an intermediary table with foreign keys, you can establish one-to-many or many-to-many relationships between the main table and associated data, overcoming the 1 GB array limit. Join tables provide better query performance and data management capabilities, with flexible indexing options that optimize search, insert, and update operations. This approach is particularly suitable for scenarios where you need to store large quantities of data like "chunk-ecc952703fbaabc97fee8f328ae8de56"

LarFii: > This is a VARCHAR so 255 unique characters. The code uses a md5 hash which is only 32 VARCHAR wide. It is unlikely that there will ever be such a large ID.

It was mentioned in the [issue 1088](https://github.com/HKUDS/LightRAG/issues/1088).

LarFii: > This is a VARCHAR so 255 unique characters. The code uses a md5 hash which is only 32 VARCHAR wide. It is unlikely that there will ever be such a large ID.

I understand the difference now. I'm not very familiar with PostgreSQL—sorry about that!

LarFii: Previously, the field in the PostgreSQL table was chunk_id. Modifying it this way will make the old data unavailable. Is there a way to add a data migration to transfer the old data to the new table?

lightrag/operate.py
```diff
@@ -223,11 +223,12 @@ async def _merge_nodes_then_upsert(
         already_source_ids.extend(
             split_string_by_multi_markers(already_node["source_id"], [GRAPH_FIELD_SEP])
         )
-        already_file_paths.extend(
-            split_string_by_multi_markers(
-                already_node["metadata"]["file_path"], [GRAPH_FIELD_SEP]
+        if "metadata" in already_node and "file_path" in already_node["metadata"]:
```
LarFii, line 226: In the [PR 1138](https://github.com/HKUDS/LightRAG/pull/1138), I have removed the metadata-related code (as it had no actual effect).

LarFii:
To avoid changing the table structure, could we consider modifying chunk_ids in the upsert to correspond to chunk_id? Since chunk_id is currently concatenated directly, the TEXT type should be sufficient.

lightrag/operate.py
OxidBurn: Perfect, thank you for that update.

CoachbotAI:
> > > To avoid changing the table structure, could we consider modifying chunk_ids in the upsert to correspond to chunk_id? Since chunk_id is currently concatenated directly, the TEXT type should be sufficient.
> > 
> > 
> > I reviewed the Postgres solution again and here are my thoughts:
> > 
> > * The current code (main branch) already uses chunk_ids as an array in both insertion logic and query logic
> > * The SQL templates specifically cast the values to varchar[] with $6::varchar[] syntax
> > * The JOIN conditions in queries use the PostgreSQL ANY operator: c.chunk_id = ANY(r.chunk_ids)
> > * The data processing in _upsert_entities and _upsert_relationships splits values into arrays
> > 
> > Schema vs. Implementation Mismatch: - In case the schema has chunk_id TEXT NULL, but the code is working with arrays, there's a fundamental mismatch
> > To use chunk_ids VARCHAR(255)[] NULL is correct and aligns the schema with the actual code usage - It makes the database reflect what the code is actually doing - There should be no migration issues as long as: 1. Existing data in the database is already being stored in array format 2. The name used in code consistently matches the name used in the schema
> > Conclusions: The change from chunk_id TEXT NULL to chunk_ids VARCHAR(255)[] NULL is actually fixing a schema-code mismatch Without this change, the code would be treating a text field as an array, which is incorrect This is not a breaking change if the code was already working with arrays (as it appears to be) Performance will actually be better with proper array types vs. implicit casting
> 
> Hey, I was just wondering—what if we use a join table to store the relations between entities and chunks instead of a chunk_ids column?

I like your suggestion.
It provided us with flexibility and potential performance boosts since we no longer need to keep this data in the same table.

LarFii:
I have merged [PR1138](https://github.com/HKUDS/LightRAG/pull/1138). I think `file_path` can also be synchronized as an array since an entity may be associated with multiple file_paths.

OxidBurn:
The PR is no longer relevant since we now have both fixes. I can release it later after refactoring with separators.

bzImage:
after git pull 5 minutes ago.. and testing with postgres backend im still facing this issue:

error:column "chunk_ids" of relation "lightrag_vdb_entity" does not exist                                               Failed to extract entities and relationships                                                                            Failed to process document doc-a2abbe75a7c91f62253147090cd48acf: column "chunk_ids" of relation "lightrag_vdb_entity" does not exist

PR #1200, opened 2025-03-27
Refactor mix prompt construction in mix_kg_vector_query function

OxidBurn:
# Description
This PR enhances the way LightRAG manages custom system prompts in the "mix" query! Before, when you provided a custom system prompt, it missed out on including the retrieval information. The only option was to stick with a predefined prompt. But now, it also takes into account situations where the custom system prompt might not have all the key parameters. In such cases, it seamlessly extends the prompt with theflexible and user-friendly. retrieved data, making it more 

# Changes Made
- Added try-except blocks to handle format string errors gracefully in all query functions:
    - `mix_kg_vector_query` - handles both knowledge graph and vector contexts
    - `kg_query` - handles knowledge graph context
    - `naive_query` - handles vector search results
  - Enhanced prompt generation to intelligently append KG and vector context when formatting keys aren't present
  - Added clear section headers (### Knowledge Graph Context, ### Vector Context) for better prompt organization
  
## Use Cases Addressed

1. **Custom prompts without format placeholders**: Users can now provide completely custom prompts without having to include specific format placeholders.
2. **Simplified prompt customization**: Reduces the complexity of creating custom prompts as users don't need to understand the internal formatting requirements.

# Checklist

  - [x] Changes tested locally
  - [x] Code reviewed

## Additional Notes

This change allows users to provide custom prompts with and without having to include specific formatting placeholders. The system will now intelligently append any available context information to the end of the prompt with proper section headers when formatting keys aren't present.

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -1079,6 +1079,8 @@ async def mix_kg_vector_query(
             sys_prompt += "\n1. From Knowledge Graph(KG):\n" + kg_context
         if vector_context:
             sys_prompt += "\n2. From Document Chunks(DC):\n" + vector_context
+        if history_context:
+            sys_prompt += "\n\n---Conversation History---\n" + history_context
 
     if query_param.only_need_prompt:
         return sys_prompt
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -1063,21 +1063,22 @@ async def mix_kg_vector_query(
     if query_param.only_need_context:
         return {"kg_context": kg_context, "vector_context": vector_context}
 
-    # 5. Construct hybrid prompt
-    sys_prompt = (
-        system_prompt
-        if system_prompt
-        else PROMPTS["mix_rag_response"].format(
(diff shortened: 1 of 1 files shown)


danielaskdd:
This is a commendable enhancement. Could you please ensure that kg_query incorporates the same functionality?

drahnreb:
> This is a commendable enhancement. 

Agree! I had a similar case.

> Could you please ensure that kg_query incorporates the same functionality?

… and `naive_query`.
I would also change the logic:
1. Adapt prompt templates to contain
a) only core aspects around context and
b) recommended default prompt for task format etc.
3. search for placeholders in custom `system_prompt`, if available.
4. if the user did not provide, add the template part a) that handles that (Some LLMs suggest to put at front of it) then add the custom prompt. Issue a warning to educate that behavior as you change the user prompt!

This could be a utility function to check for all adjustable templates as per #1401 .
@danielaskdd What do you think?

PR #1228, opened 2025-03-29
fix(useLightragGraph): Change the label of the edge from type to keyword

choizhang:
## Description

At present, all edge labels display the hard coded 'Directed', and it is unclear how they were designed and considered. Do you feel like changing it to properties Keywords are more meaningful？

Also, can we associate the weights of edge with their thickness? If so, I will develop this area

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed


LarFii:
Thanks for sharing!

Issue #1261, opened 2025-04-03
[Feature Request]: Is there a way to represent entities of the same type in the same color?


tigflanker:
### Do you need to file a feature request?

- [x] I have searched the existing feature request and this feature request is not already filed.
- [ ] I believe this is a legitimate feature request, not just a question or bug.

### Feature Request Description

For example, companies could be represented in blue, employees in green, and products in yellow. 

The specific colors don’t need to be predefined, but using the same color for the same type of entity would provide a very intuitive understanding. 💗

PR #1265, opened 2025-04-03
Feat：Nodes of the same type are represented by the same color

choizhang:
## Description

1. Optimize node color generation logic
2. Add legend components and toggle buttons

## Related Issues
#1261 

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

### DIFF
--- a/lightrag_webui/src/components/graph/Legend.tsx
+++ b/lightrag_webui/src/components/graph/Legend.tsx
@@ -1,0 +1,41 @@
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import { useGraphStore } from '@/stores/graph'
+import { Card } from '@/components/ui/Card'
+import { ScrollArea } from '@/components/ui/ScrollArea'
+
+interface LegendProps {
+  className?: string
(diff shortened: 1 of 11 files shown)


danielaskdd:
Thanks for sharing. PR is merged.

By the way, backendBaseUrl should not be fix to localhost. You can setup a vite dev environment for debug. There are config file examples in lightrag_webui folder.

danielaskdd:
@choizhang Could you also modify the node expansion feature to determine the color based on the node type?

choizhang:
> I took a look and found it relatively simple to do, so I've already completed it. Thanks. Currently, the biggest issue with the Web UI is that the Query page doesn't support streaming responses. Can you take a look at it to see if you can handle it?我看了一下，发现相对简单，所以我已经完成了。谢谢。目前，Web UI 最大的问题是查询页面不支持流式响应。你能看看是否能处理这个问题吗？

Okay, I'll try to solve it these days

danielaskdd:
That's great, thank you.

PR #1268, opened 2025-04-04
fix(postgres_impl): added file_path to queries

Keviddle:
## Description

Added 'file_path' to both get_by_id queries of the postgresql implementation, since they were still missing, which caused the new citation feature to not work when using postgres (and furthermore causes crashes when trying to query).

There is also another fix currently in PR, which adds at least a failsafe in case of doubt when not finding file paths. This fix provides the file paths when using postgres

## Related Issues

Related to PR #1264 
Fixes the same problem (at least partially since it would still crash if file_path is empty), but also adds support for acutally getting the file_paths

## Changes Made

- Modified postgres_impl.py and added 'file_path' to both get_by_id queries

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

No futher notes since all was covered in the previous texts


danielaskdd:
Thanks for sharing.

PR #1271, opened 2025-04-05
Fix the prefix or suffix error of the entity type

zjc98724:
Description
This pull request fixes the prefix or suffix errors of the entity type. Previously, when an invalid entity type was detected, the code would log a warning and potentially return None. Now, instead of interrupting the process, it directly uses regular expressions to clean up invalid characters, ensuring that the entity type is always valid.
Related Issues
None (Or fill in the associated issue number, e.g., Fixes #123)
Changes Made
src/entity_extractor.py
Modifications:
Removed the logic for checking if entity_type is empty or starts with (".
Directly used a regular expression to clean up parentheses characters:
python
# Original code
if not entity_type.strip() or entity_type.startswith('("'):
    logger.warning(...)
    # return None
    entity_type = re.sub(r'[()]', '', entity_type)

# New code
entity_type = re.sub(r'[()]', '', entity_type)

Optimization: Avoided the interruption of the entire record processing due to local exceptions.
Checklist
 Changes tested locally
 Code reviewed
 Documentation updated (if necessary)
 Unit tests added (if applicable)
Additional Notes
This modification assumes that all entity_type exceptions can be resolved by regular expression cleaning. It is necessary to verify this based on the actual business scenario.
It is recommended to add unit tests to cover the following scenarios:
python
def test_entity_type_cleaning():
    assert clean_entity_type('("invalid_type")') == 'invalid_type'
    assert clean_entity_type('valid_type') == 'valid_type'

If this involves production data, it is recommended to add logs to record the values before and after cleaning for traceability.

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -154,11 +154,8 @@ async def _handle_single_entity_extraction(
 
     # Clean and validate entity type
     entity_type = clean_str(record_attributes[2]).strip('"')
-    if not entity_type.strip() or entity_type.startswith('("'):
-        logger.warning(
-            f"Entity extraction error: invalid entity type in: {record_attributes}"
-        )
-        return None
+    # Fix the prefix or suffix error of the entity type
+    entity_type = re.sub(r'[()]', '', entity_type)
 
     # Clean and validate description
     entity_description = clean_str(record_attributes[3]).strip('"')


LarFii:
Thanks for sharing! I noticed two minor issues:

1. **Missing Empty Value Check**:  
   The updated code doesn’t check if `entity_type` is an empty string. If `entity_type` is empty, removing the parentheses will still leave it empty, which could cause issues in downstream processing.

2. **Logging**:  
   The original code logged a warning when an invalid `entity_type` was encountered, but the updated version doesn’t log anything. This might lead to a lack of useful information during debugging or monitoring.

**Suggestions**:
- After removing the parentheses, make sure to check whether `entity_type` is empty and log a warning if needed.
- Ensure that `entity_type` still holds a valid value after cleaning.

You could consider something like this:

```python
entity_type = re.sub(r'[()]', '', entity_type)
if not entity_type.strip():
    logger.warning(
        f"Entity extraction error: empty entity type after cleaning in: {record_attributes}"
    )
    return None
```

Also, there are some linting issues — please run `pre-commit run --all-files` before submitting.

Issue #1286, opened 2025-04-06
[Feature Request]:  Enhance WebUI Streaming Response Support


danielaskdd:
### Do you need to file a feature request?

- [ ] I have searched the existing feature request and this feature request is not already filed.
- [ ] I believe this is a legitimate feature request, not just a question or bug.

### Feature Request Description

The WebUI's query functionality does not adequately support streaming responses. The following improvements are needed:  

1. When selecting streaming responses for queries, the WebUI should process LightRAG's responses according to streaming response specifications.  
2. Switching tabs during a streaming response should not interrupt the response (currently, tab switching only hides the page and should not require special handling).  
3. Only session history should be maintained. The session history should be stored in the backend.

danielaskdd:
Session history is not implemented.

PR #1331, opened 2025-04-10
Fix response stream

choizhang:
## Description

Fix the issue where webui does not support stream output

## Related Issues

#1274 #1286 

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Directly connecting to the server is already supported, but using a proxy server locally is not yet possible


danielaskdd:
Thanks.

Issue #1342, opened 2025-04-10
[Feature Request]: Edit node/edge name and description by webui


danielaskdd:
### Do you need to file a feature request?

- [x] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

choizhang:
Let me try

danielaskdd:
Please note that editing an entity name may result in a naming conflict. So a name conflict check must be performed before saving any changes.

By the way, merging two or more nodes is discuss in #1343

danielaskdd:
The upsert_node and upsert_edge APIs are currently not implemented. If you encounter any challenges while implementing these APIs, please don't hesitate to reach out.

choizhang:
The API issue has been resolved, and real-time updates of graph data require some time for debugging

PR #1364, opened 2025-04-13
Edit node/edge name and description by webui 

choizhang:
## Description

Edit node/edge name and description by webui 

## Related Issues

#1342 

## Checklist

- [x] Changes tested locally
- [x] Code reviewed


danielaskdd (changes requested):
When editing nodes and edges, the local graph data needs to be updated after successful saving. When updating the local graph data, refreshing all data should not be adopted as it would trigger re-layout. Instead, the graph data should be modified directly.  

Additionally, the method for editing description fields is recommended to be changed to a pop-up window for easier editing and input of full paragraphs.

danielaskdd:
Apologies for the confusion. After retesting this PR with the default graph storage, I can confirm that the local graph data updates correctly. 

Could you please update the UI to display the description field editor in a pop-up window? This would improve usability for longer text input.

choizhang:
Yes, the user experience of refreshing all data is very poor. The UI has been updated, and your feedback is correct. Because the amount of testing data on my end is very small, I did not notice this issue before

PR #1374, opened 2025-04-15
Fix Neo4j node and edge edit problem

danielaskdd:
#### Fix Neo4j node and edge edit problem

Neo4j use unique identifier as node and edge key, which is difference from NetworkX.

#### refactor: improve graph property update mechanism

- Move graph data and UI state update logic into store
- Ensure all property updates trigger PropertiesView refresh
- Add graphDataVersion dependency to PropertiesView
- Follow React best practices by avoiding direct UI state modification in utility functions
- Fix issue where non-entity_id property changes weren't reflected in UI


danielaskdd:
@choizhang 

I’ve added property editing functionality for Neo4j storage based on your code. Note that the keys for nodes and edges returned by Neo4j are unique IDs, which differs from NetworkX where entity names are used. This distinction needs attention during subsequent graph editing operations.

Additionally, your original implementation couldn’t synchronize interface updates with real-time modifications. I’ve restructured the program following React’s conventions for your reference. Thank you.

Issue #1297, opened 2025-04-07
[Bug]: <title>"List index out of range" when a stream answer is displayed


ujnxs123:
### Do you need to file an issue?

- [ ] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

After getting the knowledge graph by uploading the document, I tried to do a Q&A test and found that everything was fine with just the only need context or prompt, while the "stream response" mode would only show "Error in stream response: list index out of range" no matter how much it was asked

### LightRAG Config Used

# Paste your config here
 The LLM model used is GPT-4O-2024-08-06

### Logs and screenshots

![Image](https://github.com/user-attachments/assets/cc528cbd-4d15-470c-816d-dc9a8e0d3f02)

### Additional Information

- LightRAG Version:1.3.1
- Operating System:
- Python Version:
- Related Issues:

danielaskdd:
The error occurs when you  do query with both stream and only_need_context enable, is that the case?

ujnxs123:
> The error occurs when you do query with both stream and only_need_context enable, is that the case?
NOI，exactly ，As soon as the stream response option is enabled, an error will be reported, regardless of whether other options are turned on.

danielaskdd:
Thanks. This issue will be fixed if we can reproduce it.

danielaskdd:
Fixex in main branch

PR #1409, opened 2025-04-19
fix streaming "list index out of range"

Enoughappens:
## Description

Fix "list index out of range" issue when using "stream response" in webui

Stated as Fixed in issues/1297, but still presents in latest version.
image

## Related Issues

https://github.com/HKUDS/LightRAG/issues/1297

## Changes Made

async def openai_complete_if_cache() -> async def inner()


danielaskdd:
Thanks for sharing

Issue #1414, opened 2025-04-19
[Feature Request]: Add an optional reranking for querying


drahnreb:
### Do you need to file a feature request?

- [x] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

#### Until today:
Standard retrieval methods (like vector similarity search) act as a first pass, retrieving an initial set of candidate documents.

#### With new feature:
Reranking introduces a second stage where a model re-evaluates these candidates to produce a refined ordering based on deeper semantic relevance to the query.

#### Suggested implementation:
Offering reranking as an optional parameter within `QueryParam`, allowing users to consciously trade speed for higher relevance, for `naive` and `mix` mode.

```
class QueryParam:
    mode: Literal["local", "global", "hybrid", "naive", "mix", "bypass"] = "global"
    top_k: Optional[int] = None  # Keep existing top_k for initial retrieval
    reranker_func: Optional[Callable] = None  # e.g. a custom reranker with model like 'bge-reranker-large' or 'cohere' and defined top_n, could be also mmr based etc.
    #... other existing or potential parameters...
```

If implemented, lightweight cross-encoders or efficient API-based solutions are preferable, integrated post-retrieval and pre-truncation. 

```
async def naive_query(
    query: str,
    chunks_vdb: BaseVectorStorage,
    text_chunks_db: BaseKVStorage,
    query_param: QueryParam,
    global_config: dict[str, str],
    hashing_kv: BaseKVStorage | None = None,
    system_prompt: str | None = None,
) -> str | AsyncIterator[str]:
    
    # existing implementation...
    # Filter out invalid chunks
    valid_chunks = [
        chunk for chunk in chunks if chunk is not None and "content" in chunk
    ]
    
    if not valid_chunks:
        logger.warning("No valid chunks found after filtering")
        return PROMPTS["fail_response"]

    reranked_chunks = valid_chunks  # Default to valid chunks if no reranker

    # reranking feature
    if query_param.reranker_func is not None:
         logger.debug(f"Applying reranker function: {query_param.reranker_func}")
        try:
            reranked_chunks = await query_param.reranker_func(chunks=valid_chunks, query=query)
        except Exception as e:
            logger.error(f"Error during reranking: {e}", exc_info=True)
            reranked_chunks = valid_chunks  # Fallback to non-reranked chunks
    
    logger.debug(
        f"Reranked chunks from {valid_chunks} to {reranked_chunks}"
[...]

PR #1400, opened 2025-04-17
Feature: Implemented an option to use DeepSeek Reasoner for intelligent semantic node re-ranking and retrieval

omdivyatej:
## Description

*Note: This feature could be extended to use other reasoning models as well down the line.*

I've added an option to use reasoning-based node re-ranking feature to LightRAG if the user wants that improves retrieval quality by using the **DeepSeek Reasoner R1** model to re-rank knowledge graph nodes based on relevance to the query. This goes beyond simple vector similarity by incorporating deep semantic understanding and multi-hop reasoning when selecting nodes.

Unlike traditional cosine similarity currently which only measures proximity in embedding space, this reasoning-based approach considers factors like information richness, contextual relevance, and centrality to the query's needs. This is particularly valuable for complex queries where the most similar embedding doesn't always contain the most useful information.

**Current Flow (Vector Similarity Only):**
1. User submits a query
2. System retrieves nodes from vector database based on embedding similarity
3. Nodes are ranked purely by cosine similarity score
4. Top-ranked nodes are used to generate the response

**New Implementation (With Reasoning Re-ranking):**
1. User submits a query
2. System retrieves nodes from vector database based on embedding similarity (initial retrieval)
3. The retrieved nodes are analyzed by the DeepSeek Reasoner model
4. The reasoner evaluates each node based on:
   - Semantic relevance to the query
   - Information richness and completeness
   - How central the node is to answering the query
   - The node's relationships with other nodes
5. Nodes are re-ranked according to the reasoner's evaluation
6. The re-ranked nodes are used to generate a more relevant response

Unlike traditional cosine similarity which only measures proximity in embedding space, this reasoning-based approach considers factors like information richness, contextual relevance, and centrality to the query's needs. This is particularly valuable for complex queries where the most similar embedding doesn't always contain the most useful information.

*Note: This feature is currently only available for `local` and `hybrid` modes.*

## Related Issues

No specific issue - this enhancement came from observing limitations in pure vector similarity retrieval when handling complex analytical queries.

## Changes Made

- Added `use_reasoning_reranking` boolean flag to `QueryParam` to enable/disable reasoning-based re-ranking
- Added `reasoning_model_name` parameter to `QueryParam` to specify which reasoning model to use
- Implemented `_rerank_nodes_with_reasoning` function in `operate.py` to handle the reasoning-based re-ranking logic
- Added prompts for reasoning re-ranking to `prompt.py`
- Enhanced `deepseek_r1_complete` function to capture and return reasoning chain of thought
[...]

PR #1415, opened 2025-04-19
feat: optional reranking for mix and naive query mode.

drahnreb:
## Description

Adds an optional reranking in `naive` and `mix` query mode.

## Related Issues

#1414

## Checklist

- [X] Changes tested locally
- [X] Code reviewed
- [X] Documentation updated (if necessary)
- [X] Unit tests added (if applicable)

### DIFF
--- a/examples/query_reranking_example.py
+++ b/examples/query_reranking_example.py
@@ -1,0 +1,151 @@
+import os
+import asyncio
+from lightrag import LightRAG, QueryParam
+from lightrag.llm.openai import gpt_4o_mini_complete
+from lightrag.kg.shared_storage import initialize_pipeline_status
+
+from typing import List, Any
+import functools
+from lightrag.utils import cosine_similarity, EmbeddingFunc
+
+#########
+# Uncomment the below two lines if running in a jupyter notebook to handle the async nature of rag.insert()
+# import nest_asyncio
+# nest_asyncio.apply()
+#########
+
+WORKING_DIR = "./dickens"
+
+if not os.path.exists(WORKING_DIR):
+    os.mkdir(WORKING_DIR)
+
+    def mmr_reranking(
+        query: str,
+        chunks: List[Any],
+        embedding_func: EmbeddingFunc,
+        lambda_param: float = 0.5,
+        top_n: int = 10,
+    ) -> List[Any]:
+        if not chunks:
+            # no chunks to rerank
+            return []
+        if len(chunks) <= top_n:
+            # fewer or equal top_n elements (set higher top_k)
+            return chunks[:]
+        embedded_query = embedding_func.func(query)
+        embedded_chunks = embedding_func.func(chunks)
+        selected_indices = []
+        remaining_indices = list(range(len(chunks)))
+        while len(selected_indices) < top_n and remaining_indices:
+            mmr_scores = {}
+            query_similarities = {
+                idx: cosine_similarity(embedded_query, embedded_chunks[idx])
+                for idx in remaining_indices
+            }
+
+            # Find the highest query similarity for the first pick
+            if not selected_indices:
+                best_idx = max(query_similarities, key=query_similarities.get)
+                selected_indices.append(best_idx)
+                remaining_indices.remove(best_idx)
+                continue  # Move to the next iteration
+
+            # Calculate MMR for remaining items
+            for idx in remaining_indices:
+                relevance_score = query_similarities[idx]
+
+                # Calculate max similarity to already selected items
+                diversity_penalty = 0.0
+                if (
+                    selected_indices
+                ):  # Should always be true here except first iteration
+                    similarities_to_selected = [
+                        cosine_similarity(
+                            embedded_chunks[idx], embedded_chunks[sel_idx]
+                        )
+                        for sel_idx in selected_indices
+                    ]
+                    diversity_penalty = (
+                        max(similarities_to_selected)
+                        if similarities_to_selected
+                        else 0.0
+                    )
+
+                mmr_score = (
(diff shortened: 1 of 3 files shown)


danielaskdd (changes requested):
Reranking the results of cos similarity search can indeed improve query quality. I have a few questions that I would like you to clarify:

1. Does the current approach consider allowing users to choose different reranker options? For example, selecting the bge-reranker-v2-m3 model deployed with Ollama, or choosing the jina-reranker-v2-base-multilingual model from Jina AI.

2. At present, using QueryParam to pass the reranker function does not seem very appropriate, because Web APIs generally should not specify reranker functions via QueryParam. Is there a better way for users to choose whether to enable reranking and which reranker to use?

3. Currently, I notice that you only modified `naive_query` to allow it to utilize the `reranker_func`. Why didn’t you similarly modify `mix_kg_vector_query`?

4. In addition, PR #1400 proposes an idea to perform reranking on nodes. I would like to hear your opinion on this approach. If it is necessary to rerank both chunks and nodes, how should the interface be designed?

drahnreb:
> 1. Does the current approach consider allowing users to choose different reranker options? For example, selecting the bge-reranker-v2-m3 model deployed with Ollama, or choosing the jina-reranker-v2-base-multilingual model from Jina AI.

As long as it follows the defined Signature https://github.com/HKUDS/LightRAG/blob/1e4ea229df5d1f17a2de22a52f66949425bed154/lightrag/base.py#L63
it accepts a query, and list of chunks to rerank. You would need to wrap it around your specific reranker (be it API or a simple filter as illustrated) just like for the other model funcs.

> 2. At present, using QueryParam to pass the reranker function does not seem very appropriate, because Web APIs generally should not specify reranker functions via QueryParam. Is there a better way for users to choose whether to enable reranking and which reranker to use?

Which Web APIs do you mean - this? https://github.com/HKUDS/LightRAG/blob/908953924a0e266e9ebb914e0c1563f942cc5d2a/lightrag/api/routers/query_routes.py#L19
Passing a callable function directly as a parameter within a web API request introduces several fundamental problems...
Client-Side code (e.g. allowing clients to send scripts  to be executed on the server) is probably not desired due to the security implications?
Otherwise we could use a configuration object and provide the set of rerankers by `langchain` and `langchain community`: https://python.langchain.com/docs/integrations/retrievers/#all-retrievers
e.g. https://api.python.langchain.com/en/latest/vertex_rank/langchain_google_community.vertex_rank.VertexAIRank.html
You would need to wrap context into langchain Documents which is not a bad thing in principle but needs some work.

> 3. Currently, I notice that you only modified `naive_query` to allow it to utilize the `reranker_func`. Why didn’t you similarly modify `mix_kg_vector_query`?

I simply missed it. You are right, it should be in both. I expected `mix_kg_vector_query ` to call `naive_query`.
- Can we simplify the code and remove `get_vector_context()` and add a flag to `naive_query` instead to handle the different returns and formatting and return with `query_param.only_need_context = True`?
- Also, if you limit to `top_k = 10` a reranking is not super useful, and in any case exposing this parameter would make sense to control it when reranking or if you do not care about context limits:
https://github.com/HKUDS/LightRAG/blob/1e4ea229df5d1f17a2de22a52f66949425bed154/lightrag/operate.py#L1127-L1128

> 4. In addition, PR [Feature: Implemented an option to use DeepSeek Reasoner for intelligent semantic node re-ranking and retrieval #1400](https://github.com/HKUDS/LightRAG/pull/1400) proposes an idea to perform reranking on nodes. I would like to hear your opinion on this approach. If it is necessary to rerank both chunks and nodes, how should the interface be designed?

I saw this PR and share the following remarks (some already mentioned): 
- reranking adds query time which is not favorable all the time. Reranking over dense vectors maybe doubles the end-to-end latency (depends on top_k, reranker_func). Generally I wouldn't LLM based reasoning over nodes to rerank, as it is suggested, because it is too time consuming.
- I do disagree to use another model, if your `llm_model_func` allows to "reason" (with an appropriate prompt) this should be sufficient to query for more relevant nodes/edges. Similar to this reranking, there is not much conceptual difference if the context comes from a vector database or a graph.
- But ideally you would want to do the _reasoning on the graph_ (that is what graphs are made for), which is partially (infant state) implemented via one-hop graph traversal retrieval. You could extend this with more sophisticated graph inference logics in pre-defined SPARQL clauses and LLM-based or do reranking via GNN. And offload the heavy work on the database, e.g. with more powerful providers like `neo4j`.

In short: In my opinion, if you want to rerank on graph, you should leverage the fact that you have a structured database and do structured reranking _before_ based on graph traversal and **not** _after_ graph retrieval from nodes (and possibly edges).

danielaskdd:
Obviously query_param is not suitable for callable object, we need to refactor the program to allow API clients to select different reranker models via query_param.

drahnreb:
> Obviously query_param is not suitable for callable object, we need to refactor the program to allow API clients to select different reranker models via query_param.

@danielaskdd, while this is a valid remark, we have a callable already in the `QueryParam` https://github.com/HKUDS/LightRAG/blob/1720771cdf687d800d013b7746d84e68d8a8b9f0/lightrag/base.py#L90

Is this going to be changed too? Or can we add the `reranker_func` for non-api usage only?

We could map rerankers to a set of supported `langchain` retrievers, but the intention of `model_func` is flexibility, which is quite the opposite what you would want in an API setting.

danielaskdd:
model_func is deprecating. There must be a new way for user to change LLM a query time.

danielaskdd:
LightRAG is incorporating a Reranker model to facilitate the sorting of text chunks.

PR #1423, opened 2025-04-21
friendly implementation of entity extraction and relationship weight extract for Low-Capability LLMs

tackhwa:
friendly implementation of entity extraction and relationship weight extract for Low-Capability LLMs

## Description
This PR addresses two critical issues when processing output from smaller LLMs:
1. These models often produce malformed entity outputs, such as:
- Missing or incorrectly placing record delimiters (e.g., placing them inside parentheses)
- Omitting opening or closing parentheses.
- Since the current entity extraction logic relies on regex, these issues lead to many valid entities being ignored, resulting in a low entity count.
2. Relationship weights are sometimes returned wrapped in quotes (e.g., "1.0" or '0.8'). The current regex does not handle this properly and defaults to returning a weight of 1.0.

## Changes Made
1. Introduces a `new extract_fixed_parenthesized_content` function that runs after `split_string_by_multi_markers`, to clean and correct malformed records caused by issue 1
2. Adds a strip() call before passing relationship weight values to is_float_regex, ensuring that quoted numeric strings are properly parsed.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -18,6 +18,7 @@ from .utils import (
     normalize_extracted_info,
     pack_user_ass_to_openai_messages,
     split_string_by_multi_markers,
+    extract_fixed_parenthesized_content,
     truncate_list_by_token_size,
     process_combine_contexts,
     compute_args_hash,
@@ -215,7 +216,7 @@ async def _handle_single_relationship_extraction(
     edge_source_id = chunk_key
     weight = (
         float(record_attributes[-1].strip('"').strip("'"))
-        if is_float_regex(record_attributes[-1])
+        if is_float_regex(record_attributes[-1].strip('"').strip("'"))
         else 1.0
     )
     return dict(
@@ -548,6 +549,8 @@ async def extract_entities(
             result,
             [context_base["record_delimiter"], context_base["completion_delimiter"]],
         )
+
+        records = extract_fixed_parenthesized_content(records)
 
         for record in records:
             record = re.search(r"\((.*)\)", record)
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -18,7 +18,6 @@ from .utils import (
     normalize_extracted_info,
     pack_user_ass_to_openai_messages,
     split_string_by_multi_markers,
-    extract_fixed_parenthesized_content,
     truncate_list_by_token_size,
     process_combine_contexts,
     compute_args_hash,
@@ -153,7 +152,7 @@ async def _handle_single_entity_extraction(
     chunk_key: str,
     file_path: str = "unknown_source",
 ):
-    if len(record_attributes) < 4 or record_attributes[0] != '"entity"':
+    if len(record_attributes) < 4 or '"entity"' not in record_attributes[0]:
         return None
 
     # Clean and validate entity name
@@ -199,7 +198,7 @@ async def _handle_single_relationship_extraction(
     chunk_key: str,
     file_path: str = "unknown_source",
 ):
-    if len(record_attributes) < 5 or record_attributes[0] != '"relationship"':
+    if len(record_attributes) < 5 or '"relationship"' not in record_attributes[0]:
         return None
     # add this record as edge
     source = clean_str(record_attributes[1])
@@ -549,8 +548,6 @@ async def extract_entities(
             result,
             [context_base["record_delimiter"], context_base["completion_delimiter"]],
         )
-
-        records = extract_fixed_parenthesized_content(records)
 
         for record in records:
             record = re.search(r"\((.*)\)", record)
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -408,33 +408,6 @@ def split_string_by_multi_markers(content: str, markers: list[str]) -> list[str]
     return [r.strip() for r in results if r.strip()]
 
 
-def extract_fixed_parenthesized_content(records: list[str]) -> list[str]:
-    """
-    Extract content that should be in parentheses from each record.
-    Ensures each extracted item has both opening and closing parentheses.
-    """
-    result = []
-
-    for record in records:
-        # First, extract properly matched pairs
-        balanced_matches = re.findall(r"\((.*?)\)", record)
-        for match in balanced_matches:
-            result.append(f"({match})")
-
-        # Process string to handle unbalanced parentheses
-        # For opening without closing
-        open_matches = re.findall(r"\(([^()]*?)$", record)
-        for match in open_matches:
-            result.append(f"({match})")
-
-        # For closing without opening
-        close_matches = re.findall(r"^([^()]*?)\)", record)
-        for match in close_matches:
-            result.append(f"({match})")
-
-    return result
-
-
 # Refer the utils functions of the official GraphRAG implementation:
 # https://github.com/microsoft/graphrag
 def clean_str(input: Any) -> str:
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -417,21 +417,21 @@ def extract_fixed_parenthesized_content(records: list[str]) -> list[str]:
 
     for record in records:
         # First, extract properly matched pairs
-        balanced_matches = re.findall(r'\((.*?)\)', record)
+        balanced_matches = re.findall(r"\((.*?)\)", record)
         for match in balanced_matches:
             result.append(f"({match})")
-        
+
         # Process string to handle unbalanced parentheses
         # For opening without closing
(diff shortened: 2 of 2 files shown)


danielaskdd (changes requested):
After testing, it was found that the number of entities and relations recognized by the modified program has decreased to some extent. It is suspected that the extract_fixed_parenthesized_content function cannot correctly handle nested parentheses and special characters. For example, some entity names themselves contain parentheses.

tackhwa:
> I guess @danielaskdd means robustness across different characteristics and issues with corner cases.
> 
> One of the examples to test:
> 
> What if `三体游戏-荒原` would be `三体游戏 (荒原)` in your source text?
> 
> For such critical changes we could need some unit tests (with `hypotheses` and fuzzing). Malformed LLM responses, where opening parentheses are missing, would need to be defined clearly and checked if even parsed by default...

The example you gave, 三体游戏 (荒原), wouldn’t work in this case, and it also wouldn’t work with the default LightRag implementation due to the following code snippet in the `async def _process_extraction_result` function. 
```
record = re.search(r"\((.*)\)", record)
            if record is None:
                continue
            record = record.group(1)
```
What I’ve implemented now is just a very simple function that takes an additional step to further split the parenthesized content after the split_string_by_multi_markers function. This ensures that entities can still be extracted correctly when the LLM fails to return the record delimiter, and it also completes any missing parentheses. It’s not a critical change at all, just a very straightforward regex tweak.

Here are some concrete examples to clarify:
1. LLM outputs the record delimiter inside the parentheses:
```
("relationship"<|>"凌晨一点"<|>"7章"<|>"时间背景"<|>"描述汪淼打电话的时间"<|>"3"##)
```
In this case, the default LightRAG will treat the output after `split_string_by_multi_markers` like this:
```
("relationship"<|>"凌晨一点"<|>"7章"<|>"时间背景"<|>"描述汪淼打电话的时间"<|>"3"
```
As a result, this entity will be ignored due to the following snippet in the code:
```
record = re.search(r"\((.*)\)", record)
            if record is None:
                continue
            record = record.group(1)
```

2. LLM does not output the record delimiter at all:
```
("entity"<|>"7章"<|>"章节"<|>"描述了汪淼与丁仪讨论宇宙背景辐射以及参加三体游戏的经历")\n("entity"<|>"凌晨一点"<|>"地点"<|>"汪淼给丁仪打电话的时间，表明时间已经很晚")
```
In this scenario, since there's no `##` record delimiter avaible in the result, the default `split_string_by_multi_markers` function will do nothing and return the result directly.
And only the first entity will be processed by LightRAG, again due to the same logic:
```
record = re.search(r"\((.*)\)", record)
            if record is None:
                continue
            record = record.group(1)
```

The case where LLM output incomplete parentheses will be similiar to `example 1`

I hope this explanation makes the issue clearer.
I can provide a test case if we really need one for this

tackhwa:
> > I guess @danielaskdd means robustness across different characteristics and issues with corner cases.
> > One of the examples to test:
> > What if `三体游戏-荒原` would be `三体游戏 (荒原)` in your source text?
> > For such critical changes we could need some unit tests (with `hypotheses` and fuzzing). Malformed LLM responses, where opening parentheses are missing, would need to be defined clearly and checked if even parsed by default...
> 
> The example you gave, 三体游戏 (荒原), wouldn’t work in this case, and it also wouldn’t work with the default LightRAG implementation due to the following code snippet in the `async def _process_extraction_result` function.
> 
> ```
> record = re.search(r"\((.*)\)", record)
>             if record is None:
>                 continue
>             record = record.group(1)
> ```
> 
> What I’ve implemented now is just a very simple function that takes an additional step to further split the parenthesized content after the `split_string_by_multi_markers` function. This ensures that entities can still be extracted correctly when the LLM fails to return the record delimiter, and it also completes any missing parentheses. It’s not a critical change at all, just a very straightforward regex tweak.
> 
> Here are some concrete examples to clarify:
> 
> 1. LLM outputs the record delimiter inside the parentheses:
> 
> ```
> ("relationship"<|>"凌晨一点"<|>"7章"<|>"时间背景"<|>"描述汪淼打电话的时间"<|>"3"##)
> ```
> 
> In this case, the default LightRAG will treat the output after `split_string_by_multi_markers` like this:
> 
> ```
> ("relationship"<|>"凌晨一点"<|>"7章"<|>"时间背景"<|>"描述汪淼打电话的时间"<|>"3"
> ```
> 
> As a result, this entity will be ignored due to the following snippet in the code (incomplete parentheses):
> 
> ```
> record = re.search(r"\((.*)\)", record)
>             if record is None:
>                 continue
>             record = record.group(1)
> ```
> 
> 2. LLM does not output the record delimiter at all:
> 
> ```
> ("entity"<|>"7章"<|>"章节"<|>"描述了汪淼与丁仪讨论宇宙背景辐射以及参加三体游戏的经历")\n("entity"<|>"凌晨一点"<|>"地点"<|>"汪淼给丁仪打电话的时间，表明时间已经很晚")
> ```
> 
> In this scenario, since there's no `##` record delimiter is available in the result, the default `split_string_by_multi_markers` function will do nothing and return the result directly. And only the first entity will be processed by LightRAG, again due to the same logic (only extract the result wrap in first parentheses):
> 
> ```
> record = re.search(r"\((.*)\)", record)
>             if record is None:
>                 continue
>             record = record.group(1)
> ```
> 
> The case where LLM output incomplete parentheses will be similiar to `example 1`
> 
> I hope this explanation makes the issue clearer. I can provide a test case if we really need one for this

drahnreb:
Sorry, I didn't want to say your proposals are not valid.
I mean, that a change in the acceptance criteria for these extractions should be generally testable to male the definitions transparent.

I think you demonstrated exactly this with your examples: As long as we don't have such test cases with expected output, it is debatable what should be accepted.

Already those delimiters and the parentheses are generally brittle because they could be easily in entity names and are no escaped.
Your malformations are edge cases even rarer.

Ideally this could be structured output, so delimiters are less critical and parsing is more robust?
But this is maybe not solving it for weaker LLMs, because it has token size implications.

@danielaskdd what was the rational to define the expected record tuples with that format in the prompt?

danielaskdd:
I am not very familiar with prompt engineering. Perhaps we should instruct the LLM to use a more simple and robust format. @LarFii

tackhwa:
Apologies for the lack of comprehensive test coverage. You're right—the current implementation in this PR does not handle nested parentheses. It’s actually a trade-off between supporting nested parentheses (as in the default LightRAG implementation) and dealing with incorrect output formats that lack proper parentheses (which this PR originally aimed to address).
And I couldn’t find a perfect solution that handles both cases at this time, unless  replacing the regex with some custom logic that compares characters one by one in a loop, but this is a critical change that would require the entire system to be thoroughly tested.

The examples I provided were real-world cases I encountered when using the `glm-4-9b-chat model`. The regex approach introduced in this PR was more tailored toward personal use cases, where entity names do not contain nested parentheses. I believe that with newer, smarter LLMs, the output format will become more consistent and correct.

Given this, I’ve already removed the regex logic from the PR, leaving only some basic formatting fixes.

I really appreciate the time and input from both @danielaskdd and @drahnreb during the review and discussion.

danielaskdd:
Thanks for sharing. I hope to find a method in the future to better perform format cleaning on the content returned by LLMs.

tackhwa:
> Thanks for sharing. I hope to find a method in the future to better perform format cleaning on the content returned by LLMs.

By the way, do you have something like a `golden_dataset` that contains all the ground truth entities and relationships? I’d like to continue developing new methods for format cleaning and test them on such a dataset. Or do you have any suggestions for a suitable dataset?

maybe cc @LarFii

drahnreb:
> Or do you have any suggestions for a suitable dataset?

https://github.com/salmon-kg/TinyButMighty
Maybe you can start with this? Stanford OpenIE has maybe more, and the paper should have more related references.
Benchmark Datasets for triplet extraction are scarce however.


> I’d like to continue developing new methods for format cleaning and test them on such a dataset.

You could (AI-) generate divers (and maybe malformed) set of examples.
But for testing/developing a better parsing, you would only need to generate the LLM triplet response **not** the source document that are the foundation of the triplets extraction.

Like I said, a better `json` structure could fix this parsing step and make this obsolete.

danielaskdd:
Entity name should support all most all kinds of character in it, including
- English symbols: !@#$%^&*"'`-+=[]{}/\
- Multi language characters like Chinese or Arabic

Powerful LLM will escape english symbols properly.

PR #1427, opened 2025-04-21
use a single parameter placeholder for the array;

OxidBurn:
## Description

Switched Postgres vector‐store “query” logic to bind document IDs as a proper text[] parameter instead of interpolating a comma‐separated string. This prevents SQL syntax errors when multiple IDs are passed into the CTE (i.e. “syntax error at or near ‘,’”).

## Changes Made

* In `lightrag/kg/postgres_impl.py`, refactored the `query(...)` method:
    * Removed inline concatenation/quoting of `ids`.
    * Always bind `ids` as the second parameter (`$2::varchar[]`) to the SQL.
    * Updated the `params` dict to include `"doc_ids": ids`.
* Updated the three RAG SQL templates (“chunks”, “entities”, “relationships”) so their CTEs use:  WHERE $2::varchar[] IS NULL OR full_doc_id = ANY($2::varchar[]) and bumped the downstream placeholders (`$3`, `$4`) accordingly.
* Retained inline formatting of the embedding vector, but no longer interpolate ID lists into the query string.
* Maintains previous behavior where `ids=None` searches all documents.

## Checklist

* [X]  Changes tested locally
* [X]  Code reviewed
* [ ]  Documentation updated (if necessary)
* [ ]  Unit tests added (if applicable)

## Additional Notes

This fix restores correct RAG lookup behavior on Postgres backends when supplying multiple document IDs and avoids introducing SQL injection vectors by using parameter binding.

### DIFF
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -2131,7 +2131,7 @@ SQL_TEMPLATES = {
     WITH relevant_chunks AS (
         SELECT id as chunk_id
         FROM LIGHTRAG_DOC_CHUNKS
-        WHERE $2::text[] IS NULL OR full_doc_id = ANY($2::text[])
+        WHERE $2::varchar[] IS NULL OR full_doc_id = ANY($2::varchar[])
     )
     SELECT source_id as src_id, target_id as tgt_id
     FROM (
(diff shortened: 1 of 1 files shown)


danielaskdd:
Thank you for your valuable contribution.

PR #1502, opened 2025-04-30
Fixed MongoDB Implementation

mbhardwajai:
Fixed MongoDB Doc Status Storage class - specifically, added its missing `delete()` method.


FeHuynhVI:
I pull code and test, same not working when delete

Issue #1530, opened 2025-05-06
[Bug]:passing a parameter called "stream" to the Bedrock API, but it's not in the list of accepted parameters


balaji-munusamy:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

When using bedrock_complete, it throws below exception. Because a parameter called "stream" is being passed to the Bedrock API, but it's not in the list of accepted parameters

```
botocore.exceptions.ParamValidationError: Parameter validation failed:
Unknown parameter in input: "stream", must be one of: modelId, messages, system, inferenceConfig, toolConfig, guardrailConfig, additionalModelRequestFields, promptVariables, additionalModelResponseFieldPaths, requestMetadata, performanceConfig
```

### LightRAG Config Used

# Paste your config here

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

PR #1531, opened 2025-05-06
Remove 'stream' param from kwargs

balaji-munusamy:
## Description

Fixes the issue #1530

### DIFF
--- a/lightrag/llm/bedrock.py
+++ b/lightrag/llm/bedrock.py
@@ -51,6 +51,7 @@ async def bedrock_complete_if_cache(
         "AWS_SESSION_TOKEN", aws_session_token
     )
     kwargs.pop("hashing_kv", None)
+    kwargs.pop("stream", None)
     # Fix message history format
     messages = []
     for history_message in history_messages:


danielaskdd:
If the 'stream' parameter is removed, how does the Bedrock API handle streaming response behavior?

balaji-munusamy:
> If the 'stream' parameter is removed, how does the Bedrock API handle streaming response behavior?

@danielaskdd AWS bedrock converse API has another method(not the same method with `stream` parameter) for streaming support and hence we'll need to call that method. Perhaps, we can create another feature request to support streaming response. Nevertheless, the current fix will unblock users from using `bedrock_complete`.

danielaskdd:
The Bedrock API's lack of streaming response support is not acceptable, resulting in diminished effectiveness for programs relying on streamed responses. It is recommended to improve the Bedrock API by referring to the implementation in openai.py. 

If the Bedrock API remains unmaintained, we may consider temporarily removing bedrock.py. Users are advised to access the Bedrock API via AWS’s official OpenAI Access Gateway: [https://github.com/aws-samples/bedrock-access-gateway ](https://github.com/aws-samples/bedrock-access-gateway).

PR #1549, opened 2025-05-08
Using keyed lock for max concurrency

maharjun:
## Description

This pull request introduces a keyed locking mechanism for entity and relationship processing in LightRAG. The aim is to ensure atomic and mutually exclusive operations on individual entities or relationships, while allowing concurrent processing of different entities and relationships.

This locking mechanism is designed with multi-processing environments in mind, although this specific compatibility has not yet been tested.

## Key motivation
It turns out that when trying to run massively parallel insertion, namely hundreds of documents at once (max_parallel_insert = 100 for instance), we run into the issue that the hard lock that is taken on the database during the merging phase becomes a significant bottleneck, especially if we're doing LLM summarization on merge. However, from my initial observation of the logic, the processing of any single entity depends only on the entries that correspond to that entity, the processing of any relation depends on only entries corresponding to that relation and possibly entries corresponding to the related entries.

The changes I have made in this pull request essentially allow for this operation to no longer require a lock of the entire database, but only a lock specific to the entity name or relation name that it seeks to modify. This allows for massively parallel document insertion that allows for full use of GPU / API capacity.

## Key Changes

1.  **Keyed Locking for Granular Control**:
    *   Introduced `get_graph_db_lock_keyed(key)`: a function/method that returns an asynchronous lock context manager specific to the provided `key` (e.g., an `entity_name` or a unique `src_id_tgt_id` string for relationships).
    *   **Mutual Exclusion**: If multiple coroutines attempt to acquire a lock for the *same* key via `get_graph_db_lock_keyed(key)`, they will operate sequentially on that specific item. One coroutine must release the lock before another can acquire it for that same key.
    *   **Concurrency**: Operations using locks for *different* keys can proceed in parallel without blocking each other.

2.  **Atomic Operations**:
    *   All operations related to a single entity (e.g., graph updates, vector DB upserts) are grouped and executed atomically under the lock obtained for that entity's key.
    *   Similarly, all operations for a single relationship are performed atomically under the lock obtained for that relationship's key.
    *   This ensures that an entity or relationship is always in a consistent state post-operation, preventing partial updates during concurrent execution.

3.  **Concurrent Processing Logic**:
    *   Utilizes `asyncio.gather()` to process merging of different entities in parallel.
    *   Utilizes `asyncio.gather()` to process merging of different relationships in parallel.
    *   The keyed locking mechanism, accessed via `get_graph_db_lock_keyed(key)`, integrates with this concurrent execution to manage access.

## Additional Changes
Introduced semaphore for merging portion because now that this portion is massively parallel it needs to be appropriately resource controlled. It is in-fact significantly more parallel than the initial portion since each entity and relation is processed in parallel potentially. In-fact in practice I have found it beneficial to increase the llm_model_max_async and embedding_func_max_async to 3x of max_parallel_insert.

## Illustrative Examples of Locking

Let's assume `self.get_graph_db_lock_keyed` is the method providing access to the new locking mechanism.

**Entity Processing**:
```python
# self.get_graph_db_lock_keyed(key) returns an async lock
# context manager specific to the 'entity_name'.
async with self.get_graph_db_lock_keyed(entity_name):
    # Any other coroutine attempting to lock the *same* entity_name
    # (i.e., calling self.get_graph_db_lock_keyed(entity_name))
    # will block here until this critical section is exited.
    # Operations within this block are atomic for this entity.
    await self.chunk_entity_relation_graph.upsert_node(...)
    await self.entities_vdb.upsert(...)
```

**Relationship Processing**:
```python
# Generate a unique key for the relationship.
relationship_key = f"{src_id}_{tgt_id}"
async with self.get_graph_db_lock_keyed(relationship_key):
    # Any other coroutine attempting to lock the *same* relationship_key
    # will block here until this critical section is exited.
    # Operations within this block are atomic for this relationship.
    await self.chunk_entity_relation_graph.upsert_edge(...)
    await self.relationships_vdb.upsert(...)
```

## Related Issues

-   Implements atomic operations for entities/relationships.
-   Addresses potential race conditions during concurrent updates to the same items.
-   Contributes to data consistency during parallel processing.

## Checklist

-   [x] Changes tested locally (for asyncio concurrency)
-   [ ] Code reviewed
-   [ ] Internal behavior, no documentation present AFAIK (maybe theres some developer docs somewhere?)
-   [ ] Unit tests added (I think this should be tested against existing tests for correctness)
-   [ ] Multi-processing compatibility tested [I have added multi-processing keyed locks, but haven't had the opportunity to test it]

## Summary

The locking strategy is fine-grained to individual data items (entities/relationships), identified by unique keys. This allows for concurrent processing of unrelated items while enforcing serial, atomic access for operations on the same item.

### DIFF
--- a/lightrag/kg/shared_storage.py
+++ b/lightrag/kg/shared_storage.py
@@ -1,9 +1,12 @@
+from collections import defaultdict
 import os
 import sys
 import asyncio
+import multiprocessing as mp
 from multiprocessing.synchronize import Lock as ProcessLock
 from multiprocessing import Manager
-from typing import Any, Dict, Optional, Union, TypeVar, Generic
+import time
+from typing import Any, Callable, Dict, List, Optional, Union, TypeVar, Generic
 
 
 # Define a direct print function for critical logs that must be visible in all processes
@@ -27,7 +30,13 @@ LockType = Union[ProcessLock, asyncio.Lock]
 _is_multiprocess = None
 _workers = None
 _manager = None
+_lock_registry: Optional[Dict[str, mp.synchronize.Lock]] = None
+_lock_registry_count: Optional[Dict[str, int]] = None
+_lock_cleanup_data: Optional[Dict[str, time.time]] = None
+_registry_guard = None
 _initialized = None
+
+CLEANUP_KEYED_LOCKS_AFTER_SECONDS = 300
 
 # shared data for storage across processes
 _shared_dicts: Optional[Dict[str, Any]] = None
@@ -40,10 +49,31 @@ _internal_lock: Optional[LockType] = None
 _pipeline_status_lock: Optional[LockType] = None
 _graph_db_lock: Optional[LockType] = None
 _data_init_lock: Optional[LockType] = None
+_graph_db_lock_keyed: Optional["KeyedUnifiedLock"] = None
 
 # async locks for coroutine synchronization in multiprocess mode
 _async_locks: Optional[Dict[str, asyncio.Lock]] = None
 
+DEBUG_LOCKS = False
+_debug_n_locks_acquired: int = 0
+def inc_debug_n_locks_acquired():
+    global _debug_n_locks_acquired
+    if DEBUG_LOCKS:
+        _debug_n_locks_acquired += 1
+        print(f"DEBUG: Keyed Lock acquired, total: {_debug_n_locks_acquired:>5}", end="\r", flush=True)
+
+def dec_debug_n_locks_acquired():
+    global _debug_n_locks_acquired
+    if DEBUG_LOCKS:
+        if _debug_n_locks_acquired > 0:
+            _debug_n_locks_acquired -= 1
+            print(f"DEBUG: Keyed Lock released, total: {_debug_n_locks_acquired:>5}", end="\r", flush=True)
+        else:
+            raise RuntimeError("Attempting to release lock when no locks are acquired")
+
+def get_debug_n_locks_acquired():
+    global _debug_n_locks_acquired
+    return _debug_n_locks_acquired
 
 class UnifiedLock(Generic[T]):
     """Provide a unified lock interface type for asyncio.Lock and multiprocessing.Lock"""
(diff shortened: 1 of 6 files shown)


maharjun:
Glad you find this useful. Let me know if you'd like some help. This week is super busy. Might find some time next week.

lightrag/kg/shared_storage.py
```diff
@@ -210,6 +240,207 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        return None
+
+    with _registry_guard:
+        combined_key = _get_combined_key(factory_name, key)
+        raw = _lock_registry.get(combined_key)
+        count = _lock_registry_count.get(combined_key)
+        if raw is None:
+            raw = _manager.Lock()
+            _lock_registry[combined_key] = raw
+            _lock_registry_count[combined_key] = 0
```
danielaskdd, line 268: _lock_registry_count[combined_key] should set to 1

danielaskdd:
Excellent work.

lightrag/operate.py
```diff
@@ -422,28 +422,32 @@ async def _merge_edges_then_upsert(
...
-                node_data={
-                    "entity_id": need_insert_id,
-                    "source_id": source_id,
-                    "description": description,
-                    "entity_type": "UNKNOWN",
-                    "file_path": file_path,
-                    "created_at": int(time.time()),
-                },
-            )
+        if (await knowledge_graph_inst.has_node(need_insert_id)):
```
danielaskdd, line 425: Move node existence check inside keyed lock to prevent race condition

lightrag/kg/shared_storage.py
```diff
@@ -210,6 +240,207 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        else:
+            if count is None:
+                raise RuntimeError(f"Shared-Data lock registry for {factory_name} is corrupted for key {key}")
+            count += 1
+            _lock_registry_count[combined_key] = count
+            if count == 1 and combined_key in _lock_cleanup_data:
```
danielaskdd, line 274: Checking count == 1 is wrong, count should be 0 when reusing a keyed lock

maharjun: Should I resolve these issues in a new PR? Or have you already done this? I've also improved typing for the storage and lock related stuff in general (checked against pyright). I can push these changes if its fine by you.

maharjun:
Okay, I see massive changes on your end as well. So I'll now need to take some time to absorb them and then contribute. The main issue is that of typing, but that'll be easy to fix once I digest whats happenning. Thanks for taking the time! Cheers!

danielaskdd:
I would appreciate a comprehensive evaluation of the revised implementation. Due to the complexity of the concurrency control logic, it's challenging to confirm all edge cases have been addressed through debugging alone.

PR #1559, opened 2025-05-11
added required parameter to the edge query results;

OxidBurn:
## Description
Add required edge property defaults to ensure consistency in graph operations.

## Related Issues
Fixes issues where missing edge properties caused downstream errors in knowledge graph operations.

## Changes Made
- Added required parameter checking to `get_edge` and `get_edges_batch` methods
- Implemented default values for critical edge properties (weight, source_id, description, keywords)
- Added warning logs when defaults are applied for better debugging and visibility

## Checklist
- [x] Changes tested locally
- [x] Code reviewed
- [ ] Unit tests

### DIFF
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -1430,20 +1430,40 @@ class PGGraphStorage(BaseGraphStorage):
             src_label,
             tgt_label,
         )
-        record = await self._query(query)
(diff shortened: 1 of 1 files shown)


danielaskdd:
This mandatory value should be stored in database. It's not a good practice to check and complete data every time it's fetched.

PR #1585, opened 2025-05-15
add feature: export and import data

morellid:
create a class for exporting and importing all the data in and out of lightrag.


morellid:
for future reference: when I started this PR it worked fine on my particular case, using local json files as data stores. But when I tried it on other data stores I realised that in some cases it didn't work: not all data stores export the necessary functions.

Issue #1573, opened 2025-05-14
[Bug]:KeyError: 'get_by_mode_id_project2llm_response_cache'


jasperchen01:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

I think there's a bug when enable enable_llm_cache_for_entity_extract.
My lightrag config as below:
```python
rag = LightRAG(
        namespace_prefix=NAMESPACE,
        working_dir=WORKING_DIR,
        enable_llm_cache=False,
        llm_model_func=_llm_model_func,
        llm_model_max_async=4,
        llm_model_max_token_size=32768,
        enable_llm_cache_for_entity_extract=True,
        embedding_func=EmbeddingFunc(
            embedding_dim=1024,
            max_token_size=8192,
            func=_embedding_func,
        ),
        embedding_batch_num=10,
        embedding_func_max_async=10,
        embedding_cache_config={
            "enabled": "true",
            "similarity_threshold": 0.95,
            "use_llm_check": False,
        },
        kv_storage="PGKVStorage",
        doc_status_storage="PGDocStatusStorage",
        graph_storage="PGGraphStorage",
        # graph_storage="NetworkXStorage",
        vector_storage="MilvusVectorDBStorage",
        auto_manage_storages_states=True,
        # llm_model_kwargs={
        #     "response_format": {"type": "json_object"},
        #     "extra_body": {"enable_search": True}
        #     },
        addon_params={
            # "language": "Simplified Chinese",
            "language": "English",
        },
    )
```

Then I get the error logs:
```bash
Traceback (most recent call last):
  File "/Users/apple/workplace/AI/matetext-ai/libs/lightrag/lightrag/operate.py", line 829, in _process_with_semaphore
    return await _process_single_content(chunk)
  File "/Users/apple/workplace/AI/matetext-ai/libs/lightrag/lightrag/operate.py", line 754, in _process_single_content
    final_result = await use_llm_func_with_cache(
  File "/Users/apple/workplace/AI/matetext-ai/libs/lightrag/lightrag/utils.py", line 1594, in use_llm_func_with_cache
    cached_return, _1, _2, _3 = await handle_cache(
  File "/Users/apple/workplace/AI/matetext-ai/libs/lightrag/lightrag/utils.py", line 1004, in handle_cache
[...]

PR #1597, opened 2025-05-19
Add workspace param for lightrag with PG storage

jasperchen01:
## Description

Add workspace param for lightrag with PG storage

## Related Issues

https://github.com/HKUDS/LightRAG/issues/1573

## Changes Made

1. Add workspace param in class StorageNameSpace(ABC)
2. Change milvus_impl.py to support the workspace collections
3. Change postgres_impl.py to support the workspace query
4. Change lightrag/lightrag.py to pass the workspace value when init the lightrag

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes


earayu (approved):

danielaskdd (changes requested):
The PGGraphStorage class itself does not directly support data isolation through workspaces. The graph data from different workspaces will be stored in the same Apache AGE graph, which can cause LightRAG RAG to malfunction.

danielaskdd:
In the current workflow, it seems that the workspace can only be set during the initialization of the LightRAG object. If you need to switch workspaces later, you have to create a new LightRAG object. Is my understanding correct? If so, is there a better way to switch workspaces?

jasperchen01:
> In the current workflow, it seems that the workspace can only be set during the initialization of the LightRAG object. If you need to switch workspaces later, you have to create a new LightRAG object. Is my understanding correct? If so, is there a better way to switch workspaces?

Yes, it can not to dynamically switch workspace for my changes. I think if we wanna the ability to dynamically switch workspaces , we need to modify all insert/query interfaces to add a workspace parameter

Issue #1640, opened 2025-05-28
[Bug]: ERROR: limit_async: Error in decorated function:


IAMJOYBO:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

```docker-compose
lightrag  | ERROR: limit_async: Error in decorated function:
lightrag  | ERROR: limit_async: Error in decorated function:
lightrag  | ERROR: limit_async: Error in decorated function:
lightrag  | ERROR: limit_async: Error in decorated function:
lightrag  | ERROR: Failed to extract entities and relationships:
lightrag  | ERROR: Traceback (most recent call last):
lightrag  |   File "/root/.local/lib/python3.11/site-packages/httpx/_transports/default.py", line 101, in map_httpcore_exceptions
lightrag  |     yield
lightrag  |   File "/root/.local/lib/python3.11/site-packages/httpx/_transports/default.py", line 394, in handle_async_request
lightrag  |     resp = await self._pool.handle_async_request(req)
lightrag  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lightrag  |   File "/root/.local/lib/python3.11/site-packages/httpcore/_async/connection_pool.py", line 256, in handle_async_request
lightrag  |     raise exc from None
lightrag  |   File "/root/.local/lib/python3.11/site-packages/httpcore/_async/connection_pool.py", line 236, in handle_async_request
lightrag  |     response = await connection.handle_async_request(
lightrag  |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lightrag  |   File "/root/.local/lib/python3.11/site-packages/httpcore/_async/connection.py", line 103, in handle_async_request
lightrag  |     return await self._connection.handle_async_request(request)
lightrag  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lightrag  |   File "/root/.local/lib/python3.11/site-packages/httpcore/_async/http11.py", line 136, in handle_async_request
lightrag  |     raise exc
lightrag  |   File "/root/.local/lib/python3.11/site-packages/httpcore/_async/http11.py", line 106, in handle_async_request
lightrag  |     ) = await self._receive_response_headers(**kwargs)
[...]

PR #1643, opened 2025-05-30
Fix: Increase Ollama timeout values to prevent ReadTimeout errors

eddiemaru-101:
resolve:  #1640

## 🐛 Problem
Users are experiencing `httpx.ReadTimeout` errors when using Ollama models, especially during entity extraction processes. The current timeout values (300s for completion, 90s for embedding) are insufficient for larger documents or slower models.

**Related Issue:** #1640

## 🔧 Solution
Increased default timeout values in `lightrag/llm/ollama.py`:
- **LLM completion timeout**: `300s` → `600s` (10 minutes)
- **Embedding timeout**: `90s` → `300s` (5 minutes)


LarFii:
Thanks for sharing!

PR #1755, opened 2025-07-08
options needs to be passed to ollama client embed() method

michele-comitini:
## Description

allow passing options to ollama embed() method.  This allows for setting correct context size by using the `num_ctx` parameter.
Without this llama process can crash as a consequence of buffer overflowing. 

## Related Issues

see: https://github.com/ollama/ollama/issues/7288#issuecomment-2591709109

## Changes Made

do a pop of  `options` element from `kwargs` and pass that to the ollama client `embed()` method as a named argument.
## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

It was tested with bge-m3 and bge-large embedding models.

### DIFF
--- a/lightrag/llm/ollama.py
+++ b/lightrag/llm/ollama.py
@@ -149,9 +149,11 @@ async def ollama_embed(texts: list[str], embed_model, **kwargs) -> np.ndarray:
     timeout = kwargs.pop("timeout", None) or 300  # Default time out 300s
 
     ollama_client = ollama.AsyncClient(host=host, timeout=timeout, headers=headers)
-
     try:
-        data = await ollama_client.embed(model=embed_model, input=texts)
+        options = kwargs.pop("options", {})
+        data = await ollama_client.embed(
+            model=embed_model, input=texts, options=options
+        )
         return np.array(data["embeddings"])
     except Exception as e:
         logger.error(f"Error in ollama_embed: {str(e)}")
--- a/lightrag/api/config.py
+++ b/lightrag/api/config.py
@@ -7,6 +7,7 @@ import argparse
 import logging
 from dotenv import load_dotenv
 from lightrag.utils import get_env_value
+from lightrag.llm.binding_options import BindingOptions
 
 from lightrag.constants import (
     DEFAULT_WOKERS,
@@ -207,6 +208,11 @@ def parse_args() -> argparse.Namespace:
         choices=["lollms", "ollama", "openai", "azure_openai"],
         help="Embedding binding type (default: from env or ollama)",
     )
+
+    # Automatically add binding options defined in binding_options module
+    # This will add command line arguments for all binding options (e.g., --option.ollama.num_ctx)
+    # and corresponding environment variables (e.g., OLLAMA_OPTION_NUM_CTX)
+    BindingOptions.add_args(parser)
 
     args = parser.parse_args()
 
--- a/lightrag/llm/binding_options.py
+++ b/lightrag/llm/binding_options.py
@@ -1,0 +1,415 @@
+"""
+Module that implements containers for specific LLM bindings.
+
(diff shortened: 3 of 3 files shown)


lightrag/llm/ollama.py
```diff
@@ -149,9 +149,9 @@ async def ollama_embed(texts: list[str], embed_model, **kwargs) -> np.ndarray:
     timeout = kwargs.pop("timeout", None) or 300  # Default time out 300s
 
     ollama_client = ollama.AsyncClient(host=host, timeout=timeout, headers=headers)
-
     try:
-        data = await ollama_client.embed(model=embed_model, input=texts)
+        options = kwargs.pop("options", {})
```
danielaskdd, line 153: The embedding_func  is missing  options in kwargs ?

michele-comitini: Not necessary if one uses the `func` parameter like this:
```python
embedding_func=EmbeddingFunc(
            embedding_dim=1024,
            max_token_size=512,
            func=lambda texts: ollama_embed(texts, embed_model="bge-large", options={"num_ctx": 512}),
        )
```

And IMHO it is only a problem of the ollama driver.

michele-comitini:
> I could not find the point at which the options parameter is passed to embedding_func.

options is only  passed to `ollama_embed()`, through a closure or partial function,  see an example of  `embedding_func` (`EmbeddingFunc` instance) in the inline comment

danielaskdd:
LightRag Server's embedding function is missing num_ctx option:

https://github.com/HKUDS/LightRAG/blob/1497f334f6f5d80784a292c07a8995012657f834/lightrag/api/lightrag_server.py#L273-L278

danielaskdd:
We need to add an EMBEDDING_NUM_CTX parameter to the .env file to ensure the LightRAG server correctly configures the embedding function.

michele-comitini:
> LightRag Server's embedding function is missing num_ctx option:
> 
> https://github.com/HKUDS/LightRAG/blob/1497f334f6f5d80784a292c07a8995012657f834/lightrag/api/lightrag_server.py#L273-L278

Oh I see now! 
I think there should a more general approach to ollama client options, and since we are there for the other drivers. It could be an Option class that is specialized for each driver.

```python
@dataclass
class DriverOptions:
.
.
.

@dataclass
class OllamaOptions(DriverOptions):
  num_ctx: int
  .
  .
  .
```
for environment then having a prefix could help making things distinct:
```
OLLAMA_OPTION_NUM_CTX
```

default of class `__init__` could be sourcing the environment values

also  command line args such as `--options.ollama.num_ctx=512` would be interpreted by `parse_args()` and used to override environment values. Then  inside the `args` object there would be always and `.options: [Options]` array property that will be useful to init the `EmbedFunc` in server's `create()`

@danielaskdd what do you think?

michele-comitini:
> . Then inside the `args` object there would be always and `.options: [Options]` array property that will be useful to init the `EmbedFunc` in server's `create()`
> 
well something more structured like a dictionary  could be cleaner

danielaskdd:
I believe you have a comprehensive understanding of the issue. I recommend modifying the code to incorporate appropriate environment variables and then updating the LightRAG Server to correctly configure the Ollama Embedding function.

michele-comitini:
OK

danielaskdd:
Has there been any progress on this PR?

michele-comitini:
No,sorry. I have been sick.  I will get to it today

michele-comitini:
> Has there been any progress on this PR?

@danielaskdd I added a new little module that helps managing binding options for Ollama and eventually for other bindings.

https://github.com/michele-comitini/LightRAG/blob/patch-1/lightrag/llm/binding_options.py

It is integrated with the global parsearg mechanism with a single line in `config.py`

PR #1768, opened 2025-07-11
fix(build): pyproject.toml setup

schmidt-marvin:
## Description
In the changes introduced from tag `1.3.9` to `1.3.10`, the setup of the package was changed from using `setup.py` to using  `pyproject.toml`. Within the pyproject.toml, there's a config error leading to only the `lightrag.api' package being installed.

## Related Issues
Installing any version of LightRAG >= 1.3.10 would only install lightrag.api, but not lightrag.kg, lightrag.llm and lightrag.tools.
image

This would in a regular use case crash the service:
image

## Changes Made
This change makes sure the automatic discovery functionality originally within `setup.py` also works in the `pyproject.toml` file.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,8 +82,10 @@ Documentation = "https://github.com/HKUDS/LightRAG"
 Repository = "https://github.com/HKUDS/LightRAG"
 "Bug Tracker" = "https://github.com/HKUDS/LightRAG/issues"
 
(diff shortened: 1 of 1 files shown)


danielaskdd:
Thanks for sharing

Issue #1800, opened 2025-07-17
[Bug]:AZURE_EMBEDDING_DEPLOYMEN different from model


Chida82:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

We are encountering an issue when attempting to insert text using the Azure OpenAI API. The configuration uses a deployment name that differs from the model name, which is expected behavior in Azure setups.
Here is how the environment variables are configured:
EMBEDDING_BINDING=azure_openai
EMBEDDING_MODEL=text-embedding-3-large
AZURE_EMBEDDING_DEPLOYMENT=embedding (this matches the deployment name used in the Azure portal and in the request URL)
Despite this, the API call fails with the following error:
"error": "Error code: 404 - {'error': {'code': 'DeploymentNotFound', 'message': 'The API deployment for this resource does not exist. If you created the deployment within the last 5 minutes, please wait a moment and try again.'}}"

We have verified that:
The deployment name is correct and active in the Azure portal.
The deployment has been created for more than 5 minutes.
The same deployment name works when tested via curl.

### Steps to reproduce

use azure_openai embending.

configure like this
### Embedding Binding type: openai, ollama, lollms, azure_openai
EMBEDDING_BINDING=azure_openai
EMBEDDING_MODEL=text-embedding-3-large
EMBEDDING_DIM=3072
EMBEDDING_BINDING_API_KEY=<add apikey>

AZURE_EMBEDDING_DEPLOYMENT=embedding
AZURE_EMBEDDING_API_VERSION=2023-05-15
AZURE_EMBEDDING_ENDPOINT=https://xxx.openai.azure.com/
AZURE_EMBEDDING_API_KEY=<add apikey>

use api to add text

### Expected Behavior

background index works

### Embedding Binding type: openai, ollama, lollms, azure_openai
EMBEDDING_BINDING=azure_openai
EMBEDDING_MODEL=text-embedding-3-large
EMBEDDING_DIM=3072
EMBEDDING_BINDING_API_KEY=<add apikey>

AZURE_EMBEDDING_DEPLOYMENT=embedding
AZURE_EMBEDDING_API_VERSION=2023-05-15
AZURE_EMBEDDING_ENDPOINT=https://xxx.openai.azure.com/
AZURE_EMBEDDING_API_KEY=<add apikey>

### Logs and screenshots

{
[...]

PR #1801, opened 2025-07-17
fix Azure deployment

Chida82:
## Description

###Purpose of the Change
The update introduced in the commit ensures proper handling of the distinction between model and azure_deployment when using Azure OpenAI APIs for embeddings.
###Why This Change Was Needed
Azure OpenAI requires both the model name (e.g., "text-embedding-3-small") and the deployment name (the name of the configured deployment resource in Azure). These two values can differ: the deployment is an alias defined in Azure that points to a specific model and is used to build the API endpoint.
###Previous Issue
In the earlier implementation, if a model value was provided, it was also used as the deployment. This caused errors when the deployment name in Azure differed from the model name. As a result, the deployment variable could be incorrectly set, leading to failed or misrouted API calls.
###Impact on AsyncAzureOpenAI
The AsyncAzureOpenAI class now accepts both model and azure_deployment as separate parameters. If not handled correctly, the API call may fail or use the wrong model.

## Related Issues

[#1800](https://github.com/HKUDS/LightRAG/issues/1800)

## Changes Made

By prioritizing the AZURE_EMBEDDING_DEPLOYMENT or AZURE_OPENAI_DEPLOYMENT environment variables, the update ensures correct override and disambiguation between model and deployment names—especially when they are identical—while the renaming of the variable improves clarity and intent.

## Checklist

- [ x] Changes tested locally
- [x ] Code reviewed

## Additional Notes

The fix is actually logical: the explicit value comes first, then the fallbacks.

### DIFF
--- a/lightrag/llm/azure_openai.py
+++ b/lightrag/llm/azure_openai.py
@@ -47,7 +47,7 @@ async def azure_openai_complete_if_cache(
     api_version: str | None = None,
     **kwargs,
 ):
-    model = model or os.getenv("AZURE_OPENAI_DEPLOYMENT") or os.getenv("LLM_MODEL")
+    deployment = os.getenv("AZURE_OPENAI_DEPLOYMENT") or model or os.getenv("LLM_MODEL")
     base_url = (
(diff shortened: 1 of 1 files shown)


danielaskdd:
Thanks for sharing.

Issue #1723, opened 2025-06-29
[Bug]: code example in front  page is wrong  use ainsert but not insert


Sandy4321:
### Do you need to file an issue?

- [ ] I have searched the existing issues and this bug is not already filed.
- [ ] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

code example in front  page is wrong 
RuntimeWarning: coroutine 'LightRAG.ainsert' was never awaited
  print(f"An error occurred: {e}")
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
INFO: Storage Initialization completed!
INFO: Storage Finalization completed!

you need to  use 
            await rag.ainsert(f.read())        
            but not await rag.insert(f.read())  

use new file
import os, asyncio, aiofiles
from lightrag import LightRAG, QueryParam
from lightrag.llm.openai import gpt_4o_mini_complete, openai_embed
from lightrag.kg.shared_storage import initialize_pipeline_status
from lightrag.utils import setup_logger
from dotenv import load_dotenv

load_dotenv(r"e:\api_keys_sander.env", override=True)
setup_logger("lightrag", level="INFO")

WORKING_DIR = "./rag_storage"
os.makedirs(WORKING_DIR, exist_ok=True)

async def initialize_rag() -> LightRAG:
    rag = LightRAG(
        working_dir=WORKING_DIR,
        embedding_func=openai_embed,
        llm_model_func=gpt_4o_mini_complete,
    )
    await rag.initialize_storages()
    await initialize_pipeline_status()
    return rag

async def main() -> None:
    rag = await initialize_rag()

    async with aiofiles.open(
        "./book.txt",
        "r", encoding="utf-8"
    ) as f:
        text = await f.read()

    await rag.ainsert(text)

    answer = await rag.aquery(
        "What are the top themes in this story?",
        param=QueryParam(mode="hybrid")
    )
    print(answer)

    await rag.finalize_storages()

if __name__ == "__main__":
    asyncio.run(main())

### LightRAG Config Used

# Paste your config here

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

Sandy4321:
LightRAG Version: latest 
Operating System: windows
Python Version: Python 3.10.11
Related Issues:

zztzztzztzzt:
I also encountered this problem. What solutions are there to fix it?

PR #1844, opened 2025-07-22
Fix "A Simple Program" example in README.md

AkosLukacs:
## Description

The example should use ainsert and aquery.

## Related Issues

#1723 

## Changes Made

Tiny change in the readme - asynchronous code does work


danielaskdd:
Thanks for sharing.

PR #1913, opened 2025-08-05
Feat: Change embedding formats from float to base64 for efficiency

danielaskdd:
Feat: Change embedding formats from float to base64 for efficiency

- Add base64 support for Jina embeddings
- Add base64 support for OpenAI embeddings
- Update env.example with new embedding options
- Improve Jina API error handling to show clean messages instead of HTML

### DIFF
--- a/env.example
+++ b/env.example
@@ -121,6 +121,16 @@ LLM_MODEL=gpt-4o
 LLM_BINDING_HOST=https://api.openai.com/v1
 LLM_BINDING_API_KEY=your_api_key
 
+### Optional for Azure
+# AZURE_OPENAI_API_VERSION=2024-08-01-preview
+# AZURE_OPENAI_DEPLOYMENT=gpt-4o
+
+### Openrouter example
+# LLM_MODEL=google/gemini-2.5-flash
+# LLM_BINDING_HOST=https://openrouter.ai/api/v1
+# LLM_BINDING_API_KEY=your_api_key
+# LLM_BINDING=openai
+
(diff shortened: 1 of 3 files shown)


lucky-verma:
@danielaskdd This PR introduced a regression for PostgreSQL storage backends. 

**Issue**: The change to `encoding_format="base64"` works for direct OpenAI API calls, but breaks when using PostgreSQL storage implementations (PGKVStorage, PGVectorStorage, PGGraphStorage) because they return embeddings as arrays rather than base64 strings.

**Evidence**: 
- v1.4.5: Works perfectly ✅
- v1.4.6+: `TypeError: argument should be a bytes-like object or ASCII string, not 'list'` ❌

**Impact**: Any users with PostgreSQL storage configurations are unable to process documents.

I can provide a fix that maintains backward compatibility. Should I create a separate issue to track this?

**Reproduction**
1. Use PostgreSQL storage configuration
2. Upload any document
3. Error occurs during embedding processing

**Root Cause Analysis**
PR #1913 changed OpenAI embeddings to request `encoding_format="base64"` but PostgreSQL storage implementations still return embeddings as arrays/lists, causing a format mismatch in the decoding logic.

danielaskdd:
Based on the error message, the issue arises because the embedding response returns a list instead of the expected base64-encoded string. This may be due to the model not adhering to the required format.

PR #1928 introduced a format compatibility check to address this. Please pull the latest code to verify whether the issue has been resolved. Thank you.

PR #2009, opened 2025-08-26
fix mismatch of 'error' and 'error_msg' in MongoDB

LinkinPony:
## Description

In this commit 6f958d5aeecc8735bd7e2db4ec43e5e3f0f136ef, the data field `error` was changed to `error_msg` without the right fallback in MongoDB. It will cause `ERROR: Error during migration check: DocProcessingStatus.__init__() got an unexpected keyword argument 'error'` if any record has a legacy field `error`.

## Related Issues

None.

## Changes Made

Add backward-compatible mapping: if `error` exists and `error_msg` is empty, move value to `error_msg`; otherwise drop `error`.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Maybe other impls have the same issue. I'm not testing them yet.

### DIFF
--- a/lightrag/kg/mongo_impl.py
+++ b/lightrag/kg/mongo_impl.py
@@ -280,6 +280,30 @@ class MongoDocStatusStorage(DocStatusStorage):
     db: AsyncDatabase = field(default=None)
     _data: AsyncCollection = field(default=None)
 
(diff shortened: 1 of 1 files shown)


danielaskdd:
Thanks for sharing

PR #2085, opened 2025-09-09
Feature predefined relationships

mertz1999:
## Description

Added pre defining relationships which can be important for some cases. it's completely like the entity_types  in terms of defenition, but I used relation_types

## Related Issues

Not issue I found

## Changes Made

lightrag/lightrag.py  : define input parameters setup for it.
lightrag/prompt.py  : define it's needed prompt.
lightrag/api/config.py :define basic configs.
lightrag/operate.py : operation of that logic 

## Checklist

- [x] Changes tested locally
- [x] Code reviewed


gurunathg:
Was about to raise a question for this exact feature. Will be very helpful to define custom relationships similar to custom entities 👍

mertz1999:
> Was about to raise a question for this exact feature. Will be very helpful to define custom relationships similar to custom entities 👍

Yes. In one of my projects I need to define relationship and entities to make sure that the system get enough information.

danielaskdd:
LightRAG does not utilize the "relationship type" feature in its query recall process, and I fail to see how this feature is helpful to LightRAG.

mertz1999:
> LightRAG does not utilize the "relationship type" feature in its query recall process, and I fail to see how this feature is helpful to LightRAG.

defining the relationship types helps model to make sure get specific types of relation that is important in some cases, for example in my project which is belong to a medical document retrieval project I need to define specific relation between drugs and diagnosis, I get your point that these types is not using in the retireval part but it will help model to first check those relations and then go to other types of relation that will be found.

danielaskdd:
"I understand your scenario. It's true that 'relationship type' can enhance query results in certain use cases. However, we still need to explore optimal strategies, such as standardizing relationship types, ensuring their reliable generation by LLMs, and how query should utilize them.  As LightRAG prepares to offer prompt template customization in next release, your input on these aspects would be highly valued.

Considering our current development roadmap, I believe the full integration of 'relationship type' functionality should be deferred until after LightRAG officially supports prompt template switching. Therefore, bringing this feature into the current Pull Request would be premature."

mertz1999:
> "I understand your scenario. It's true that 'relationship type' can enhance query results in certain use cases. However, we still need to explore optimal strategies, such as standardizing relationship types, ensuring their reliable generation by LLMs, and how query should utilize them. As LightRAG prepares to offer prompt template customization in next release, your input on these aspects would be highly valued.
> 
> Considering our current development roadmap, I believe the full integration of 'relationship type' functionality should be deferred until after LightRAG officially supports prompt template switching. Therefore, bringing this feature into the current Pull Request would be premature."

Thanks I got it the issue.

Issue #2365, opened 2025-11-16
[Bug]: dim error


gptmaas:
### Do you need to file an issue?

- [ ] I have searched the existing issues and this bug is not already filed.
- [ ] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

2025-11-16 20:25:14,788 [ERROR][handler]: RPC error: [upsert_rows], <MilvusException: (code=65535, message=the length(106496) of float data should divide the dim(3072))>, <Time:{'RPC start': '2025-11-16 20:25:14.209261', 'RPC error': '2025-11-16 20:25:14.788743'}> (decorators.py:140)

### LightRAG Config Used

# Paste your config here

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

EightyOliveira:
You need to ensure that the dimension of the Milvus collection exactly matches the dimension of the embedding vectors you are actually inserting.

Let LightRAG automatically create the table, or manually create it with the correct dimension.

danielaskdd:
The error message indicates inconsistent vector dimensions in the data being written to Milvus. This is most likely due to one of the vector dimension change from 3072 to 2048 (The remainder of 6496 divided by 3072 is 2048). Please check the backend logs for any errors or warnings related to the embedding service. 

Additionally, kindly provide your embedding model-related environment variables for further analysis.

PR #2368, opened 2025-11-17
Refact: Add Embedding Dimension Validation in EmbeddingFunc

danielaskdd:
### Problem Statement

When using custom OpenAI-compatible embedding endpoints (or other embedding providers), dimension mismatches between expected and actual embedding outputs could cause runtime errors at the storage layer. This was particularly problematic with Milvus, where errors like:

```
MilvusException: (code=65535, message=the length(106496) of float data should divide the dim(3072))
```

would occur only when data reached the vector database, making debugging difficult and potentially causing data corruption.

### Root Cause

Embedding providers may return results in varying formats or with incorrect dimensions due to:

- API misconfiguration
- Model version mismatches
- Data format inconsistencies (base64 vs raw arrays)
- Batch processing errors during concatenation

Previously, dimension validation only happened implicitly at the storage layer, meaning invalid embeddings could propagate through the system before being detected.

### Solution

This PR implements **centralized dimension validation** in the `EmbeddingFunc` class (`lightrag/utils.py`), ensuring all embedding results are validated immediately after generation, before reaching any storage backend.

#### Implementation Details

**Validation Logic** (in `EmbeddingFunc.__call__`):

```python
# Validate using total element count (efficient O(1) check)
total_elements = result.size
if total_elements % expected_dim != 0:
    raise ValueError(
        f"Embedding dimension mismatch detected: "
        f"total elements ({total_elements}) cannot be evenly divided by "
        f"expected dimension ({expected_dim})."
    )

    raise ValueError(
        f"Vector count mismatch: expected {expected_vectors} vectors "
        f"but got {actual_vectors} vectors."
    )
```

#### Key Advantages

1. **Early Detection** - Catches dimension errors at the source, not at storage time
2. **Universal Coverage** - Applies to ALL storage backends (Milvus, PostgreSQL, MongoDB, Neo4j, etc.)
3. **Performance** - Single modulo operation (`O(1)`) instead of shape inspections
4. **Shape Agnostic** - Works with any array shape (1D, 2D, flattened)
5. **Clear Error Messages** - Provides actionable debugging information

#### Modified Files

1. **`lightrag/utils.py`**
   - Added dimension validation in `EmbeddingFunc.__call__` method
   - Validates total elements divisibility by expected dimension
   - Validates vector count matches input text count

2. **`lightrag/kg/milvus_impl.py`**
   - Reverted to original `np.concatenate()` implementation
   - Removed temporary workarounds that were masking the root cause

### Benefits

✅ **Data Integrity** - Prevents invalid embeddings from entering the system  
✅ **Better Debugging** - Clear error messages at the point of failure  
✅ **Storage Agnostic** - Protects all vector storage implementations  
✅ **Zero Performance Impact** - Minimal overhead (single modulo check)  
✅ **Backward Compatible** - No breaking changes to existing functionality  

### Testing Recommendations

- ✅ Test with various embedding providers (OpenAI, custom endpoints, local models)
- ✅ Verify error messages are clear when dimension mismatches occur
- ✅ Confirm no performance regression in embedding generation
- ✅ Test with different batch sizes to ensure validation works correctly

### Breaking Changes

**None** - This change is fully backward compatible. Valid embeddings continue to work as before; only invalid embeddings now fail with clearer error messages.

### Migration Guide

No migration required. This change adds validation without modifying any APIs or data formats.

---

### Example Error Output

**Before this PR:**

```
MilvusException: (code=65535, message=the length(106496) of float data should divide the dim(3072))
```

**After this PR:**

```
ValueError: Embedding dimension mismatch detected: total elements (106496) 
cannot be evenly divided by expected dimension (3072).
```

Much clearer! 🎉

---

**Related Issues:** #2365
**Type:** Bug Fix / Improvement
**Component:** Core - Embedding System

### DIFF
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -363,6 +363,7 @@ class EmbeddingFunc:
         max_token_size: Optional token limit for the embedding model
         send_dimensions: Whether to inject embedding_dim as a keyword argument
     """
+
     embedding_dim: int
     func: callable
     max_token_size: int | None = None  # Token limit for the embedding model


danielaskdd:
@codex review

PR #2395, opened 2025-11-20
issue-2394: use deployment variable instead of model for embeddings API call

Amrit75:
Description

This pull request fixes a critical bug in azure_openai_embed() where the Azure embeddings API is called with the wrong model value.

The function correctly resolves the Azure deployment name into the deployment variable, but it mistakenly passes the unused model parameter (typically None) to embeddings.create(). This causes Azure to look for a deployment literally named "None" and results in a 404 DeploymentNotFound error.

This PR updates the call to:

model=deployment

which is the correct behavior and aligns with how Azure OpenAI routing works.

Related Issues

Fixes: #<insert your issue number here>
(If you haven’t created the issue yet, you can add it afterward.)

Changes Made

Updated lightrag/llm/azure_openai.py in azure_openai_embed():

Replaced model=model with model=deployment

Added in-line comment explaining why Azure requires model=deployment

Verified the change through local testing with Azure embeddings

Checklist

 Changes tested locally

 Code reviewed

 Documentation updated (not required for this change)

 Unit tests added (optional — no tests currently exist for this module)

Additional Notes

This fix resolves a blocking issue for all Azure OpenAI embedding users.

Without this change, embedding requests always fail with 404, even with correct Azure configuration.

Happy to make any adjustments requested by maintainers.

### DIFF
--- a/lightrag/llm/azure_openai.py
+++ b/lightrag/llm/azure_openai.py
@@ -172,6 +172,6 @@ async def azure_openai_embed(
     )
 
     response = await openai_async_client.embeddings.create(
-        model=model, input=texts, encoding_format="float"
+        model=deployment, input=texts, encoding_format="float"
     )
     return np.array([dp.embedding for dp in response.data])


danielaskdd:
@codex review

danielaskdd:
Using deployment directly as the model parameter leads to:

- The incoming model parameter being ignored (unless the environment variable AZURE_EMBEDDING_DEPLOYMENT is unset)
- Inconsistent handling compared to the azure_openai_complete_if_cache function in the same file

We should prioritize the `model` parameter if provided, falling back to the deployment name only when necessary. I have updated the code on the main branch—please verify that it behaves as expected.

```python
    response = await openai_async_client.embeddings.create(
        model=model or deployment, input=texts, encoding_format="float"
    )
```

PR #2432, opened 2025-11-28
Doc: Update README examples to prevent double-wrapping of embedding functions

danielaskdd:
### Fix: Update README examples to prevent double-wrapping of embedding functions

**Problem:**
The `EmbeddingFunc.__call__` method now contains processing logic that prevents functions from being wrapped twice. This causes issues when embedding functions already decorated with `@wrap_embedding_func_with_attrs` are wrapped again using `EmbeddingFunc()`.

**Root Cause:**
Since `openai_embed`, `ollama_embed`, and similar functions are already decorated with `@wrap_embedding_func_with_attrs`, wrapping them again with `EmbeddingFunc()` results in double decoration, which breaks the expected behavior.

**Solution:**
Update all README examples that incorrectly use `EmbeddingFunc()` to wrap pre-decorated embedding functions. The correct approach is to either:

1. Use `@wrap_embedding_func_with_attrs` decorator to create a custom embedding function that calls `xxx_embed.func` (the unwrapped function)
2. Pass the decorated embedding function directly to `LightRAG` without additional wrapping

**Before (Incorrect - Double Wrapping):**

```python
embedding_func=EmbeddingFunc(
    embedding_dim=768,
    func=lambda texts: ollama_embed(texts, ...)  # ollama_embed is already decorated!
)
```

**After (Correct):**

```python
@wrap_embedding_func_with_attrs(embedding_dim=768, max_token_size=8192)
async def embedding_func(texts: list[str]) -> np.ndarray:
    return await ollama_embed.func(texts, ...)  # Call .func to access the unwrapped function

rag = LightRAG(
    ...
    embedding_func=embedding_func,  # Pass directly, no EmbeddingFunc wrapper needed
)
```

**Key Points:**

- Call `xxx_embed.func` instead of `xxx_embed` to avoid triggering double decoration
- The `.func` attribute accesses the underlying function while preserving the built-in retry logic
- Pass decorated functions directly to `embedding_func` parameter without additional `EmbeddingFunc()` wrapper

### DIFF
--- a/README-zh.md
+++ b/README-zh.md
@@ -445,6 +445,10 @@ async def initialize_rag():
     return rag
 ```
 
+> **关于嵌入函数封装的重要说明：**
+>
+> `EmbeddingFunc` 不能嵌套封装。已经被 `@wrap_embedding_func_with_attrs` 装饰过的嵌入函数（如 `openai_embed`、`ollama_embed` 等）不能再次使用 `EmbeddingFunc()` 封装。这就是为什么在创建自定义嵌入函数时，我们调用 `xxx_embed.func`（底层未封装的函数）而不是直接调用 `xxx_embed`。
+
 </details>
 
 <details>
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #2585, opened 2026-01-15
Fix binding options not loaded from environment variables

danielaskdd:
### Summary

This PR fixes a critical bug where environment variables for LLM binding options (e.g., `OPENAI_LLM_TEMPERATURE`, `OPENAI_LLM_MAX_TOKENS`) were silently ignored, causing the server to always use default values.

### Problem

When using OpenAI/Azure OpenAI bindings, environment variables like `OPENAI_LLM_TEMPERATURE=0.7` had no effect. The startup log would show:

```
INFO: OpenAI LLM Options: {}
```

**Root Cause**: The `_GlobalArgsProxy` class did not properly support `vars()` calls. When `binding_options.py` called `vars(args)` on the proxy object, it returned an empty dict instead of the underlying `argparse.Namespace` attributes, causing all binding-specific options to be lost.

### Changes

**Commit 1: Refactor argument parsing** (`a000bdf0`)

- Consolidated duplicate `--llm-binding` and `--embedding-binding` parsing logic
- Added edge case handling for CLI arguments starting with `-`
- Improved code maintainability by reducing duplication

**Commit 2: Fix GlobalArgsProxy** (`a555690c`)

- Replaced `__getattr__` with `__getattribute__` to intercept `__dict__` access
- When `vars(proxy)` is called, it now correctly returns the underlying namespace's `__dict__`
- Enables `binding_options.options_dict()` to properly extract provider-specific options

### Testing

Verified with:

```python
from lightrag.api.config import global_args
from lightrag.llm.binding_options import OpenAILLMOptions

options = OpenAILLMOptions.options_dict(global_args)
print(options)  # {'temperature': 0.7, 'max_tokens': 9000}
```

### Impact

- **Before**: All OpenAI/Gemini/Ollama LLM binding options were ignored
- **After**: Environment variables are properly loaded and applied to LLM calls

### DIFF
--- a/lightrag/api/config.py
+++ b/lightrag/api/config.py
@@ -525,19 +525,44 @@ class _GlobalArgsProxy:
 
     This maintains backward compatibility with existing code while
     allowing programmatic control over initialization timing.
+
+    The proxy fully delegates to the underlying argparse.Namespace,
+    including support for vars() calls which is used by binding_options
+    to extract provider-specific configuration options.
     """
 
-    def __getattr__(self, name):
+    def __getattribute__(self, name):
+        """Override attribute access to support vars() and regular attribute access.
+
+        This method intercepts __dict__ access (used by vars()) and delegates
+        to the underlying _global_args namespace, ensuring binding options
+        can be properly extracted.
+        """
+        global _initialized, _global_args
+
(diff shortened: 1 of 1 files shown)


danielaskdd:
@codex review

PR #2599, opened 2026-01-17
fix: resolve entity resolution race condition in parallel document processing

cleo-ia:
## Summary
- Fix entity resolution race condition when `max_parallel_insert > 1`
- Add post-processing consolidation step after all documents are processed
- Fix tuple handling bug in edge reconnection (get_node_edges returns tuples not dicts)

## Problem
When documents are processed in parallel, entity resolution only sees entities from the current batch, not from other parallel batches. This causes duplicates like "THALIE" vs "Thalie" to persist.

## Solution
Added `consolidate_graph_entities()` function that runs AFTER all documents complete. It:
1. Gets all nodes from the graph
2. Groups by type and finds similar names using fuzzy matching
3. Merges duplicates under canonical names
4. Reconnects edges to canonical entities
5. Updates all storages (graph, VDB, entity_chunks, etc.)

## Test plan
- [x] All 85 entity resolution and conflict detection tests pass
- [x] Production testing showed 2153 → 1828 entities merged (325 duplicates fixed)
- [x] Fixed tuple handling bug caught in production logs

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -2733,11 +2733,14 @@ async def consolidate_graph_entities(
                         )
 
                 # Get edges connected to old node and reconnect to canonical
+                # get_node_edges returns list[tuple[str, str]] - (source_id, target_id)
                 old_edges = await knowledge_graph_inst.get_node_edges(old_name)
                 if old_edges:
-                    for edge in old_edges:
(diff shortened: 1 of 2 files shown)


cleo-ia:
Closed - PR created on wrong repository by mistake

PR #2633, opened 2026-01-26
perf: optimize get_knowledge_graph with stored procedure

cleo-ia:
## Summary

- Add support for PostgreSQL stored procedure `lightrag_get_knowledge_graph` for optimized graph retrieval
- Performs BFS server-side in a single round-trip, eliminating N+1 query problem
- Falls back to Python-based BFS if stored procedure is not available
- Supports all max_nodes values (250, 500, 1000, 3000, etc.)

## Changes

- New method `_get_knowledge_graph_stored_proc()` that calls the PostgreSQL function
- Existing BFS implementation moved to `_get_knowledge_graph_bfs()` as fallback
- Main `get_knowledge_graph()` method tries stored proc first, falls back gracefully

## Prerequisites

The stored procedure `lightrag_get_knowledge_graph(p_workspace, p_node_label, p_max_depth, p_max_nodes)` must be created in PostgreSQL/Supabase for optimization to take effect. Without it, the existing BFS fallback is used.

## Test plan

- [ ] Deploy to Render
- [ ] Verify graph displays correctly with stored procedure
- [ ] Check logs for "Stored proc returned X nodes" message
- [ ] Compare response times vs previous BFS implementation

### DIFF
--- a/lightrag/kg/pg_graph_simple_impl.py
+++ b/lightrag/kg/pg_graph_simple_impl.py
@@ -560,9 +560,11 @@ class PGGraphStorageSimple(BaseGraphStorage):
                     props = row["properties"]
                     if isinstance(props, str):
                         props = json.loads(props)
+                    # Extract entity_type for labels, default to "entity"
+                    entity_type = (props or {}).get("entity_type", "entity")
                     nodes_dict[node_id] = KnowledgeGraphNode(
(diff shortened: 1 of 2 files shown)


cleo-ia:
Opened on wrong repo by mistake

PR #2646, opened 2026-01-28
perf: reduce worker pool size and LLM timeout

cleo-ia:
## Summary

- **MAX_ASYNC**: 512 → 128 (reduces asyncio tasks from ~1600 to ~500 with 3 workspaces)
- **LLM_TIMEOUT**: 300 → 120 seconds (faster failure detection, less inertia)

## Problem

The worker pool creates persistent asyncio tasks per workspace that stay alive waiting for work. With:
- 512 LLM workers per workspace
- 32 embedding workers per workspace
- 3 active workspaces

This caused **1097+ asyncio tasks** even when idle (0 documents processing).

## Solution

Reduce `MAX_ASYNC` to 128 workers per workspace. This is still more than sufficient for concurrent LLM calls since provider rate limits kick in well before 128 concurrent requests.

| Config | Before | After | Impact |
|--------|--------|-------|--------|
| MAX_ASYNC | 512 | 128 | ~75% fewer worker tasks |
| LLM_TIMEOUT | 300s | 120s | Faster failure detection |

## Test plan

- [ ] Deploy and verify asyncio tasks metric is reduced
- [ ] Verify document processing still works correctly
- [ ] Monitor LLM latency with new timeout

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -99,7 +99,7 @@ services:
       - key: LLM_BINDING_API_KEY
         sync: false  # Set in Render Dashboard as secret
       - key: LLM_TIMEOUT
-        value: "300"
+        value: "120"
 
       #############################
       # Embedding Configuration (Vercel AI Gateway)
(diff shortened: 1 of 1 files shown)


cleo-ia:
Opened on wrong repo by mistake. Closing.

PR #2663, opened 2026-01-31
Postgress/pgvector backend now allows index large embeddings with HALFVEC

Daggle24:
## Description

When the embedding model has a size of more than 2000 dimensions (e.g., `text-embedding-3-large` with 3072d), the PostgreSQL `pgvector` plugin encounters an error when creating HNSW indexes: `column cannot have more than 2000 dimensions for hnsw index`.

This PR introduces support for `HALFVEC` (16-bit floats) for PostgreSQL vector storage, which allows for embedding dimensions up to 4000 for HNSW indexes.

## Issue

```bash
NFO: PostgreSQL: New table 'LIGHTRAG_VDB_CHUNKS_text_embedding_3_large_3072d' created successfully
INFO: Creating HNSW index idx_1645fa932b57_hnsw_cosine on table LIGHTRAG_VDB_CHUNKS_text_embedding_3_large_3072d
ERROR: PostgreSQL database,
sql:
                CREATE INDEX idx_1645fa932b57_hnsw_cosine
                ON LIGHTRAG_VDB_CHUNKS_text_embedding_3_large_3072d USING hnsw (content_vector vector_cosine_ops)
                WITH (m = 16, ef_construction = 200)
            ,
data:None,
error:column cannot have more than 2000 dimensions for hnsw index
ERROR: Failed to create vector index on table LIGHTRAG_VDB_CHUNKS_text_embedding_3_large_3072d, Got: column cannot have more than 2000 dimensions for hnsw index
INFO: Ensure this new table creation is caused by new workspace setup and not an unexpected embedding model change.
```

### PostgreSQL Implementation

- Added `HNSW_HALFVEC` as a supported vector index type in `PostgreSQLDB`.
- Modified table creation logic to use `HALFVEC` column type when `HNSW_HALFVEC` is selected.
- Updated index creation SQL to use `halfvec_cosine_ops`.
- Standardized the error message for unsupported vector types.

### Configuration

- Updated env.example to include `HNSW_HALFVEC` in the supported types.
- Added a recommendation to use `HNSW_HALFVEC` for embeddings with more than 2000 dimensions.

## How to Use

Set the following variable in the .env

```bash
POSTGRES_VECTOR_INDEX_TYPE=HNSW_HALFVEC
```

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated env.example

## Additional Notes

This change is essential for users working with large embedding models (like OpenAI's large models or Cohere's default settings 1024-4096d) when using the PostgreSQL backend.

### DIFF
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -3321,6 +3321,10 @@ class PGVectorStorage(BaseVectorStorage):
                         # Handle numpy arrays from pgvector
                         elif hasattr(vector_data, "tolist"):
                             vectors_dict[result["id"]] = vector_data.tolist()
+                        elif hasattr(vector_data, "to_list") and callable(
+                            vector_data.to_list
+                        ):
+                            vectors_dict[result["id"]] = vector_data.to_list()
                     except (json.JSONDecodeError, TypeError) as e:
                         logger.warning(
                             f"[{self.workspace}] Failed to parse vector data for ID {result['id']}: {e}"


danielaskdd:
There are conflicts to be resolved in the PR.

Daggle24:
> There are conflicts to be resolved in the PR.

Sorry, Is my first PR here, I forgot the pre-commit, it's now fixed.

hzywhite:
1. The query path is still hard-coded to `::vector`
The PR updates table creation and index creation to use `HALFVEC` / `halfvec_cosine_ops`, but the search SQL still uses:

- `content_vector <=> '[...]'::vector`

for relationships, entities, and chunks.

That means the new `HNSW_HALFVEC` path is not fully wired through. Even if the index can now be created successfully, queries are still not adapted to `halfvec`. Based on pgvector guidance, the search side also needs to cast appropriately in order to use the halfvec index. Otherwise this can fall back to a sequential scan, or potentially fail depending on operator/type compatibility in the target environment.

2. There is no pgvector extension version guard
`halfvec` support was added in pgvector 0.7.0. This PR introduces a new config option that depends on that capability, but the code only does `CREATE EXTENSION IF NOT EXISTS vector` and does not validate the installed extension version or provide a clear error path for older installations.

So on databases with an older pgvector extension, `POSTGRES_VECTOR_INDEX_TYPE=HNSW_HALFVEC` will fail at runtime when altering the column type or creating the index.

Also, the new tests only verify DDL generation. They do not cover the query path, which is where the main remaining gap is.

I think the feature direction is good, but I would recommend updating the query SQL for the `HNSW_HALFVEC` path and adding either a pgvector version check or a clearer failure message before merging.

Daggle24:
Thanks for the excellent feedback, the addressed issues were fixed in my last commit, previously we could've be falling back into sequential scans impacting the performance, now we make sure that the queries perform Index Scans with hnsw_cosine using the HNSW index.

lightrag/kg/postgres_impl.py
```diff
@@ -1584,11 +1620,15 @@ async def _create_vector_index(self, table_name: str, embedding_dim: int):
...
+        if self.vector_index_type == "HNSW_HALFVEC":
+            column_type = "HALFVEC"
+        else:
+            column_type = "VECTOR"
         try:
             vector_index_exists = await self.query(check_vector_index_sql)
             if not vector_index_exists:
                 # Only set vector dimension when index doesn't exist
-                alter_sql = f"ALTER TABLE {k} ALTER COLUMN content_vector TYPE VECTOR({embedding_dim})"
+                alter_sql = f"ALTER TABLE {k} ALTER COLUMN content_vector TYPE {column_type}({embedding_dim})"
```
danielaskdd, line 1643: When `setup_table()` runs on an existing PostgreSQL collection, it always calls `_create_vector_index()`. In the `HNSW_HALFVEC` path this only checks for the new `..._hnsw_halfvec_cosine` name, then immediately executes `ALTER TABLE ... ALTER COLUMN content_vector TYPE HALFVEC(...)`. Any pre-existing `HNSW`/`IVFFLAT` index on `content_vector` is left in place, so deployments switching from `HNSW` or `IVFFLAT` to `HNSW_HALFVEC` will try to keep a `vector_*_ops` index on a `halfvec` column and fail before the replacement index is created.

danielaskdd:
This change starts creating real `HALFVEC` columns, but the migration guard in `setup_table()` still infers `legacy_dim` only from `list`/`tuple`, objects with `__len__`, or strings. With pgvector's asyncpg codec, selecting from a `halfvec` column returns a `HalfVector`, which does not satisfy any of those branches, so `legacy_dim` stays `None`. As a result, model upgrades from one halfvec-backed table to another can skip the explicit dimension-mismatch check and fail later with a generic migration/insert error instead of the intended early `DataMigrationError`.

Daggle24:
Thank for the feedback and very nice catch, here is the summary of the fixes:

**Index switching conflict fix:** 
```python
# All known vector index suffixes, used to drop conflicting indexes when switching types
_VECTOR_INDEX_SUFFIXES = [
    "hnsw_cosine",
    "hnsw_halfvec_cosine",
    "ivfflat_cosine",
    "vchordrq_cosine",
]
```

Added a _VECTOR_INDEX_SUFFIXES constant listing all known vector index suffixes. In _create_vector_index(), before the ALTER + CREATE INDEX block, the method now iterates over all other suffixes and issues DROP INDEX IF EXISTS for each one. This ensures that switching from any index type (HNSW, IVFFLAT, VCHORDRQ) to another (e.g. HNSW_HALFVEC) won't fail due to a leftover index on the old column type.

```python
if self.vector_index_type == "HNSW_HALFVEC":
            column_type = "HALFVEC"
        else:
            column_type = "VECTOR"
        try:
            vector_index_exists = await self.query(check_vector_index_sql)
            if not vector_index_exists:
                for suffix in _VECTOR_INDEX_SUFFIXES:
                    if suffix == index_suffix:
                        continue
                    old_name = _safe_index_name(k, suffix)
                    await self.execute(f"DROP INDEX IF EXISTS {old_name}")
                alter_sql = f"ALTER TABLE {k} ALTER COLUMN content_vector TYPE {column_type}({embedding_dim})"
                await self.execute(alter_sql)
```

**HalfVector dimension detection:**

Added a new branch in setup_table()'s legacy dimension check (after __len__ check, before str check) that handles objects with a dimensions() method -- which is how pgvector's HalfVector class exposes its size. This ensures dimension mismatch is properly detected when migrating between halfvec-backed tables.

Something simmilar was happening with:

**tolist() method**

```python
 # Handle numpy arrays from pgvector
                        elif hasattr(vector_data, "tolist"):
                            vectors_dict[result["id"]] = vector_data.tolist()
                        elif hasattr(vector_data, "to_list") and callable(
                            vector_data.to_list
                        ):
                            vectors_dict[result["id"]] = vector_data.to_list()
 ```
**Added 4 new tests:**

test_create_vector_index_drops_old_indexes_when_switching -- verifies old indexes are dropped when switching to HNSW_HALFVEC
test_create_vector_index_no_drop_when_index_exists -- verifies no drops/creates when the target index already exists
test_setup_table_detects_halfvector_dimension_mismatch -- verifies DataMigrationError is raised when a HalfVector column has mismatched dimensions
test_setup_table_accepts_matching_halfvector_dimension -- verifies no error when HalfVector dimensions match

danielaskdd:
@codex review

PR #2673, opened 2026-02-05
Fix deletion cleanup for failed documents

### LINKED ISSUE #2442: [Bug]: Document Deletion Not Cleaning Up Associated Chunks and LLM Cache Leading to Re-ingestion Failures

## Bug Description

I'm encountering a critical issue with LightRAG's document deletion and re-ingestion workflow when using Neo4j as the graph storage backend. When a document fails during the merging stage due to corrupted LLM responses (specifically a `CypherSyntaxError`), deleting and re-ingesting the document does not resolve the issue because:

1. **Chunks are not deleted** when the parent document is deleted
2. **Corrupted LLM cache persists** and gets reused during re-ingestion
3. This creates a cycle where the same error occurs repeatedly, preventing successful document ingestion

### Initial Ingestion Error

The document fails during the merging stage with a `CypherSyntaxError` related to invalid entity type handling:

```python
Error processing entity `projectId`: RetryError[<Future at 0xeb19c0d75f70 state=finished raised CypherSyntaxError>]

neo4j.exceptions.CypherSyntaxError: {neo4j_code: Neo.ClientError.Statement.SyntaxError} 
{message: Invalid input 'attribute': expected 'FOREACH', ',', ':', 'ORDER BY', 'CALL', 'CREATE', 'LOAD CSV', 'DELETE', 'DETACH', 'FINISH', 'INSERT', 'LIMIT', 'MATCH', 'MERGE', 'NODETACH', 'OFFSET', 'OPTIONAL', 'REMOVE', 'RETURN', 'SET', 'SKIP', 'UNION', 'UNWIND', 'USE', 'WITH' or <EOF> (line 4, column 29 (offset: 131))
"                    SET n:``attribute``"
                             ^}

tenacity.RetryError: RetryError[`projectId`: <Future at 0xeb19c0d75f70 state=finished raised CypherSyntaxError>]
Merging stage failed in document 2/3: ENG-82583
```

The root cause appears to be that the LLM extracted `attribute` as an entity type, which Neo4j cannot accept as a valid label due to it being a reserved keyword or invalid identifier.

### Deletion Behavior

When attempting to delete the failed document:

```
Starting document deletion process
LLM cache cleanup requested for this deletion job
Deleting document 1/1: doc-2b11ded29e2288ac1cf5ec8957803b6c
Starting deletion process for document doc-2b11ded29e2288ac1cf5ec8957803b6c
Deleting doc-2b11ded29e2288ac1cf5ec8957803b6c ENG-82583(previous status: FAILED)
Document deleted without associated chunks: doc-2b11ded29e2288ac1cf5ec8957803b6c
Document deleted 1/1: doc-2b11ded29e2288ac1cf5ec8957803b6c[ENG-82583]
File deletion skipped, missing or unsafe file: ENG-82583
Deletion completed: 1 successful, 0 failed
```

**Key observations:**
- The log message "Document deleted without associated chunks" indicates that chunks were not found/deleted
- Despite "LLM cache cleanup requested," the corrupted LLM extraction results persist

### Re-ingestion Problem

When re-ingesting the same document after deletion, the exact same `CypherSyntaxError` occurs, suggesting that:
- The LLM cache contains the corrupted entity extraction (`attribute` as entity type)
- Orphaned chunks from the previous failed ingestion may still exist in the vector store
- The system reuses this cached data instead of performing fresh extraction

## Steps to Reproduce

1. Ingest a document that triggers LLM to extract invalid entity types (e.g., reserved keywords like `attribute`)
2. Document fails during merge stage with `CypherSyntaxError`
3. Delete the failed document using the deletion API/functionality
4. Attempt to re-ingest the same document
5. Observe that the same error occurs, indicating cached corrupted data is being reused

## Expected Behavior

When deleting a failed document, the system should:

1. **Delete all associated chunks** from the vector storage
2. **Clear LLM cache entries** specifically related to that document (extraction results, embeddings cache)
3. **Remove any partial graph entities/relationships** created before the failure
4. Allow fresh re-ingestion without reusing corrupted cached data

## Current Workaround Needed

Currently, the only way to resolve this is to:
- Clear the entire LLM cache (affecting all documents)
- Delete all documents and start fresh
- Manually clean up orphaned chunks and cache entries

This is not practical for production systems with large knowledge bases where only specific documents need to be re-ingested.

## Questions

1. **Is there a way to selectively clear LLM cache for specific documents** without wiping the entire cache?
2. **Why are chunks not being deleted** when their parent document is deleted? Is this a bug or expected behavior?
3. **How can I handle LLM extraction errors** (like reserved keywords as entity types) more gracefully to prevent ingestion failures?
4. **Is there a cleanup utility** to identify and remove orphaned chunks that no longer have parent documents?
5. **Should there be validation** of extracted entity types against Neo4j reserved keywords before attempting to create nodes?

## Additional Context

This issue significantly impacts production deployments where document re-ingestion is necessary due to:
- LLM extraction improvements
- Configuration changes
- Initial ingestion failures
- Document updates

The inability to cleanly delete and re-ingest specific documents without affecting the entire knowledge base limits the system's maintainability and reliability.

***

Would appreciate any guidance on resolving this without full cache/document wipeout. Happy to provide additional logs or test cases if needed.

liqiongyu:
Fixes #2442

### What
- Preserve doc_status.chunks_list/chunks_count when a document fails during extract or merge, so later deletion can remove associated chunks.
- Add offline regression tests covering extract-failure and merge-failure deletion + optional per-doc LLM cache cleanup.

### Why
Some DocStatusStorage adapters default missing chunks_list to an empty list on upsert. When a doc fails, the FAILED status update previously omitted chunks_list, which could overwrite a previously-recorded chunk list. Deletion then saw no chunks and skipped chunk cleanup.

### Tests
- uv run ruff check lightrag/lightrag.py tests/test_document_deletion_cleanup.py
- uv run pytest -m offline tests/test_document_deletion_cleanup.py

### DIFF
--- a/lightrag/base.py
+++ b/lightrag/base.py
@@ -400,6 +400,19 @@ class BaseKVStorage(StorageNameSpace, ABC):
             bool: True if storage contains no data, False otherwise
         """
 
+    async def get_ids_by_doc_id(self, doc_id: str) -> list[str]:
+        """Get record IDs associated with a document.
+
+        This is an optional capability used by deletion cleanup to discover chunk
+        IDs when a document status record does not contain `chunks_list`.
+
+        Implementations are expected to return IDs where the stored record has
+        `full_doc_id == doc_id` (commonly for `text_chunks` storage).
+        """
+        raise NotImplementedError(
+            f"{self.__class__.__name__} does not support get_ids_by_doc_id"
+        )
+
 
 @dataclass
 class BaseGraphStorage(StorageNameSpace, ABC):
--- a/lightrag/kg/json_kv_impl.py
+++ b/lightrag/kg/json_kv_impl.py
@@ -134,6 +134,14 @@ class JsonKVStorage(BaseKVStorage):
                     results.append(None)
             return results
 
+    async def get_ids_by_doc_id(self, doc_id: str) -> list[str]:
+        async with self._storage_lock:
+            return [
+                record_id
+                for record_id, record in self._data.items()
+                if isinstance(record, dict) and record.get("full_doc_id") == doc_id
+            ]
+
     async def filter_keys(self, keys: set[str]) -> set[str]:
         async with self._storage_lock:
             return set(keys) - set(self._data.keys())
--- a/lightrag/kg/mongo_impl.py
+++ b/lightrag/kg/mongo_impl.py
@@ -169,6 +169,10 @@ class MongoKVStorage(BaseKVStorage):
             ordered_results.append(doc_map.get(str(id_value)))
         return ordered_results
 
+    async def get_ids_by_doc_id(self, doc_id: str) -> list[str]:
+        cursor = self._data.find({"full_doc_id": doc_id}, {"_id": 1})
+        return [str(item["_id"]) async for item in cursor]
+
     async def filter_keys(self, keys: set[str]) -> set[str]:
         cursor = self._data.find({"_id": {"$in": list(keys)}}, {"_id": 1})
         existing_ids = {str(x["_id"]) async for x in cursor}
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -2149,6 +2149,25 @@ class PGKVStorage(BaseKVStorage):
 
         return _order_results(results)
 
+    async def get_ids_by_doc_id(self, doc_id: str) -> list[str]:
+        if not is_namespace(self.namespace, NameSpace.KV_STORE_TEXT_CHUNKS):
+            return []
+
+        table_name = namespace_to_table_name(self.namespace)
+        sql = f"SELECT id FROM {table_name} WHERE workspace=$1 AND full_doc_id=$2"
+        params = {"workspace": self.workspace, "full_doc_id": doc_id}
+        try:
+            rows = await self.db.query(sql, list(params.values()), multirows=True)
+        except Exception as e:
+            logger.error(
+                f"[{self.workspace}] PostgreSQL database,\nsql:{sql},\nparams:{params},\nerror:{e}"
+            )
+            raise
+
+        if not rows:
+            return []
+        return [str(row.get("id")) for row in rows if row and row.get("id") is not None]
+
     async def filter_keys(self, keys: set[str]) -> set[str]:
         """Filter out duplicated content"""
         if not keys:
--- a/lightrag/kg/redis_impl.py
+++ b/lightrag/kg/redis_impl.py
@@ -304,6 +304,50 @@ class RedisKVStorage(BaseKVStorage):
                 logger.error(f"[{self.workspace}] JSON decode error in batch get: {e}")
                 return [None] * len(ids)
 
+    @redis_retry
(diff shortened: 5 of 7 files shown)


danielaskdd:
@code review

danielaskdd:
@codex review

danielaskdd:
## Suggestion: Look up chunks from text_chunks storage instead of adding chunks_list to doc_status
                                                                                                                                                                       
Thanks for this PR — the set → list type fix is critical for PostgreSQL users (asyncpg's ANY($2) doesn't accept set), and the chunks: dict = {} initialization  correctly prevents UnboundLocalError.                                                                                                                                
                                                                                                                                                                       
However, I'd like to suggest an alternative approach for the orphaned chunks cleanup problem.

Observation: During document processing, chunks are already persisted to text_chunks and chunks_vdb in Stage 1 (lines 1909-1914), before entity extraction begins.  Each chunk carries a full_doc_id field linking it back to the document. So when extraction or merge fails later, the chunks are already in storage and can be  discovered by querying on full_doc_id.

Suggestion: Instead of storing a chunks_list field in doc_status, query text_chunks storage by full_doc_id at deletion time to discover the chunk IDs.

Advantages:
1. Backward compatibility — Works for FAILED documents created before this fix, which won't have chunks_list in their status record
2. Data consistency — Queries what actually exists in storage, rather than relying on a snapshot that may be stale (e.g., if chunk upsert partially failed)
3. Simpler error handling — No need for the extra doc_status.upsert inside the except blocks, avoiding the risk of the upsert failing and masking the original error
4. Less storage overhead — No need to redundantly store the full chunk ID list in doc_status

What this would require: A get_ids_by_doc_id(doc_id) method on BaseKVStorage (or equivalent). For PG this is a simple query:
```
SELECT id FROM LIGHTRAG_DOC_CHUNKS WHERE workspace=$1 AND full_doc_id=$2
```

For the JSON backend it's a dict scan filtering on full_doc_id.

Worth keeping from this PR: The set → list conversion (lines 3436-3438) and the chunks: dict = {} initialization (line 1789) are still valuable regardless of  approach.

liqiongyu:
Thanks for the thoughtful suggestion — agreed.

I kept the set → list conversion (PostgreSQL `ANY($2)` compatibility) and the `chunks: dict = {}` init, but I also implemented the deletion-time fallback you described:

- If `doc_status.chunks_list` is missing/empty, `adelete_by_doc_id()` now discovers chunk IDs by querying `text_chunks` for records where `full_doc_id == doc_id`, then deletes those chunks from both `text_chunks` and `chunks_vdb`.
- Added an optional `BaseKVStorage.get_ids_by_doc_id(doc_id)` capability and implemented it for the JSON / PostgreSQL / MongoDB / Redis KV backends (for `text_chunks`).
- Added a regression test that simulates a legacy FAILED status overwrite that drops `chunks_list`, and verifies deletion still removes orphaned chunks.

The PR branch has been updated (HEAD `8236b1b4`). Note: the latest Actions runs are currently `action_required` (fork PR policy), so a maintainer will need to approve/re-run workflows to validate CI.

PR #2703, opened 2026-02-22
fix(webui): make build runtime-agnostic by fixing Bun-only imports in…

Pranavh-2004:
Description

WebUI build fails silently (exit code 1 or 127) in environments where Bun is unavailable, incompatible, or not on PATH. Two root causes in vite.config.ts make the build Bun-only despite Vite itself being runtime-agnostic:
	•	@/lib/constants path alias import — Node.js cannot resolve tsconfig path aliases at config load time; Bun masks this by resolving them natively
	•	import.meta.env usage — only populated inside Bun’s runtime; Node.js leaves it undefined, crashing the config

This PR fixes both issues and makes the default build scripts runtime-agnostic so npm run build works out of the box.

Related Issues

Fixes: [Bug]: bun run build error: script “build” exited with code 1 (Nov 2025)
Affects users on CentOS/older Linux, restricted temp dir permissions, or environments where Bun is not installed

Changes Made

vite.config.ts
	•	Replace import { webuiPrefix } from '@/lib/constants' with relative import './src/lib/constants' — the @ alias is defined inside this config, so it can’t resolve itself
	•	Replace import.meta.env references with Vite’s official loadEnv() API using functional defineConfig(({ mode }) => ...) form
	•	Import loadEnv from vite

package.json
	•	Make default scripts runtime-agnostic: "build": "vite build" instead of "build": "bunx --bun vite build"
	•	Rename old *-no-bun scripts to *:bun variants (build:bun, dev:bun, preview:bun) for explicit Bun runtime usage
	•	dev, build, preview, lint now work with both bun run and npm run

README.md
	•	Add Using Node.js / npm (alternative) installation section
	•	Add troubleshooting section for common build failures
	•	Update script commands table to show both bun and npm usage

FrontendBuildGuide.md
	•	Update “Key Principle” to note npm is fully supported as fallback
	•	Expand FAQ: Can I use npm? → Yes, with instructions
	•	Add FAQ entry for Cannot find package '@/lib' error

Checklist
	•	Changes tested locally
	•	Code reviewed
	•	Documentation updated (if necessary)
	•	Unit tests added (if applicable)

Additional Notes

Verified all three build paths succeed after changes: bun run build, npm run build, npx vite build
No new lint errors introduced (pre-existing lint warnings are unrelated)
Tests (bun test) still require Bun — only build/dev/preview scripts are made agnostic
Zero architecture risk: only config-level changes, no application code modified

### DIFF
--- a/docs/FrontendBuildGuide.md
+++ b/docs/FrontendBuildGuide.md
@@ -8,7 +8,7 @@ The LightRAG project includes a React-based WebUI frontend. This guide explains 
 
 - **Git Repository**: Frontend build results are **NOT** included (kept clean)
 - **PyPI Package**: Frontend build results **ARE** included (ready to use)
-- **Build Tool**: Uses **Bun** (not npm/yarn)
+- **Build Tool**: **Bun** is recommended, but **Node.js/npm** is fully supported as a fallback
 
 ## Installation Scenarios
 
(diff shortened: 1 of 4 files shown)


danielaskdd:
@codex review

PR #2725, opened 2026-02-28
perf: use deque for BFS queue in get_knowledge_subgraph()

giulio-leone:
## Problem

`get_knowledge_subgraph()` in `NetworkXStorage` uses `list.pop(0)` to drain the BFS queue, which is **O(n)** per removal. For large knowledge graphs with many nodes, the traversal becomes **O(n²)**.

## Solution

Switch to `collections.deque` with `.popleft()` for **O(1)** front removal while keeping identical `.append()` semantics.

## Changes

- `lightrag/kg/networkx_impl.py`:
  - Import `deque` from `collections`
  - Initialize queue as `deque([...])` instead of list
  - `.pop(0)` → `.popleft()`

## Testing

- Module imports cleanly
- Syntax verified via `ast.parse()`

### DIFF
--- a/lightrag/kg/networkx_impl.py
+++ b/lightrag/kg/networkx_impl.py
@@ -1,4 +1,5 @@
 import os
+from collections import deque
 from dataclasses import dataclass
 from typing import final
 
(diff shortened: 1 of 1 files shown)


giulio-leone:
Friendly ping — CI is green and this is ready for review. Happy to address any feedback. Thanks!

danielaskdd:
@codex review

PR #2731, opened 2026-03-01
fix: add opt-in allow_extra_vectors flag for vector count mismatch

### LINKED ISSUE #2549: [Bug]: Bug in utils.py

### Describe the bug

When indexing a PDF using LightRAG with RAGAnything the system crashes because of a bug in utils.py.

Before the Dec 22 update to utils.py everything was working perfectly.

"After the Dec 22 update of utils.py, the strict validation causes a ValueError when embedding results return double the expected vectors. This occurs frequently with multimodal parsers (like MinerU) where text and metadata may be triggering a dual-pass in the OpenAI utility.

ERROR: Embedding func: Error in decorated function for task 140018899390464_373646.95697091: Vector count mismatch: expected 6 vectors but got 12 vectors (from embedding result).

With the help of a.i., the following snippet in utils.py was swapped out with the following working code and then indexing proceeded without error but with warnings raised by the new working code.

**********offending code in utils.py**************
        # actual_vectors = total_elements // expected_dim
        # if args and isinstance(args[0], (list, tuple)):
        #     expected_vectors = len(args[0])
        #     if actual_vectors != expected_vectors:
        #         raise ValueError(
        #             f"Vector count mismatch: "
        #             f"expected {expected_vectors} vectors but got {actual_vectors} vectors (from embedding result)."
        #         )

**************replacement code runs with warnings***************
        if args and isinstance(args[0], (list, tuple)):
            expected_vectors = len(args[0])
                # WORKAROUND: Handle the doubling bug in multimodal/mineru pipelines
                if actual_vectors == 2 * expected_vectors:
                    logger.warning(f"Vector count doubling detected ({actual_vectors} vs {expected_vectors}). Slicing result to match.")
                    # Reshape if it's a 2D array, slice, then return
                    if len(result.shape) > 1:
                        return result[:expected_vectors]
                    else:
                        # Fallback for flat arrays
                        new_size = expected_vectors * expected_dim
                        return result[:new_size]
                else:
                    raise ValueError(
                        f"Vector count mismatch: "
                        f"expected {expected_vectors} vectors but got {actual_vectors} vectors (from embedding result)."
                    )

***********This is the warning raised by the new code as it manages the error**************
WARNING: Vector count doubling detected (2 vs 1). Slicing result to match.

### Steps to reproduce

This is the python script I was running which used RAGAnything to index a PDF with lots of tables and images.
https://github.com/johnshearing/LightRAG/blob/main/jrs/_1_raganything_test.py

Br1an67:
## Description

When embedding functions return more vectors than expected (common with multimodal parsers like MinerU), the strict validation in `EmbeddingFunc.__call__` raises a `ValueError` that crashes the entire pipeline.

Instead of silently slicing (which risks silent data corruption if vectors are reordered/interleaved), this adds an explicit **opt-in flag** so the behavior is auditable:

- **Default** (`allow_extra_vectors=False`): raises `ValueError` with a detailed message naming the provider, batch size, and actual vs expected count. The error suggests setting `allow_extra_vectors=True` if appropriate.
- **Opt-in** (`allow_extra_vectors=True`): logs a warning and slices to expected count, for providers known to safely append extra padding vectors.

## Related Issues

Fixes #2549

## Changes Made

- **`lightrag/utils.py`**: Added `allow_extra_vectors: bool = False` field to `EmbeddingFunc`. Vector count mismatch now raises `ValueError` by default with detailed diagnostics. When opted in, reshapes to 2D and slices overflow vectors.
- **`tests/test_vector_count_mismatch.py`**: 5 tests covering default raise, opt-in slicing, underflow always raises, match passes, and error message content.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Addresses review feedback: the core approach is now **fail-fast by default** with an explicit opt-in escape hatch, rather than silently slicing at the shared `EmbeddingFunc.__call__` layer. This eliminates the silent data corruption risk across all six storage backends while still providing a documented path for providers known to return extra padding vectors.

### DIFF
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -436,6 +436,9 @@ class EmbeddingFunc:
         max_token_size: Enable embedding token limit checking for description summarization(Set embedding_token_limit in LightRAG)
         send_dimensions: Whether to inject embedding_dim argument to underlying function
         model_name: Model name for implementing workspace data isolation in vector DB
+        allow_extra_vectors: When True, silently slice overflow vectors instead of raising.
+            Only enable this for providers known to safely append extra padding vectors
+            (e.g. multimodal parsers that split images into extra embeddings). Default False.
     """
 
     embedding_dim: int
@@ -445,6 +448,7 @@ class EmbeddingFunc:
     model_name: str | None = (
         None  # Model name for implementing workspace data isolation in vector DB
     )
+    allow_extra_vectors: bool = False
 
     def __post_init__(self):
         """Unwrap nested EmbeddingFunc to prevent double wrapping issues.
@@ -514,15 +518,29 @@ class EmbeddingFunc:
                 f"expected dimension ({expected_dim}). "
             )
 
-        # Optional: Verify vector count matches input text count
+        # Verify vector count matches input text count
         actual_vectors = total_elements // expected_dim
         if args and isinstance(args[0], (list, tuple)):
             expected_vectors = len(args[0])
             if actual_vectors != expected_vectors:
-                raise ValueError(
-                    f"Vector count mismatch: "
-                    f"expected {expected_vectors} vectors but got {actual_vectors} vectors (from embedding result)."
-                )
+                provider = self.model_name or "unknown"
+                if actual_vectors > expected_vectors and self.allow_extra_vectors:
+                    logger.warning(
+                        f"Vector count mismatch (provider={provider}): "
+                        f"expected {expected_vectors} vectors for "
+                        f"{expected_vectors} inputs but got {actual_vectors}. "
+                        f"Slicing to first {expected_vectors} vectors "
+                        f"(allow_extra_vectors=True)."
+                    )
+                    result = result.reshape(-1, expected_dim)[:expected_vectors]
+                else:
+                    raise ValueError(
+                        f"Vector count mismatch (provider={provider}): "
+                        f"expected {expected_vectors} vectors for "
+                        f"{expected_vectors} inputs but got {actual_vectors}. "
+                        f"If this provider is known to return extra padding "
+                        f"vectors, set allow_extra_vectors=True on EmbeddingFunc."
+                    )
 
         return result
 
--- a/tests/test_vector_count_mismatch.py
+++ b/tests/test_vector_count_mismatch.py
@@ -1,0 +1,94 @@
+"""
+Tests for vector count mismatch handling in EmbeddingFunc.
+
+Verifies that vector count mismatches raise ValueError by default,
+and that the opt-in allow_extra_vectors flag enables safe slicing
+for providers known to return extra padding vectors.
+"""
+
+import numpy as np
+import pytest
+
+from lightrag.utils import EmbeddingFunc
+
+
+@pytest.mark.offline
+@pytest.mark.asyncio
+async def test_overflow_raises_by_default():
+    """Overflow raises ValueError when allow_extra_vectors is False (default)."""
+
+    async def mock_embed(texts, **kwargs):
+        dim = kwargs.get("embedding_dim", 128)
+        return np.random.rand(len(texts) * 2, dim).astype(np.float32)
+
+    func = EmbeddingFunc(embedding_dim=128, func=mock_embed, model_name="test-model")
+    with pytest.raises(ValueError, match="allow_extra_vectors"):
+        await func(["hello", "world"])
+
+
+@pytest.mark.offline
+@pytest.mark.asyncio
+async def test_overflow_slices_with_opt_in():
+    """Overflow slices to expected count when allow_extra_vectors=True."""
+
+    async def mock_embed(texts, **kwargs):
+        dim = kwargs.get("embedding_dim", 128)
+        return np.random.rand(len(texts) * 2, dim).astype(np.float32)
+
+    func = EmbeddingFunc(
+        embedding_dim=128,
+        func=mock_embed,
+        model_name="test-model",
+        allow_extra_vectors=True,
+    )
+    result = await func(["hello", "world"])
+    assert result.shape == (2, 128)
+
+
+@pytest.mark.offline
+@pytest.mark.asyncio
+async def test_underflow_raises_even_with_opt_in():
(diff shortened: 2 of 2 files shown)


danielaskdd:
Findings

1. High: vector-count underflow is now silently accepted, which can crash or corrupt upserts downstream.
    - In [lightrag/utils.py (line 521)], any mismatch (actual_vectors != expected_vectors) now goes through “slice to expected count”.
    - If actual_vectors < expected_vectors, slicing does nothing meaningful and returns too few vectors instead of failing fast.
    - Multiple backends assume strict 1:1 and index by i, so this can raise runtime errors or inconsistent behavior
    - Suggested fix: only slice when actual_vectors > expected_vectors; keep raising ValueError when actual_vectors < expected_vectors.

2. Medium: tests cover only overflow mismatch, not underflow mismatch.
    - New tests in [tests/test_vector_count_mismatch.py] validate “too many vectors” slicing, but there is no test asserting failure on “too few vectors”.
    - This gap would miss the regression above.

Br1an67:
Thanks for the review! Both points are spot-on — pushed a fix in 3ec1f621:

1. **Underflow now raises `ValueError`** — slicing only happens when `actual_vectors > expected_vectors` (overflow). When fewer vectors are returned, we fail fast since backends assume strict 1:1 mapping.
2. **Added underflow test** — `test_vector_count_underflow_raises` verifies that returning too few vectors raises `ValueError`.

danielaskdd:
@codex review

Br1an67:
Addressed the codex review — pushed 0d43c281:

**Per-vector dimension guard**: Before slicing a 2D result, we now check `result.shape[-1] == expected_dim`. If a provider returns `(N, 2*dim)` instead of `(2*N, dim)`, the slice would be a no-op and silently pass wrong-width embeddings to downstream stores (e.g. FAISS). Now raises `ValueError` in that case.

Added `test_wrong_dimension_per_vector_raises` to cover this scenario.

danielaskdd:
@codex review

Br1an67:
Fixed in d8a785bf — >2D results (e.g. `(1, 4, dim)`) are now reshaped to `(actual_vectors, dim)` before slicing, so the trim always operates on the vector axis. Added `test_nested_batch_overflow_flattened` to cover this case.

danielaskdd:
@codex review

Br1an67:
Addressed the latest codex P1 — pushed 7791da6d:

After slicing a flat 1D overflow result, it now reshapes to `(expected_vectors, expected_dim)` to preserve the 2D batch contract. Previously a single-text input would return shape `(dim,)` instead of `(1, dim)`, breaking downstream code that indexes `result[0]`.

Updated the 1D test to assert `result.shape == (1, 128)` instead of just checking element count.

danielaskdd:
Critical Issues:                                                                                            
                                                                                                                                                                                                       
1. Silent data corruption: wrong vectors stored for wrong documents
                                                                                                                                                                                                                                                                                                                                                                                                                
The PR assumes the first N vectors in an overflowing result correspond 1:1 to the first N input texts. This assumption is undocumented and unverifiable. If the embedding provider returns extra vectors 
that are prepended, interleaved, or reordered (not appended), then result[:expected_vectors] silently stores the wrong embeddings for each document. Every storage backend that uses positional indexing
(embeddings[i] → list_data[i]) — Milvus, Mongo, Postgres, Qdrant, Faiss, NanoVDB — will index the wrong vector without any error at any layer.                                                           
              
This is silent, permanent data corruption. Queries return wrong results with no error surfaced anywhere.                                                                                                 
              
2. 1D branch silently assembles vectors from flat bytes with no semantic basis                                                                                                                           

A 1D result for a batch embedding is never a legitimate return value from any provider in this codebase (all return 2D arrays). The code silently takes the first expected_vectors * expected_dim        
elements of a flat array and reshapes them into vectors. There is no basis for believing these chunks correspond to actual document embeddings. The warning message doesn't even flag that the result was
1D — a developer reading logs has no way to know.                                                                                                                                                       
              
3. The result.shape[-1] dimension check is insufficient for multi-dimensional arrays                                                                                                                                   
For a 3D array of shape (2, 5, 1536) with expected_dim=1536, shape[-1] == 1536 passes the check. The code then reshapes to (10, 1536) and slices to [:expected_vectors] — but those first N rows came    
from the first batch element, not from N distinct input documents. The guard needed is a full shape assertion, not just the last axis.


Recommended Action:

The core approach of silently slicing vectors at the shared EmbeddingFunc.__call__ layer is the wrong level to fix this. The silent data corruption risk across all six storage backends outweighs the usability benefit.  

1. Raise with a detailed ValueError (as the original code did) that names the provider, batch size, and actual vs. expected count.                                                                       

2. If a specific provider is known to safely return extra padding vectors, add an explicit opt-in allow_extra_vectors: bool = False flag to EmbeddingFunc, documented clearly, so the behavior is
auditable rather than a generic silent fallback.

Br1an67:
Thanks for the thorough review — all three points are valid. The silent slicing approach was the wrong level to fix this.

Rewrote the PR (squashed to a single commit):

1. **Default behavior is now fail-fast** — `ValueError` with detailed message naming the provider, batch size, and actual vs expected count. No more silent slicing by default.
2. **Explicit opt-in** — added `allow_extra_vectors: bool = False` to `EmbeddingFunc`. Users with providers known to safely append extra padding vectors can set this to `True`, making the behavior auditable rather than a generic silent fallback.
3. **Removed the 1D/3D reshape heuristics** — when opted in, the code does a simple `reshape(-1, dim)[:expected]` which is straightforward and reviewable.

Tests now cover: default raise, opt-in slicing, underflow always raises, match passes, and error message content validation.

danielaskdd:
Forcefully truncating vectors may mask underlying issues. It is recommended to implement dedicated embedding functions for the problematic models to ensure they return accurate results directly.

johnshearing:
This is working well for me but I will implement dedicated embedding functions for RAG_Anything as recommended by @danielaskdd at a later time. In the meantime, I would like to say thanks to @Br1an67 for this excellent workaround.

PR #2793, opened 2026-03-18
fix(api): normalize missing document file paths

danielaskdd:
## Description

Normalize missing document `file_path` values to `unknown_source` so document status APIs always return a non-null string and text-ingested documents no longer serialize with `None` file paths.

## Related Issues

#2762

## Changes Made

- Added a shared `normalize_file_path()` helper in the document routes layer to coerce `None`, blank strings, and legacy `no-file-path` values to `unknown_source`
- Normalized `/documents/text` and `/documents/texts` file sources before enqueueing, and added a second normalization pass in `LightRAG.apipeline_enqueue_documents()` so stored document records do not keep `None` file paths
- Restored `DocStatusResponse.file_path` to a required string and applied normalization in document status, track status, and paginated document responses
- Added regression tests covering text ingestion without `file_source` and non-null `DocStatusResponse.file_path` serialization

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed
- [ ] Documentation updated (if necessary)
- [x] Unit tests added (if applicable)

## Additional Notes

Local verification run:
- `./scripts/test.sh tests/test_document_file_path_normalization.py`
- `ruff check lightrag/api/routers/document_routes.py lightrag/lightrag.py tests/test_document_file_path_normalization.py`

PR #2796, opened 2026-03-19
fix: prevent None file_path from propagating as unknown_source

he-yufeng:
## Summary

Fixes the root cause of #2764 where some documents show "unknown_source" instead of their actual filenames.

### Root Cause

`getattr(status_doc, "file_path", "unknown_source")` only returns the fallback when the **attribute doesn't exist**. When `file_path` is present but `None`, it returns `None` — which then permanently overwrites valid filenames during reprocessing or consistency validation.

Same pattern in `JsonDocStatusStorage`: `"file_path" not in data` passes when the key exists with a null value, creating `DocProcessingStatus(file_path=None)`.

### Changes

**`lightrag/lightrag.py`** (4 sites):
```python
# Before — returns None when file_path attribute is None
file_path = getattr(status_doc, "file_path", "unknown_source")

# After — catches both missing and None
file_path = getattr(status_doc, "file_path", None) or "unknown_source"
```

**`lightrag/kg/json_doc_status_impl.py`** (3 sites):
```python
# Before — only catches missing key, not null value
if "file_path" not in data:
    data["file_path"] = "no-file-path"

# After — catches missing, None, and empty string
if not data.get("file_path"):
    data["file_path"] = "no-file-path"
```

Complements #2793 which fixed the API input layer — this PR fixes the processing pipeline and storage layer.

## Test plan

- [x] 275 tests pass, 33 skipped (pre-existing skip/xfail)
- [x] 1 pre-existing failure in `test_interactive_setup_outputs.py` unrelated to this change

### DIFF
--- a/lightrag/kg/json_doc_status_impl.py
+++ b/lightrag/kg/json_doc_status_impl.py
@@ -113,8 +113,8 @@ class JsonDocStatusStorage(DocStatusStorage):
                         data = v.copy()
                         # Remove deprecated content field if it exists
                         data.pop("content", None)
-                        # If file_path is not in data, use document id as file path
-                        if "file_path" not in data:
+                        # Normalize missing or null file_path
+                        if not data.get("file_path"):
                             data["file_path"] = "no-file-path"
                         # Ensure new fields exist with default values
                         if "metadata" not in data:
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

Issue #2264, opened 2025-10-26
[Question]: Why is LightRAG insert/init much slower than direct vLLM batching even with async tuning?


cameron-925:
### Do you need to ask a question?

- [x] I have searched the existing question and discussions and this question is not already answered.
- [x] I believe this is a legitimate question, not just a bug or feature request.

### Your Question

I’m inserting 1,000 docs into LightRAG and seeing overall init/insert times much longer than generating the same prompts directly via vLLM with a batch size of 8. After tuning LightRAG async params I improved total time from 7h08m → 1h45m, but it’s still far slower than my 10-20 min vLLM test for pure generation on equivalent prompts. I understand LightRAG performs extra steps (chunking, embeddings, keywords, storage), but I want to confirm whether my concurrency is configured as intended and whether other bottlenecks are expected.

## Questions

1. Does LightRAG serialize any stages per document that would limit throughput even when *llm_model_max_async* is higher?
2. Is max_parallel_insert interacting with *insert_batch_size* and *llm_model_max_async* in a way that limits GPU utilization?
3. Any guidance on recommended ratios for:
    - max_parallel_insert
    - llm_model_max_async
    - embedding_batch_num
    - Embedding_func_max_async
    - relative to long prompts (~2k tokens) and a single 48 GB GPU?
4. Any other recommendations?

## Environment

- Hardware: Single node with multiple NVIDIA RTX 6000 Ada GPUs (currently using 1 GPU for LLM). GPU memory not fully utilized.
- OS / Python: Ubuntu 22.04, Python 3.12.11
- LightRAG version/commit: 1.4.7
- vLLM version: 0.10.0
- Ollama version (for embeddings): 0.6.0
- Backends:
    - LLM: vLLM serving Qwen/Qwen2.5-7B-Instruct
    - Embeddings: nomic-embed-text via Ollama
- Default storage backends
    - `*.json`            # KV & doc-status storage
    - `*.json`           # Vector store entries
    - `*.graphml`      # Knowledge graph storage

## Workload
- Corpus: 1,000 docs
- Prompt token lengths (LightRAG ent/rel extraction prompts + input text):
    - Min 1933, P25 1955, P50 1970.5, P75 2002, Max 2683

## LightRAG Initialization Code and Parameters

```python
LLM_MODEL = "Qwen/Qwen2.5-7B-Instruct"
EMBEDDING_MODEL = "nomic-embed-text"
EMBEDDING_DIM = "768"

async def initialize_rag():
[...]

Issue #1957, opened 2025-08-14
[Question]: Merge scale performance issues


Yannholo:
### Do you need to ask a question?

- [x] I have searched the existing question and discussions and this question is not already answered.
- [x] I believe this is a legitimate question, not just a bug or feature request.

### Your Question

We have been trying to scale lightrag to around 1000 documents and we are struggling with the ingestion process.

We are using a custom GCP Spanner storage (We would like to contribute it at some point)

In our stuggle we took some time to add telemetry on different parts of lightrag to figure out where the issues came from. (Our first ideal being out custom storage)

Our conclusion seams to be that the way the graph merging process is not scaling up as we would expect.
In our current setup we have : 
- ~1000 documents
- ~4000 chunks
- ~140000 entities
- ~160000 relationships

Adding a new document can take more than 20minutes. (We are considering disabling the graph altogether until we have a solution)

The main issue seams to come down to the hundreds of new embbedding calls that have to be made to reembed the merged entities / relations, most of these calls contain a single text and can get quite long, especially when we start hitting up rate limits slowing everything down.

Would it be considered to have the possibility to batch (multiple in a single call) embeddings / upserts during the merge process ? That would definitelly help optimise these steps.

danielaskdd:
It appears your issue stems from the node and relationship merging phase. If newly inserted documents have extensive interconnections with existing ones, this can trigger a high volume of merge operations. To enhance performance during this stage, consider the following recommendations:

1.  **Upgrade LightRAG to v1.4.x:** The latest version includes improvements that increase LLM concurrency during the merging phase.
2.  **Deploy Embedding Models Locally:** Given that the new version performs node merging and updates concurrently, requiring immediate data persistence after node merges, batching of Embedding calls is not feasible. Deploying your Embedding model locally can significantly accelerate embedding operations.

Yannholo:
We are already using v1.4.6.

I've done more tests, i've run embeddings in a synthetic test using asyncio and the performances are good (I can hit the maximum in the ratelimiter) so i don't think this is the actual issue.

One thing that was slowing us was the tokenizer, as i understand it is only used to count and split. I've created a "unicode tokenizer" that is super fast, and had to increase token limits so it still create reasonably sized chunks.

Our theory for the limitation now is the single threadingness of python. Could that be an issue ?

Yannholo:
I don't think we need the graph to be (at all time) up to date, there could be a queue of updates to do (add or delete) for graph elements. 

[...]

Issue #1648, opened 2025-06-02
[Question] : Performance Bottleneck with Large Scale Ingestion (50K Documents) - LightRAG FastAPI + Ray Integration


pankaj-2k01:
### Do you need to ask a question?

- [x] I have searched the existing question and discussions and this question is not already answered.
- [x] I believe this is a legitimate question, not just a bug or feature request.

### Your Question

Issue Description
We are currently building a FastAPI application that integrates LightRAG for document ingestion and querying, leveraging Ray for distributed processing. While the system works effectively for smaller datasets, we are encountering significant performance degradation and extended processing times when attempting to ingest a large volume of documents (approximately 50,000 text-heavy files, primarily PDFs).

We suspect the bottleneck lies within the distributed ingestion pipeline, specifically concerning the interaction between Ray actors, external API calls for embeddings, and LightRAG's internal storage mechanisms. We are seeking insights, potential optimizations, or best practices for handling such large-scale ingestion with LightRAG and Ray.

System Configuration
FastAPI Version: (Assumed latest, please specify if different)
LightRAG Version: (Assumed latest, please specify if different)
Ray Version: (Assumed latest, please specify if different)
Python Version: 3.9+
Operating System: Linux (e.g., Ubuntu)
Hardware: (Mention CPU cores, RAM, SSD/HDD, e.g., "16 vCPUs, 64GB RAM, NVMe SSD")
OpenAI-Compatible API: Self-hosted (http://**********) using gpt-4o for LLM and text-embedding-ada-002 for embeddings.
Working Directory: Local disk path (./test_storage)
Observed Behavior with 50,000 Documents
Ingestion Time: Extends from minutes (for ~100 documents) to many hours or potentially days (for 50,000 documents).
Resource Utilization:
CPU: Appears to be utilized by Ray workers during text extraction and RAG ingestion, but not consistently at 100% across all cores.
Memory: Ray object store and individual actor memory usage grow, but don't typically hit OOM errors.
Disk I/O: High during ingestion as LightRAG writes data.
Network I/O: Significant during embedding calls to the external API.
Error Rates: Occasional HTTP 429 (Too Many Requests) from the embedding API, handled by retry logic, but this indicates API saturation and contributes to delays.
Responsiveness: The /ingest endpoint blocks until the entire ingestion process (including Ray tasks) completes.
Current Implementation Details (Code Snippets for Context)
[...]

Issue #1387, opened 2025-04-16
[Question]: Why the change to merging and upserting nodes and edges serially?


alineberry:
### Do you need to ask a question?

- [x] I have searched the existing question and discussions and this question is not already answered.
- [x] I believe this is a legitimate question, not just a bug or feature request.

### Your Question

On Release 1.3.1 the `extract_entities` had a change where nodes and edges went from being merged+upserted in parallel to serial. This seems to slow things down; what was the reason for this change?

danielaskdd:
Too many concurrent requests can generate excessive database connections, consuming a large amount of resources and leading to a performance decline. The correct approach is to use batch mode for data queries. The main branch has already been optimized following this principle. Currently, the performance of PostgreSQL and Neo4j has been significantly improved.

danielaskdd:
The bottleneck in node merging occurs during LLM calls, and simply increasing Python's concurrency isn't the right solution. In v1.3.2, we've optimized the process by deferring node merging until after each document's processing phase, followed by a comprehensive merge of all nodes with identical names across the entire document. This approach has significantly improved node merging speed.

alineberry:
Yes, we have identified that optimization as well. Seems it would double the throughput for our use case. 

I am also curious what is the status of using dense retrieval + LLM during the entity merging step on your end? We see massive duplication (~60-80% compression rate) with the v1.3.0 we work from.

danielaskdd:
> Yes, we have identified that optimization as well. Seems it would double the throughput for our use case.
> 
> I am also curious what is the status of using dense retrieval + LLM during the entity merging step on your end? We see massive duplication (~60-80% compression rate) with the v1.3.0 we work from.

I can't get what's the point of your question. What's dense retrieval and  massive duplication meaning?

alineberry:
I've seen other discussions about using the entity vdb to do a better job of merging than exact string matching.

Another question... do your recent changes collate extractions across all documents in the batch and do a single large merge/upsert step? In contrast to merging/upserting data serially for each document.

danielaskdd:
#1323 is already put in pending stage. I don't think merging across all documents is not a good idea.

danielaskdd:
Parallel merging is implemented after v1.4.0

PR #2816, opened 2026-03-21
feat: batch Neo4j graph writes with UNWIND queries

ndcorder:
Currently every entity and relationship gets its own Neo4j session + transaction + MERGE query during the merge phase. For a document producing 80 entities and 120 relations, that's 200+ round-trips.

This adds batch_upsert_nodes/batch_upsert_edges to the graph storage interface with a default loop fallback so other backends aren't affected. Neo4JStorage overrides them with UNWIND-based Cypher. The merge phase in operate.py now collects results and writes them in bulk instead of one-at-a-time.

Should help with #1387, #1957, #1648, #2264.

### DIFF
--- a/lightrag/base.py
+++ b/lightrag/base.py
@@ -565,6 +565,28 @@ class BaseGraphStorage(StorageNameSpace, ABC):
             result[node_id] = edges if edges is not None else []
         return result
 
+    async def batch_upsert_nodes(
+        self, nodes: list[tuple[str, dict[str, str]]]
+    ) -> None:
+        """Batch insert/update nodes. Default loops over upsert_node."""
+        for node_id, node_data in nodes:
+            await self.upsert_node(node_id, node_data)
+
+    async def batch_upsert_edges(
+        self, edges: list[tuple[str, str, dict[str, str]]]
+    ) -> None:
+        """Batch insert/update edges. Default loops over upsert_edge."""
+        for src, tgt, edge_data in edges:
+            await self.upsert_edge(src, tgt, edge_data)
+
+    async def has_nodes_batch(self, node_ids: list[str]) -> set[str]:
+        """Return the subset of node_ids that exist. Default loops over has_node."""
+        result = set()
+        for node_id in node_ids:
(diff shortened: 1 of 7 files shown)


danielaskdd:
Thanks for your interesting in LightRAG and contributions. However I noticed a **critical concurrency issue (Race Condition) introduced by deferring the graph writes**.

In the current implementation of lightrag/operate.py, the _locked_process_entity_name (and similarly _locked_process_edges) function acquires an application-level distributed lock using async with get_storage_keyed_lock:

```
async with get_storage_keyed_lock([entity_name], namespace=namespace, enable_logging=False):
    entity_data = await _merge_nodes_then_upsert(...) 
    # This reads the old node from DB, merges data, but DOES NOT write it yet!
# Lock is released here!
```

By deferring the upsert_node out of the loop and putting it into a batch at the end of the phase (after all locks are released):

```
# Batch-write all entity nodes to the graph in one call
if nodes_to_upsert:
    await knowledge_graph_inst.batch_upsert_nodes(nodes_to_upsert)
```

This breaks the "Read-Modify-Write" atomic cycle:

1. Task A acquires the lock for Apple, reads from the DB (finds it missing or gets old state), calculates the merged state, and releases the lock.
2. Task B acquires the lock for Apple, reads from the DB (still finds it missing/old state, because Task A hasn't executed the batch upsert yet!), calculates the merged state, and releases the lock.
3. Both tasks append their result to their respective nodes_to_upsert lists.
4. When batch_upsert_nodes is finally called, Task B will simply overwrite Task A's result (or trigger a duplication/constraint error).

This could lead to severe data loss and inconsistencies when multiple documents are processed concurrently and contain the same entities/edges.

To maintain concurrent safety while benefiting from batching, we might need to rely entirely on DB-level atomic merge operations (e.g., using ON CREATE SET and ON MATCH SET in Cypher) and skip application-level reading before writing. (Though this might break compatibility with other GraphDB implementations).

This PR will not be merged until you find a way to work it out. Thanks!

PR #2817, opened 2026-03-21
feat: skip LLM summary for entities with unchanged descriptions

ndcorder:
When re-ingesting a document or doing incremental updates, most entities already have the same descriptions. Right now we still call the LLM to re-summarize them every time, which is wasteful.

This adds an early return in _merge_nodes_then_upsert — if all incoming descriptions are already present on the existing node, we skip the summary call and just update source tracking.

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -1770,6 +1770,31 @@ async def _merge_nodes_then_upsert(
         )
         description_list = [fallback_description]
 
+    # Skip LLM summary if no new descriptions were added (re-ingestion optimisation)
+    if already_node and already_description:
+        existing_descriptions = set(
+            d.strip()
+            for d in already_description
+            if d.strip()
+        )
+        incoming_descriptions = set(
+            d.strip()
+            for d in sorted_descriptions
+            if d.strip()
+        )
+        if incoming_descriptions and incoming_descriptions.issubset(existing_descriptions):
+            logger.debug(
+                f"Entity '{entity_name}': no new descriptions, skipping LLM summary"
+            )
+            node_data = dict(already_node)
+            node_data["source_id"] = source_id
+            node_data["file_path"] = GRAPH_FIELD_SEP.join(
+                list(dict.fromkeys(already_file_paths + [dp.get("file_path") for dp in nodes_data if dp.get("file_path")]))
+            )
+            await knowledge_graph_inst.upsert_node(entity_name, node_data=node_data)
+            node_data["entity_name"] = entity_name
+            return node_data
+
     # Check for cancellation before LLM summary
     if pipeline_status is not None and pipeline_status_lock is not None:
         async with pipeline_status_lock:
--- a/tests/test_skip_unchanged_summary.py
+++ b/tests/test_skip_unchanged_summary.py
@@ -1,0 +1,144 @@
+# test_skip_unchanged_summary.py
+
+import pytest
+from unittest.mock import AsyncMock, patch
+from lightrag.utils import GRAPH_FIELD_SEP
+
+
+def _make_global_config(**overrides):
+    cfg = {
+        "source_ids_limit_method": "FIFO",
+        "max_source_ids_per_entity": 100,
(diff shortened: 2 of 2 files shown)


danielaskdd:
Hi, thanks for the contribution!

However, we have some concerns regarding the effectiveness of this optimization. Since `incoming_descriptions` are generated by LLM during the extraction phase, the wording of these descriptions will almost certainly vary slightly each time, even for the same entity in the same context. 

Because of these variations, an exact string comparison like `incoming_descriptions.issubset(existing_descriptions)` is highly unlikely to evaluate to `True` in a real-world scenario (unless the exact same LLM extraction cache is hit).

Could you provide any practical testing or metrics showing that this optimization actually triggers and skips the summarization step effectively in your use cases?

An alternative approach might be to use vector embeddings to check for semantic similarity between the new and existing descriptions: 
* **Pros**: It would correctly identify when a new description adds no new information, regardless of wording changes.
* **Cons**: It would introduce additional overhead by requiring vector database queries/embedding calculations during the merge step, which might negate the performance benefits of skipping the LLM summarization call.

We'd love to hear your thoughts on this!

ndcorder:
Good point — I ran some numbers to see how this actually plays out.

**Test setup:** Extracted entities from the same security policy text using `google/gemini-3-flash-preview` via OpenRouter, 5 runs each, compared descriptions character-for-character.

| Scenario | Exact match rate |
|-|-|
| `enable_llm_cache_for_entity_extract=True` (default) | 100% — cache key is MD5(prompt), same chunk = same hash = same response |
| Cache off, temperature=0.0 | 100% (14/14 entities identical across all 5 runs) |
| Cache off, temperature=0.7 | 27% (4/15 entities identical) |

So you're right that without cache at nonzero temperature, exact matching only catches about a quarter of entities. The optimization is mainly useful in two scenarios:

1. **Re-ingestion with cache enabled (the default)** — extraction cache returns byte-identical results, so issubset() triggers for every unchanged entity. This is the primary use case I had in mind.
2. **Temperature=0 without cache** — some models (Gemini at least) are fully deterministic here, so it still works.

For the temp>0 no-cache case, the semantic similarity approach you mentioned would catch more, but the embedding overhead on every merge probably isn't worth it. The nice thing about the current approach is zero overhead when it doesn't trigger — it's just a set comparison.

I'm open to scoping this more explicitly in the docstring/comments as a "cache-aware re-ingestion optimization" rather than a general-purpose skip, if that makes the intent clearer. Or if you'd prefer a different approach entirely, happy to hear it.

<details>
<summary>Raw test output (temp=0.7, showing the kind of variation you described)</summary>

```
Entity: Access Control Policy (4 unique descriptions across 5 runs)
  Run 1: A security policy, also known as AC-01, that mandates multi-factor authentication for the TracCloud production environment.
  Run 2: A policy designated as AC-01 that mandates multi-factor authentication for accessing specific environments.
  Run 3: A document designated as AC-01 that mandates multi-factor authentication for accessing specific environments.

Entity: Jane Martinez (3 unique descriptions across 5 runs)
  Run 1: The Information Security Officer at Redrock Software who oversees policy implementation and incident response.
  Run 2: The Information Security Officer at Redrock Software who oversees policy implementation and coordinates incident response.
  Run 4: The Information Security Officer at Redrock Software who oversees policy implementation and incident response coordination.

Entity: TracCloud (3 unique descriptions across 5 runs)
  Run 1: A production environment that stores CUI and requires multi-factor authentication for access.
  Run 2: A production environment that stores CUI and requires compliance with security standards.
  Run 3: The production environment that stores CUI and requires multi-factor authentication for access.
```

</details>

ndcorder:
I think I was too tunnel-visioned when I originally made this PR because it's not necessary.

If the cache is enabled, the summary LLM call itself would also be cached, so the skip logic is redundant. And if the cache is off, descriptions vary too much for exact matching to help. Either way it doesn't add real value. I'm sorry for wasting your time!

danielaskdd:
You are welcome.

PR #2819, opened 2026-03-21
fix: defer KG rebuild during batch document deletion

### LINKED ISSUE #2795: [Performance]: batch document deletion triggers N redundant knowledge graph rebuilds

### Describe the bug

When deleting multiple documents via the batch delete endpoint, `background_delete_documents()` calls `adelete_by_doc_id()` in a loop. Each call internally runs `rebuild_knowledge_from_chunks()` — meaning the knowledge graph is fully rebuilt **N times** for N documents.

For shared entities/relations that span multiple documents, the same entity summaries are recomputed over and over, generating redundant LLM calls and chunk re-parses that produce the exact same result.

### Code path

```
background_delete_documents()                    # document_routes.py:1820
  for doc_id in doc_ids:                         # sequential loop
    await rag.adelete_by_doc_id(doc_id)          # lightrag.py:3040
      → rebuild_knowledge_from_chunks(...)       # lightrag.py:3644 ← runs N times
```

### Measured impact (85-document batch deletion, PostgreSQL storage)

| Metric | Current (N rebuilds) | Optimal (1 rebuild) | Waste factor |
|--------|---------------------|---------------------|-------------|
| Rebuild operations | 85 | **1** | 85x |
| LLM summary calls | 1,593 | ~21 | ~75x |
| Cached chunk re-parses | 135,987 | ~1,800 | ~75x |
| Warning log lines (tuple parsing) | 15,110 | ~200 | ~75x |

The 75x overhead comes from shared entities being rebuilt repeatedly as each document is deleted. In production this translates to hours of unnecessary processing time and significant LLM API costs.

### Proposed fix

Add a `skip_rebuild` parameter to `adelete_by_doc_id()` and defer the rebuild to a single pass after all deletions complete:

```python
# In adelete_by_doc_id():
async def adelete_by_doc_id(
    self, doc_id: str, delete_llm_cache: bool = False, skip_rebuild: bool = False
) -> DeletionResult:
    ...
    # Step 8: conditionally skip rebuild
    if not skip_rebuild and (entities_to_rebuild or relationships_to_rebuild):
        await rebuild_knowledge_from_chunks(...)
    
    # Return rebuild targets so caller can aggregate them
    result.entities_to_rebuild = entities_to_rebuild
    result.relationships_to_rebuild = relationships_to_rebuild
    return result
```

```python
# In background_delete_documents():
all_entities_to_rebuild = {}
all_relationships_to_rebuild = {}

for doc_id in doc_ids:
    result = await rag.adelete_by_doc_id(doc_id, skip_rebuild=True)
    if result.status == "success":
        all_entities_to_rebuild.update(result.entities_to_rebuild)
        all_relationships_to_rebuild.update(result.relationships_to_rebuild)

# Single rebuild pass for all affected entities/relations
if all_entities_to_rebuild or all_relationships_to_rebuild:
    await rebuild_knowledge_from_chunks(
        entities_to_rebuild=all_entities_to_rebuild,
        relationships_to_rebuild=all_relationships_to_rebuild,
        ...
    )
```

Key design points:
- `skip_rebuild` defaults to `False` — **zero breaking changes** for single-document deletion or direct API callers
- `DeletionResult` gains optional `entities_to_rebuild` / `relationships_to_rebuild` fields so callers can aggregate
- Graph consistency is preserved — same end state, just computed once instead of N times
- The existing `rebuild_knowledge_from_chunks()` function already supports the aggregated dict pattern

### Files affected

| File | Change |
|------|--------|
| `lightrag/lightrag.py` | `adelete_by_doc_id()`: add `skip_rebuild` param, conditionally skip rebuild, return rebuild targets |
| `lightrag/api/routers/document_routes.py` | `background_delete_documents()`: aggregate rebuild targets across loop, single rebuild at end |

### LightRAG Config Used

PostgreSQL + pgvector + AGE, OpenAI API.

### Additional context

We implemented and validated this fix in our production fork. The deferred rebuild produces the exact same graph state as N individual rebuilds, with ~75x fewer LLM calls. Happy to contribute a PR if there's interest.

he-yufeng:
## Summary

When deleting N documents via `background_delete_documents()`, each `adelete_by_doc_id()` call triggers a full `rebuild_knowledge_from_chunks()`. For entities shared across documents, this means the same summaries get recomputed N times. On an 85-document batch, the issue author measured ~75x redundant LLM calls.

This PR adds a `skip_rebuild` parameter to `adelete_by_doc_id()` (defaults to `False` for backwards compat). The batch deletion loop passes `skip_rebuild=True`, collects rebuild targets from each result, and does a single combined rebuild at the end.

**Changes:**
- `lightrag/base.py` — `DeletionResult` gains optional `entities_to_rebuild` / `relationships_to_rebuild` fields
- `lightrag/lightrag.py` — `adelete_by_doc_id()` conditionally skips rebuild when `skip_rebuild=True`, populates rebuild targets on the result
- `lightrag/api/routers/document_routes.py` — `background_delete_documents()` aggregates targets across loop, single `rebuild_knowledge_from_chunks()` call at the end
- `tests/test_batch_delete_deferred_rebuild.py` — 4 tests covering both modes

Single-doc deletion (direct `adelete_by_doc_id()` calls) is completely unaffected — same code path as before.

Closes #2795

### DIFF
--- a/lightrag/api/routers/document_routes.py
+++ b/lightrag/api/routers/document_routes.py
@@ -1854,6 +1854,9 @@ async def background_delete_documents(
     total_docs = len(doc_ids)
     successful_deletions = []
     failed_deletions = []
+    # Aggregate rebuild targets across all deletions so we only rebuild once
+    all_entities_to_rebuild: dict[str, list] = {}
+    all_relationships_to_rebuild: dict[tuple, list] = {}
 
     # Double-check pipeline status before proceeding
     async with pipeline_status_lock:
@@ -1906,13 +1909,22 @@ async def background_delete_documents(
             file_path = "#"
             try:
                 result = await rag.adelete_by_doc_id(
-                    doc_id, delete_llm_cache=delete_llm_cache
+                    doc_id,
+                    delete_llm_cache=delete_llm_cache,
+                    skip_rebuild=True,
                 )
                 file_path = (
                     getattr(result, "file_path", "-") if "result" in locals() else "-"
                 )
                 if result.status == "success":
                     successful_deletions.append(doc_id)
+                    # Collect deferred rebuild targets
+                    if result.entities_to_rebuild:
+                        all_entities_to_rebuild.update(result.entities_to_rebuild)
+                    if result.relationships_to_rebuild:
+                        all_relationships_to_rebuild.update(
+                            result.relationships_to_rebuild
+                        )
                     success_msg = (
                         f"Document deleted {i}/{total_docs}: {doc_id}[{file_path}]"
                     )
@@ -2048,6 +2060,44 @@ async def background_delete_documents(
                 async with pipeline_status_lock:
                     pipeline_status["latest_message"] = error_msg
                     pipeline_status["history_messages"].append(error_msg)
+
+        # Single deferred rebuild for all affected entities/relations
+        if all_entities_to_rebuild or all_relationships_to_rebuild:
+            from dataclasses import asdict
+
+            from lightrag.operate import rebuild_knowledge_from_chunks
+
+            rebuild_msg = (
+                f"Rebuilding knowledge graph: {len(all_entities_to_rebuild)} entities, "
+                f"{len(all_relationships_to_rebuild)} relations"
+            )
+            logger.info(rebuild_msg)
+            async with pipeline_status_lock:
+                pipeline_status["latest_message"] = rebuild_msg
+                pipeline_status["history_messages"].append(rebuild_msg)
+            try:
+                await rebuild_knowledge_from_chunks(
+                    entities_to_rebuild=all_entities_to_rebuild,
+                    relationships_to_rebuild=all_relationships_to_rebuild,
+                    knowledge_graph_inst=rag.chunk_entity_relation_graph,
+                    entities_vdb=rag.entities_vdb,
+                    relationships_vdb=rag.relationships_vdb,
+                    text_chunks_storage=rag.text_chunks,
+                    llm_response_cache=rag.llm_response_cache,
+                    global_config=asdict(rag),
+                    pipeline_status=pipeline_status,
+                    pipeline_status_lock=pipeline_status_lock,
+                    entity_chunks_storage=rag.entity_chunks,
+                    relation_chunks_storage=rag.relation_chunks,
+                )
+                await rag._insert_done()
+            except Exception as rebuild_err:
+                rebuild_error_msg = f"Failed to rebuild knowledge graph after batch deletion: {rebuild_err}"
+                logger.error(rebuild_error_msg)
+                logger.error(traceback.format_exc())
+                async with pipeline_status_lock:
+                    pipeline_status["latest_message"] = rebuild_error_msg
+                    pipeline_status["history_messages"].append(rebuild_error_msg)
 
     except Exception as e:
         error_msg = f"Critical error during batch deletion: {str(e)}"
--- a/lightrag/api/routers/document_routes.py
+++ b/lightrag/api/routers/document_routes.py
@@ -2062,42 +2062,96 @@ async def background_delete_documents(
                     pipeline_status["history_messages"].append(error_msg)
 
         # Single deferred rebuild for all affected entities/relations
+        rebuild_ok = True
         if all_entities_to_rebuild or all_relationships_to_rebuild:
             from dataclasses import asdict
 
             from lightrag.operate import rebuild_knowledge_from_chunks
 
-            rebuild_msg = (
-                f"Rebuilding knowledge graph: {len(all_entities_to_rebuild)} entities, "
-                f"{len(all_relationships_to_rebuild)} relations"
-            )
-            logger.info(rebuild_msg)
-            async with pipeline_status_lock:
-                pipeline_status["latest_message"] = rebuild_msg
-                pipeline_status["history_messages"].append(rebuild_msg)
-            try:
-                await rebuild_knowledge_from_chunks(
-                    entities_to_rebuild=all_entities_to_rebuild,
-                    relationships_to_rebuild=all_relationships_to_rebuild,
-                    knowledge_graph_inst=rag.chunk_entity_relation_graph,
-                    entities_vdb=rag.entities_vdb,
-                    relationships_vdb=rag.relationships_vdb,
-                    text_chunks_storage=rag.text_chunks,
-                    llm_response_cache=rag.llm_response_cache,
-                    global_config=asdict(rag),
-                    pipeline_status=pipeline_status,
-                    pipeline_status_lock=pipeline_status_lock,
-                    entity_chunks_storage=rag.entity_chunks,
-                    relation_chunks_storage=rag.relation_chunks,
+            # --- Issue 1: filter out entities/relations that were already fully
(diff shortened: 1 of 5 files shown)


danielaskdd:
Deferring the refactoring of entity relationships is a sound approach; however, there are several issues in the current PR that need to be addressed:

1. **Redundant Tasks in Overlapping Pools**: If multiple documents scheduled for deletion share common entities, those entities might be tagged for both "Refactor" and "Delete" pools simultaneously. In such cases, the refactoring task becomes redundant and should be invalidated or skipped.
2. **Lack of Robust Error Handling**: The LLM-driven KG rebuild stage is prone to failure. If an error occurs during this phase, the pipeline should abort without removing the file from doc_status. This ensures the system maintains state and allows the user to re-trigger the deletion process later, and idempotency for deletion operation should be ensured.

he-yufeng:
Addressed the feedback — latest push includes:

- Pruned stale entities/relations from the rebuild pool (checks graph existence before queueing)
- Added error handling around the deferred rebuild so a failure there doesn't take down the whole delete operation
- Cleaned up the pool merging logic to avoid duplicates across overlapping batches

danielaskdd:
This PR conflicts with PR #2826 ("make document deletion retry-safe"). Please resolve the conflicts before we proceed with the review.

he-yufeng:
Rebased onto main, conflicts with #2826 resolved.

danielaskdd:
@codex  review

he-yufeng:
@danielaskdd Thanks for the review — both points were solid catches. Here's what I did:

**1. Redundant tasks in overlapping pools:**

Before the deferred rebuild runs, I now check whether each entity/relation still exists in the graph. If doc A tagged an entity for "rebuild" but doc B's deletion later removed it entirely, the rebuild for that node gets skipped. The pruning loop sits right before `rebuild_knowledge_from_chunks` in `document_routes.py` — it calls `has_node()` / `has_edge()` and drops anything that's already gone.

**2. Error handling / idempotency:**

When `skip_rebuild=True`, `adelete_by_doc_id` now keeps `doc_status` alive intentionally. The batch handler in `document_routes.py` only removes `doc_status` records *after* the deferred rebuild succeeds. If the rebuild blows up, doc_status stays put so the user can re-trigger deletion later — the retry-safe machinery from #2826 handles the rest.

**3. Conflict with #2826:**

Rebased on latest main (which includes #2826). Resolved cleanly, no merge conflicts. The `in_final_delete_stage` guard and `_update_delete_retry_state` from #2826 work as-is with the new `skip_rebuild` path.

danielaskdd:
@codex review

he-yufeng:
﻿Hi, checking in on this PR. I addressed the earlier feedback around overlapping rebuild/delete pools and deferred rebuild error handling, and GitHub still shows the branch clean with lint/offline tests green. Happy to adjust if there is anything else blocking review.

he-yufeng:
Rebased this onto current upstream/main and resolved the conflict in `adelete_by_doc_id` around the rebuild step.

Kept the PR's `skip_rebuild=True` deferred rebuild behavior, while preserving upstream's current `_build_global_config()` call for the normal per-doc rebuild path.

Validated locally on Windows:
- python -m pytest tests/test_batch_delete_deferred_rebuild.py -q (4 passed)
- python -m ruff check lightrag/lightrag.py lightrag/api/routers/document_routes.py lightrag/base.py tests/test_batch_delete_deferred_rebuild.py
- python -m ruff format --check lightrag/lightrag.py lightrag/api/routers/document_routes.py lightrag/base.py tests/test_batch_delete_deferred_rebuild.py
- python -m py_compile lightrag/lightrag.py lightrag/api/routers/document_routes.py lightrag/base.py tests/test_batch_delete_deferred_rebuild.py
- git diff --check upstream/main..HEAD

he-yufeng:
Follow-up after the lint job: CI's pinned pre-commit ruff-format wanted one additional assert wrapping change in `tests/test_batch_delete_deferred_rebuild.py` that my local direct `ruff format` did not catch.

Revalidated locally on Windows:
- python -m pytest tests/test_batch_delete_deferred_rebuild.py -q (4 passed)
- python -m pre_commit run ruff-format --files tests/test_batch_delete_deferred_rebuild.py lightrag/api/routers/document_routes.py lightrag/lightrag.py lightrag/base.py
- python -m pre_commit run ruff --files tests/test_batch_delete_deferred_rebuild.py lightrag/api/routers/document_routes.py lightrag/lightrag.py lightrag/base.py
- python -m py_compile lightrag/lightrag.py lightrag/api/routers/document_routes.py lightrag/base.py tests/test_batch_delete_deferred_rebuild.py
- git diff --check upstream/main..HEAD

he-yufeng:
Fixed the failing offline test after the rebase follow-up. The failure was from the document-route test double still using the old adelete_by_doc_id signature; it now records the skip_rebuild flag and asserts the batch/background path passes skip_rebuild=True. Validation: pytest tests/api/routes/test_document_routes_docx_archive.py::test_background_delete_removes_parser_hint_file_variants tests/test_batch_delete_deferred_rebuild.py -q passed (5 passed); pre-commit ruff passed on touched files; pre-commit ruff-format passed on touched files; py_compile on touched files passed; git diff --check passed.

danielaskdd:
Closing this PR — but not the idea. Deferring the per-document KG rebuild to a single aggregate rebuild per batch remains the direction we want, and this PR validated the core mechanism (`skip_rebuild=True`, aggregate-then-rebuild-once, keeping `doc_status` alive until the rebuild lands) as well as the overlapping-pool pruning. Thank you for the work and the responsiveness across the review rounds — the design is being carried forward with credit to this PR and #2795.

Three things force a ground-up redo rather than another iteration here:

1. **The two Codex P1 findings are structural.** With `skip_rebuild=True`, step 10 still deletes `full_entities`/`full_relations` and step 11 deletes `full_docs` per document, while the aggregated rebuild targets exist only in process memory. If the final rebuild fails (or the process dies before it), re-triggering deletion finds no anchors, silently no-ops, and drops the remaining `doc_status` rows — the shared entities keep summaries mixing deleted content permanently, and the "re-trigger to retry" promise cannot hold. Relatedly, a mid-batch per-doc failure can leave an earlier doc's stale chunk list in the aggregate, and `_rebuild_single_entity` writes that list verbatim into the authoritative `entity_chunks`/`relation_chunks` tracking. Fixing both requires durable rebuild intent and rebuild-time re-validation — a different persistence design, not a patch.

2. **The code this PR modifies no longer exists.** Since the last rebase, deletion on main was rewritten into a journaled, fail-closed purge primitive (`_purge_kg_contributions`, #3400/#3416) with recovery-anchor proofs and a phase journal in `doc_status.metadata`. Deferral now has to be a mode of that primitive with an extended journal phase; the branch also no longer shares history with main, so there is no incremental path.

3. The redesign is specified in **RFC #3659**, which keeps this PR's validated mechanism and closes the durability gaps by re-deriving rebuild targets from the recovery anchors at batch finalization (anchors retained until the rebuild succeeds), so retries converge on every backend without a separate journal.

If you'd like to pick up the implementation of #3659, you'd be very welcome — the RFC lists the hard invariants and the test matrix a new PR needs to satisfy.

PR #2825, opened 2026-03-22
fix: make document deletion retry-safe

danielaskdd:
## Description

This pull request makes document deletion retry-safe under the normal flow. It keeps `doc_status` as the final success marker, preserves retry metadata when deletion fails during graph rebuild, and continues best-effort cleanup even when `chunks_list` is missing but doc-level metadata is still available.

## Related Issues

N/A

## Changes Made

- Remove the early-success branch for documents with an empty `chunks_list` so deletion can still clean graph, vector, and KV state from doc-level metadata.
- Persist deletion retry metadata in `doc_status`, including backup chunk IDs, failure stage, and the latest deletion error when rebuild fails.
- Merge chunk-tracking storage with graph `source_id` during dependency analysis so retry deletes can rebuild or clean entities and relations correctly after partial failures.
- Add regression coverage for rebuild failure, retry success after failure, and deletion with metadata-backed chunk recovery.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -3185,21 +3185,12 @@ class LightRAG:
                 if not isinstance(metadata, dict):
                     metadata = {}
 
-                backup_chunk_ids = _normalize_string_list(
-                    metadata.get("deletion_chunk_ids", [])
-                )
-                current_chunk_ids = _normalize_string_list(
-                    doc_status_data.get("chunks_list", [])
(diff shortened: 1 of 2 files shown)


danielaskdd:
Superseded by #2826, which contains the same changes on a single squashed commit for cleaner review history.

PR #2847, opened 2026-03-27
perf(storage): add cooperative yielding to prevent event loop blocking

danielaskdd:
## Description

Add cooperative yielding (`asyncio.sleep(0)`) at periodic intervals inside CPU-heavy async loops across the storage layer and entity/relation processing pipeline. This prevents event loop starvation in single-worker deployments where long synchronous-style iterations can block other coroutines from running.

Also fixes an off-by-one bug in `RedisKVStorage.upsert` introduced by the original implementation that caused `IndexError` or silent timestamp corruption on every upsert call.

## Related Issues

N/A

## Changes Made

- **`lightrag/utils.py`**: Add shared `_cooperative_yield(iteration, every)` helper (replaces 7 duplicated local copies)
- **`lightrag/kg/redis_impl.py`**: Fix critical off-by-one bug — `exists_results[i]` → `exists_results[i - 1]` when using `enumerate(..., start=1)`; add cooperative yielding to upsert pipeline loops
- **`lightrag/kg/json_kv_impl.py`**: Import centralized helper; add comment explaining why yield inside `NamespaceLock` is intentional (reads `self._data`, must stay in lock scope)
- **`lightrag/kg/json_doc_status_impl.py`**: Import centralized helper; move data-preparation loop outside lock scope so cooperative yield benefits waiting writers
- **`lightrag/kg/mongo_impl.py`**: Import centralized helper; add `assert len(embeddings) == len(list_data)` to catch embedding batch size mismatches before silent wrong-vector storage
- **`lightrag/kg/opensearch_impl.py`**: Same assertion and centralized helper import
- **`lightrag/kg/postgres_impl.py`**: Same assertion and centralized helper import
- **`lightrag/operate.py`**: Import centralized helper; use `every=32` (tighter interval) for entity/relation processing loops; remove redundant yield in reduce phase (LLM `await` already yields naturally)

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

- Storage operations yield every 64 iterations (`every=64`, default)
- Entity/relation processing in `operate.py` yields every 32 iterations (`every=32`) for more responsive scheduling during graph merge phases
- The `json_kv_impl.py` yield intentionally remains inside the lock because the loop reads `self._data` to determine create vs update timestamps; a comment explains this to future maintainers

### DIFF
--- a/lightrag/api/routers/document_routes.py
+++ b/lightrag/api/routers/document_routes.py
@@ -1250,7 +1250,8 @@ async def pipeline_enqueue_file(
 
         # Get file size for error reporting
         try:
-            file_size = file_path.stat().st_size
+            stat = await asyncio.to_thread(file_path.stat)
+            file_size = stat.st_size
         except Exception:
             file_size = 0
 
@@ -1341,8 +1342,8 @@ async def pipeline_enqueue_file(
                     | ".less"
                 ):
                     try:
-                        # Try to decode as UTF-8
-                        content = file.decode("utf-8")
+                        # Try to decode as UTF-8 (offloaded to thread to avoid blocking the event loop)
+                        content = await asyncio.to_thread(file.decode, "utf-8")
 
                         # Validate content
                         if not content or len(content.strip()) == 0:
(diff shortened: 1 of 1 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

PR #2867, opened 2026-04-01
examples: add AG2 multi-agent demo with LightRAG retrieval

faridun-ag2:
## Summary

Adds a new example demonstrating [AG2](https://ag2.ai) multi-agent orchestration on top of LightRAG. Three specialized agents collaborate to answer complex questions using LightRAG's retrieval modes as tools.

AG2 is an open-source multi-agent framework with 500K+ monthly PyPI downloads, 4,300+ GitHub stars, and 400+ contributors.

### Architecture
```
User question → AG2 GroupChat → LightRAG queries
  ├── Researcher: hybrid/local search for specific facts
  ├── Analyst: global search for themes and relationships
  └── Writer: synthesizes findings into final answer
```

### What it demonstrates
- AG2 tool registration pattern with LightRAG's `aquery` as a callable tool
- Multiple agents using different LightRAG retrieval modes (local, global, hybrid)
- GroupChat orchestration with automatic speaker selection
- Clean separation: agents reason (via OpenAI), LightRAG retrieves (via knowledge graph)

### How it differs from existing examples
- Existing `lightrag_*_demo.py` files show single-user, single-query usage
- This demo shows **multi-agent collaboration** where agents independently query LightRAG and build on each other's findings

### Requirements
```
pip install lightrag-hku "ag2[openai]>=0.11.4,<1.0"
export OPENAI_API_KEY="..."
```

## Test plan
- [x] Script runs end-to-end with `python examples/lightrag_ag2_multiagent_demo.py`
- [x] All three agents participate and use tools correctly
- [x] Working directory cleaned up after execution
- [x] No hardcoded API keys
- [x] Passes ruff format and lint checks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ ignore_this.txt
 # Project-specific files
 /dickens*/
 /book.txt
+/ag2_demo_workdir/
 download_models_hf.py
 
 # Frontend build output (built during PyPI release)
--- a/examples/lightrag_ag2_multiagent_demo.py
+++ b/examples/lightrag_ag2_multiagent_demo.py
@@ -205,11 +205,12 @@ def create_agents():
 def run_multiagent_query(user_proxy, researcher, analyst, writer, question: str):
     """Run a multi-agent GroupChat to answer a question using LightRAG."""
     # Enforce pipeline: Researcher -> Analyst -> Writer.
-    # - Researcher and Analyst hand off to User for tool execution
-    # - User returns results to the same agent or advances to the next
-    # - Writer has no tool access, so it only follows Analyst
+    # func_call_filter (default True) automatically routes tool calls
+    # to/from user_proxy, so transitions only govern non-tool handoffs.
+    # User can only start with Researcher; Researcher advances to Analyst;
+    # Analyst advances to Writer. Writer terminates the conversation.
     allowed_transitions = {
-        user_proxy: [researcher, analyst],
+        user_proxy: [researcher],
         researcher: [user_proxy, analyst],
         analyst: [user_proxy, writer],
         writer: [],
--- a/examples/lightrag_ag2_multiagent_demo.py
+++ b/examples/lightrag_ag2_multiagent_demo.py
@@ -6,8 +6,8 @@ questions over indexed documents.
 
 Architecture:
     User -> AG2 GroupChat (Researcher + Analyst + Writer) -> LightRAG queries
-    - Researcher: uses LightRAG hybrid/local search to gather facts
-    - Analyst: examines entity relationships via global search
+    - Researcher: uses LightRAG hybrid search to gather facts
+    - Analyst: uses LightRAG naive (vector) search for complementary results
     - Writer: synthesizes findings into a final answer
 
 Requires:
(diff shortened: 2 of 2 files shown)


faridun-m:
Hi @danielaskdd @LarFii — would love your review on this example when you get a chance!

This adds an [AG2](https://ag2.ai) multi-agent demo that uses LightRAG's `aquery` as a tool. It's a self-contained example following the existing `lightrag_*_demo.py` pattern, with no changes to LightRAG core.

Happy to adjust anything based on your feedback. Thanks!

danielaskdd:
@codex review

danielaskdd:
Suggested Revisions:

1. Query Strategy: Configure the writer to combine results from both hybrid and naive query modes. This is because LightRAG’s hybrid mode typically encompasses the scope of the global mode.
2. Environment Setup: Update `WORKING_DIR` to `./ag2_demo_workdir` and ensure this directory is added to `.gitignore`.

danielaskdd:
#### Move termination detection off the user proxy
AG2 only stops when the agent about to reply sees a terminating message. Here the Writer is told to emit `TERMINATE`, but the only `is_termination_msg` hook is on User; with `speaker_selection_method="auto"`, the manager can legally pick `Researcher` or `Analyst` next instead. In those runs the chat keeps consuming extra turns and model calls after the final answer instead of ending cleanly.

danielaskdd:
#### Constrain speaker order before allowing Writer to answer

This demo assumes `Researcher` and `Analyst` will query LightRAG before Writer synthesizes the result, but `speaker_selection_method="auto"` does not enforce that sequence. If the manager selects `Writer` early, the example can finish without any retrieval step because Writer has no tool access and is explicitly told to end with TERMINATE, so the advertised multi-agent RAG flow is not reliable.

faridun-m:
@danielaskdd Thanks for the detailed review! All four points have been addressed:

1. **Hybrid already covers global** — Analyst now uses `naive` (vector) mode instead of `global`, giving genuinely complementary results
2. **WORKING_DIR renamed + .gitignore** — Changed to `./ag2_demo_workdir` and added to `.gitignore`
3. **Termination detection** — Added `is_termination_msg` to both UserProxy and GroupChatManager so the chat stops cleanly regardless of which agent is selected next
4. **Speaker order enforced** — Using `allowed_or_disallowed_speaker_transitions` to guarantee Researcher → Analyst → Writer pipeline. Also strengthened Analyst's prompt to always call the tool before responding

Tested end-to-end — all three agents now participate with distinct retrieval modes (hybrid vs naive) and the conversation terminates cleanly.

haoyedangkong:
@codex review

faridun-ag2:
@danielaskdd PR updated with one more fix based on automated review:

- **Tightened speaker transitions**: `user_proxy` can now only transition to `researcher` (was `[researcher, analyst]`). This prevents the manager from skipping Researcher and going directly to Analyst after the initial message. Tool call routing between agents and proxy is handled independently by AG2's `func_call_filter`, so tool execution still works correctly for both Researcher and Analyst.

All four original review points + this transition fix are verified end-to-end. Ready for re-review when you have a chance!

PR #2931, opened 2026-04-12
fix(postgres): parameterize Cypher write queries to prevent injection

sjhddh:
## Summary
- Extend parameterized query pattern from read paths to write paths in PostgreSQL AGE graph storage, preventing potential Cypher injection through crafted entity names or property values
- Remove the TODO comment that acknowledged this gap ("extending parameterization to those paths is tracked as a follow-up task")
- Add defence-in-depth improvements to `_normalize_node_id` for remaining interpolation paths

## Problem
`upsert_node` and `upsert_edge` interpolated entity IDs and properties into Cypher via f-strings:

```python
# BEFORE — entity_id and properties interpolated into Cypher string
cypher_query = f"""MERGE (n:base {{entity_id: "{label}"}})
                 SET n += {properties}
                 RETURN n"""
query = f"SELECT * FROM cypher({_dollar_quote(self.graph_name)}, {_dollar_quote(cypher_query)}) AS (n agtype)"
```

The existing `_normalize_node_id` only escaped `\` and `"`, which is insufficient for full Cypher safety. The code's own TODO comment in `_query`'s docstring acknowledged this gap.

Meanwhile, read paths (e.g., `get_node_edges` at line ~4975) already used proper AGE parameterization with `$entity_id` / `$1::agtype`.

## Changes

```python
# AFTER — entity_id and properties passed as bound parameters
cypher_query = """MERGE (n:base {entity_id: $entity_id})
                 SET n += $props
                 RETURN n"""
query = (
    f"SELECT * FROM cypher("
    f"{_dollar_quote(self.graph_name)}::name, "
    f"{_dollar_quote(cypher_query)}::cstring, "
    f"$1::agtype) AS (n agtype)"
)
pg_params = {"params": json.dumps({"entity_id": node_id, "props": node_data}, ensure_ascii=False)}
```

- `upsert_node`: entity_id and node properties now passed via `$1::agtype`
- `upsert_edge`: source/target entity IDs and edge properties now passed via `$1::agtype`
- `_query` write path: now forwards `params` to `db.execute` via its `data` kwarg
- `_normalize_node_id`: added null-byte stripping (defence-in-depth for delete/remove paths)
- Also fixes a minor issue where `upsert_edge` had a duplicate `SET r += {edge_properties}` line

## Test plan
- [x] 13 new tests covering injection payloads, special characters, unicode, dollar signs
- [x] All 82 existing postgres tests pass (6 skipped — integration tests requiring live DB)
- [x] Code compiles cleanly with `from lightrag.kg.postgres_impl import *`

### DIFF
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -4577,15 +4577,25 @@ class PGGraphStorage(BaseGraphStorage):
         """
         Normalize node ID to ensure special characters are properly handled in Cypher queries.
 
+        Used by write paths that still embed entity IDs in Cypher strings
+        (delete_node, remove_nodes, remove_edges).  The upsert paths now use
+        parameterized Cypher instead.
+
+        Within a Cypher double-quoted string the only recognised escape
+        sequences are ``\\"`` and ``\\\\``.  We also strip null bytes which
+        could truncate the string in some PostgreSQL/AGE code paths.
+
         Args:
             node_id: The original node ID
 
         Returns:
-            Normalized node ID suitable for Cypher queries
+            Normalized node ID suitable for embedding in a Cypher double-quoted string
         """
-        # Escape backslashes
-        normalized_id = node_id
+        # Strip null bytes that could truncate the string
+        normalized_id = node_id.replace("\x00", "")
+        # Escape backslashes first (order matters)
         normalized_id = normalized_id.replace("\\", "\\\\")
+        # Escape double quotes
         normalized_id = normalized_id.replace('"', '\\"')
         return normalized_id
 
@@ -4825,10 +4835,7 @@ class PGGraphStorage(BaseGraphStorage):
             upsert (bool): passed through to db.execute for write operations.
             params (dict | None): AGE agtype parameters for parameterized Cypher
                 (e.g. ``{"params": json.dumps({"entity_id": "..."})}``).
-                Only honoured when ``readonly=True``. Write paths (upsert_node,
-                upsert_edge, delete_node, remove_nodes, remove_edges) still
-                interpolate entity IDs via _normalize_node_id; extending
-                parameterization to those paths is tracked as a follow-up task.
+                Honoured for both read and write paths.
             timing_label (str | None): optional label for performance logging.
 
         Returns:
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
Please resolve all conflicts prior to initiating any review.

sjhddh:
Update after debugging this against the actual PostgreSQL/Apache AGE write path:

The original write-path parameterization idea was only partially compatible with AGE. In particular, AGE accepts bound scalar IDs in the Cypher template, but it rejects a bound agtype object on the right-hand side of `SET +=`.

Observed runtime failure:

```text
FeatureNotSupportedError: SET clause expects a map
```

This was triggered by:

```cypher
MERGE (n:base {entity_id: $entity_id})
SET n += $props
RETURN n
```

and similarly for `upsert_edge` with `SET r += $props`.

Final fix applied:

- Keep graph identifiers parameterized:
  - `upsert_node`: `entity_id` still goes through `$1::agtype`
  - `upsert_edge`: `src_id` / `tgt_id` still go through `$1::agtype`
- Stop binding the whole property map as `$props` on write queries.
- Render node/edge property maps as safely escaped Cypher literals via `_format_properties(...)`, because that is what AGE actually accepts for `SET ... += ...` in this path.
- Exclude `entity_id` from the inlined node property map so the identifier is only supplied through the bound `MERGE` key instead of being duplicated in `SET`.
- Harden `_format_properties(...)` to escape backticks in property keys and preserve Unicode with `ensure_ascii=False`.
- Update regression tests to match AGE’s real constraint and add coverage for special characters / Unicode / backticks in property keys.

Net result:

- entity IDs remain parameterized, so crafted IDs are still treated as data rather than Cypher
- property updates no longer hit the AGE `SET clause expects a map` error
- the write path now matches what AGE can actually execute

Verification run:

```text
./scripts/test.sh tests/test_postgres_cypher_injection.py tests/test_postgres_upsert_edge_cypher.py
17 passed
```

So the final implementation is not “fully parameterized properties on writes”; it is the compatible hybrid approach required by AGE: parameterized IDs + escaped literal property maps.

danielaskdd:
@codex review

PR #2941, opened 2026-04-15
fix: prevent orphan graph edges when VDB upsert fails during merge

### LINKED ISSUE #2917: [Bug/Investigation] AGE_edges count drifts from vdb_relation count after delete/merge operations
## Symptom

After running a series of `DELETE /documents/delete_entity` and `POST /graph/entities/merge` operations against a LightRAG instance using Apache AGE on PostgreSQL, we end up with **more edges in the AGE graph than relations in the vector DB**:

\`\`\`
SELECT 'AGE_vertices', COUNT(*) FROM chunk_entity_relation.\"_ag_label_vertex\"
UNION ALL SELECT 'AGE_edges', COUNT(*) FROM chunk_entity_relation.\"_ag_label_edge\"
UNION ALL SELECT 'vdb_entity', COUNT(*) FROM lightrag_vdb_entity_nomic_embed_text_768d
UNION ALL SELECT 'vdb_relation', COUNT(*) FROM lightrag_vdb_relation_nomic_embed_text_768d;

   ?column?   | count
--------------+-------
 AGE_vertices |  1201
 AGE_edges    |  1616
 vdb_relation |  1244
\`\`\`

`AGE_edges (1616)` vs `vdb_relation (1244)` = **+372 orphan edges** in graph storage with no corresponding VDB row, accumulated over time.

We can reproduce orphans by querying:

\`\`\`sql
SELECT COUNT(*) FROM chunk_entity_relation.\"_ag_label_edge\" e
 WHERE NOT EXISTS (SELECT 1 FROM chunk_entity_relation.\"_ag_label_vertex\" v WHERE v.id = e.start_id)
    OR NOT EXISTS (SELECT 1 FROM chunk_entity_relation.\"_ag_label_vertex\" v WHERE v.id = e.end_id);
\`\`\`

## What we checked

`PGGraphStorage.delete_node` already uses Cypher `DETACH DELETE`:

\`\`\`python
# kg/postgres_impl.py:5128-5147
async def delete_node(self, node_id: str) -> None:
    label = self._normalize_node_id(node_id)
    cypher_query = f\"\"\"MATCH (n:base {{entity_id: \"{label}\"}})
                 DETACH DELETE n\"\"\"
    query = f\"SELECT * FROM cypher({_dollar_quote(self.graph_name)}, {_dollar_quote(cypher_query)}) AS (n agtype)\"
    await self._query(query, readonly=False)
\`\`\`

So the delete code path itself is correct — `DETACH DELETE` should remove the node and its edges atomically. The orphans must come from somewhere else.

## Possible causes (we did not isolate)

- `merge_entities` writes new edges with `upsert_edge` (line 1384 in `utils_graph.py`) and then deletes the source nodes with `delete_node`. If a write succeeds in graph storage but the corresponding VDB row never lands (e.g. transient embedder failure), the graph edge survives but the VDB row never appears.
- `relationships_vdb.delete_entity_relation(entity_name)` may not delete every related VDB row when the entity has many neighbours.
- We hit a separate bug where `merge_entities` aborted mid-flight on Ollama context-length errors (see #2916). It's possible those aborted operations left partial state.

## Workaround we use

Direct SQL cleanup of orphan edges, run when `lightrag.service` is idle:

\`\`\`sql
DELETE FROM chunk_entity_relation.\"_ag_label_edge\" e
 WHERE NOT EXISTS (SELECT 1 FROM chunk_entity_relation.\"_ag_label_vertex\" v WHERE v.id = e.start_id)
    OR NOT EXISTS (SELECT 1 FROM chunk_entity_relation.\"_ag_label_vertex\" v WHERE v.id = e.end_id);
\`\`\`

This removes orphans safely (they cannot be reached by any traversal, since at least one endpoint is missing from the vertex table). After cleanup, `AGE_edges` matches `vdb_relation`.

## Ask

I am **not opening a PR** for this one because I haven't isolated the root cause — `delete_node` already does `DETACH DELETE`, so the regression is somewhere else and I'd be guessing. Filing this as a research issue so maintainers familiar with the merge/upsert paths can take a look. Happy to provide more state from our deployment if it helps.

## Environment

- LightRAG v1.4.13
- PostgreSQL 16 + pgvector 0.6.0 + Apache AGE 1.5.0
- Ollama nomic-embed-text (768d)
- LLM via OmniRoute → Cerebras Qwen 3 235B / Gemini Flash 2.5 / etc.
- Workspace: `chunk_entity_relation`

korbonits:
## Description

Fixes the edge-count drift bug reported in #2917, where the graph-edge count drifts above the VDB-relation count after merge or delete operations.

## Root Cause

In `_merge_entities_impl`, edges are written to the graph (step 6) and then embedded into the relationships VDB one-by-one (step 7). If the VDB upsert fails partway through — e.g. embedder crash, context-length exceeded with a high-degree hub entity, network timeout — the graph already holds the new edges but the VDB does not. The subsequent `DETACH DELETE` of source entities (step 10) then removes the *original* edges, leaving the new edges as orphans with no VDB counterpart. Over time this causes visible drift (reported: 372 orphan edges on a live instance).

## Changes Made

**`lightrag/utils_graph.py`**
- Wrap each per-edge `relationships_vdb.upsert()` call in `try/except` inside `_merge_entities_impl`
- Collect failed edges in `vdb_failed_edges`; after the loop call `remove_edges(vdb_failed_edges)` to roll back the matching graph writes — ensuring graph and VDB are always in sync (either both have the edge, or neither does)
- Fix long-standing typo `"updatign"` → `"updating"` in log message
- Add `check_graph_consistency(graph, vdb)` — compares all graph edges against VDB entries via `get_all_edges()` + `get_by_ids()`, returns `{orphan_graph_edges, total_graph_edges, total_vdb_relations}`; handles PostgreSQL/AGE double-quoted `entity_id` values transparently
- Add `repair_graph_consistency(graph, vdb, *, dry_run=False)` — calls the check then removes orphans via `remove_edges`; `dry_run=True` for safe inspection before committing changes

**`lightrag/lightrag.py`**
- Expose both utilities as `rag.acheck_graph_consistency()`, `rag.arepair_graph_consistency(dry_run=True)`, and their synchronous wrappers on the `LightRAG` class

**`tests/test_graph_consistency.py`** (new file)
- 8 unit tests covering: no-edges, all-present, orphan detection, AGE quote stripping, dry-run, repair, no-op repair, and VDB-failure rollback in `_merge_entities_impl`

## Related Issues

Closes #2917

## Usage (for existing drift)

```python
# Inspect without changing anything
report = await rag.acheck_graph_consistency()
print(f"Orphan edges: {len(report['orphan_graph_edges'])}")

# Fix
report = await rag.arepair_graph_consistency(dry_run=True)  # preview
report = await rag.arepair_graph_consistency()               # apply
print(f"Repaired: {report['repaired']}")
```

## Checklist

- [x] Changes tested locally
- [x] Pre-commit checks pass (`pre-commit run --all-files`)
- [x] Unit tests added
- [x] Documentation updated (docstrings + `LightRAG` method docstrings with usage examples)


danielaskdd:
@codex review

Jasonleonardvolk:
Related question: beyond orphan edges from failed upserts, do you
see cases where the merge itself creates structurally contradictory
edges (e.g., two entities with conflicting claims merged into the
same graph)? I've been working on deterministic structural
contradiction detection for graph databases:
https://github.com/Jasonleonardvolk/sigma-guard

danielaskdd:
Thank you for this PR — the root cause analysis is correct and was genuinely helpful: the per-edge VDB upserts in `_merge_entities_impl` step 7 can fail after the graph has already been updated in step 6, and the source-entity deletion in step 10 then turns the new edges into orphans with no vector counterpart. That is exactly the mechanism behind the drift in #2917.

After careful consideration we've decided not to pursue the runtime rollback / incremental repair approach, and are closing this PR in favor of a different strategy:

1. **Runtime rollback cannot guarantee consistency.** A merge is a multi-step write across two stores (node deletion, edge re-pointing, attribute merging). The space of partial-failure states is combinatorial — e.g. the entity vector upsert succeeds while an edge vector upsert fails — and the rollback writes themselves run against the same storage that just failed, so they can fail too. Without a distributed transaction there is no reliable runtime recovery point.
2. **Removing graph edges is the wrong direction.** The knowledge graph is the authoritative, complete data source — `local`-mode queries read edges directly from the graph. Rolling back graph edges to match a failed VDB write deletes authoritative data to match a derived store, and `remove_edges` on the merge target can also remove pre-existing healthy edges that were merged into the new edge set.
3. **Single-direction ID matching in the consistency check misjudges.** Relation VDB ids depend on endpoint order; legacy custom-KG imports hashed the original order, so a check that probes only the normalized id reports false orphans (and a repair would then delete healthy graph edges).

The new direction (implemented in #3243) is **fail loud + offline full rebuild**: a VDB write failure during merge now raises a clear `VectorStorageConsistencyError` (graph kept authoritative, source entities never deleted on the failure path), and a new offline tool `lightrag-rebuild-vdb` drops and rebuilds the vector storages from the graph and the `text_chunks` KV store — which also clears stale VDB-only records that incremental repair cannot safely handle, and includes a read-only consistency check mode similar in spirit to the one proposed here (probing both candidate relation ids).

Thanks again for the diagnosis and the report — closing this PR in favor of #3243.

---

PR #2943, opened 2026-04-16
feat(api): expose split_by_character in /documents/text and /documents/texts

### LINKED ISSUE #2942: Expose split_by_character parameter in HTTP API /documents/text endpoint
## Problem

The Python API `LightRAG.ainsert()` supports `split_by_character` and `split_by_character_only` parameters for custom chunk splitting. However, the HTTP API endpoints `/documents/text` and `/documents/texts` in `lightrag/api/document_routes.py` do not expose these parameters — they are not included in the Pydantic request models and not passed through to `ainsert()`.

This forces HTTP API users to rely solely on the built-in token-based chunker, even when they have pre-chunked content with a known separator.

## Use Case

We pre-chunk documents with a semantic chunker (heading-aware, with breadcrumbs and atomic blocks) before sending to LightRAG. We join chunks with a unique separator and want LightRAG to split on it, preserving our chunk boundaries as-is.

Without `split_by_character` in the HTTP API, the only options are:
1. Send each chunk as a separate document (`/documents/texts` with N items) — creates N `doc_id`s per file, breaks deletion, deduplication, and `doc_status` tracking.
2. Use the Python API directly — not possible when LightRAG runs as a separate service.

## Proposed Change

Add `split_by_character` and `split_by_character_only` fields to `InsertTextRequest` and `InsertTextsRequest` in `document_routes.py`, and pass them through to `rag.ainsert()`.

### InsertTextRequest

```python
class InsertTextRequest(BaseModel):
    text: str
    # ... existing fields ...
    split_by_character: Optional[str] = Field(
        default=None,
        description="Character(s) to split the text on instead of token-based chunking",
    )
    split_by_character_only: bool = Field(
        default=False,
        description="If True, split only on split_by_character without token-based fallback",
    )
```

### Route handler

```python
await rag.ainsert(
    request.text,
    split_by_character=request.split_by_character,
    split_by_character_only=request.split_by_character_only,
)
```

Same for `InsertTextsRequest` / `/documents/texts`.

## Notes

- Fully backward compatible — both fields are optional with defaults matching current behavior.
- We have a working patch in production (LightRAG v1.4.14) and can submit a PR.

gn1024:
## Summary

Expose `split_by_character` and `split_by_character_only` parameters in the HTTP API endpoints `/documents/text` and `/documents/texts`.

These parameters are already supported by the Python API (`ainsert`, `apipeline_process_enqueue_documents`) but were not exposed in the HTTP route models, forcing HTTP clients to use the built-in token-based chunker.

## Changes

- Added `split_by_character` and `split_by_character_only` fields to `InsertTextRequest` and `InsertTextsRequest` Pydantic models
- Pass these fields through `pipeline_index_texts` to `apipeline_process_enqueue_documents`

## Use Case

External chunkers that pre-chunk documents with semantic awareness (heading-based splitting, breadcrumbs, atomic blocks) can join chunks with a unique separator and have LightRAG split on it — preserving chunk boundaries as-is while maintaining a single `doc_id` per file.

## Backward Compatibility

Both fields are `Optional` with defaults matching current behavior (`None` / `False`). No breaking changes.

Closes #2942

### DIFF
--- a/lightrag/api/routers/document_routes.py
+++ b/lightrag/api/routers/document_routes.py
@@ -240,6 +240,18 @@ class InsertTextRequest(BaseModel):
     file_source: Optional[str] = Field(
         default=None, min_length=0, description="File Source"
     )
+    split_by_character: Optional[str] = Field(
+        default=None,
+        description="Character(s) to split the text on instead of token-based chunking. "
+        "When set, the text is split on this separator and each piece becomes a chunk. "
+        "Useful when sending pre-chunked content joined by a known delimiter.",
+    )
+    split_by_character_only: bool = Field(
+        default=False,
+        description="If True, split only on split_by_character without token-based fallback. "
+        "If False (default), chunks that exceed chunk_token_size after character splitting "
+        "are further split by the token-based chunker.",
+    )
 
     @field_validator("text", mode="after")
     @classmethod
@@ -256,6 +268,8 @@ class InsertTextRequest(BaseModel):
             "example": {
                 "text": "This is a sample text to be inserted into the RAG system.",
                 "file_source": "Source of the text (optional)",
+                "split_by_character": None,
+                "split_by_character_only": False,
             }
         }
     )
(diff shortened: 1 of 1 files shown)


danielaskdd:
This issue was resolved by v1.5.0rc2.

PR #2949, opened 2026-04-18
perf(postgres): use binary parameter for vector similarity queries

wkpark:
## Description

Replace string-interpolated embedding literals in vector similarity SQL queries with a positional `$4` binary parameter. The query embedding is now transmitted via asyncpg's `register_vector` binary codec instead of being serialized to a comma-separated text literal and parsed back by PostgreSQL on every query.

## Changes Made

- **`lightrag/kg/postgres_impl.py`**
  - Removed `embedding_string` interpolation from `SQL_TEMPLATES` (`relationships`, `entities`, `chunks`)
  - Replaced `'[{embedding_string}]'::{vector_cast}` with `$4::{vector_cast}` in all three query templates
  - Added `"embedding": embedding` to the params dict passed to `db.query()`
  - Removed per-query `",".join(map(str, embedding))` serialization call
  - Cleaned up unnecessary table alias prefixes (`e.`, `r.`, `c.`)

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Additional Notes

With the old approach, a 1024-dim embedding was serialized to ~7 KB of text per query, sent to PostgreSQL, and re-parsed into a vector on every call. With `$4` binary parameter + `register_vector` codec, asyncpg transmits the raw float bytes directly — no text serialization, no server-side text→vector parsing.

Informal benchmarking showed a measurable latency improvement per query, particularly at higher vector dimensions.

Initial version drafted with Gemini-cli; refined with Claude code.

### DIFF
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -3493,22 +3493,22 @@ class PGVectorStorage(BaseVectorStorage):
             )  # higher priority for query
             embedding = embeddings[0]
 
-        embedding_string = ",".join(map(str, embedding))
-
+        # Use positional $4 parameter instead of string-interpolated literal.
+        # asyncpg sends the embedding via register_vector binary codec, avoiding
+        # per-query text serialization and PostgreSQL text-to-vector parsing.
         vector_cast = (
             "halfvec"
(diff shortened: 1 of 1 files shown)


danielaskdd:
@codex review

Issue #1353, opened 2025-04-11
[Feature Request]: Multiple Prompt.py files for different use-cases


frederikhendrix:
### Do you need to file a feature request?

- [ ] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

### Feature Request Description

Currently I am planning on using the LightRAG implementation for multiple different use-cases. To guide the AI a bit better for my specific use-case I change the prompt, the examples, the rag_response and this ultimately leads to a better result.

But when I want to use it for a different use-case I'm not able to have a second or third prompt.py file which I can insert at different endpoints.

Maybe it is an idea to add to query_param variable the name of the prompt.py file you want to use, also to the settings and set it to default there?

And then do something like this: 

**Directory structure:**

```
my_application/
└── prompts/
    ├── __init__.py
    ├── default.py
    └── alternative.py
```

**prompts/__init__.py:**

```python
import os
from typing import Any, Dict
from . import default, alternative

def get_prompts(prompt_type: str = None) -> Dict[str, Any]:
    # Use the provided prompt_type, or fall back to the environment variable if not provided.
    style = (prompt_type if prompt_type is not None else os.getenv("PROMPTS_STYLE", "default")).lower()
    
    if style == "alternative":
        return alternative.PROMPTS
    return default.PROMPTS

GRAPH_FIELD_SEP = default.GRAPH_FIELD_SEP  # You can also decide based on style if needed.
```

Now you can simply import from the package:

```python
from prompts import GRAPH_FIELD_SEP, get_prompts

# Using default (either through env or as fallback)
PROMPTS = get_prompts()

# Explicitly using the alternative configuration
PROMPTS_alt = get_prompts(prompt_type="alternative")
```

This way depending on your use-case you can opt for a specific prompt.py file with desciptive prompts.

danielaskdd:
Multiple prompt templates also applied during the document indexing stage. Do we have a comprehensive and user-friendly solution that allows users to select different templates during both indexing and query phases?

danielaskdd:
**Proposal: Three Methods for Prompt Template Selection**  

I recommend implementing the following three approaches to select prompt templates by name:  

1. **Environment Variable**: Configure via `DEFAULT_PROMPT_TEMPLATE`.  
2. **API Endpoint**: Expose an API to allow users to dynamically switch the active template.  
[...]

PR #2951, opened 2026-04-18
feat: add per-instance prompt overrides via LightRAG.prompts field

wkpark:
## Description

Add per-instance prompt override support to LightRAG. Callers can now pass `prompts={...}` to the `LightRAG` constructor to override specific prompt templates without mutating the global `lightrag.prompt.PROMPTS` dict.
This eliminates cross-contamination between multiple `LightRAG` instances running in the same process with different prompt configurations.

## Related Issues

See also https://github.com/HKUDS/LightRAG/issues/1353

## Changes Made

- **`lightrag/lightrag.py`**: Add `prompts: dict[str, Any]` field to the `LightRAG` dataclass (`default_factory=dict`). Included in `global_config` automatically via `asdict(self)`.
- **`lightrag/operate.py`**: Add `_get_prompt(key, global_config=None)` helper. Checks `global_config["prompts"]` first; falls back to global `PROMPTS` if no override is set. Replace all direct `PROMPTS["..."]` accesses in extraction and query paths with `_get_prompt("...", global_config)`.

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Additional Notes

**Backward compatible.** Instances that do not pass `prompts=` behave identically to before — `_get_prompt()` falls back to the global `PROMPTS` dict unconditionally when the per-instance override dict is empty.

Two delimiter constants (`DEFAULT_TUPLE_DELIMITER`, `DEFAULT_COMPLETION_DELIMITER`) in `_rebuild_from_extraction_result()` still reference `PROMPTS` directly, as that function has no access to `global_config`. Overriding these two keys via `prompts=` is therefore not supported and would cause a parsing mismatch against cached extraction results. All other prompt keys are fully overridable.

Note: prepared with claude code

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -610,6 +610,9 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
         repr=False,
     )
 
+    prompts: dict[str, Any] = field(default_factory=dict)
+    """Per-instance prompt overrides. Keys matching PROMPTS dict take precedence over the global PROMPTS."""
+
     # Storages Management
     # ---
 
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -56,7 +56,13 @@ from lightrag.base import (
     QueryContextResult,
 )
 from lightrag.chunk_schema import strip_internal_multimodal_markup_for_extraction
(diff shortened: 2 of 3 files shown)


wkpark:
@codex review

PR #2972, opened 2026-04-24
fix(anthropic): make Anthropic/Claude provider usable out-of-box (three bugs)

### LINKED ISSUE #2955: Anthropic/Claude provider: 3 bugs prevent out-of-box usage (v1.4.14)
# Anthropic/Claude provider: 3 bugs prevent out-of-box usage (v1.4.14)

_Ready-to-paste issue for https://github.com/HKUDS/LightRAG/issues_

---

### Environment

- `lightrag-hku==1.4.14` (from PyPI)
- `anthropic` Python SDK (current at time of filing)
- Python 3.11+
- macOS (reproducible on Linux too — not platform-specific)

### Summary

`lightrag/llm/anthropic.py` has three separate defects that each independently prevent `anthropic_complete` from working with the current Anthropic Python SDK and LightRAG's own pipeline. End-to-end the provider is unusable without a wrapper.

---

### Bug 1 — `voyageai` imported unconditionally at module load

[`lightrag/llm/anthropic.py#L19-L21`](https://github.com/HKUDS/LightRAG/blob/main/lightrag/llm/anthropic.py#L19-L21):

```python
if not pm.is_installed("voyageai"):
    pm.install("voyageai")
import voyageai
```

The `pipmaster` auto-install silently no-ops inside a uv-managed venv (and in any env where the process doesn't have site-packages write permission), so the subsequent `import voyageai` fails with `ModuleNotFoundError`. Users who only want the LLM side (not Voyage embeddings) are blocked.

**Fix:** move `import voyageai` inside `anthropic_embed` so it's only required when embeddings are actually used, or list `voyageai` as an explicit extra/dep in `pyproject.toml`.

---

### Bug 2 — `max_tokens` is required by the Anthropic SDK but never set

[`lightrag/llm/anthropic.py#L113`](https://github.com/HKUDS/LightRAG/blob/main/lightrag/llm/anthropic.py#L113):

```python
create_params = {"model": model, "messages": messages, "stream": True, **kwargs}
```

`anthropic.messages.create()` requires `max_tokens`, but the wrapper neither sets a default nor documents a required kwarg. Every call fails with:

```
TypeError: Missing required arguments; Expected either ('max_tokens', 'messages' and 'model')
or ('max_tokens', 'messages', 'model' and 'stream') arguments to be given
```

The error message is confusing because `model` and `messages` ARE present — only `max_tokens` is missing.

**Fix:** `kwargs.setdefault("max_tokens", 8192)` (or similar) inside `anthropic_complete_if_cache` before building `create_params`.

---

### Bug 3 — `stream=True` hardcoded; function returns async generator even when LightRAG expects a string

[`lightrag/llm/anthropic.py#L113, L144-L163`](https://github.com/HKUDS/LightRAG/blob/main/lightrag/llm/anthropic.py#L113):

```python
create_params = {"model": model, "messages": messages, "stream": True, **kwargs}
...
async def stream_response():
    async for event in response:
        ...
return stream_response()
```

`anthropic_complete` always returns an `AsyncIterator[str]`. LightRAG's own KG extraction pipeline (called via `rag.ainsert(...)`) expects a plain `str`. Result:

```
TypeError: expected string or bytes-like object, got 'async_generator'
```

Additionally: when caller passes `stream=False`, the `**kwargs` spread overrides the hardcoded `stream=True`, the SDK returns a `Message` object (not an iterator), and `stream_response` crashes with:

```
TypeError: 'async for' requires an object with __aiter__ method, got Message
```

So both the default path (iterator when string expected) and the override path (non-iterator when iterator expected) are broken.

**Fix:** make streaming opt-in. Either (a) default to `stream=False` and return `response.content[0].text`, or (b) detect the caller's intent and return the right shape. Match `openai_complete`'s behavior for consistency.

---

### Minimal reproducer

```python
import asyncio
from lightrag import LightRAG
from lightrag.llm.anthropic import anthropic_complete
from lightrag.utils import EmbeddingFunc
import numpy as np

async def dummy_embed(texts):
    return np.zeros((len(texts), 1024), dtype=np.float32)

async def main():
    rag = LightRAG(
        working_dir="./scratch",
        llm_model_func=anthropic_complete,
        llm_model_name="claude-sonnet-4-5-20250929",
        embedding_func=EmbeddingFunc(embedding_dim=1024, max_token_size=8192, func=dummy_embed),
    )
    from lightrag.kg.shared_storage import initialize_pipeline_status
    await rag.ainsert("LightRAG is a graph-augmented RAG system.")

asyncio.run(main())
```

Set `ANTHROPIC_API_KEY`, run, and observe the `TypeError: Missing required arguments` (Bug 2) followed by `expected string or bytes-like object, got 'async_generator'` (Bug 3).

### Workaround (community)

I published a drop-in wrapper that bypasses the broken path by calling the Anthropic SDK directly:
https://gist.github.com/peterchoi1014/3bb3cf1fa316619673174165d7bbc14c

Happy to open a PR fixing all three against `main` if the maintainers agree on the approach — let me know which shape you prefer for Bug 3 (default stream=False vs dual-mode).

---

### Also: undocumented initialization

Not a bug, but a sharp edge: `rag.ainsert()` fails with `StorageNotInitializedError("JsonDocStatusStorage")` unless the caller also runs:

```python
from lightrag.kg.shared_storage import initialize_pipeline_status
```

The `LightRAG(...)` snippets in the README do not show this, and the failure mode is opaque. A README update or an auto-init in `__init__` / `__post_init__` would save a lot of first-use pain.

he-yufeng:
Fixes #2955.

The Anthropic/Claude provider in \`lightrag/llm/anthropic.py\` had three independent defects that each blocked \`anthropic_complete\` end-to-end under the current Anthropic SDK and LightRAG's own KG extraction pipeline. All three are addressed here in one file.

## 1. \`voyageai\` imported unconditionally at module load

\`pipmaster\`'s auto-install silently no-ops inside a uv-managed venv (or anywhere without site-packages write permission), so the subsequent \`import voyageai\` at module top level raised \`ModuleNotFoundError\` for users who never touched Voyage embeddings at all. Moved the \`install + import voyageai\` pair into \`anthropic_embed\` so it's only required when embeddings are actually used.

## 2. \`max_tokens\` missing

\`anthropic.messages.create()\` requires \`max_tokens\`, but the wrapper never set a default and never surfaced it as a documented kwarg. Every call failed with:

\`\`\`
TypeError: Missing required arguments; Expected either ('max_tokens', 'messages' and 'model')
or ('max_tokens', 'messages', 'model' and 'stream') arguments to be given
\`\`\`

Added \`kwargs.setdefault("max_tokens", 8192)\` — matches the other providers' generous default and callers can still override via \`**kwargs\`.

## 3. \`stream=True\` hardcoded

The wrapper always built \`create_params\` with \`"stream": True\` and returned an \`AsyncIterator[str]\`, so LightRAG's KG extraction (which calls the LLM expecting a \`str\`) crashed with:

\`\`\`
TypeError: expected string or bytes-like object, got 'async_generator'
\`\`\`

A caller that overrode with \`stream=False\` via \`**kwargs\` landed on the reverse failure — the SDK returned a \`Message\` and the code's \`async for event in response\` blew up with \`'async for' requires an object with __aiter__ method, got Message\`.

Made streaming opt-in: pop \`stream\` out of \`kwargs\`, thread it through explicitly, and return concatenated \`content[*].text\` when \`stream=False\` (matching \`openai_complete\`). The existing async-generator path stays intact for \`stream=True\`.

## Test plan

- Repro from the issue (\`rag.ainsert(...)\` with \`anthropic_complete\`) now completes; previously raised at step 2 (\`max_tokens\` missing) and step 3 (iterator vs string).
- \`stream=True\` still yields chunks in the pre-existing shape.
- \`anthropic_embed\` still installs/imports voyageai on first use; LLM-only callers no longer hit \`ModuleNotFoundError\`.
- \`ruff check .\` clean on the touched file.

### DIFF
--- a/lightrag/llm/anthropic.py
+++ b/lightrag/llm/anthropic.py
@@ -104,6 +104,9 @@ async def anthropic_complete_if_cache(
         )
     kwargs.pop("response_format", None)
     timeout = kwargs.pop("timeout", None)
+    # Anthropic requires this, while LightRAG callers usually don't pass it.
+    kwargs.setdefault("max_tokens", 8192)
+    stream_requested = kwargs.pop("stream", False)
 
     anthropic_async_client = (
         AsyncAnthropic(
@@ -129,7 +132,12 @@ async def anthropic_complete_if_cache(
     verbose_debug(f"System prompt: {system_prompt}")
 
     try:
-        create_params = {"model": model, "messages": messages, "stream": True, **kwargs}
+        create_params = {
+            "model": model,
+            "messages": messages,
+            "stream": stream_requested,
+            **kwargs,
+        }
         if system_prompt:
             create_params["system"] = system_prompt
         response = await anthropic_async_client.messages.create(**create_params)
@@ -159,6 +167,19 @@ async def anthropic_complete_if_cache(
             f"Anthropic API Call Failed,\nModel: {model},\nParams: {kwargs}, Got: {e}{extra}"
         )
         raise
+
+    if not stream_requested:
+        # Non-streaming: return the plain string LightRAG's KG pipeline expects.
+        content_blocks = getattr(response, "content", None) or []
+        text_parts = [
+            getattr(block, "text", "")
+            for block in content_blocks
+            if getattr(block, "type", None) == "text"
+        ]
+        text = "".join(text_parts)
+        if r"\u" in text:
+            text = safe_unicode_decode(text.encode("utf-8"))
+        return text
 
     async def stream_response():
         try:
--- a/tests/test_anthropic_llm.py
+++ b/tests/test_anthropic_llm.py
@@ -1,0 +1,118 @@
+from __future__ import annotations
+
+import asyncio
+import importlib
+import sys
+from types import ModuleType, SimpleNamespace
+
+import pytest
+
+
+class _FakeAnthropicError(Exception):
+    pass
+
+
+class _FakeMessages:
+    def __init__(self, calls: list[dict], response):
+        self._calls = calls
+        self._response = response
+
+    async def create(self, **kwargs):
+        self._calls.append(kwargs)
+        return self._response
+
+
+class _FakeStream:
+    def __init__(self, chunks: list[str]):
+        self._chunks = chunks
+
+    async def __aiter__(self):
+        for chunk in self._chunks:
+            yield SimpleNamespace(delta=SimpleNamespace(text=chunk))
+
+
+def _load_anthropic_module(monkeypatch, response):
+    calls: list[dict] = []
+    clients: list[dict] = []
+
+    class FakeAsyncAnthropic:
+        def __init__(self, **kwargs):
+            clients.append(kwargs)
+            self.messages = _FakeMessages(calls, response)
+
+    fake_pm = SimpleNamespace(is_installed=lambda name: True, install=lambda name: None)
+    fake_tenacity = SimpleNamespace(
+        retry=lambda **_kwargs: lambda func: func,
+        stop_after_attempt=lambda *_args, **_kwargs: None,
+        wait_exponential=lambda *_args, **_kwargs: None,
+        retry_if_exception_type=lambda *_args, **_kwargs: None,
+    )
+    fake_anthropic = ModuleType("anthropic")
+    fake_anthropic.AsyncAnthropic = FakeAsyncAnthropic
+    fake_anthropic.APIConnectionError = _FakeAnthropicError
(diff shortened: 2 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
The `dev` branch has undergone significant changes related to LLM invocation. Please rebase your current modifications onto `dev`, conduct thorough debugging and testing, and then submit a pull request targeting the `dev` branch.

PR #2977, opened 2026-04-25
Fix docker-compose compatibility with Podman

tears710:
## Summary

Replace Docker-specific compose syntax with Podman-compatible alternatives while maintaining full backward compatibility with Docker Compose.

## Changes

- **Replace deploy.restart_policy with top-level restart directive**
  Podman does not support the deploy block for restart policies. The restart: on-failure:10 syntax is part of the Compose specification and works with both Docker and Podman.

- **Remove extra_hosts with host-gateway special value**
  Docker auto-resolves host.docker.internal on macOS/Windows without explicit configuration. Podman uses host.containers.internal and fails on the host-gateway special value, producing: unable to replace host-gateway of host entry: host containers internal IP address is empty. Removing this entry works for both runtimes since each provides its own host-to-container DNS resolution by default.

- **Parameterize MEMGRAPH_URI host via MEMGRAPH_HOST env var**
  Changed from hardcoded host.docker.internal to ${MEMGRAPH_HOST:-host.docker.internal}, allowing Podman users to set MEMGRAPH_HOST=host.containers.internal without modifying the compose file.

## Affected Files
- docker-compose.yml
- docker-compose-full.yml

## Testing
- Verified with podman-compose up -d — container starts successfully
- Verified with docker compose config — YAML validates correctly
- Health check returns healthy status

## Operational Impact
No breaking changes. Existing Docker Compose users are unaffected — all modified values maintain the same default behavior.

### DIFF
--- a/docker-compose-full.yml
+++ b/docker-compose-full.yml
@@ -16,12 +16,7 @@ services:
       - ./data/inputs:/app/data/inputs
       - ./config.ini:/app/config.ini
       - ./.env:/app/.env
-    deploy:
-      restart_policy:
-        condition: on-failure
-        max_attempts: 10
-    extra_hosts:
-      - "host.docker.internal:host-gateway"
+    restart: on-failure:10
     environment:
       HOST: "0.0.0.0"
       PORT: "9621"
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #2990, opened 2026-04-27
fix: preserve custom node/edge attributes during merge operations

abc-lee:
## Summary

Fix data loss bug where custom node/edge attributes are silently discarded during merge operations.

## The Bug

`_merge_nodes_then_upsert` and `_merge_edges_then_upsert` reconstruct `node_data`/`edge_data` from scratch with only 7 hardcoded fields. This silently discards any custom attributes (e.g. `brain_meta_*`, `community_id`, or any user-extended fields) added by downstream users.

This is inconsistent with:
- `aedit_entity` which uses `{**node_data, **updated_data}` pattern
- `amerge_entities` which explicitly collects all keys via `_merge_attributes`

## The Fix

Instead of building a new dict from scratch:
1. Start from the existing node/edge data dict
2. Update with standard fields (source_type, updated_at, chunk_id_list, etc.)
3. Custom attributes are preserved

This approach is minimal, targeted, and maintains backwards compatibility.

## Testing

- [x] All existing tests pass
- [x] The fix has been validated in the niu-agent downstream project

### DIFF
--- a/lightrag/kg/networkx_impl.py
+++ b/lightrag/kg/networkx_impl.py
@@ -25,6 +25,16 @@ load_dotenv(dotenv_path=".env", override=False)
 @dataclass
 class NetworkXStorage(BaseGraphStorage):
     @staticmethod
+    def _normalize_node_id(node_id: str) -> str:
+        """Normalize node ID for case-insensitive matching.
+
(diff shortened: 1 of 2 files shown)


Konsilion:
+1

PR #3018, opened 2026-05-04
Bedrock - pull default environment for running bedrock without external environment variables

curtislisle:
## Description

This change request modifies Dockerfile.lite and adds an env.bedrock.dev file so the default configuration running the container comes up using the Bedrock Nova2-lite model from AWS.  The Dockerfile was modified to copy over the minimal environment file to .env inside the container.


curtislisle:
pulling in environment definitions so default container runs inside an AWS instance properly configured for bedrock access via IAM roles.

PR #3020, opened 2026-05-05
fix(api): prevent cross-workspace info disclosure via LIGHTRAG-WORKSPACE header on /health

sebastiondev:
## Summary

The `/health` endpoint accepts a `LIGHTRAG-WORKSPACE` request header and uses it to look up the pipeline status for an arbitrary workspace name. Because `/health` is in the default `WHITELIST_PATHS` (`/health,/api/*`), it is reachable without authentication. An unauthenticated attacker can:

1. Probe whether arbitrary workspace names exist on the server (a non-existent or uninitialized workspace returns a 500 with `PipelineNotInitializedError`; an existing one returns 200).
2. Read the `pipeline_busy` boolean of any other workspace's pipeline.

Additionally, the existing sanitizer replaces non-alphanumeric characters with `_`, so a header value like `secret-internal` is silently rewritten to `secret_internal`. Any naive `equals` check on the raw header would have been bypassable by this near-miss mapping. The fix correctly compares **after** sanitization.

- **CWE:** CWE-285 (Improper Authorization) / CWE-200 (Information Exposure). The original triage label of CWE-22 (path traversal) is a misnomer — there is no filesystem traversal involved.
- **Severity:** Low. The leaked data is limited to workspace existence and a single boolean; no documents, configuration, or credentials are exposed via this path.
- **Affected code:** `get_workspace_from_request` and the `/health` handler in `lightrag/api/lightrag_server.py`.
- **Data flow:** unauthenticated HTTP request → `LIGHTRAG-WORKSPACE` header → `get_workspace_from_request` → `get_namespace_data("pipeline_status", workspace=...)` → response body.

## Fix

In `get_workspace_from_request`, after sanitizing the header value, compare it to the server's configured workspace (`args.workspace`, falling back to `get_default_workspace()`). If they differ, log the attempt and return HTTP 403. The `/health` handler is updated to re-raise `HTTPException` so the 403 isn't swallowed and converted to a 500 by the existing generic exception handler.

Why this scope: `get_workspace_from_request` is only consumed by `/health`. All other routers operate on the server-bound `rag.workspace` set at startup, so they cannot be steered by this header. The change is therefore intentionally narrow — 18 lines in the server module.

## Tests

Added `tests/test_workspace_header_authz.py` with three assertions against a real `TestClient`:

1. `LIGHTRAG-WORKSPACE: secret_internal` → **403** (direct cross-workspace request rejected).
2. `LIGHTRAG-WORKSPACE: secret-internal` → **403** (sanitization-based bypass rejected; the dash would otherwise normalize to underscore and collide with `secret_internal`).
3. `LIGHTRAG-WORKSPACE: production` (the configured workspace) → **200** with `pipeline_busy: false` (legitimate use still works).

The test pre-creates a `secret_internal` pipeline namespace with `busy=True` to confirm the attacker cannot read it after the fix. All three pass locally.

## Security analysis

We verified exploitability against the default configuration: `WHITELIST_PATHS` includes `/health`, so no API key or JWT is needed to reach the vulnerable code. With network access alone, an attacker can enumerate workspace names (binary signal via 200 vs 500) and observe pipeline busy state — useful for reconnaissance and timing of further attacks against a multi-tenant deployment.

The fix closes the only call site that consumes the attacker-controlled header. Other endpoints already use the server-bound `rag.workspace` and are unaffected.

### Adversarial review

Before submitting, we tried to disprove this finding. We checked whether the auth middleware would block the request (it doesn't — `/health` is whitelisted by default), whether the sanitization layer alone was sufficient (it wasn't — it permits lossy collisions like `secret-internal` → `secret_internal`), and whether other routers might offer the same primitive (they don't — they ignore the header and use the server-bound workspace). We also considered whether an operator could simply remove `/health` from the whitelist as a workaround; they can, but the default config is the realistic threat model and shouldn't leak cross-tenant state. The leaked information is admittedly small (existence + a boolean), which is why we've labeled this low severity rather than something more dramatic.

cc @lewiswigmore

### DIFF
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -471,6 +471,10 @@ def create_app(args):
 
         Returns:
             Workspace identifier (may be empty string for global namespace)
+
+        Raises:
+            HTTPException: 403 if the requested workspace does not match the
+                server's configured workspace.
         """
         # Check custom header first
         workspace = request.headers.get("LIGHTRAG-WORKSPACE", "").strip()
@@ -485,6 +489,18 @@ def create_app(args):
                     f"Sanitized to '{sanitized}'."
                 )
                 workspace = sanitized
+
+            server_workspace = args.workspace or get_default_workspace() or ""
+            if workspace != server_workspace:
+                logger.warning(
+                    "Unauthorized workspace access attempt: "
+                    f"requested '{workspace}', "
+                    f"server workspace is '{server_workspace}'."
+                )
+                raise HTTPException(
+                    status_code=403,
+                    detail="Unauthorized workspace",
+                )
 
         return workspace
 
@@ -1384,6 +1400,8 @@ def create_app(args):
                 "webui_title": webui_title,
                 "webui_description": webui_description,
             }
+        except HTTPException:
+            raise
         except Exception as e:
             logger.error(f"Error getting health status: {str(e)}")
             raise HTTPException(status_code=500, detail=str(e))
--- a/tests/test_workspace_header_authz.py
+++ b/tests/test_workspace_header_authz.py
@@ -1,0 +1,102 @@
+"""Regression test for cross-workspace access via LIGHTRAG-WORKSPACE."""
+
+import argparse
+import asyncio
+import os
+from unittest.mock import AsyncMock, patch
+
+import pytest
+from fastapi.testclient import TestClient
+
+from lightrag.api.config import initialize_config
+
+pytestmark = pytest.mark.offline
+
+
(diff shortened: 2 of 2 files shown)


lewiswigmore:
Closing this to reduce the open-PR pile-up — we have multiple outstanding security contributions to this repo and that volume is not fair on your review queue. Keeping #3026 (fix(opensearch): escape wildcard metacharacters in search_labels to prevent DoS ) as the primary one to focus attention on.

Happy to revisit this finding separately later if it is still relevant. Apologies for the noise.

PR #3025, opened 2026-05-06
Replace status tooltip with details modal

g2424303264-code:
## Description

Replace the document status hover tooltip with a click-triggered details dialog. The dialog displays the existing status information in a scrollable content area and includes a dedicated copy button for long error messages or metadata.

## Related Issues

N/A

## Changes Made

- Added a document status details dialog opened from the status icon.
- Added a scrollable content area for status details.
- Added a dedicated copy button with success and failure feedback.
- Added localized strings for the new dialog title, tooltips, and copy messages.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Frontend checks were run with `bun test` and `bun run build` in `lightrag_webui/`.


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
Pushed a follow-up commit (0496d44b) addressing the review points:

1. **Speculative type-cast removed** — `as DocStatusResponse & { extraction_format?: unknown; engine?: unknown }` is gone. These fields aren't declared on `DocStatusResponse` and no backend route emits them; if they ever start being sent, prefer adding them to the type.
2. **Tooltip flash on dialog close fixed** — added `onCloseAutoFocus={(e) => e.preventDefault()}` to `DialogContent`. Root cause: Radix `Dialog` restores focus to the trigger button on close, and Radix `TooltipTrigger` listens to `focus` for keyboard a11y, so the focus return triggered the tooltip every time. Suppressing the auto-focus restore fixes the flash without breaking keyboard nav (Tab still works).
3. **Trigger nesting simplified** — collapsed `<TooltipProvider><Tooltip><TooltipTrigger asChild><DialogTrigger asChild><Button>` into `<DialogTrigger asChild><Button tooltip={openLabel} side=\"top\">`, reusing the existing `Button.tooltip` pattern (same as `UploadDocumentsDialog`/`ClearDocumentsDialog`). One fewer per-row `TooltipProvider`.
4. **Styled tooltips for file/summary cells** — replaced native `title=` with `<Tooltip>` components for visual consistency. Lifted a single `<TooltipProvider>` to the `<Table>` root so it's shared across rows instead of created per-row.
5. **Unreachable guard removed** — `if (!details) return null` inside `DocumentStatusDetailsDialog` can't fire because the call site already checks `hasDocumentDetails(doc)`.

Verified with \`bun run build\` (✓ 1.67s). Existing \`react-hooks/set-state-in-effect\` lint errors at lines 1131/1150/1156 pre-date this PR and are unrelated.

danielaskdd:
Pushed another follow-up (7499017d) addressing two display issues observed when opening the modal:

**1. Copy button tooltip auto-shows on dialog open**
Radix `FocusScope` (inside `DialogContent`) focuses the first tabbable child on open. In our layout the copy button is the first focusable element (DialogTitle/DialogDescription aren't focusable, and the built-in X close button is rendered after `children`). `TooltipTrigger` listens to `focus` for keyboard a11y, so the auto-focus immediately fired the tooltip. Fix: redirect initial focus to `DialogContent` itself via `onOpenAutoFocus`:

```tsx
onOpenAutoFocus={(e) => {
  e.preventDefault()
  ;(e.currentTarget as HTMLElement | null)?.focus()
}}
```

`DialogContent` already carries `tabIndex={-1}`, so it can hold focus. Tab still traps inside the dialog, ESC still closes it.

**2. Empty band below the last line of details**
The inner box used `min-h-[7.5em]` (≈105px). With typical content (Track ID + 2 metadata lines ≈ 80px) the min-height padded the box taller than its content, producing the visible gap below the last line. Removed `min-h-[7.5em]` so the box sizes to content; `max-h-[60vh]` still caps long content for scrolling, and `p-3` keeps the existing inner padding even for very short content.

PR #3029, opened 2026-05-07
feat: expose doc_id in query responses

eeeetttt:
## Summary

Carry `full_doc_id` from the storage layer through the retrieval pipeline
so it appears as `doc_id` in all query response references and chunks.

## Problem

`full_doc_id` is already stored in both KV store and vector database,
but the retrieval pipeline drops it at multiple points.

## Solution

4-point pass-through (17 lines added, 3 files changed):

- `operate.py`: extract `full_doc_id` in `_get_vector_context()` and
  propagate through all three branches of `_merge_all_chunks()`
- `utils.py`: add `doc_id` to `convert_to_user_format()` chunks and
  `generate_reference_list_from_chunks()` references
- `query_routes.py`: add optional `doc_id` field to `ReferenceItem` model

`doc_id` is `Optional[str]`, fully backward compatible.
All three query endpoints (`/query`, `/query/stream`, `/query/data`) gain `doc_id`.

### DIFF
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -432,7 +432,16 @@ def create_app(args):
             )
         else:
             # For other endpoints, return the default FastAPI validation error
-            return JSONResponse(status_code=422, content={"detail": exc.errors()})
+            # Pydantic model_validator errors may contain unserializable objects
+            # in ctx['error'], so convert them to strings before JSON encoding
+            errors = []
+            for error in exc.errors():
+                error_dict = dict(error)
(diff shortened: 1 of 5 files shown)


danielaskdd:
@codex review

eeeetttt:
@codex review

eeeetttt:
never mind

eeeetttt:
Users can now bring their own doc_id via the API instead of relying on auto-generated MD5 hashes. This is especially useful when integrating an existing knowledge base — keeping document IDs consistent across systems.

dianatofficial (commented):
Looks solid. Documentation and typing annotations are up to date.

PR #3038, opened 2026-05-08
Add offline sample retrieval check for RAGAS evaluation

FU-max-boop:
## Summary

This PR adds a small offline retrieval sanity check for the bundled LightRAG RAGAS evaluation samples.

Changes:

- add `sample_retrieval_oracle.json` to map each sample question to its expected sample document(s)
- add `offline_retrieval_check.py`, a deterministic lexical check that does not start LightRAG or call API/model/RAGAS services
- add regression coverage for the checker and bundled sample oracle
- document the optional check in the RAGAS evaluation README

## Why

The existing evaluation flow can require a running LightRAG server, indexed documents, embeddings, LLM calls, and RAGAS. Before spending that setup time or API budget, it is useful to verify that the bundled sample questions are structurally aligned with the bundled sample documents.

This check catches a narrower failure mode: sample questions whose expected documents are not retrievable even under a simple deterministic baseline. It also documents that the final bundled sample question expects two documents, so top-1 retrieval only gives partial recall while top-2 gives full recall.

## Validation

From the repo root:

```bash
python3 -m py_compile lightrag/evaluation/offline_retrieval_check.py tests/test_evaluation_offline_retrieval_check.py
python3 lightrag/evaluation/offline_retrieval_check.py --strict
python3 -m unittest tests/test_evaluation_offline_retrieval_check.py
git diff --check
```

Observed result for the default strict check:

```text
Queries: 6
Top-k: 2
Average recall@k: 1.000
Mean reciprocal rank: 1.000
Full-recall queries: 6/6
No-hit queries: 0
```

I also checked `--top-k 1 --strict`; it exits non-zero because the multi-document sample has recall@1 = 0.500, which is the intended strict behavior.

`pytest` and `ruff` are not installed in my local environment, so I could not run those commands here.

## Scope / limits

This PR does not change LightRAG retrieval behavior, the API server, embeddings, LLM calls, or RAGAS scoring. It only adds an offline sample-data sanity check and test coverage for that check.

### DIFF
--- a/lightrag/evaluation/README_EVALUASTION_RAGAS.md
+++ b/lightrag/evaluation/README_EVALUASTION_RAGAS.md
@@ -54,6 +54,15 @@ pip install -e ".[evaluation]"
 
 ### 2. Run Evaluation
 
+**Optional offline sample retrieval check (no API/model calls):**
+```bash
+python lightrag/evaluation/offline_retrieval_check.py --strict
+```
+
+This checks whether the bundled sample questions can lexically retrieve their
+expected sample documents before running LightRAG, embeddings, LLM calls, or
+RAGAS.
+
 **Basic usage (uses defaults):**
 ```bash
 cd /path/to/LightRAG
--- a/lightrag/evaluation/offline_retrieval_check.py
+++ b/lightrag/evaluation/offline_retrieval_check.py
@@ -157,10 +157,18 @@ def audit_samples(
             raise ValueError(f"No oracle entry for question: {question}")
 
         query_tokens = tokenize(question)
-        ranked = sorted(
-            documents,
-            key=lambda doc: (-score_query(query_tokens, doc, idf), doc.name),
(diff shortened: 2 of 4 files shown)


danielaskdd:
@codex review

FU-max-boop:
Thanks for the review. I pushed `b33eeab` to exclude zero-score documents from the ranked retrieval list, so an expected document cannot pass purely because it sorts first by filename. I also added a regression test for the no-overlap case.

Re-ran:
- `python3 -m py_compile lightrag/evaluation/offline_retrieval_check.py tests/test_evaluation_offline_retrieval_check.py`
- `python3 -m unittest tests/test_evaluation_offline_retrieval_check.py`
- `python3 lightrag/evaluation/offline_retrieval_check.py --strict`
- `git diff --check`

danielaskdd:
@codex review

PR #2822, opened 2026-03-22
perf: batch OpenSearch graph/KV/vector operations to reduce HTTP roundtrips

he-yufeng:
## Summary

Fixes #2785

The profiling data in #2785 shows that individual HTTP roundtrips dominate OpenSearch wall time compared to Neo4j:

- **Graph**: `upsert_node` (252s / 266 calls) and `upsert_edge` (224s / 244 calls) each did individual `client.index()` calls
- **KV**: `upsert` called 651 times with small payloads (429s total), each triggering a separate `async_bulk`
- **Vector**: `delete` called 244 times (128s total), same per-call overhead

This PR buffers write operations in memory and flushes them in a single `async_bulk` call during `index_done_callback()`, which is already called at the end of each processing stage.

## What changed

**Graph storage (`OpenSearchGraphStorage`)**
- `upsert_node()` and `upsert_edge()` now buffer docs to `_pending_nodes` / `_pending_edges` dicts instead of calling `client.index()` per operation
- `_flush_pending_ops()` flushes both buffers via one `async_bulk` call
- `has_node()` and `get_node()` check the pending buffer first, so callers see buffered-but-unflushed data (read-your-writes consistency)
- `upsert_edge()` checks the buffer for reverse edge IDs too, avoiding a server roundtrip when both directions are written in the same batch

**KV storage (`OpenSearchKVStorage`)**
- `upsert()` buffers docs to `_pending_upserts` dict
- `_flush_pending_upserts()` flushes via one `async_bulk` call
- `get_by_id()` checks the pending buffer first

**Vector storage (`OpenSearchVectorDBStorage`)**
- `delete()` buffers IDs to `_pending_deletes` list
- `_flush_pending_deletes()` deduplicates and flushes via one `async_bulk` call

**Safety**
- `finalize()` flushes pending ops before releasing the client connection
- `drop()` discards pending buffers (no point writing to indices being deleted)
- Error handling preserved -- `async_bulk(raise_on_error=False)` logs partial failures

## Backwards compatibility

No API changes. The public interface (`upsert_node`, `upsert_edge`, `upsert`, `delete`, `index_done_callback`) is unchanged. The only behavioral difference is that writes are deferred until the next `index_done_callback()` or `finalize()` call, which matches how all other LightRAG storage backends work (the base class docstrings explicitly state "changes will be persisted during the next index_done_callback").

## Test plan

- [x] All 109 existing OpenSearch tests pass
[...]

Issue #2785, opened 2026-03-16
[Performance]: Performance improvement for OpenSearch storage


LantaoJin:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

https://github.com/HKUDS/LightRAG/pull/2739 introduced an OpenSearch storage backend. We noticed a significant performance discrepancy: OpenSearch indexing took 8 minutes, while the default storage took only 1.5 minutes (a 5.3x overhead). This was tested locally via docker-compose-3.x.yml (512MB heap size cluster).

### Profiling on `examples/lightrag_openai_demo.py`
```
================================================================================
STORAGE PROFILING REPORT
================================================================================

  [DOC_STATUS_STORAGE]  total: 0.0645s  calls: 8
    upsert                                      0.0472s  (3 calls)
    index_done_callback                         0.0166s  (1 calls)
    get_docs_by_status                          0.0005s  (3 calls)
    filter_keys                                 0.0003s  (1 calls)

  [GRAPH_STORAGE]  total: 0.5220s  calls: 1492
    upsert_edge                                 0.1228s  (245 calls)
    upsert_node                                 0.1101s  (252 calls)
    get_node                                    0.1065s  (735 calls)
    index_done_callback                         0.0517s  (1 calls)
    has_edge                                    0.0470s  (245 calls)
    get_edges_batch                             0.0291s  (4 calls)
    edge_degrees_batch                          0.0232s  (2 calls)
    get_nodes_batch                             0.0152s  (4 calls)
    node_degrees_batch                          0.0095s  (2 calls)
    get_nodes_edges_batch                       0.0069s  (2 calls)

  [KV_STORAGE]  total: 1.0799s  calls: 1672
    get_by_id                                   0.6464s  (1114 calls)
    upsert                                      0.3611s  (543 calls)
    index_done_callback                         0.0720s  (12 calls)
    get_by_ids                                  0.0005s  (3 calls)

  [VECTOR_STORAGE]  total: 263.6970s  calls: 754
    upsert                                    262.7471s  (498 calls)
    query                                       0.3988s  (5 calls)
    delete                                      0.3905s  (245 calls)
    index_done_callback                         0.1551s  (3 calls)
    get_vectors_by_ids                          0.0055s  (3 calls)

  ────────────────────────────────────────────────────────────
[...]

PR #3043, opened 2026-05-09
perf(opensearch): batch vector upserts to reduce HTTP roundtrips

Saswatsusmoy:
## Summary

Refs #2785.

`OpenSearchVectorDBStorage.upsert()` previously issued one `helpers.async_bulk` call per invocation. The profile in #2785 shows the upstream pipeline calling `upsert()` 506 times for the demo workload, so the vector path was dominated by 506 sequential HTTP roundtrips even though every individual call is technically a "bulk" of 1.

This PR buffers writes and deletes in memory and flushes them in a single `async_bulk` call inside `index_done_callback()` / `finalize()`, matching the contract documented on `BaseVectorStorage` ("changes will be persisted during the next `index_done_callback`"). All buffer reads / writes and destructive server mutations are serialised through a `get_namespace_lock`, the same primitive used by `NanoVectorDBStorage`.

This change is complementary to #2822 (KV / graph batching); the two PRs touch different code paths on the vector side and should be merged sequentially with a rebase.

## Profiling (single-node OpenSearch 3.0, 512MB heap, localhost)

Synthetic 60-paragraph doc, deterministic embeddings, mock LLM.

|                                            | main      | this PR   | delta            |
| ------------------------------------------ | --------: | --------: | ---------------: |
| Wall (insert + finalize)                   | 3.145s    | 3.237s    | +0.09s (noise)   |
| `VECTOR_STORAGE` total                     | 0.294s    | 0.229s    | **-22.4%**       |
| `VECTOR_STORAGE.upsert` total              | 0.283s    | 0.018s    | **-93.5%**       |
| `VECTOR_STORAGE.upsert` calls              | 20        | 21        | (same workload)  |
| `VECTOR_STORAGE.index_done_callback` total | 0.011s    | 0.210s    | flush moved here |

Per-call cost on `upsert()` drops from ~14.2ms to ~0.87ms (16.3× faster per call). Net storage time saved is small at this scale because localhost RTT is ~0.5ms; on cloud-hosted OpenSearch (10–50ms RTT) the saving scales linearly with `(N_upserts - N_flushes) * RTT`.

### Batched writes
- `upsert()` generates embeddings (outside the lock to avoid blocking reads) then buffers documents into `_pending_vector_docs` under `_flush_lock`. An upsert cancels any pending tombstone for the same id.
- `delete()` / `delete_entity()` buffer ids into `_pending_vector_deletes`. A delete cancels any pending upsert for the same id.
- `_flush_pending_vector_ops()` issues a single `async_bulk` for all pending upserts + deletes, lock-guarded.
- `index_done_callback()` flushes pending ops before refreshing the index, so k-NN search sees the buffered writes.
- `finalize()` flushes pending ops before releasing the client.
- `drop()` runs entirely under `_flush_lock`, clearing both buffers and serialising with any in-flight flush.

### Concurrency
- `_flush_lock` is a `get_namespace_lock(namespace, workspace=...)` instance (multi-process aware), matching `NanoVectorDBStorage._storage_lock`.
- All buffer mutations (`upsert` write loop, `delete`, `delete_entity`), buffer reads (`get_by_id`, `get_by_ids`, `get_vectors_by_ids`), the flush itself, and destructive server mutations (`delete_entity_relation`'s `delete_by_query`, `drop`'s `indices.delete` + recreate) acquire this lock.
- Embedding computation and `mget` calls run outside the lock so remote RTTs don't block flushes.

### Read-your-writes
- `get_by_id` / `get_by_ids` / `get_vectors_by_ids` check the buffer under the lock first; if the id isn't buffered they fall through to OpenSearch. Pending tombstones return `None`. The lock-everywhere model means an in-flight flush is observed as either completely-before or completely-after — no half-committed snapshot is ever visible.
- `get_by_id` and `get_by_ids` strip the `vector` field on the mget fallback path (via `_source_excludes=["vector"]` + defensive `pop`) to match `NanoVectorDBStorage` / Milvus / Qdrant / Faiss / PGVector / MongoDB. `get_vectors_by_ids` still returns the embedding (its sole purpose).

### Failure handling
- Per-doc bulk failures are classified by HTTP status by `_extract_bulk_failed_ids`:
  - Retryable (408 / 429 / 5xx / missing status): kept in the buffer for the next flush.
  - Non-retryable (most 4xx, mapping errors): dropped from the buffer. A sample of `(op, id, status, error)` is logged at WARNING with the dict-shape `error` field summarised via `reason` / `type`, capped at 200 chars. Remaining failures are dumped at DEBUG.
  - `404` on a delete is treated as success-equivalent (the row is already gone).
- If `helpers.async_bulk` itself raises, the buffers are left untouched and the next flush retries — OpenSearch index ops are idempotent on `_id`.

### `index_done_callback` early-return removed
Previously the callback returned early when `_index_ready=False`, skipping the flush even with a non-empty buffer. Now the flush runs first; `_flush_pending_vector_ops` calls `_ensure_index_ready` which recreates the index if missing. The refresh step is skipped only when the index is still not ready after the flush attempt.

## Backwards compatibility

No public API changes. Method signatures, return types, and error paths are unchanged.

Behavioural changes worth noting:

1. **Vector writes are deferred until `index_done_callback()` / `finalize()`** — matches `BaseVectorStorage`'s documented contract and the behaviour of every other LightRAG storage backend.

2. **Cross-worker read-after-write visibility now requires `index_done_callback()`** (multi-worker deployments, e.g. `lightrag-gunicorn`). Before this PR, OpenSearch's per-call `async_bulk` made writes visible to other workers via mget / translog within milliseconds. After this PR, the in-memory buffer is **process-local**, so Worker B's `get_by_id` will not see Worker A's buffered writes until A has flushed. The `BaseVectorStorage` docstring now states this explicitly. Application code that relies on cross-worker sub-second visibility must `await rag.entities_vdb.index_done_callback()` (or wait for the pipeline's own `_insert_done()`) before reading from another worker. The pipeline itself is unaffected because it holds `pipeline_status.busy` for the duration of ingestion.

3. **`get_by_id` / `get_by_ids` no longer return the `vector` field** on the mget fallback path. Previously OpenSearch was the only backend that leaked this field; the new behaviour matches `NanoVectorDBStorage`, Milvus, Qdrant, Faiss, PGVector and MongoDB. Callers needing the embedding must use `get_vectors_by_ids` (existing API, returns `{id: list[float]}` for all backends). No call site in `operate.py` / `lightrag.py` / `pipeline.py` reads the `vector` key off these results.

## Test plan

- [x] `ruff check .` passes.
- [x] `pytest tests/test_opensearch_storage.py` — 151 passed (17 new + 4 updated).
- [x] `TestVectorStorageBatching` covers:
  - repeated `upsert()` calls collapse into a single `async_bulk`
  - upserting the same id twice keeps the latest payload
  - `delete()` cancels a pending upsert; `upsert()` cancels a pending delete
  - `get_by_id` reads the buffer (no `mget`)
  - `get_by_id` returns `None` for pending tombstones
  - `get_by_ids` merges buffer + `mget` for unbuffered ids only (with `_source_excludes=["vector"]`)
  - `get_vectors_by_ids` uses the buffer; mget passes `_source_includes=["vector"]`
  - `finalize()` flushes before releasing the client
  - `drop()` discards both buffers
  - per-doc retryable failures (5xx) are retained for retry
  - per-doc non-retryable failures (4xx) are dropped and logged with sample (op, id, status, error)
  - `delete_entity_relation` prunes matching buffered docs
  - `_extract_bulk_failed_ids` shape robustness + dict-shape error handling + long-error truncation
  - `get_by_id` / `get_by_ids` strip `vector` on the mget fallback path
  - `index_done_callback` still flushes when the index was marked missing after writes were buffered
  - `delete_entity_relation` and `drop` are serialised with an in-flight flush
  - concurrent `upsert` / `delete` during flush are blocked by the lock and complete after release
- [x] Integration run against a real `opensearchproject/opensearch:3.0.0` container shows the per-call upsert latency drop reported above.


Saswatsusmoy:
Both issues addressed in 552719fb:

**1. Race in `_flush_pending_vector_ops` (Codex P1, Copilot)** — fixed via snapshot-and-flush. Inside `_flush_lock`, the live buffers are captured into locals and immediately swapped for fresh empty containers *before any await*. Concurrent `upsert` / `delete` calls during the bulk await now land in the new buffers and survive the flush. On failure, the snapshot is merged back via `_merge_back_snapshot`, which never overwrites a newer concurrent write — a pending delete shadows a retried upsert and vice versa.

New test `test_concurrent_writes_during_flush_are_preserved` reproduces the original race using an event-gated `async_bulk` mock and asserts the new write survives. `test_concurrent_delete_during_flush_supersedes_retried_upsert` covers the merge-back precedence rules.

**2. Non-retryable failures looping forever (Copilot)** — `_extract_bulk_failed_ids` now returns `(retryable, non_retryable)`, classified by HTTP status. Retryable: 408 / 429 / 5xx, plus missing status (network error). Non-retryable: everything else (most 4xx). `404` on a delete is dropped from both sets since the row is already gone.

New test `test_failed_flush_drops_non_retryable_entries` verifies a 400 is dropped while a 503 is retained. Existing `test_failed_flush_entries_retained_for_retry` updated to use 503 for clarity. `test_extract_bulk_failed_ids_classifies_by_status` covers all the classification cases.

134 tests pass; ruff clean. No public API change.

LantaoJin (commented):
Should we merge https://github.com/HKUDS/LightRAG/pull/2822 and this PR. Seems #2822 handles 3 storage types but only batches vector `deletes`. #3043 specifically closes the vector `upsert` gap.

Since they are all buffering writes in memory then flushing, it could extract a common pattern:

  1. A generic `_BulkWriteBuffer` helper (or mixin) that encapsulates:
    - The pending dict/list
    - `buffer(id, doc)` / `buffer_delete(id)`
    - `flush(client, index_name)` with the `async_bulk` call, logging, and error handling
    - `read_from_buffer(id)` for read-your-writes
    - `clear()` for drop
  2. PR #3043 's more sophisticated version adds: snapshot-and-swap for concurrency safety, retryable vs non-retryable failure classification, and upsert/delete mutual cancellation. If that logic were in the shared helper, all three storage types (KV, Graph, Vector) would benefit from it rather than only the vector path having the robust implementation.

The practical gap: #2822's flush is simpler (clear after success, retry entire buffer on failure), while #3043 classifies per-doc failures and only retries retryable ones. If merged independently, you'd have two different error-handling philosophies in the same file. Consolidating them into a shared helper would force alignment on the retry semantics.

danielaskdd:
## Iteration: review feedback addressed

This iteration replaces the original snapshot-and-flush concurrency model with a single namespace-lock model and tightens five related concerns surfaced in review.

### 1. Concurrency model: snapshot-and-flush → namespace lock

The original implementation kept `upsert` / `delete` outside `_flush_lock` and used snapshot-and-swap inside the flush to avoid dropping concurrent writes. Codex (P1) and Copilot both flagged subtle races around the merge-back logic. The new model picks a simpler invariant: a single `get_namespace_lock(namespace, workspace=...)` serialises all buffer reads, all buffer writes, the flush itself, and destructive server mutations. This:

- Eliminates the snapshot-and-swap state machine including `_merge_back_snapshot` (~100 lines removed).
- Closes the in-flight RYW window: a concurrent `get_by_id` either sees the state strictly before or strictly after the flush, never an "in-between" where the snapshot has been swapped out but the bulk hasn't yet been observable in OpenSearch.
- Brings the locking model in line with `NanoVectorDBStorage._storage_lock` (also a namespace lock), so multi-process OpenSearch deployments serialise flushes against the same index across workers.

Performance impact is bounded: `_insert_done()` runs ≈ once per processed file, the lock-held window is one `async_bulk` RTT (tens of ms), and `upsert`'s embedding computation runs outside the lock so reads aren't blocked on remote embedding providers.

### 2. `get_by_id` / `get_by_ids`: strip `vector` to match other backends

The buffer path already stripped `vector` from the returned dict, but the mget fallback path leaked it. This was the only backend with this inconsistency — `NanoVectorDBStorage`, Milvus, Qdrant, Faiss, PGVector and MongoDB all strip it. Fixed by passing `_source_excludes=["vector"]` to mget (saves network) plus a defensive `pop("vector", None)` (covers misconfigured indices that ignore `_source_excludes`). No call site in `operate.py` / `lightrag.py` / `pipeline.py` reads the `vector` key off `get_by_id`'s result, so this is a safe alignment.

### 3. `delete_entity_relation` and `drop`: serialised with flush

Previously these methods pruned the in-memory buffer without holding the flush lock and then issued a server-side mutation (`delete_by_query` / `indices.delete`). A concurrent flush could send an `index v1` action while `delete_by_query` was sweeping matching docs — net result was timing-dependent. Now both methods run their entire body under `_flush_lock`, so destructive server mutations strictly serialise with in-flight indexing.

### 4. Non-retryable bulk failures: structured logging

`_extract_bulk_failed_ids` previously returned `(set[str], set[str])` and the warning only included the count, so 100 mapping-error drops produced one log line and zero diagnostic context. The helper now returns `(set[str], list[_FailedBulkOp])` with structured `op` / `doc_id` / `status` / `error` fields; the `error` summary handles dict-shape errors via `reason` / `type` and is capped at 200 chars. The warning samples the first 5 failures with full context; the rest go to DEBUG. Retryable-vs-non-retryable classification is unchanged.

### 5. `index_done_callback`: removed unnecessary early-return

`if not self._index_ready: return` previously skipped flush entirely when a prior read had marked the index missing. The flush itself calls `_ensure_index_ready` which recreates the index, so the early-return was just delaying buffered writes by one cycle. The refresh step still short-circuits if the index isn't ready after the flush attempt.

### 6. Cross-worker visibility caveat (documentation)

The in-memory buffer is process-local; gunicorn workers cannot see each other's pending writes. **Any lock — including a multi-process `get_namespace_lock` — cannot fix this**, because the buffer itself is per-process dict state. The fix is documentation, not code:

- `BaseVectorStorage.upsert` / `delete` / `delete_entity` / `delete_entity_relation` docstrings now carry a "Multi-worker note" pointing callers to `index_done_callback()` when they depend on cross-worker visibility.
- The PR's "Backwards compatibility" section calls out that pre-PR translog-based sub-second cross-worker `get_by_id` is no longer available.

Pipeline workloads are unaffected because the document pipeline holds a `pipeline_status.busy` cross-worker lock during ingestion; only one worker processes documents at a time, and that worker's own `_insert_done()` runs per file (≈ tens of seconds), bounding buffer lifetime.

### Test coverage delta

17 new tests, 4 updated. New coverage:

- `vector` stripping on the mget fallback path (`test_get_by_id_strips_vector_from_mget_path`, `test_get_by_ids_strips_vector_from_mget_path`)
- Structured failure logging (`test_non_retryable_logs_sample_ids`)
- `_extract_bulk_failed_ids` dict-shape `error` handling + long-error truncation
- `index_done_callback` flushes when the index was marked missing after buffered writes (`test_index_done_callback_flushes_when_index_recreated`)
- `delete_entity_relation` and `drop` blocked by in-flight flush (`test_delete_entity_relation_serialised_with_flush`, `test_drop_serialised_with_flush`)
- Concurrent `upsert` / `delete` blocked by the flush lock and completing after release (rewritten from the old snapshot-and-flush tests)

The test fixture now patches `get_namespace_lock` to return real `asyncio.Lock` instances cached per `(namespace, workspace)` — using a no-op lock would silently disable the new serialisation invariants under test.

### Out of scope

- PR #2822 (KV / graph batching): should rebase against this PR after merge and reuse `_extract_bulk_failed_ids` + the same namespace-lock pattern.
- Cross-process shared buffer (Redis / shared memory): architectural change beyond the scope of #2785.

PR #3048, opened 2026-05-09
fix(utils): make write_json atomic and concurrency-safe

Zhehua-Hu:
## Description

The current `write_json` truncates the destination via `open(path, 'w')` and writes in place. When two writers (typically a stale, slowly-shutting-down `lightrag-server` and a freshly started one) both flush the same KV/doc-status file concurrently, one writer's mid-flight output gets cut short by the other's truncation, leaving a corrupt half-written JSON on disk. The next server start then fails with `JSONDecodeError` and the entire pipeline state is effectively lost (`kv_store_doc_status.json` is the worst victim because its records are large).

## Reproduction

1. Start `lightrag-server`, begin a long ingestion.
2. While it is mid-flush, run `pkill -f lightrag-server` and immediately start a new server.
3. Result: the doc-status file is sometimes truncated mid-record. The next start raises `json.JSONDecodeError: Expecting ',' delimiter`.

This was hit during a real ~7000-document ingestion against `data/inputs/`. After the corruption, ~6300 doc-status records were lost permanently and had to be reconstructed manually from `kv_store_full_docs.json` plus `kv_store_text_chunks.json`.

## Changes Made

`lightrag/utils.py::write_json` now uses the standard write-temp-then-rename pattern:

1. Write to `<file>.tmp.<pid>`
2. `fcntl.LOCK_EX` during the write (advisory cross-process lock; gracefully degrades on systems without `fcntl`)
3. `flush()` + `os.fsync()` to push kernel buffers to disk
4. `os.replace()` to atomically swap into the final path (POSIX `rename` guarantees no partial file is ever visible to a reader)

The two-stage sanitization fallback from the original implementation is preserved.

## Checklist

- [x] Changes tested locally — long ingestion run no longer corrupts `kv_store_doc_status.json` even when the server is restarted while mid-flush
- [x] Backwards-compatible — same function signature, same return semantics
- [ ] Unit tests added — happy to add a test case if maintainers want one

## Additional Notes

The `fcntl` import is wrapped in a `try/except` so the patch still works on platforms where it is unavailable; on those platforms the atomic-rename guarantee alone is sufficient (a writer can never be observed mid-flush, only the previous or new full file).

### DIFF
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -1288,39 +1288,58 @@ class SanitizingJSONEncoder(json.JSONEncoder):
 
 def write_json(json_obj, file_name):
     """
-    Write JSON data to file with optimized sanitization strategy.
+    Atomic + concurrency-safe JSON write.
 
-    This function uses a two-stage approach:
-    1. Fast path: Try direct serialization (works for clean data ~99% of time)
-    2. Slow path: Use custom encoder that sanitizes during serialization
+    Concurrency safety strategy (防止 doc_status.json 截断):
+    1. Write to <file>.tmp.<pid> first
+    2. fcntl LOCK_EX during write (cross-process advisory lock)
+    3. fsync to flush kernel buffer
+    4. os.replace() atomically swap into target path
+       (POSIX guarantees rename is atomic — no half-written state visible)
 
-    The custom encoder approach avoids creating a deep copy of the data,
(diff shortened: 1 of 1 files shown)


Zhehua-Hu:
Closing — submitted by mistake on my side, no review needed.

PR #3052, opened 2026-05-10
fix(PGGraphStorage): edge properties lost on upsert with Apache AGE

oleksandr-kushnir:
Hey! 👋 I ran into a frustrating issue where the LightRAG knowledge graph web UI was rendering nodes correctly but showing **no labels on any edges** after uploading documents with the PostgreSQL backend. Spent a while debugging and traced it down to `upsert_edge` in `PGGraphStorage`.

## What's happening

`SET r += {map_literal}` silently drops all properties on **edges** in Apache AGE — it works fine for nodes, but for relationships the write just disappears without an error. So every `DIRECTED` edge ends up with `properties = {}`, and relation keywords/descriptions/weights are never stored.

The comment in the code already flags that the parameterized form (`SET r += $props`) doesn't work, and the inlined literal was meant as the fix — but unfortunately the inlined form has the same silent failure for edges.

I tried a couple of alternatives before finding one that works:

| Pattern | Result on AGE 1.5.0 |
|---------|---------------------|
| `SET r += {map_literal}` | ❌ silently ignored |
| `ON CREATE SET / ON MATCH SET` | ❌ syntax error |
| `SET r.key = value` | ❌ no error, but not persisted |
| Inline props in `CREATE` clause | ✅ works |

## The fix

The only reliable way to write edge properties in AGE is to inline them directly in a `CREATE` statement. Since you can't do `MERGE + CREATE`, I use `OPTIONAL MATCH` to find and delete the old edge first, then `CREATE` the new one with all properties inlined:

```cypher
MATCH (source:base {entity_id: $src_id})
WITH source
MATCH (target:base {entity_id: $tgt_id})
WITH source, target
OPTIONAL MATCH (source)-[old:DIRECTED]-(target)
DELETE old
WITH source, target
CREATE (source)-[r:DIRECTED {props_literal}]->(target)
RETURN r
```

`_format_properties` (already in the codebase) produces exactly the inlined map literal needed here — so it's a small, self-contained change.

## Verification

```sql
LOAD 'age';
SET search_path = ag_catalog, $user, public;
SELECT * FROM cypher('chunk_entity_relation', $$
  MATCH (a)-[r:DIRECTED]->(b)
  RETURN a.entity_id, b.entity_id, properties(r)
  LIMIT 5
$$) AS (src agtype, tgt agtype, props agtype);
```

Before: `props` = `{}` for every edge.
After: `props` = `{keywords: ..., weight: 1.0, description: ...}` ✅

Tested on Apache AGE 1.5.0 / PostgreSQL 16. Hope this helps someone else who's been staring at a graph with invisible edge labels! 😄

### DIFF
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -5162,14 +5162,21 @@ class PGGraphStorage(BaseGraphStorage):
             edge_data (dict): dictionary of properties to set on the edge
         """
         # AGE does not support binding a full agtype map in ``SET r += $props``
-        # (verified on AGE 1.5.0). Keep endpoint identifiers parameterized and
-        # inline a safely escaped property map literal for the relationship payload.
-        props_literal = self._format_properties(edge_data)
+        # (verified on AGE 1.5.0), and the inlined literal form ``SET r += {map}``
+        # is also silently ignored for edges (though it works for nodes). Individual
+        # ``SET r.key = value`` assignments run without error but also do not persist.
+        # The only reliable way to write edge properties in AGE is to inline them
+        # directly in a CREATE clause. We use OPTIONAL MATCH to delete any existing
+        # edge first so the operation remains idempotent.
+        props_literal = self._format_properties(edge_data) if edge_data else "{}"
         cypher_query = f"""MATCH (source:base {{entity_id: $src_id}})
                      WITH source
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3054, opened 2026-05-10
fix(embed): normalize text for OpenAI embeddings; demo safety

Abdeltoto:
## Context

Follow-up to the Unicode/ascii-codec reports seen around real extracted text with typographic punctuation, including the downstream RAG-Anything discussion in #184.

I do **not** have the original full `UnicodeEncodeError` traceback available anymore, so this PR no longer treats smart quotes as a proven OpenAI SDK failure mode. The OpenAI client should send UTF-8 correctly; the error is more likely to come from surrounding runtime paths (Windows console/logging, proxy, older tokenization stack, etc.).

## Changes

- Add an opt-in compatibility flag: `OPENAI_EMBED_NORMALIZE_UNICODE=true`.
- When enabled, `openai_embed` normalizes input text with NFC and maps common smart quotes to straight quotes before prefixing/truncation/API submission.
- Keep default behavior unchanged: normalization is **off by default**, so embedded text matches stored text unless callers explicitly opt in.
- `examples/lightrag_openai_demo.py`: initialize `rag = None` so the `finally` block cannot raise if startup exits early (for example, missing API key).

## Compatibility note

This is intentionally a narrow escape hatch, not a root-cause fix. If the project prefers ingestion-boundary normalization instead, I’m happy to close or rework this toward that direction.

### DIFF
--- a/examples/lightrag_openai_demo.py
+++ b/examples/lightrag_openai_demo.py
@@ -89,6 +89,7 @@ async def initialize_rag():
 
 
 async def main():
+    rag = None
     # Check if OPENAI_API_KEY environment variable exists
     if not os.getenv("OPENAI_API_KEY"):
         print(
--- a/lightrag/llm/openai.py
+++ b/lightrag/llm/openai.py
@@ -2,6 +2,7 @@ from ..utils import verbose_debug, VERBOSE_DEBUG
 import os
 import logging
 import warnings
+import unicodedata
 
 from collections.abc import AsyncIterator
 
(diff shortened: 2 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
Hi @Abdeltoto, thanks for the fix! Before merging, could you paste the original `UnicodeEncodeError` traceback into the PR description? A couple of concerns we'd like to clarify first:

1. **Root cause uncertainty.** The OpenAI SDK transports request bodies as UTF-8 over HTTP, so smart quotes alone shouldn't surface a `UnicodeEncodeError` from the embedding call itself. The error is more commonly raised by surrounding code paths — e.g. `print`/`logging` on Windows defaulting to `cp1252`, an older tiktoken build, or a proxy in front of the API. Without the traceback, this patch may be papering over the real issue, and the next non-ASCII character (em dash `U+2014`, ellipsis `U+2026`, NBSP `U+00A0`, …) will reproduce it.

2. **Should this be fixed at ingestion instead?** If the actual problem is encoding-related, the ingestion/extraction layer feels like a more appropriate place to normalize text — once, consistently, for *all* downstream consumers (embedding, LLM completion, chunk storage). Fixing it only inside `openai_embed` means the `openai_complete_if_cache` path (and other embedding backends) remain vulnerable to the same class of error.

3. **Embedded text diverging from stored text.** Right now the patch silently rewrites the user's content before embedding, while the original (smart-quoted) text is what gets stored in KV/vector metadata and returned to the LLM at query time. That divergence is small in practice, but it's a behavior change worth being explicit about — and it argues again for handling this at the ingestion boundary rather than inside the embedding function.

Would an env-gated flag (off by default) be acceptable as an interim, while we investigate whether the underlying issue belongs further upstream?

RedEye1605 (commented):
Nice fix! Normalizing Unicode for embeddings is a solid defensive practice.

The text normalization function is well-documented and handles the common smart quote issues. Good catch on the demo safety issue too - initializing  prevents the uninitialized variable error in the  function.

One small suggestion: you might want to consider also normalizing other common Unicode whitespace characters like non-breaking spaces (U+00A0) or en/em dashes if those show up in real-world documents. But the current scope is appropriate for the reported issue.

RedEye1605 (commented):
Nice fix! Normalizing Unicode for embeddings is a solid defensive practice.

The text normalization function is well-documented and handles the common smart quote issues. Good catch on the demo safety issue too - initializing rag = None prevents the uninitialized variable error in the main() function.

One small suggestion: you might want to consider also normalizing other common Unicode whitespace characters like non-breaking spaces (U+00A0) or en/em dashes if those show up in real-world documents. But the current scope is appropriate for the reported issue.

Abdeltoto:
Thanks for the careful review — your concerns make sense.

I don’t have the original full traceback anymore, so I reworked this to avoid presenting smart quotes as a proven OpenAI SDK issue. The branch is now rebased, and the normalization is **opt-in only** via `OPENAI_EMBED_NORMALIZE_UNICODE=true`; default behavior stays unchanged so embedded text does not silently diverge from stored text.

That leaves this as a narrow compatibility escape hatch for environments where punctuation trips a surrounding runtime path (console/logging/proxy/tokenizer), not as a root-cause fix. If you’d rather keep normalization at the ingestion boundary instead, I’m happy to close this PR or rework it in that direction.

PR #3056, opened 2026-05-11
fix(PGGraphStorage): serialise concurrent upsert_edge with advisory lock

oleksandr-kushnir:
## Summary

Follow-up to #3052 that addresses the codex bot's P2 concurrency concern.

`PGGraphStorage.upsert_edge` now uses `OPTIONAL MATCH + DELETE + CREATE`, but that pattern is not atomic against concurrent writers: two transactions upserting the same pair can both observe no existing edge and both `CREATE` one, leaving duplicate `DIRECTED` rows that inflate degree counts and duplicate relations.

This PR serialises per logical edge with a transaction-scoped advisory lock keyed on `(graph_name, ordered (src_id, tgt_id))` so `{A,B}` and `{B,A}` collide on the same lock (the `OPTIONAL MATCH` is undirected), while independent AGE graphs in the same database remain free of each other.

## Why two statements instead of a CTE

AGE refuses to plan a join against a `cypher()` call that contains a `CREATE` clause:

```
ERROR:  cypher create clause cannot be rescanned
HINT:  its unsafe to use joins in a query with a Cypher CREATE clause
```

So the lock cannot live in a CTE alongside the cypher upsert. Instead, `upsert_edge` now opens an explicit transaction via `db._run_with_retry` and runs two statements on the same connection: the lock acquisition first, then the cypher upsert. The transaction-scoped lock is released on commit.

## Lock key

```sql
hashtextextended(
  $1::text || E'\x01' ||
  LEAST($2::text, $3::text) || E'\x01' || GREATEST($2::text, $3::text),
  0
)
```

Positional params: `$1 = graph_name`, `$2 = src_id`, `$3 = tgt_id`.

- `$1 = graph_name` prefixes the hash so the same `(Alice, Bob)` in different AGE graphs / workspaces does **not** collide. `pg_advisory_xact_lock` is database-wide, and we don't want independent tenants to serialise each other's ingestion.
- `LEAST`/`GREATEST` make the endpoint portion order-independent so `{A,B}` and `{B,A}` map to the same lock.
- `E'\x01'` (SOH) separator avoids ambiguity between `(AB,C)` and `(A,BC)`. `E'\x00'` would be cleaner but PostgreSQL TEXT cannot contain null bytes.
- Raw IDs (and `graph_name`) flow as positional `text` params; they are never interpolated into the SQL.

## Retry semantics

`upsert_edge` is decorated with `@retry(... retry=retry_if_exception(_is_transient_graph_write_error) ...)`, which only re-fires when the exception is a `PGGraphQueryException` whose `__cause__` is one of `DeadlockDetectedError` / `SerializationError` / `LockNotAvailableError` / `QueryCanceledError`.

Previously `upsert_edge` went through `self._query`, which wrapped raw asyncpg errors into `PGGraphQueryException(...) from e`. Routing it through `db._run_with_retry` to run the lock + cypher in one transaction bypassed that wrapping, so a query-level transient error would surface as a raw asyncpg exception, fail the predicate's `isinstance` check, and skip retries — silently degrading concurrent ingestion under contention.

`upsert_edge` now re-wraps non-`PGGraphQueryException` errors into `PGGraphQueryException(...) from e` itself, mirroring what `_query` does for `upsert_node` and the other AGE write paths. The `isinstance(e, PGGraphQueryException)` guard avoids double-wrapping when an error was already wrapped deeper down.

## Test plan

- [x] All 7 `tests/test_postgres_upsert_edge_cypher.py` tests pass — including new regressions:
  - `test_upsert_edge_lock_key_includes_graph_name` — different `graph_name`s produce different lock keys.
  - `test_upsert_edge_lock_key_is_endpoint_order_independent` — `{A,B}` and `{B,A}` share the same lock-SQL template.
  - `test_upsert_edge_wraps_transient_errors_for_retry` — `DeadlockDetectedError` is re-raised as `PGGraphQueryException`, passes the predicate, and the `@retry` loop fires 3 times (proves the wrapping engages retries, not just shape).
  - `test_upsert_edge_does_not_retry_non_transient_errors` — `ValueError` is wrapped but fails fast after a single attempt.
- [x] `tests/test_postgres_cypher_injection.py` injection suite updated for the 3-arg lock signature; all 14 tests pass.
- [x] `tests/test_postgres_performance_timing.py::test_graph_upsert_edge_passes_timing_label` rewired to mock `db._run_with_retry` (the new execution path) instead of `_query`; the upsert-node timing test is unchanged.
- [x] `pre-commit` (ruff + ruff-format) clean.
- [x] End-to-end smoke against live Postgres + AGE: lock + cypher in one transaction both execute cleanly; verified `cypher CREATE` does not error when run as a separate statement (it does when joined with a CTE).
- [x] Order-independent lock key verified: `{Alice,Bob}` and `{Bob,Alice}` produce the same `hashtextextended` value within one graph.
- [x] Cross-graph isolation verified: `(Alice, Bob)` in `graph_a` and `graph_b` produce **different** `hashtextextended` values and do not contend on `pg_advisory_xact_lock`.

## Updates since first review

- **217ddec9** — Scope advisory locks by `graph_name` to address [codex P2 on r3260254786](https://github.com/HKUDS/LightRAG/pull/3056#discussion_r3260254786). Independent AGE graphs sharing one PostgreSQL database no longer serialise each other's edges.
- **749fb863** — Fix `test_graph_upsert_edge_passes_timing_label`: it was still mocking `storage._query`, but the new path drives the cypher via `db._run_with_retry`. Updated the mock to match the actual execution path.
- **9ad5d89f** — Restore the `PGGraphQueryException` wrapping that `_query` used to provide, so the outer `@retry` predicate can still recognise transient asyncpg errors via `__cause__` and retry them. Added two regression tests pinning both the retry-on-transient and fail-fast-on-non-transient behaviours.

Generated with [Claude Code](https://claude.com/claude-code)


danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3064, opened 2026-05-13
Refactor multimodal: status semantics, nested chunk schema, per-chunk entity injection

danielaskdd:
## Summary

- **Modality-aware analysis**: drawings now route to the VLM role while tables and equations route to the EXTRACT role, each with its own template in [lightrag/prompt_multimodal.py](lightrag/prompt_multimodal.py). `llm_analyze_result` switches to explicit `status / message / analyze_time` semantics, replacing the old `grounded / conservative` branches. Image format checks and a 32px pixel floor short-circuit to `status="skipped"` without an LLM call; required-field failures raise `MultimodalAnalysisError` so the document fails instead of writing an unusable result.
- **Cache lifecycle closed**: analysis cache ids are written back to each sidecar item's `llm_cache_list` and merged into the multimodal chunk's `llm_cache_list`. The existing document-delete cleanup path now naturally removes `cache_type="analysis"` entries that previously leaked when documents were deleted.
- **Chunk schema upgrade + per-chunk entity injection**: multimodal and paragraph-semantic chunks both adopt the new nested `heading = {level, heading, parent_headings}` + `sidecar = {type, id, refs}` schema. P-chunks propagate `blockid` through Stage B/C splits and merge them in Stage D so each chunk traces back to its source blocks. Multimodal entity injection moves from the centralized post-pass in `utils_pipeline.augment_chunk_results_with_mm_entities` (deleted) to per-chunk dispatch inside `operate._process_single_content`, keyed by `chunk.sidecar.type`. Extraction prompts are pre-cleaned of `<cite>` / `<drawing>` / `<equation>` internal identifiers without mutating the stored chunk content.

## Test plan

- [x] `./scripts/test.sh tests` — **1240 passed / 1 skipped / 1 xfailed**
- [x] `ruff check lightrag tests` — clean
- [ ] End-to-end smoke on a .docx with image / table / equation modalities enabled
  - sidecar JSON shows `llm_analyze_result.status="success"` and `llm_cache_list` populated for each modality
  - multimodal chunks carry the correct nested `sidecar.type` and inherited analysis cache ids
  - `drawing-* / table-* / equation-*` entities + relations land in the graph
  - second run hits the analysis cache for every modality (no model calls)
  - `await rag.adelete_by_doc_id(...)` clears the `default:analysis:*` entries

### DIFF
--- a/lightrag/chunk_schema.py
+++ b/lightrag/chunk_schema.py
@@ -1,0 +1,201 @@
+"""Chunk schema helpers shared across the chunking + extraction pipeline.
+
+Three responsibilities live here so chunker implementations and the pipeline
+both consume identical normalization rules:
+
+- :func:`normalize_chunk_heading` collapses the legacy flat
+  ``heading``/``parent_headings``/``level`` triple and the new nested form
+  into the canonical ``{"level", "heading", "parent_headings"}`` dict.
+- :func:`normalize_chunk_sidecar` validates the new ``sidecar`` payload and
+  ensures ``refs`` is always present as a list (single-source items may omit
+  it before normalization; we materialize a single-element list for the
+  storage layer).
+- :func:`strip_internal_multimodal_markup_for_extraction` rewrites
+  ``<cite>`` / ``<drawing>`` / ``<equation>`` markup so the entity-extraction
+  LLM sees a clean text body. The original ``chunk["content"]`` is never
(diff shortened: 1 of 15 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3083, opened 2026-05-16
write_nx_graph: atomic write via per-writer .tmp + os.replace

RooseveltAdvisors:
## Problem

`NetworkXStorage.write_nx_graph` currently writes the graph directly to the destination path:

```python
@staticmethod
def write_nx_graph(graph: nx.Graph, file_name, workspace="_"):
    logger.info(
        f"[{workspace}] Writing graph with {graph.number_of_nodes()} nodes, {graph.number_of_edges()} edges"
    )
    nx.write_graphml(graph, file_name)
```

`nx.write_graphml` opens the destination in write-mode (truncate-first) and streams XML into it. If the process is killed during the flush — `systemctl stop` mid-checkpoint, OOM kill, restart while a save is in flight, or even just a non-graceful shutdown — the graphml file on disk is truncated and partially written. The next call to `nx.read_graphml(file_name)` raises an XML parse error, and the service cannot reload its graph state.

The window scales with graph size. For non-trivial knowledge graphs (six-figure node counts) the write takes long enough that hitting the bad window under normal ops is reliable rather than rare.

## Fix

Atomic write via temp + rename:

```python
tmp = (
    f"{file_name}.tmp.{os.getpid()}"
    f".{threading.get_ident()}.{time.time_ns()}"
)
nx.write_graphml(graph, tmp)
os.replace(tmp, file_name)
```

`os.replace` is atomic on the same filesystem on both POSIX (`rename(2)`) and Windows (`MoveFileEx` with `MOVEFILE_REPLACE_EXISTING`), so the production file is always either the prior snapshot or the new one — never a torn middle.

The per-writer `.tmp` name (PID + thread id + ns timestamp) means two writers running concurrently each get their own sibling tmp. Without this, two concurrent calls would write into the same sentinel filename: the first `os.replace` consumes the tmp, the second crashes with `FileNotFoundError: ... .tmp -> ... .graphml`. Concurrent writers arise from:

- Two LightRAG instances sharing a working dir (server + a batch indexer running in another process).
- **Defense-in-depth against partial writes**: even though `index_done_callback` is already serialized inside a single process by `self._storage_lock`, the truncate-then-stream behavior of `nx.write_graphml` means any crash mid-write leaves the destination unreadable regardless of whether another writer exists. The atomic temp+rename closes that window.

## Orphan tmp reaper

A crash *between* `nx.write_graphml(tmp)` and `os.replace(tmp, dst)` leaves a `.tmp.<pid>.<tid>.<ns>` sibling on disk. Over repeated crashes these would accumulate.

`NetworkXStorage.__post_init__` now invokes `_reap_orphan_tmp_files`, which deletes any `graph_<namespace>.graphml.tmp.*` sibling whose mtime is older than `_TMP_REAP_AGE_SECONDS` (1 hour). The threshold is large enough that an in-flight write from another live process cannot plausibly still be running, so the reaper will not touch a tmp file that belongs to a concurrent live writer.

## Verified

`tests/test_networkx_atomic_write.py` adds three offline regression tests:

- **`test_write_nx_graph_concurrent_writers_no_race`**: 5 threads racing on the same destination — all complete without exception, the final file parses cleanly via `nx.read_graphml`, and zero `.tmp.*` siblings remain.
- **`test_write_nx_graph_crash_preserves_prior_snapshot`**: a v1 snapshot is committed; a v2 write is then issued with `os.replace` mocked to raise (simulating crash between `write_graphml` and rename); the destination is still v1, and the orphan tmp from the simulated crash is present (and will be reaped on next startup).
- **`test_reap_orphan_tmp_files_respects_age_threshold`**: a tmp file aged past the threshold is reaped; a fresh tmp file and an unrelated path are left untouched.

All three pass under `pytest tests/test_networkx_atomic_write.py`.

Without the patch, the concurrent-writer test reproduces `FileNotFoundError` on the slower threads' `os.replace`.

## Behavior change

| Before | After |
|---|---|
| Crash mid-write → graphml truncated, service can't reload | Crash mid-write → prior snapshot intact on disk |
| Concurrent writers → race; one or more fail with FileNotFoundError | Concurrent writers → each writes its own .tmp, each renames atomically |
| Crash between write_graphml and replace → orphan .tmp file accumulates over time | Startup reaper removes orphan .tmp files older than 1h |
| Success path | Unchanged |

No API change. Diff: `lightrag/kg/networkx_impl.py` (atomic write + orphan reaper, ~50 lines including comments) + `tests/test_networkx_atomic_write.py` (regression suite).


danielaskdd:
Thank you for your interest in contributing to LightRAG. Please resolve the existing merge conflicts before we proceed with the PR review.

RedEye1605 (commented):
Good improvement for data integrity! The atomic write pattern with os.replace() is the right approach here.

A few small suggestions:
- Consider adding a brief docstring update to write_nx_graph to document the atomic write behavior
- The inline comment is well-written, but could be slightly more concise for readability

Overall, this is a solid defensive programming improvement that prevents corrupted graph files on crashes.

PR #3086, opened 2026-05-17
fix: guard against IndexError on empty LLM choices list

qizwiz:
## Problem

Three call sites access `choices[0]` without first checking whether the API returned any choices. When the LLM:

- applies a **content-policy filter** and returns `choices=[]`
- hits a **token limit** before generating the first choice
- returns a non-standard error body

the code raises a bare `IndexError` that crashes the calling pipeline.

## Changes

| File | Fix |
|------|-----|
| `lightrag/llm/zhipu.py` | `if not response.choices: return ""` before accessing `choices[0].message` |
| `examples/generate_query.py` | same pattern for OpenAI API |
| `examples/lightrag_azure_openai_demo.py` | same pattern for Azure OpenAI |

Returning `""` on empty choices matches the existing `content = message.content or ""` pattern in `zhipu.py` and keeps the calling code consistent.

---
*Found by [pact](https://github.com/qizwiz/pact) static analysis — `llm_response_unguarded` mode.*

### DIFF
--- a/examples/generate_query.py
+++ b/examples/generate_query.py
@@ -17,7 +17,7 @@ def openai_complete_if_cache(
     response = openai_client.chat.completions.create(
         model=model, messages=messages, **kwargs
     )
-    if not response.choices:
+    if not response.choices or response.choices[0].message is None:
         return ""
     return response.choices[0].message.content
 
--- a/examples/generate_query.py
+++ b/examples/generate_query.py
@@ -17,6 +17,8 @@ def openai_complete_if_cache(
     response = openai_client.chat.completions.create(
         model=model, messages=messages, **kwargs
     )
+    if not response.choices:
+        return ""
     return response.choices[0].message.content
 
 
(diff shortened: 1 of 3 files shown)


qizwiz:
**Update: guard strengthened after live reproduction (commit 957145b)**

While building a proof script for this fix, we discovered that the original guard (`if not response.choices`) is incomplete.

**Gemini 2.5 Flash** (via its OpenAI-compatible endpoint) returns HTTP 200 with `choices[0].message = None` when its safety filter triggers — `finish_reason: 'content_filter: PROHIBITED_CONTENT'`. The list has one element, so the empty-list guard passes, but accessing `.message.content` crashes with `AttributeError: 'NoneType' object has no attribute 'content'`.

Reproduced live:
```python
# A: unfixed
response.choices[0].message.content
# → AttributeError: 'NoneType' object has no attribute 'content'

# B: previous guard (submitted in original commit) — still crashes
if not response.choices:
    return ""
return response.choices[0].message.content
# → AttributeError (choices has 1 element, but message is None)

# C: comprehensive fix (this commit)
if not response.choices or response.choices[0].message is None:
    return ""
return response.choices[0].message.content
# → "" (no crash)
```

This affects any OpenAI-compatible provider that signals content filtering via a null message rather than an empty choices list. Updated all three files accordingly.

danielaskdd:
@codex review

PR #3098, opened 2026-05-20
feat(redis): add basename and content_hash lookups for doc status

danielaskdd:
## Summary

Adds two dedup-lookup methods to `RedisDocStatusStorage` so the Redis backend is feature-equivalent with `JsonDocStatusStorage` and `PostgreSQLDocStatusStorage`:

- `get_doc_by_file_basename(basename) -> tuple[doc_id, doc_data] | None`
- `get_doc_by_content_hash(content_hash) -> tuple[doc_id, doc_data] | None`

## Motivation

After the recent two-stage dedup work in the ingestion pipeline (basename match → content-hash match), `JsonDocStatusStorage` and `PostgreSQLDocStatusStorage` both expose these helpers. The Redis backend did not — meaning any deployment configured with `LIGHTRAG_DOC_STATUS_STORAGE=RedisDocStatusStorage` silently degrades: the dedup path hits a missing-attribute fallback and re-ingests already-known documents.

This PR closes that backend gap without changing public behavior on other backends or altering the storage schema.

## What's Changed

`lightrag/kg/redis_impl.py` (+95 lines, no deletions):
- `RedisDocStatusStorage.get_doc_by_file_basename(basename)`:
  - Early-returns `None` for empty input and for the `"unknown_source"` sentinel (matches JSON behavior — otherwise every unsourced doc would collide).
  - SCAN + pipeline `GET` across `{namespace}:*`, exact-match on `file_path`.
  - Returns `(doc_id, doc_data)` so callers can update the existing record without a second round-trip.
- `RedisDocStatusStorage.get_doc_by_content_hash(content_hash)`:
  - Early-returns `None` for empty input — important for legacy rows that pre-date the `content_hash` field, so an empty query never collides with a missing field.
  - Same SCAN pattern, exact-match on `content_hash`.

Both methods follow the existing `get_doc_by_file_path` template (no new indexing, no schema change). Complexity is O(n) over the namespace — identical to the JSON backend's full-scan, and intentionally not optimized in this PR to keep the diff minimal. RediSearch / secondary-index optimization is a separate concern that should be tackled across backends together.

`tests/test_redis_doc_status_lookup.py` (new, 271 lines):
- 8 unit tests using an in-memory `_FakeRedis` + `_FakePipeline` (no live Redis service required; marked `pytest.mark.offline`).
- Covers: hit, miss, empty input, `unknown_source` sentinel, legacy rows without `content_hash` not being matched by empty or non-empty queries.

Other Redis methods (`upsert`, `get_by_id`, `get_by_ids`) are intentionally untouched: the new doc-status fields (`content_hash`, `sidecar_location`, `parse_format`, `parse_engine`, `process_options`, `chunk_options`, `heading`, `sidecar`) are transparent dicts that `json.dumps` / `json.loads` already handle, and read-side `.get()` returns `None` for legacy rows.

## What's Broken & How It Works

Nothing intentionally broken. Callers that previously relied on Redis silently lacking these methods (and falling back to a no-dedup path) will now actually deduplicate against existing records. If you have a Redis-backed deployment and have not been deduplicating, **this PR will start dedup-ing as designed** — that's the intended fix.

How the new lookup works at runtime:
1. Pipeline computes the canonical basename and content_hash of an incoming doc.
2. It calls `get_doc_by_file_basename(basename)` first. Hit → reuse the existing `doc_id` and skip re-parse.
3. Miss → calls `get_doc_by_content_hash(content_hash)` for a content-level dedup.
4. Both miss → enqueue as a new doc.

Edge cases handled:
- Empty basename / hash → guard returns `None` (no scan).
- Stored `file_path == "unknown_source"` → basename query for the sentinel never matches (avoids cross-contaminating unrelated unsourced docs).
- Legacy rows missing `content_hash` → `.get("content_hash")` is `None`, never matches a non-empty query.

## Tests

```
$ pytest tests/test_redis_doc_status_lookup.py -v
8 passed in 0.55s
```

Also verified: `ruff format --check` and `ruff check` pass on changed files.

### DIFF
--- a/lightrag/kg/redis_impl.py
+++ b/lightrag/kg/redis_impl.py
@@ -1110,6 +1110,101 @@ class RedisDocStatusStorage(DocStatusStorage):
                 logger.error(f"[{self.workspace}] Error in get_doc_by_file_path: {e}")
                 return None
 
+    async def get_doc_by_file_basename(
+        self, basename: str
+    ) -> Union[tuple[str, dict[str, Any]], None]:
+        """Find an existing record whose canonical basename matches.
+
+        The caller is responsible for passing an already-canonical basename.
+        Stored ``file_path`` values are canonicalized by the business layer, so
+        this lookup intentionally performs an exact match only.
+        """
+        if not basename:
+            return None
+        if basename == "unknown_source":
+            return None
+
+        async with self._get_redis_connection() as redis:
+            try:
+                cursor = 0
+                while True:
+                    cursor, keys = await redis.scan(
+                        cursor, match=f"{self.final_namespace}:*", count=1000
+                    )
+                    if keys:
+                        pipe = redis.pipeline()
+                        for key in keys:
+                            pipe.get(key)
+                        values = await pipe.execute()
+
+                        for key, value in zip(keys, values):
+                            if not value:
+                                continue
+                            try:
+                                doc_data = json.loads(value)
+                            except json.JSONDecodeError as e:
+                                logger.error(
+                                    f"[{self.workspace}] JSON decode error in get_doc_by_file_basename: {e}"
+                                )
+                                continue
+                            if doc_data.get("file_path") == basename:
+                                doc_id = key.split(":", 1)[1]
+                                return doc_id, doc_data
+
+                    if cursor == 0:
+                        break
+
+                return None
+            except Exception as e:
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3099, opened 2026-05-20
refactor(mongo): align doc-status storage with JSON storage for parity

danielaskdd:
## Summary

Bring the MongoDB storage to the same behavioral contract as the recent JsonKVStorage / JsonDocStatusStorage updates (and the Redis PR #3098 / PG parity refactor):

- **MongoDocStatusStorage**
  - Add `get_doc_by_file_basename`: indexed `find_one` on `file_path` with exact match. Honors the empty-string and `"unknown_source"` sentinel short-circuits required by the base contract.
  - Add `get_doc_by_content_hash`: indexed `find_one` on `content_hash`. Empty hash is a guaranteed miss, mirroring the PG partial-index semantics.
  - Add a partial index on `content_hash` in `create_and_migrate_indexes_if_not_exists` (filter expression: `$exists` + `$type:string` + `$gt:""`) so the index stays small and legacy / empty values are skipped. Wire `partialFilterExpression` through the `create_index` dispatch. Legacy unprefixed `content_hash` index name added to the workspace-prefix migration list.
- **MongoKVStorage (full_docs / text_chunks) needs no schema changes**: BSON is schemaless, and the existing `$set` upsert path persists whatever pipeline-derived fields the caller writes:
  - `full_docs`: `sidecar_location` / `parse_format` / `parse_engine` / `content_hash` / `process_options` / `chunk_options`
  - `text_chunks`: `heading` / `sidecar`
- **Backward compatibility**: `DocProcessingStatus.content_hash` already defaults to `None`, so legacy doc_status rows that lack the field still deserialize cleanly through `_prepare_doc_status_data` + `DocProcessingStatus(**data)`.

## Test plan

- [x] Extend `tests/test_mongo_storage.py` with `TestMongoDocStatusLookup` (7 cases: hit, miss, empty-input short-circuit, `"unknown_source"` sentinel — for both new methods).
- [x] `python -m pytest tests/test_mongo_storage.py -v` → 8 passed (including the pre-existing graph test, runs fully offline).
- [ ] Verify on a real MongoDB instance: confirm the `content_hash` partial index is created (`getIndexes()`) and that `find_one({"content_hash": ...})` / `find_one({"file_path": ...})` hit the indexes.

https://claude.ai/code/session_01Fwxqj1RkesaTpHqsHUuK9e

PR #3100, opened 2026-05-20
feat(opensearch): add basename and content_hash lookups for doc status

danielaskdd:
## Summary

Mirror the JSON / MongoDB (#3099) / Redis (#3098) storage parity work for the OpenSearch backend so the new pipeline dedup paths can resolve documents by canonical basename or content_hash without falling back to the per-status full-scan default implementation in `DocStatusStorage`.

## Motivation

`JsonDocStatusStorage` recently grew two new lookup helpers — `get_doc_by_file_basename` and `get_doc_by_content_hash` — used by the document ingestion pipeline for filename- and content-hash-based deduplication. The base class ships a default implementation that scans every status via `get_docs_by_statuses`, which is acceptable for tiny deployments but pathological for any real OpenSearch index. The Mongo PR (#3099) added native overrides for the same two methods backed by a `content_hash` index; this PR does the equivalent for `OpenSearchDocStatusStorage`.

The pipeline also adds new fields to `full_docs` (`sidecar_location`, `parse_format`, `parse_engine`, `content_hash`, `process_options`, `chunk_options`) and `text_chunks` (`heading`, `sidecar`). Because `OpenSearchKVStorage` already uses `"dynamic": True` mappings, new fields land naturally with no schema migration required, and legacy rows without those keys are read back as missing — consumers already use `.get()` with defaults. The `content_hash` attribute newly added to `DocProcessingStatus` also defaults to `None`, so old doc_status rows deserialize cleanly.

## What's changed

- `OpenSearchDocStatusStorage._create_index_if_not_exists` declares `content_hash` as a `keyword` field for new indices.
- New `_ensure_content_hash_mapping` helper: on startup, when the doc status index already exists, inspect its mapping and `put_mapping` the `content_hash` keyword field if missing. Idempotent — safe to call every initialize. Failures are logged but do not abort startup, mirroring the `_verify_mirrored_id_mapping` posture.
- `get_doc_by_file_basename(basename)` — term query on the existing `file_path` keyword field. Empty input and the `unknown_source` sentinel short-circuit before any cluster round-trip.
- `get_doc_by_content_hash(content_hash)` — term query on the new `content_hash` keyword field. Empty input short-circuits so legacy rows missing the field cannot accidentally match via coercion.
- Both new methods reuse the existing `_is_missing_index_error` / `_mark_index_missing` recovery pattern used by the other read methods.

Nothing breaks. For pre-existing indices created by older LightRAG releases:

- The first `initialize()` after upgrade triggers `_ensure_content_hash_mapping`, which adds the keyword mapping to the live index before any reads hit `content_hash`.
- Existing doc status rows without `content_hash` are unaffected; OpenSearch does not require a reindex for new fields.
- `DocProcessingStatus(**data)` continues to work for rows without `content_hash` because the dataclass field defaults to `None`.

## Test plan

- [x] All 128 existing OpenSearch tests pass unchanged.
- [x] New tests cover hits, misses, empty/short-circuit input, `unknown_source` sentinel for `get_doc_by_file_basename`; hits, misses, empty/short-circuit input for `get_doc_by_content_hash`; mapping added when missing on existing index; mapping skipped when already present.
- [x] `ruff check` clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WcV3wXFusz6Qq3SeRxgbQv)_

### DIFF
--- a/lightrag/kg/opensearch_impl.py
+++ b/lightrag/kg/opensearch_impl.py
@@ -632,6 +632,7 @@ class OpenSearchDocStatusStorage(DocStatusStorage):
                             "status": {"type": "keyword"},
                             "file_path": {"type": "keyword"},
                             "track_id": {"type": "keyword"},
+                            "content_hash": {"type": "keyword"},
                             "created_at": {"type": "date"},
                             "updated_at": {"type": "date"},
                         },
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3104, opened 2026-05-21
feat: expose split_by_character parameter in /documents/text API endpoint

kennedydqz-del:
Exposes the split_by_character and split_by_character_only parameters in the /documents/text and /documents/texts HTTP API endpoints, allowing API consumers to control text chunking behavior when inserting documents.

Previously these parameters were only available through the Python SDK (insert() / ainsert()) and CLI. This PR adds them to the REST API.

Changes:
- Add split_by_character (optional string) and split_by_character_only (bool, default false) to InsertTextRequest
- Add the same fields to InsertTextsRequest
- Thread them through pipeline_index_texts() to apipeline_process_enqueue_documents()
- Update OpenAPI schema examples


danielaskdd:
Thanks for this, @kennedydqz-del — exposing chunk-splitting control on the REST API is a real gap and your PR is what surfaced it. 🙏

We've landed a broader version of this in #3142, so this need is now covered. Rather than flat top-level fields, #3142 adds a single `chunking` request object to `/documents/text` and `/documents/texts` that:

- lets clients pick a chunking **strategy** (`fixed_token` / `recursive_character` / `semantic_vector` / `paragraph_semantic`), and
- passes strategy-specific params under `chunking.params`, with full **synchronous** validation (malformed requests get HTTP 422 *before* the background task is scheduled, instead of failing the document after a 200).

Your `split_by_character` / `split_by_character_only` map directly onto the `fixed_token` strategy's params. The equivalent of this PR's request is:

```json
{
  "text": "content...",
  "file_source": "manual.md",
  "chunking": {
    "strategy": "fixed_token",
    "params": {
      "split_by_character": "\n\n",
      "split_by_character_only": true
    }
  }
}
```

We deliberately went with a nested `chunking` object instead of flat top-level fields so the API can grow (R/V/P strategies, per-strategy `chunk_token_size` / overlap / separators, etc.) without accumulating one-off top-level knobs — that trade-off is captured in the #3142 design and noted explicitly as "the PR #3104 flat fields are intentionally not adopted."

Closing as superseded by #3142. Thank you again for the contribution and for pushing this forward — please do review #3142 and shout if the `chunking.params` form doesn't cover your use case.

PR #3107, opened 2026-05-21
fix: omit references section in LLM response when include_references=false

### LINKED ISSUE #2832: [Bug]: References are included in the response even if include_references=false
### Describe the bug

We are using LightRAG 1.4.9.4 with PostgreSQL and pgvector extension. We uploaded a bunch of documents in markdown format.

We call LightRAG via the API.

Request: `POST http://<BASE_URL>:9631/query`

```json
{
    "query": "What is wProgDevice?",
    "mode": "mix",
    "include_references": false
}
```

Response:

```json
{
    "response": "Based on the provided knowledge graph data, **wProgDevice** is a parameter used in **umiController** methods to specify the device type.\n\n### References\n\n- [2] tickets_45.md\n- [1] tickets_52.md",
    "references": null
}
```

As you can see, `references` field is `null` in the response, which is expected based on the `include_references": false` in the request.

However, why are the references mentioned in the `response` field of the response (`### References\n\n- [2] tickets_45.md\n- [1] tickets_52.md`)? We would expect that `"include_references": false` parameter would omit them as well.

kennedydqz-del:
Fixes #2832

## Problem

When `include_references=false` is passed to the `/query` API, the API layer correctly strips the `references` array from the JSON response. However, the LLM still generates a `### References` section in its response text (e.g., `### References\n\n- [2] tickets_45.md\n- [1] tickets_52.md`) because the system prompt always instructs the LLM to include citations from the context.

## Root Cause

The system prompt templates (`rag_response` and `naive_rag_response` in `prompt.py`) always contain detailed instructions for the LLM to generate a references section. The `include_references` parameter only affected the API response envelope, not the prompt sent to the LLM.

## Changes

- **`prompt.py`**: Added two new prompt templates:
  - `rag_response_no_refs` — KG query prompt without citation instructions
  - `naive_rag_response_no_refs` — Naive/vector query prompt without citation instructions
- **`operate.py`**: Select the appropriate prompt variant based on `query_param.include_references`
  - `kg_query()`: Use `rag_response_no_refs` when `include_references=False`
  - `naive_query()`: Use `naive_rag_response_no_refs` when `include_references=False`

When `include_references=True` (or a custom `system_prompt` is provided), behavior is unchanged.


danielaskdd:
Thanks for the contribution, @kennedydqz-del, and for the clear write-up.

After review, the goal here — suppressing the `### References` section in the LLM's text output — is already achievable at runtime via the existing `user_prompt` query parameter, with no template changes. For that reason we won't be merging this PR.

**Why it already works**

In both `rag_response` and `naive_rag_response`, the `{user_prompt}` placeholder is injected as instruction **#6 "Additional Instructions"** ([prompt.py#L380](https://github.com/HKUDS/LightRAG/blob/main/lightrag/prompt.py#L380) / [#L434](https://github.com/HKUDS/LightRAG/blob/main/lightrag/prompt.py#L434)), which comes *after* the **#4 "References Section Format"** instruction ([L363-L369](https://github.com/HKUDS/LightRAG/blob/main/lightrag/prompt.py#L363-L369) / [L417-L423](https://github.com/HKUDS/LightRAG/blob/main/lightrag/prompt.py#L417-L423)). Because the user prompt is positioned last, it can override the earlier references directive. `kg_query()` / `naive_query()` inject the value verbatim via `.format(...)`.

So passing the following as `user_prompt` — either through the `/query` API parameter or the **User Prompt** field in the Web UI retrieval panel — reliably drops the references section:

> Ignore the `References Section Format` instruction in the system prompt, and do not include a `References` section in the response.

This requires no code change, works per-request, and is controllable by every user. We've also seeded this exact prompt as a built-in suggestion in the Web UI prompt history (see #3266), so it's available out of the box.

**On `include_references`**

`include_references` is intentionally an API-response-envelope flag only — it strips the `references` array from the returned JSON ([query_routes.py#L452](https://github.com/HKUDS/LightRAG/blob/main/lightrag/api/routers/query_routes.py#L452)) and is deliberately decoupled from prompt content. We'd rather keep prompt behavior driven by `user_prompt` than branch the templates into `_no_refs` variants, which adds maintenance surface and a second source of truth for the same prompts.

Issue #2832 is addressed by the `user_prompt` approach above. Really appreciate the effort — please feel free to continue the discussion if you find a case `user_prompt` can't cover.

PR #3128, opened 2026-05-24
fix(server): normalize scope.path so Mount works in proxy-strip mode

Xaverrrrr:
## Description

`LIGHTRAG_API_PREFIX` exists to support deployments where one host serves multiple LightRAG sites behind a reverse proxy. `docs/MultiSiteDeployment.md` documents two compatible flows:

- **proxy-strip mode** (the nginx example at L138-150): the proxy strips `/site01/` and forwards `/webui/` to the backend.
- **verbatim mode** (the Vite dev flow at L279-315): the proxy forwards `/site01/webui/` unchanged.

Both flows must work because Vite dev relies on verbatim forwarding while production nginx examples rely on the strip form.

### The actual bug

Without uvicorn's `root_path` set, **strip mode breaks the WebUI Mount** even though plain Routes still work:

- Backend receives `scope["path"]="/webui/"`, `scope["root_path"]=""`.
- FastAPI's `__call__` overrides `scope["root_path"]="/site01"` (since 0.108).
- Outer `Mount.matches` resolves via `get_route_path`'s fallback (path that does not start with root_path is returned unchanged → `/webui/`), regex hits, and mutates the child scope to `root_path="/site01/webui"` while leaving `scope["path"]` as `/webui/`.
- Inner `StaticFiles.get_path` calls `get_route_path` again: `/webui/` does not start with `/site01/webui` → returns `/webui/` unchanged → `os.path.normpath` turns it into the filename `webui` looked up inside the static dir → 404.

Plain Routes do not chain into a nested `get_route_path`, which is why `/openapi.json` works in strip mode while `/webui/` does not.

The originally reported "redirect drops the prefix" symptom is a different but related issue, and is real on FastAPI < 0.108 (no `scope["root_path"]` override in `__call__`).

### Why PR #3128's original direction was incomplete

Setting uvicorn's `root_path` would have fixed strip mode (it reconstructs the canonical ASGI form where `scope["path"]` always includes `root_path`), but it breaks verbatim mode: uvicorn builds `scope["path"] = root_path + <incoming>`, so a verbatim-forwarded `/site01/webui/` becomes `/site01/site01/webui/` → 404. The two modes are incompatible at the uvicorn config layer.

### What this PR does

A small ASGI middleware (`_RootPathNormalizationMiddleware`) runs before routing and prepends `root_path` to `scope["path"]` iff it does not already start with `root_path`. Both modes converge to the canonical ASGI form without touching uvicorn's config — Mount routing works in both, no doc rewrite needed.

## Changes

- Add `_RootPathNormalizationMiddleware` in `lightrag/api/lightrag_server.py`; wire it into `create_app` before CORS when `api_prefix` is set
- Hoist `_normalize_api_prefix` to module scope (folded in from `7522a909`'s cleanup)
- Pin `fastapi>=0.108` in `pyproject.toml` so the redirect handler's `scope["root_path"]` lookup is always populated
- Drop the original `root_path` line from `lightrag_server.main()`'s uvicorn kwargs
- Add `TestUvicornRootPathSemantics` (7 tests) in `tests/test_path_prefixes.py` covering Route and Mount routing in both deployment modes, plus launcher kwargs and gunicorn worker_class guards

## Manual verification

`LIGHTRAG_API_PREFIX=/site01`, server on `localhost:9621`:

| GET | Before | After |
|---|---|---|
| `/webui/` | 404 | 200 |
| `/site01/webui/` | 200 | 200 |
| `/openapi.json` | 200 | 200 |
| `/site01/openapi.json` | 200 | 200 |

## Related Issues

- [#3039](https://github.com/HKUDS/LightRAG/pull/3039)

## Testing

`tests/test_path_prefixes.py`: 26/26 passing (19 existing + 7 new in `TestUvicornRootPathSemantics`).

## Checklist

- [x] Changes were tested
- [x] Code reviewed
- [x] Documentation reflects the actual fix
- [x] Unit tests added

### DIFF
--- a/docs/MultiSiteDeployment.md
+++ b/docs/MultiSiteDeployment.md
@@ -73,11 +73,20 @@ and in-app links.
 
 | Variable | Default | Meaning |
 | --- | --- | --- |
-| `LIGHTRAG_API_PREFIX` | `""` | Reverse-proxy prefix that the upstream proxy strips before forwarding to FastAPI. Passed to FastAPI as `root_path`. |
+| `LIGHTRAG_API_PREFIX` | `""` | Reverse-proxy mount prefix. The backend accepts both strip and verbatim forwarding — pick whichever fits your proxy stack. Passed to FastAPI as `root_path`. |
 
 The WebUI is always mounted at `/webui` server-side. `window.__LIGHTRAG_CONFIG__.webuiPrefix` is computed as `LIGHTRAG_API_PREFIX + "/webui/"` and injected for the SPA — you do **not** set it yourself.
 
 There are no longer any frontend `VITE_API_PREFIX` / `VITE_WEBUI_PREFIX` variables. Setting them has no effect (they are ignored by the build).
+
+### Forwarding modes: strip and verbatim both work
+
+After setting `LIGHTRAG_API_PREFIX=/site01`, the backend resolves all routes correctly under either forwarding style:
+
+- **Strip** — proxy removes the prefix, backend sees `/webui/` and `/documents/foo`. The nginx example below uses this style.
+- **Verbatim** — proxy forwards the request unchanged, backend sees `/site01/webui/` and `/site01/documents/foo`. The Vite dev flow ([Scenario 2](#scenario-2--simulate-a-site-prefix)) and any non-rewriting proxy use this style.
+
+A small ASGI middleware in `create_app` prepends `root_path` to `scope["path"]` whenever the path does not already include it, so plain Routes and Mount sub-apps (the WebUI's `StaticFiles`) both resolve identically in either mode. You do not need to standardize on one — both coexist on the same backend without configuration toggles.
 
 ---
 
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -2566,6 +2566,7 @@ def main():
         "host": global_args.host,
         "port": global_args.port,
         "log_config": None,  # Disable default config
+        "root_path": global_args.api_prefix if global_args.api_prefix else "", #properly pass api prefix to uvicorn
     }
 
     if global_args.ssl:
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -385,6 +385,60 @@ def _inject_swagger_theme(html: str, theme: str) -> str:
 WEBUI_PATH = "/webui"
 
 
+def _normalize_api_prefix(value: str | None) -> str:
+    """Canonicalize an API prefix before handing it to FastAPI's ``root_path``.
+
+    Strips surrounding whitespace, ensures a leading slash, drops a trailing
+    slash, and treats empty/"/" as "no prefix". Raw CLI/env input like
+    ``"site01"`` or ``"/site01/"`` would otherwise feed an invalid form to
+    FastAPI and to the WebUI prefix injection.
+    """
+    if value is None:
+        return ""
+    value = value.strip()
+    if not value or value == "/":
+        return ""
+    if not value.startswith("/"):
+        value = "/" + value
+    return value.rstrip("/")
+
+
+class _RootPathNormalizationMiddleware:
+    """Make Mount sub-apps work when the reverse proxy strips the API prefix.
+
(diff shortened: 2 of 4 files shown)


danielaskdd:
@codex review

danielaskdd:
Hi @Xaverrrrr — I dug into this and I'm afraid the fix in this PR (even after the `7522a909` follow-up that normalizes the prefix) actually introduces a regression in one of the documented deployment modes. Force-pushed a revert + an alternative fix that addresses the original root cause; details below so you can verify.

## Why setting uvicorn `root_path` breaks `verbatim` mode

Uvicorn builds the ASGI scope with `scope["path"] = root_path + <incoming http path>` (see `uvicorn/protocols/http/h11_impl.py:199-211`, also the WebSocket protocols). With `--api-prefix /site01` and the PR's change:

| Deployment mode | Incoming path | scope["path"] after uvicorn | Starlette strips `root_path` → | Result |
|---|---|---|---|---|
| Reverse proxy strips prefix (`nginx` rewrite) | `/health` | `/site01/health` | `/health` | ✅ 200 |
| Verbatim forwarding (Vite dev / passthrough proxy) | `/site01/health` | `/site01/site01/health` | `/site01/health` | ❌ 404 |

Verbatim mode is an explicitly supported flow per `docs/MultiSiteDeployment.md`:

> SPA requests like `fetch("/site01/documents/foo")` are forwarded verbatim to whatever `VITE_BACKEND_URL` points at. The upstream — local backend or production nginx — is responsible for understanding the prefix.

So the moment uvicorn `root_path` is set, the Vite dev story breaks. The existing `tests/test_path_prefixes.py` didn't catch this because `TestClient` doesn't simulate uvicorn's scope construction.

## Why the original symptom was real, but only on old FastAPI

`request.scope.get("root_path", "")` returning empty is real — **on FastAPI < 0.108**. Since FastAPI 0.108 (Dec 2023), `FastAPI.__call__` overrides `scope["root_path"]` from `app.root_path` (`fastapi/applications.py:1131-1134`):

```python
async def __call__(self, scope, receive, send):
    if self.root_path:
        scope["root_path"] = self.root_path
    await super().__call__(scope, receive, send)
```

So the existing `app_kwargs["root_path"]` is sufficient on its own — no uvicorn-side change is needed on modern FastAPI. The real fix is to pin the supported FastAPI floor.

## What the revert does

- Drop the uvicorn `root_path` kwarg
- Pin `fastapi>=0.108` in `pyproject.toml` (the actual root cause)
- Keep `_normalize_api_prefix` hoisted to module level (your `7522a909` cleanup — folded in)
- Add `TestUvicornRootPathSemantics` (5 tests) that lock in both deployment modes, prove the bug we avoided is real, and guard against re-adding `root_path` to either uvicorn or gunicorn launchers

Tests: 24/24 passing.

Let me know if you can reproduce the original redirect-loses-prefix on FastAPI ≥ 0.108 — if so, there's something else going on and we can dig further. Thanks for raising this!

danielaskdd:
Follow-up to my earlier review — I was wrong that the proxy-strip mode worked end-to-end without uvicorn `root_path`. Plain Routes worked, but the WebUI Mount silently 404s. Reproduced locally with `LIGHTRAG_API_PREFIX=/site01` and `GET /webui/` → 404.

Root cause is a Starlette interaction: outer `Mount.matches` mutates `scope["root_path"]` to `/site01/webui` but leaves `scope["path"]="/webui/"`. The inner `StaticFiles.get_path` then sees a non-overlapping pair and falls through to a literal `"webui"` filename lookup. Plain Routes never re-derive a route_path from the mutated scope, which is why they were unaffected and why TestClient missed this.

Force-pushed an updated commit replacing the simple revert with a small ASGI middleware that normalizes `scope["path"]` to always include `root_path` before routing. Both deployment modes (nginx strip + Vite verbatim) now route correctly; no uvicorn `root_path` needed.

Updated the PR title and description to reflect the new shape. 7 regression tests now cover Route and Mount in both modes plus launcher guards. Sorry for the back-and-forth — yours was a real bug, just not the one the original patch was reaching for.

Xaverrrrr:
Thank you @danielaskdd for your quick response!
While my solution did the trick on my system, yours is of course much more robust and sophisticated.
Kind regards
Xaver

PR #3129, opened 2026-05-24
fix(concurrency): wrap ainsert_custom_kg writes with keyed lock (follow-up to #3056)

danielaskdd:
## Summary

Follow-up to #3056. That PR added a PostgreSQL advisory lock inside `PGGraphStorage.upsert_edge` as the **storage-layer** safeguard against duplicate `DIRECTED` rows under concurrent writers. This PR fills in the one real business-layer gap left open at the time:

**`ainsert_custom_kg → upsert_edges_batch` had no business-layer keyed lock.** Two concurrent `insert_custom_kg` calls touching overlapping entities could race on the same `(src, tgt)` pair without the business layer noticing.

### `lightrag/lightrag.py` — wrap `ainsert_custom_kg` writes in a keyed lock

After dedup, compute `lock_key_set` = union of every entity name and every relationship endpoint in the batch. Acquire `get_storage_keyed_lock(sorted(lock_key_set), namespace=f\"{workspace}:GraphDB\")` once, then run all graph + vdb writes (entity upserts, missing-node fill, edge upserts, vdb upserts, legacy-vdb deletes) inside the same critical section. Namespace matches what `operate.py:_locked_process_edges` and `_locked_process_entity_name` use, so the per-key mutexes (`get_storage_keyed_lock` acquires one mutex per key in the namespace) collide across paths — a concurrent custom-KG insert waits behind an in-flight document ingest instead of racing it. Empty batches skip the lock entirely.

### `lightrag/kg/postgres_impl.py` — deterministic batch iteration

`upsert_edges_batch` now iterates `sorted(deduped_edges)` instead of `deduped_edges.values()`. The dedup key is already `tuple(sorted((src, tgt)))`, so iterating by sorted key gives every caller the same global order. **Not a deadlock fix** — `upsert_edge` opens an independent transaction per call and releases the advisory lock on commit, so the batch never holds two advisory locks simultaneously. The change is purely for reproducible logs / replays.

## Test plan

- [x] **`tests/test_graph_keyed_locks.py`** (new, 5 cases) pins the keyed-lock contracts:
  - `test_aedit_entity_rename_locks_old_and_new_names` — rename locks `{old, new}`, no pre-fetch of incident edges.
  - `test_aedit_entity_non_rename_locks_single_entity_name` — non-rename locks `{entity}`.
  - `test_adelete_by_entity_locks_single_entity_name` — delete locks `{entity}`.
  - `test_ainsert_custom_kg_locks_every_entity_and_endpoint` — verify the new coarse-grained lock includes every entity name + every relationship endpoint, sorted.
  - `test_ainsert_custom_kg_empty_batch_skips_keyed_lock` — empty batch acquires no lock at all.
- [x] **`tests/test_postgres_upsert_edge_cypher.py`** additions:
  - `test_upsert_edges_batch_iterates_in_sorted_order` — deterministic iteration order regardless of insertion order.
  - `test_upsert_edges_batch_dedupes_last_write_wins` — reciprocal duplicates collapse to one upsert with the latest `edge_data`.
- [x] Full `pytest tests/` — **1601 passed, 1 skipped, 31 deselected**.
- [x] `pre-commit run` clean.

## Review notes

The first commit on this branch attempted to also extend the lock set in `aedit_entity` / `adelete_by_entity` by pre-fetching incident edges, and framed the `upsert_edges_batch` sort as an AB-BA deadlock fix. **Both rationales were wrong:**

- `get_storage_keyed_lock` acquires one mutex per key in a shared namespace. Identical key strings already mutually exclude across callers, so `[entity_name]` already collides with `sorted([src, tgt])` whenever the edge names the entity — no need to enumerate incident edges.
- `upsert_edges_batch` calls `upsert_edge` one edge at a time, and each call commits before the next starts, so the batch never holds two advisory locks at once. There is no AB-BA risk to fix.

The second commit on this branch (`ce60a7a4`) reverts those over-reaches and rewrites the corresponding comments / tests to reflect the correct concurrency model. The final scope is the single `ainsert_custom_kg` lock + a cosmetic deterministic-iteration tweak on `upsert_edges_batch`.

### DIFF
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -5747,11 +5747,12 @@ class PGGraphStorage(BaseGraphStorage):
             deduped_edges.pop(edge_key, None)
             deduped_edges[edge_key] = (src, tgt, edge_data)
 
-        # Iterate in canonical (LEAST, GREATEST) order so concurrent
-        # upsert_edges_batch calls always acquire the per-edge advisory locks
-        # in the same global order. Without this, dict insertion-order could
-        # let two batches acquire locks for {A,B} and {C,D} in opposite
-        # sequences and deadlock on each other in PostgreSQL.
+        # Iterate in canonical (LEAST, GREATEST) order rather than dict
+        # insertion order. upsert_edge opens an independent transaction per
+        # call and releases the advisory lock on commit, so this is not a
+        # deadlock fix — but a deterministic iteration order makes logs and
+        # replays reproducible across callers, and matches the dedup key
+        # already used above.
         for edge_key in sorted(deduped_edges):
             src, tgt, edge_data = deduped_edges[edge_key]
             await self.upsert_edge(src, tgt, edge_data=edge_data)
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -5747,7 +5747,13 @@ class PGGraphStorage(BaseGraphStorage):
             deduped_edges.pop(edge_key, None)
             deduped_edges[edge_key] = (src, tgt, edge_data)
 
-        for src, tgt, edge_data in deduped_edges.values():
+        # Iterate in canonical (LEAST, GREATEST) order so concurrent
+        # upsert_edges_batch calls always acquire the per-edge advisory locks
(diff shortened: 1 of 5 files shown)


danielaskdd:
@claude review once

Issue #3126, opened 2026-05-24
[Bug]: POSTGRES user, password, db are not configurable


KD-MM2:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

As title, postgres user, password, db are not configurable.

Found how to fix:

- Checking the postgres image: https://hub.docker.com/r/gzdaniel/postgres-for-rag

> Default admin user and password: rag/rag

- then change env, compose:

```
      POSTGRES_USER: "rag"
      POSTGRES_PASSWORD: "rag"
      POSTGRES_DB: "rag"
```

remove postgres docker volume, force-recreate then OK.

### Steps to reproduce

```
make env-base
make env-storage
docker compose -f docker-compose.final.yml up -d
```

### LightRAG Config Used

env: 
```
############################
### Data storage selection
############################
### Default storage: JSON/Nano/NetworkX (Recommended for test deployment)
LIGHTRAG_KV_STORAGE=PGKVStorage
# LIGHTRAG_KV_STORAGE=JsonKVStorage
LIGHTRAG_DOC_STATUS_STORAGE=PGDocStatusStorage
# LIGHTRAG_DOC_STATUS_STORAGE=JsonDocStatusStorage
LIGHTRAG_GRAPH_STORAGE=Neo4JStorage
# LIGHTRAG_GRAPH_STORAGE=NetworkXStorage
LIGHTRAG_VECTOR_STORAGE=QdrantVectorDBStorage
# LIGHTRAG_VECTOR_STORAGE=NanoVectorDBStorage

### Wizard metadata used to preserve env-storage Docker deployment defaults across setup reruns
LIGHTRAG_SETUP_POSTGRES_DEPLOYMENT=docker
LIGHTRAG_SETUP_NEO4J_DEPLOYMENT=docker
# LIGHTRAG_SETUP_MONGODB_DEPLOYMENT=docker
# LIGHTRAG_SETUP_MONGODB_DEPLOYMENT=atlas-capable
# LIGHTRAG_SETUP_REDIS_DEPLOYMENT=docker
# LIGHTRAG_SETUP_MILVUS_DEPLOYMENT=docker
LIGHTRAG_SETUP_QDRANT_DEPLOYMENT=docker
# LIGHTRAG_SETUP_MEMGRAPH_DEPLOYMENT=docker
[...]

PR #3132, opened 2026-05-25
fix issue #3126 POSTGRES user, password, db are not configurable

akash1922:
approach to address this:
> Force 'rag' under Docker: When PostgreSQL runs via local Docker, the wizard enforces rag / rag / rag and skips credential prompts to prevent configuration mismatches.

> Preserve External DB Customization: When running on an external or host PostgreSQL instance, the wizard still prompts for custom credentials.

>Updated Tests: Adjusted the unit tests in tests/test_interactive_setup/test_collect.py to align with this logics.


akash1922:
ok working on it will make it resolve

danielaskdd:
Thank you @akash1922 for digging into this and opening the PR — much appreciated! 🙏

Issue #3126 has now been addressed in #3141, which takes a slightly different approach: rather than keeping the forced `rag` / `rag` / `rag` credentials under Docker, it relies on the fact that the latest `gzdaniel/postgres-for-rag:pg18-age-pgvector` image no longer ships hardcoded credentials — it creates the user/password/database from the `POSTGRES_USER` / `POSTGRES_PASSWORD` / `POSTGRES_DB` environment variables on first start. So the wizard now lets you customize the credentials for Docker deployments as well (defaulting to `rag` / `rag` / `lightrag`, matching the host path).

Since #3141 fully covers this, I'm closing this PR. Thanks again for surfacing the problem and for your contribution!

PR #3133, opened 2026-05-25
perf(opensearch): batch KV operations under namespace lock

danielaskdd:
## Summary

Refs #2785. Supersedes the KV-storage portion of #2822 (Vector was already covered by #3043; Graph is deferred to a separate follow-up).

`OpenSearchKVStorage.upsert()` and `delete()` each previously issued an `async_bulk` call per invocation. The profile in #2785 shows the upstream pipeline calling KV `upsert` 651 times for the demo workload, dominating the KV path with sequential HTTP roundtrips even though every call is technically a "bulk" of 1.

This PR buffers writes and deletes in memory and flushes them in a single `async_bulk` call inside `index_done_callback()` / `finalize()`, matching the contract documented on `BaseKVStorage` ("changes will be persisted during the next `index_done_callback`"). All buffer reads / writes are serialised through `get_namespace_lock`, mirroring the model introduced for the Vector backend in #3043.

Credit goes to @he-yufeng for the original KV batching implementation in #2822; this PR is a rebase + redesign for the post-#3043 lock model with a fix for the resurrect bug (see below).

### Batched writes
- `upsert()` time-stamps + injects `__mirrored_id` eagerly outside the lock; the buffer-write loop is the only step that holds `_flush_lock`. An upsert cancels any pending tombstone for the same id.
- `delete()` buffers ids into `_pending_kv_deletes`. A delete cancels any pending upsert for the same id. `_index_ready` is intentionally NOT checked — the cancellation invariant must hold even if the index has been marked missing.
- `_flush_pending_kv_ops()` issues a single `async_bulk` for all pending deletes + upserts (deletes first), lock-guarded. Reuses the `_extract_bulk_failed_ids` / `_FailedBulkOp` / `_summarize_bulk_error` helpers added by #3043.
- `index_done_callback()` flushes pending ops before refreshing the index.
- `finalize()` flushes pending ops, releases the client unconditionally to avoid connection leaks, **and raises `RuntimeError` if any ops remained buffered after the final flush** (e.g. due to retryable 5xx failures). Without this last step, `LightRAG.finalize_storages()` would log the storage as successfully finalized while buffered writes silently failed to reach OpenSearch. The same fix is applied to `OpenSearchVectorDBStorage.finalize()`, which had the identical bug since #3043.
- `drop()` runs entirely under `_flush_lock`, clearing both buffers and serialising with any in-flight flush.

### Concurrency
- `_flush_lock` is a `get_namespace_lock(namespace, workspace=...)` instance (multi-process aware), set in `initialize()`. Same primitive as `NanoVectorDBStorage._storage_lock` and `OpenSearchVectorDBStorage._flush_lock`.
- All buffer mutations and reads acquire this lock; remote IO (mget, `client.count`) runs outside the lock so RTT does not block flushes.

### Read-your-writes, with consistent shape
- `get_by_id` / `get_by_ids` follow a three-tier priority under the lock: pending delete (return `None`) → pending upsert (return buffered source) → fall through to mget.
- The buffered path strips `__mirrored_id` (the PIT-sort sentinel injected during upsert) and attaches `_id` / `create_time` / `update_time`, matching the mget read path. Without this strip, callers would see a `__mirrored_id` field for buffered docs but not for persisted ones.
- `filter_keys` treats buffered upserts as existing (excluded from the missing set) and buffered deletes as missing (the tombstone id is NOT sent to mget — otherwise a persisted-but-pending-delete row would be misclassified as existing).
- `is_empty` short-circuits to False when the buffer has pending upserts, avoiding the counterintuitive "I just upserted but is_empty returned True" outcome.

### Failure handling
- Per-doc bulk failures are classified by HTTP status via `_extract_bulk_failed_ids`:
  - Retryable (408 / 429 / 5xx / missing status): kept in the buffer for the next flush.
  - Non-retryable (most 4xx, mapping errors): dropped from the buffer. A sample of `(op, id, status, error)` is logged at WARNING with dict-shape errors summarised via `reason` / `type`, capped at 200 chars; remaining failures dump at DEBUG.
  - `404` on a delete is treated as success-equivalent.
- If `helpers.async_bulk` raises, the buffers are left intact — KV ops are idempotent on `_id`.

## The resurrect bug (and why KV delete also buffers)

Without delete buffering, mixing upsert + delete on the same id would silently re-index the row:

```
T0: upsert("k1", v1)          → _pending_upserts["k1"] = v1
T1: delete(["k1"])            → directly async_bulk delete; buffer untouched
T2: get_by_id("k1")           → buffer hit returns v1 (the doc has been deleted)
T3: index_done_callback flush → re-indexes the buffered v1, reviving the row
```

The fix is the upsert↔delete cancellation pair: each buffer entry is the authoritative pending state, and a delete actively pops the upsert from `_pending_upserts` (not the other way around). The flush then sees the correct intent and emits one delete action.

## Backwards compatibility

No public API changes. Method signatures, return types, and error paths are unchanged.

Behavioural changes worth noting:

1. **KV writes are deferred until `index_done_callback()` / `finalize()`** — matches `BaseKVStorage`'s documented contract and the behaviour of every other LightRAG storage backend.

2. **Cross-worker read-after-write visibility now requires `index_done_callback()`** (multi-worker deployments, e.g. `lightrag-gunicorn`). Before this PR, OpenSearch's per-call `async_bulk` made writes visible to other workers via mget/translog within milliseconds. After this PR, the in-memory buffer is **process-local**, so Worker B's `get_by_id` will not see Worker A's buffered writes until A has flushed. The `BaseKVStorage` docstring now states this explicitly. Application code that relies on cross-worker sub-second visibility must `await rag.text_chunks.index_done_callback()` (or wait for the pipeline's own `_insert_done()`) before reading from another worker. The pipeline itself is unaffected because it holds `pipeline_status.busy` for the duration of ingestion.

## Out of scope

- **Graph storage batching**: deferred to a separate PR. Graph has ~12 read paths that would need to consult the buffer (`has_node` / `get_node` / `get_edge` / `get_node_edges` / `get_edges_batch` / `get_nodes_batch` / `node_degree` / `edge_degree` / `get_knowledge_graph` / `get_all_labels` / `get_popular_labels` / `search_labels`) plus inter-entity cascades (delete_node invalidating buffered edges). Single PR risk too high; cleaner to split.
- **Vector storage**: already covered by #3043, untouched here.
- **Sharing the failure-handling helper across backends**: `_extract_bulk_failed_ids` is reused; full base-class extraction can wait until Graph lands.

## Test plan

- [x] `ruff check .` passes.
- [x] `pytest tests/test_opensearch_storage.py` — 168 passed (17 new + 3 updated).
- [x] `TestKVStorageBatching` covers:
  - repeated `upsert()` calls collapse into a single `async_bulk`
  - upserting the same id twice keeps the latest payload
  - `delete()` cancels a pending upsert; `upsert()` cancels a pending delete
  - `delete()` still cancels pending upsert when `_index_ready` is False (resurrect-bug guard)
  - `get_by_id` reads the buffer (no `mget`)
  - `get_by_id` returns `None` for pending tombstones (no `mget`)
  - `get_by_id` strips `__mirrored_id` from the buffer path
  - `get_by_ids` merges buffer + mget for unbuffered ids only
  - `filter_keys` excludes buffered upserts from the missing set
  - `filter_keys` treats buffered deletes as missing AND skips them in mget
  - `is_empty` returns False with a pending upsert (no `client.count`)
  - `finalize()` flushes before releasing the client
  - `finalize()` raises when retryable bulk failures left ops buffered (KV + Vector)
  - `finalize()` wraps a flushing `OpenSearchException` in a RuntimeError that names the unflushed buffer counts (KV + Vector)
  - `drop()` discards both buffers AND is serialised with an in-flight flush
  - per-doc retryable failures (5xx) are retained for retry
  - per-doc non-retryable failures (4xx) are dropped
  - concurrent `upsert` during flush is blocked by the lock and lands after release
- [x] All 151 pre-existing OpenSearch tests still pass.

### DIFF
--- a/lightrag/base.py
+++ b/lightrag/base.py
@@ -404,6 +404,17 @@ class BaseKVStorage(StorageNameSpace, ABC):
         Importance notes for in-memory storage:
         1. Changes will be persisted to disk during the next index_done_callback
         2. update flags to notify other processes that data persistence is needed
+
+        Multi-worker note:
+            Backends that buffer writes in process memory (e.g.
+            OpenSearchKVStorage as of the KV-batching change derived from
+            #2822) keep the buffer process-local. In a multi-worker
+            deployment (e.g. lightrag-gunicorn) other workers will not
+            observe these writes until the writing worker has called
+            index_done_callback(). Callers that depend on cross-worker
+            read-after-write visibility must explicitly await
+            index_done_callback() before relying on reads from another
+            worker.
         """
 
     @abstractmethod
@@ -413,6 +424,9 @@ class BaseKVStorage(StorageNameSpace, ABC):
         Importance notes for in-memory storage:
         1. Changes will be persisted to disk during the next index_done_callback
         2. update flags to notify other processes that data persistence is needed
+
+        Multi-worker note: see ``upsert`` -- buffered tombstones are
+        process-local until index_done_callback() runs.
 
         Args:
             ids (list[str]): List of document IDs to be deleted from storage
--- a/lightrag/kg/opensearch_impl.py
+++ b/lightrag/kg/opensearch_impl.py
@@ -458,13 +458,45 @@ class OpenSearchKVStorage(BaseKVStorage):
             raise
 
     async def finalize(self):
-        """Flush pending writes and release the OpenSearch client connection."""
+        """Flush pending writes and release the OpenSearch client connection.
+
+        Raises ``RuntimeError`` if the buffer still contains pending ops
+        after the final flush attempt -- e.g. because retryable bulk
+        failures (5xx) left rows behind. The client is released either
+        way so we don't leak a connection on shutdown, but the caller
+        must learn that buffered data did not reach OpenSearch.
+        """
+        flush_error: BaseException | None = None
         try:
             await self._flush_pending_kv_ops()
-        finally:
-            if self.client is not None:
-                await ClientManager.release_client(self.client)
-                self.client = None
+        except BaseException as e:
+            # _flush_pending_kv_ops leaves the buffers intact on raise;
+            # capture the error so we can re-surface it after the client
+            # is released (we still must not leak the connection).
+            flush_error = e
+
+        if self.client is not None:
+            await ClientManager.release_client(self.client)
+            self.client = None
+
+        # Snapshot remaining buffer state to report concrete counts.
+        pending_upserts = len(self._pending_upserts)
+        pending_deletes = len(self._pending_kv_deletes)
+
+        if flush_error is not None:
+            raise RuntimeError(
+                f"[{self.workspace}] OpenSearchKVStorage.finalize() flush "
+                f"raised; {pending_upserts} pending upserts and "
+                f"{pending_deletes} pending deletes were left buffered "
+                f"(client released, data lost)"
+            ) from flush_error
+        if pending_upserts or pending_deletes:
+            raise RuntimeError(
+                f"[{self.workspace}] OpenSearchKVStorage.finalize() left "
+                f"{pending_upserts} pending upserts and {pending_deletes} "
+                f"pending deletes buffered after final flush attempt "
+                f"(transient bulk failure); these writes have been lost"
+            )
 
     async def _iter_raw_docs(
         self, batch_size: int = 1000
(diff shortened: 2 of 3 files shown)


danielaskdd:
Pushed `1a92b63` to address the silent-data-loss path in `finalize()`:

> `_flush_pending_kv_ops` keeps retryable per-doc failures in the buffers and returns normally, so the original `finalize()` released the client without any signal that ops were unflushed. `finalize_storages()` would then log the storage as successfully finalized while the writes never reached OpenSearch.

Fix:

- `OpenSearchKVStorage.finalize()` and `OpenSearchVectorDBStorage.finalize()` now:
  1. Capture any exception from the flush attempt (the flush itself leaves buffers intact on raise).
  2. **Release the client unconditionally** — we'd rather know the data is lost than leak a connection.
  3. Snapshot the post-flush pending counts.
  4. Raise `RuntimeError` if either (a) the flush raised, or (b) the buffers still contain ops. The error message names the unflushed `upserts` / `deletes` counts; the flush exception is chained via `raise ... from e`.

`LightRAG.finalize_storages()` already wraps each `storage.finalize()` call in a per-storage try/except (lines 1174–1183 of `lightrag.py`), so the error is logged against the storage name and the other storages still finalize. Net result: the loss is visible without aborting shutdown of the rest.

The Vector fix is in scope because the bug pattern was identical (introduced together with the buffer model in #3043); applying the fix in both backends keeps the contract consistent.

Tests added (4 total): for each of KV / Vector,
- `test_*_finalize_raises_when_retryable_buffer_remains`: 503 keeps an op buffered → `finalize()` raises `RuntimeError("…pending upserts…")`; client released.
- `test_*_finalize_propagates_flush_exception`: `async_bulk` raises `OpenSearchException` → `finalize()` raises `RuntimeError` whose `__cause__` is the original exception; client released.

172 tests passing, ruff clean.

danielaskdd:
Good catch — pushed `dacf5e2` for the cancellation-swallowing problem.

The previous `finalize()` used `except BaseException as e` and then `raise RuntimeError(...) from e`, which turned `asyncio.CancelledError` (and `KeyboardInterrupt` / `SystemExit`) into a plain `Exception`. `LightRAG.finalize_storages()` only catches `Exception` and logs the storage as failed to finalize, so a cancellation arriving during shutdown would never reach the shutdown coordinator — the parent task continued running as if it hadn't been cancelled.

Fix:

- Narrow the catch to `except Exception as e` — regular flush failures (`OpenSearchException`, etc.) still get wrapped into the `RuntimeError("…pending upserts…")` we already raise.
- Move the client release into a `finally` block. Now whether the flush succeeds, raises an `Exception`, or is cancelled, the connection is released and we never leak.
- `CancelledError` / `KeyboardInterrupt` / `SystemExit` propagate out of the `finally` and continue to the caller, so shutdown cancellation is honoured.

Both `OpenSearchKVStorage.finalize()` and `OpenSearchVectorDBStorage.finalize()` use the new pattern.

Two new tests:

- `test_kv_finalize_propagates_cancellation`
- `test_vector_finalize_propagates_cancellation`

Each mocks `helpers.async_bulk` to raise `asyncio.CancelledError`, then asserts `pytest.raises(asyncio.CancelledError)` from `finalize()` AND that the client was released first (via `release_client.assert_awaited_once()`).

174 tests passing, ruff clean.

PR #3136, opened 2026-05-25
🐛 fix(setup): treat opensearch-dashboards as opensearch sub-service

danielaskdd:
## Summary

- `make server` was no longer prompting users to "switch LightRAG to host mode" when their `.env` no longer declared any `_DEPLOYMENT=docker` services
- Root cause: orphan `dashboards` blocks left in `docker-compose.final.yml` (e.g. after a user disables OpenSearch docker deployment) made `compose_has_non_wizard_services` return true, short-circuiting `resolve_compose_output_action` at branch B (`setup.sh:767`) before ever reaching the host-mode prompt at `setup.sh:774`
- Fix: collapse `dashboards` under the `opensearch` root in `_managed_service_root_name`, matching how `milvus-{etcd,minio}` collapse to `milvus`

## Test plan

- [x] `pytest tests/test_interactive_setup/` — 258 passed (was 256), includes two new regression tests:
  - `test_detect_managed_root_services_groups_opensearch_dashboards_under_opensearch`
  - `test_compose_has_non_wizard_services_ignores_orphan_dashboards`
- [x] End-to-end verified by sourcing `scripts/setup/lib/file_ops.sh` and calling `compose_has_non_wizard_services` against a real-world `docker-compose.final.yml` containing only `lightrag` + orphan `dashboards`: returns false after the fix (returned true before), confirming control flow now reaches the host-mode prompt
- [ ] Manual: run `make server` with a compose file that contains only `lightrag` + `dashboards` and verify the prompt `All wizard-managed services have been removed. Remove LightRAG from Docker and switch to host mode?` appears

## Notes

- The wizard does not generate `dashboards` blocks (no `dashboards` matches anywhere under `scripts/setup/` or in `docker-compose-full.yml`). After this change, if a user re-enables OpenSearch docker deployment and the wizard rewrites the opensearch service, the orphan `dashboards` block will be stripped along with it — this is consistent with current behavior for the wizard-managed roots and avoids stale references to a service the wizard does not maintain
- `_managed_volume_root_name` was reviewed; no dashboards-specific volume exists in any wizard-generated compose, so no mapping is needed there


danielaskdd:
Follow-up commit `7fdb1404`: now that `dashboards` collapses under the `opensearch` root, the template generates them as a pair (mirroring `milvus.yml` carrying `milvus-etcd`/`milvus-minio`).

Without this, the previous commit's behaviour was: enabling OpenSearch docker would only emit the `opensearch` block, and a pre-existing user `dashboards` block would be stripped on rewrite — silently breaking a working setup. Now both services are wizard-managed together, so the lifecycle is symmetric: both come in, both go out.

Dashboards service details:
- pinned to `opensearchproject/opensearch-dashboards:3` (matches opensearch image)
- reuses `OPENSEARCH_PASSWORD` and the `admin` user from the demo certs the wizard provisions
- `OPENSEARCH_SSL_VERIFICATIONMODE: none` to accept the self-signed demo cert
- waits on `opensearch` `service_healthy` before starting
- port 5601 commented by default, consistent with other templates — uncomment to expose the UI

Added `test_generate_docker_compose_pairs_dashboards_with_opensearch` to lock this in (259 tests pass).

PR #3139, opened 2026-05-26
feat(graph): block graph mutations while pipeline is busy

danielaskdd:
## Summary

- Hide the entity/relation edit affordances in the WebUI while the document pipeline is busy, and keep any open edit dialog mounted with its draft preserved (save button disabled, banner shown).
- Reject every graph-mutation API call with HTTP 409 when the pipeline is busy, covering both edit/create/merge in `graph_routes.py` and the entity/relation delete endpoints in `document_routes.py`.

## Motivation

`update_entity` / `update_relation` / `delete_*` write the same graph storages that the ingestion pipeline mutates during document processing, scanning, and `/documents/clear`/`/documents/{id}` deletion. Until now the API accepted those mutations unconditionally, so a user editing a node while a re-index ran could overwrite freshly-extracted properties, silently lose merges, or surface confusing partial-success responses. The WebUI had no signal that editing was unsafe either.

### WebUI (`53df4985`)

- [`PropertiesView.tsx`](lightrag_webui/src/components/graph/PropertiesView.tsx) subscribes to `useBackendState.use.pipelineBusy()` (already maintained by the health-check poller) and forwards it down to `PropertyRow` → `EditablePropertyRow` and into a new lock indicator on the node/relation header rows.
- [`EditablePropertyRow.tsx`](lightrag_webui/src/components/graph/EditablePropertyRow.tsx) hides the pencil icon and guards `handleEditClick` while busy. The component stays mounted, so an in-flight edit dialog is **not** unmounted when the pipeline flips to busy — the draft survives.
- [`PropertyEditDialog.tsx`](lightrag_webui/src/components/graph/PropertyEditDialog.tsx) gains a `disableSave` prop. When set, it shows an amber banner and disables the save button (cancel stays available); when the pipeline returns to idle the banner disappears and save re-enables automatically.
- New i18n key `graphPanel.propertiesView.editLockedByPipeline` added to all 11 locales (`ar/de/en/fr/ja/ko/ru/uk/vi/zh/zh_TW`).

### Backend (`fb5a3c56`)

- New helper [`check_pipeline_busy_or_raise`](lightrag/api/routers/document_routes.py) in `document_routes.py`: read-only check on `pipeline_status['busy']` under the namespace lock, raises `HTTPException(409)` on contention. Crucially it does **not** acquire `busy` or `destructive_busy` itself — holding a flag for every UI edit would serialise document ingestion behind interactive graph work. The narrow race between the check and the underlying storage write is intentionally tolerated; the storage layer's per-edge/-node locks are the last line of defense.
- The guard is invoked at the entry of all 7 graph-mutation endpoints:
  - `POST /graph/entity/edit`, `POST /graph/relation/edit`
  - `POST /graph/entity/create`, `POST /graph/relation/create`
  - `POST /graph/entities/merge`
  - `DELETE /documents/delete_entity`, `DELETE /documents/delete_relation`
- Each `graph_routes.py` endpoint also gains `except HTTPException: raise` ahead of `except Exception`, aligning with `delete_entity`'s existing pattern so the 409 isn't swallowed and re-thrown as a 500.
- `PipelineNotInitializedError` is caught silently, matching `_acquire_destructive_busy` so test rigs without a bootstrapped shared-storage Manager keep working.

### Tests

- New [`tests/api/routes/test_graph_routes_pipeline_busy.py`](tests/api/routes/test_graph_routes_pipeline_busy.py):
  - 7 parametrised integration tests verify every endpoint returns 409 with `"Pipeline is busy"` in `detail` and that the underlying `aedit_*`/`acreate_*`/`amerge_*`/`adelete_*` is never awaited;
  - 1 idle pass-through test confirms the request reaches `rag.aedit_entity` when the guard permits;
  - 3 unit tests for the helper itself: raises 409 when `busy=True`, returns silently when `busy=False`, and stays no-op when `pipeline_status` was never bootstrapped.
- Marked `pytest.mark.offline`, no live services required.

## What this does not cover

- A user keeping an old edit dialog open and clicking save the moment busy flips on is still racy at the UI layer (the dialog is already open so the click handler executes); the backend 409 catches it. Likewise a programmatic client bypassing the WebUI is caught by the backend guard.
- Concurrency between two ingestion-pipeline writers (already serialised by `_PipelineMixin`) is unaffected.

## Test plan

- [x] `cd lightrag_webui && bun run lint && bun run build`
- [x] `ruff check lightrag/api/routers/graph_routes.py lightrag/api/routers/document_routes.py tests/api/routes/test_graph_routes_pipeline_busy.py`
- [x] `./scripts/test.sh tests/api/` — 138 passed, 14 skipped (integration/online), 0 failed
- [ ] Manual: trigger ingestion, open a node, confirm pencil icons disappear and a 🔒 indicator with localized tooltip appears in the header
- [ ] Manual: open an edit dialog with a long description draft, trigger ingestion mid-edit, confirm dialog stays open, draft survives, save button disables and re-enables when idle
- [ ] Manual: while pipeline is busy, `curl -X POST .../graph/entity/edit` returns 409 with the expected detail

### DIFF
--- a/lightrag/api/routers/document_routes.py
+++ b/lightrag/api/routers/document_routes.py
@@ -1063,6 +1063,54 @@ async def _reserve_enqueue_slot(rag: LightRAG) -> bool:
     return True
 
 
+async def check_pipeline_busy_or_raise(rag: LightRAG) -> None:
+    """Refuse the request with HTTP 409 when the document pipeline is busy.
+
+    Intended for short, fine-grained graph mutations (entity/relation
+    edit/create/delete/merge). Reads ``pipeline_status['busy']`` under
+    the namespace lock and raises immediately on contention -- it does
+    NOT set any flag, so it cannot block the pipeline itself.
+
+    ``busy`` is set by the processing loop and by destructive jobs
+    (``/documents/clear`` / per-doc delete). Both paths concurrently
+    write the same graph storages that these endpoints mutate, so a
+    409 here mirrors the existing UI guard and tells clients to wait.
+
+    A narrow race remains between this check and the underlying graph
+    write: if the pipeline transitions to busy in that window, the
+    per-edge/-node locks inside the storage layer are the last line of
+    defense. That trade-off is deliberate -- holding ``busy`` here
+    would serialise every UI edit against document ingestion, which is
+    a worse user-visible failure mode than tolerating the race.
+
+    No-op (returns silently) when ``pipeline_status`` was never
+    bootstrapped, matching the behaviour of ``_acquire_destructive_busy``
+    so test rigs without a real shared-storage Manager keep working.
+    """
+    from lightrag.exceptions import PipelineNotInitializedError
+    from lightrag.kg.shared_storage import get_namespace_data, get_namespace_lock
+
+    try:
+        pipeline_status = await get_namespace_data(
+            "pipeline_status", workspace=rag.workspace
+        )
+    except PipelineNotInitializedError:
+        return
+    pipeline_status_lock = get_namespace_lock(
+        "pipeline_status", workspace=rag.workspace
+    )
+    async with pipeline_status_lock:
+        if pipeline_status.get("busy"):
+            raise HTTPException(
+                status_code=409,
+                detail=(
+                    "Pipeline is busy with another operation. "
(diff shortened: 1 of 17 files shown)


danielaskdd:
@codex review

PR #3142, opened 2026-05-27
feat(api): add chunking strategy selection to /documents/text(s)

danielaskdd:
## Summary

Adds a stable `chunking` request object to `POST /documents/text` and `POST /documents/texts` so API clients can choose a chunking strategy and pass strategy-specific parameters, with **fully synchronous validation** (HTTP 422 before any background work). Fully backward compatible — omitting `chunking` preserves today's default fixed-token behavior.

Along the way it makes the fixed-token (F) strategy symmetric with R/V/P for `chunk_token_size` (including a new `CHUNK_F_SIZE` env), closing a latent gap where a per-strategy F size could not take effect (and would `TypeError` if forced into the sub-dict).

## Request shape

```json
{
  "text": "content...",
  "file_source": "manual.md",
  "chunking": {
    "strategy": "fixed_token",
    "params": { "chunk_token_size": 1200, "chunk_overlap_token_size": 100,
                "split_by_character": "\n\n", "split_by_character_only": true }
  }
}
```

`strategy` ∈ `fixed_token | recursive_character | semantic_vector | paragraph_semantic`. All strategy-specific options (including `chunk_token_size`) live under `params`; no new top-level flat fields are added.

## What changed

**API layer** (`document_routes.py`)
- Per-strategy **typed params models** (Pydantic v2, `strict=True` + `extra="forbid"`). This rejects, synchronously at request-parse time → **HTTP 422**: unknown keys, wrong types (`true`/`"5"`/`1.5` for ints, `123` for str, `"abc"` for `list[str]`), bad enum (`breakpoint_threshold_type`), out-of-range values (`buffer_size`, `breakpoint_threshold_amount`). Without this, those slip past parsing and explode later in the background chunker.
- `_resolve_text_chunking` freezes the request into `(process_options, chunk_options)` and additionally validates the **effective** `chunk_overlap_token_size < chunk_token_size` against the resolved snapshot (overlap is usually inherited from `addon_params`/env, so the request model can't see it). The route handlers call it **synchronously before scheduling** the background task, so this also fails with 422 — not a silent 200-then-background-failure.

**Pipeline** (`pipeline.py`)
- The explicit-F branch and the legacy `chunking_func` branch now read `fixed_token.chunk_token_size` first (falling back to the top-level size), so F is symmetric with R/V/P. `LightRAG.ainsert` (which intentionally omits `process_options` to keep a user's custom `chunking_func`) now honors `CHUNK_F_SIZE` too. No behavior change when the override is unset.

**Config** (`routing.py`, `env.example`)
- New `CHUNK_F_SIZE` env seeds `fixed_token.chunk_token_size`, symmetric with `CHUNK_R_SIZE` / `CHUNK_V_SIZE`. Same "slot-absent → inherit top-level" convention.

**Docs**
- Document the `chunking` request and `CHUNK_F_SIZE`; clarify `CHUNK_SIZE` / `CHUNK_OVERLAP_SIZE` are the **global** fallback shared by all strategies (not F-specific); add `fixed_token.chunk_token_size` to the chunker shape; note that `LightRAG.ainsert` is fixed-token-only and the REST API ingests via the pipeline enqueue path with per-document `process_options`.

## Backward compatibility

- `chunking` is optional; omitting it reproduces current behavior exactly (default F + `addon_params['chunker']`).
- No request fields removed or renamed; no migration required; existing payloads remain valid.
- F-size resolution change is a no-op unless `CHUNK_F_SIZE` / an explicit `fixed_token.chunk_token_size` is set.

## Test plan

- `TextChunkingConfig` validation matrix (type/value/cross-field/unknown-key → 422).
- `_resolve_text_chunking`: strategy → `process_options` mapping, size routed into the strategy sub-dict for all four strategies, params merge, `CHUNK_R_SIZE` override, and **effective overlap < size** rejection (inherited-overlap case).
- Route-level: `request.chunking` forwarded to `pipeline_index_texts` (and `None` when omitted); malformed `chunking` and size-below-inherited-overlap both return **422 before any background task is scheduled** (spy never fires).
- Pipeline F-size regressions: explicit-F honors caller `fixed_token.chunk_token_size`; `CHUNK_F_SIZE` env seeds it and beats the legacy ctor field; `ainsert` legacy path honors `CHUNK_F_SIZE` while still invoking the user's `chunking_func`.

`tests/chunker tests/pipeline tests/api/routes` → **409 passed, 4 skipped** (skips are optional `langchain` deps); `ruff` clean.


danielaskdd:
@codex review

danielaskdd:
Fixed in 87b7d2c4.

`normalize_addon_params` only runs `default_chunker_config()` when the whole `chunker` key is absent, so a partial `addon_params['chunker']` skipped the strategy size-env seeding, and `_apply_chunk_size_overlay` only backfilled the top-level size + overlaps + P size — dropping `CHUNK_F_SIZE` (and, with the same latent gap, `CHUNK_R_SIZE` / `CHUNK_V_SIZE`) for partial configs.

The overlay now mirrors the env backfill for F/R/V: it fills `<strategy>.chunk_token_size` from `CHUNK_F_SIZE` / `CHUNK_R_SIZE` / `CHUNK_V_SIZE` **only when the env is set and the slot is unset** — explicit `addon_params` still wins (tier 1 > tier 2), and absent + no env stays absent so the strategy inherits the top-level size at consumption time (no behavior change for that case).

Added regression tests for: partial-config picks up the size env (F/R/V), explicit value beats env, and no-env leaves the slot absent. Full `tests/chunker tests/pipeline tests/api/routes` → 412 passed, ruff clean.

danielaskdd:
@codex review

danielaskdd:
Fixed in 3e288afb.

Same two-layer shape as the overlap fix: the `(0, 100]` ceiling is percentile/gradient-specific, so it can't be decided at request-parse time when the type is inherited.

- **Model layer** (`SemanticVectorChunkParams`): keeps the type-independent `amount > 0` check, but applies the `(0, 100]` ceiling **only when `breakpoint_threshold_type` is supplied in the same request**. When the type is omitted it no longer assumes `percentile`.
- **Effective layer** (`_resolve_text_chunking` → `_validate_effective_semantic_amount`): applies the ceiling for `percentile`/`gradient` against the **merged** type (request param over `addon_params`/`CHUNK_V_BREAKPOINT_THRESHOLD_TYPE`), run synchronously by the handlers.

Result: a request overriding only `breakpoint_threshold_amount` (e.g. `150`) is accepted when the inherited type is `standard_deviation`/`interquartile`, and still rejected (synchronous 422) when the effective type is `percentile`/`gradient`.

Tests added: model defers the cap when type omitted; explicit `percentile` + >100 still 422 at parse; `_resolve_text_chunking` allows >100 with inherited std and rejects with inherited percentile; route accepts the partial std override (200) and rejects the percentile case (422, no enqueue). Full `tests/chunker tests/pipeline tests/api/routes` → 417 passed, ruff clean.

danielaskdd:
@codex review

danielaskdd:
Fixed in 62884199.

Every param field is `Optional[...] = None` where None already means "inherit the addon_params/env default", so an explicit `null` is semantically the same as omitting the field. The normalization in `TextChunkingConfig._validate_params` now uses:

```python
self.params = typed.model_dump(exclude_unset=True, exclude_none=True)
```

So an explicit `"chunk_token_size": null` (or any nullable field) is dropped rather than merged over the resolved default — no more `int(None)` failing the document in the background, and a real sibling value in the same params is preserved.

Tests added: model drops an explicit null (keeps a real sibling value); `_resolve_text_chunking` keeps the inherited size when the request sends null; route returns 200 with the null filtered from the forwarded params. Full `tests/chunker tests/pipeline tests/api/routes` → 421 passed, ruff clean.

danielaskdd:
@codex review

danielaskdd:
Fixed in cfd3cafa.

`sentence_split_regex` is now compiled in the request model via a field validator on `SemanticVectorChunkParams`:

```python
@field_validator("sentence_split_regex")
@classmethod
def _valid_sentence_split_regex(cls, v):
    if v is None:
        return v
    try:
        re.compile(v)
    except re.error as exc:
        raise ValueError(f"sentence_split_regex is not a valid regular expression: {exc}") from exc
    return v
```

Regex validity is a self-contained per-value check (no inherited/effective context needed), so it belongs at the parse layer — a malformed pattern like `"("` now returns 422 synchronously instead of 200-then-background-failure. `None` (inherit) still passes.

Tests added: malformed regex rejected at the model and via the route (422, no enqueue); a valid regex passes. Full `tests/chunker tests/pipeline tests/api/routes` → 424 passed, ruff clean.

danielaskdd:
@codex review

danielaskdd:
Fixed in 20ba8a89.

Confirmed against `chunking_by_token_size`: `chunk_overlap_token_size` is consumed only in the normal token-window loop and the oversize-segment re-split loop. In delimiter-only mode (`split_by_character` set **and** `split_by_character_only=True`) it validates each segment against `chunk_token_size` and raises on an oversize one — overlap is never touched.

`_validate_effective_chunk_overlap` now short-circuits for `fixed_token` when the **effective** `split_by_character` is truthy **and** `split_by_character_only` is True (read from the merged snapshot, so an inherited delimiter counts). Both conditions are required because `split_by_character_only` is a no-op without a delimiter — the chunker then falls back to token windowing, which does use overlap, so the check must still fire there.

Tests added: delimiter-only with a small `chunk_token_size` passes (`_resolve_text_chunking` + route 200); the only-flag without a delimiter still rejects overlap >= size. Full `tests/chunker tests/pipeline tests/api/routes` → 427 passed, ruff clean.

danielaskdd:
@codex review

danielaskdd:
Verified on pydantic 2.13.4 — this is already the behavior, so no functional change was needed (83323171 adds a clarifying comment + regression tests to lock it).

A strict `float` field accepts an `int` and widens it losslessly to `float`; strict only rejects `str`/`bool`. Reproduced across all three paths:

```
SemanticVectorChunkParams (python):  int 95 -> 95.0 OK | "95" -> REJ(float_type) | True -> REJ(float_type)
SemanticVectorChunkParams (json):    int 95 -> 95.0 OK | true -> REJ(float_type)
TextChunkingConfig (json):           {"breakpoint_threshold_amount": 95} -> {"breakpoint_threshold_amount": 95.0}
```

The existing route test `test_insert_text_allows_amount_override_inheriting_std_type` already sends integer `150` through the full FastAPI/JSON path and returns 200 with `params == {"breakpoint_threshold_amount": 150.0}`.

I avoided the two suggested alternatives because both would regress: relaxing strict (`strict=False`) would let numeric **strings** through, and `int | float` would stop normalizing ints to float. The strict-`float` field already does exactly "accept ints, reject strings/bools."

Added tests: int widened to float (python + JSON modes); `"95"` and `true` rejected. Full `tests/chunker tests/pipeline tests/api/routes` → 430 passed, ruff clean. If you're seeing a 422 for integer amounts on a different pydantic version, let me know the version and I'll dig in.

danielaskdd:
@codex review

PR #3147, opened 2026-05-28
perf(faiss): defer vector embedding to flush time

danielaskdd:
## Summary

`FaissVectorDBStorage.upsert()` embedded **eagerly** on every call, so repeated upserts of the same id — and many small upsert calls — re-triggered the embedding model. This applies the deferred-embedding pattern already proven for `NanoVectorDBStorage` (PR #3145) to the Faiss backend: buffer upserts and embed **once per id at flush time**.

Closes the Faiss-backend case of the repeated-embedding problem in #2785.

## Motivation

- Same as #2785 / #3145: per-call embedding wastes model calls when the same id is upserted multiple times or content arrives in many small batches.
- `NanoVectorDBStorage` already solved this in this codebase. Faiss is the other in-process / file-backed vector backend, so the same pattern fits cleanly. The Nano protocol is the authoritative reference; this PR is a careful 1:1 port plus three Faiss-specific cases the Nano playbook didn't need to cover.

## What's changed (`lightrag/kg/faiss_impl.py`)

- New `_PendingFaissDoc` dataclass + `self._pending_upserts: dict[str, _PendingFaissDoc]` buffer + `self._index_dirty` flag.
- `upsert()` now buffers `_PendingFaissDoc(vector=None)` under the lock; **no embedding inline**. Re-upserting the same id overwrites the prior pending doc (clearing any temp vector a previous `get_vectors_by_ids` had cached).
- New `_flush_pending_locked()` (caller must hold the **non-reentrant** `_storage_lock`):
  - Batch-embeds all `vector is None` pending docs.
  - One in-place `faiss.normalize_L2(arr)` on the whole batch (Faiss requires 2D); the per-row normalized vector is cached back to `_PendingFaissDoc.vector` (invariant: `pdoc.vector` is always **already L2-normalized float32**).
  - One `_remove_faiss_ids_locked` rebuild that strips every existing fid for any re-upserted custom id (uses the new plural helper — see below), then one `index.add(matrix)` for the whole pending batch.
  - Raises on embedding / count-mismatch failure; pending stays intact, `_index_dirty` is not touched.
- `index_done_callback()` rewritten: `reload-if-stale → flush → save → notify`. Flush / save **both raise** on failure (no silent log + return False). If only the save fails, `_pending_upserts` is empty (flush succeeded) and `_index_dirty=True` so a later `finalize` retries the save without re-embedding.
- New `finalize()` safety net: flushes any pending and saves. Skips the reload when `_index_dirty=True` so materialized-but-unsaved rows aren't dropped.
- Read-your-writes for `get_by_id` / `get_by_ids` / `get_vectors_by_ids`: consult `_pending_upserts` first; fall back through `_get_index()` (reload-if-stale) + `_id_to_meta`. `get_vectors_by_ids` lazy-embeds pending docs (same batch-normalize, cache-back invariant as flush). The pre-existing "Caveat — methods read `_id_to_meta` without `_get_index`" is now eliminated for async reads (still documented for the synchronous `client_storage` property).
- `delete` / `delete_entity` / `delete_entity_relation` / `drop` cancel matching pending entries under the lock, reload-if-stale, then mutate. **`delete` / `delete_entity_relation` propagate exceptions** rather than swallowing them — this intentionally diverges from Nano's swallow-and-log so document deletion / status-update flows abort if vectors weren't actually removed.
- `_remove_faiss_ids` → synchronous `_remove_faiss_ids_locked` (caller holds the lock). Single rebuild API used by both delete paths and flush.
- New plural helper `_find_faiss_ids_by_custom_id(custom_id) -> list[int]`. The singular variant stays for read paths; the plural variant is used in `_flush_pending_locked` and `delete` to defend against legacy / externally corrupted stores where multiple fids share a `__id__`.
- `_load_faiss_index` now logs a warning on `index > meta` cross-file skew (a crash between the two atomic_writes). `query()` skips orphan faiss hits so callers never see `{"id": None, ...}`. **Auto-repair of the on-disk skew is out of scope and left to a follow-up PR** (semantics decision: truncate index vs rebuild meta).

## Behavior notes

- **`query()` and `client_storage` see only flushed (materialized) data** — unflushed pending upserts are intentionally not searchable / visible. Same contract as Nano; matches the OpenSearch implementation.
- A flush failure now **raises** through `_insert_done` instead of silently logging — same lift in failure visibility as PR #3145 made for Nano.
- The `_PendingFaissDoc.vector` invariant (always normalized after the first time it is set) ensures the flush never re-normalizes a vector that `get_vectors_by_ids` has already cached.

## Test plan

- [x] New `tests/kg/faiss_impl/test_faiss_deferred_embedding.py`:
  - **14 tests** ported from the Nano deferred-embedding suite: no embed on upsert; embed on `index_done_callback`; same-id coalesced; `get_vectors_by_ids` lazy embed + flush reuse; re-upsert clears cached vector; delete cancels pending + removes materialized; stale-client reload paths for `upsert` / `delete` / `finalize`; read-your-writes + query-after-flush; `delete_entity_relation` cancels pending; embed-failure raises + preserves pending; `drop` discards pending; `finalize` retries save after flush failure.
  - **6 Faiss-specific tests**: `reupsert_after_flush_replaces_single_fid`, `index_done_callback_save_failure_raises`, `reload_warns_on_index_meta_skew`, `query_skips_orphan_faiss_hits`, `reupsert_cleans_duplicate_custom_id_rows`, `delete_propagates_errors`.
  - Every state-mutating test asserts `_index.ntotal == len(_id_to_meta)` via the local `_assert_consistent` helper — Faiss's root failure mode is the two structures diverging, so single-side assertions aren't enough.
- [x] `./scripts/test.sh tests/kg/faiss_impl` → **26 passed** (20 new + 6 existing `test_atomic_write_faiss` / `test_faiss_meta_inconsistency`).
- [x] `./scripts/test.sh tests/kg/nano_impl` → 17 passed (no cross-backend regression).
- [x] `ruff check lightrag/kg/faiss_impl.py tests/kg/faiss_impl/test_faiss_deferred_embedding.py` clean.

---

### DIFF
--- a/lightrag/kg/faiss_impl.py
+++ b/lightrag/kg/faiss_impl.py
@@ -21,6 +21,25 @@ from .shared_storage import (
 import faiss  # type: ignore
 
 
+@dataclass
+class _PendingFaissDoc:
+    """A buffered upsert waiting for deferred embedding and materialization.
+
+    ``record`` holds ``__id__`` / ``__created_at__`` plus the ``meta_fields``
+    (which always include ``content`` for the entity/relation/chunk vdbs), so
+    the content needed for deferred embedding lives in the record itself — no
+    separate copy is kept. ``vector`` starts as ``None`` and is filled either
+    during the lock-held flush or by a lazy ``get_vectors_by_ids`` embedding;
+    once set it is always an **already-L2-normalized float32 1D ndarray**, so
+    the next flush can ``vstack`` and ``index.add`` without re-normalizing.
+    ``__vector__`` is materialized into the metadata dict only at flush time,
+    right before ``self._index.add``.
+    """
+
+    record: dict[str, Any]
+    vector: np.ndarray | None = None
+
+
 @final
 @dataclass
 class FaissVectorDBStorage(BaseVectorStorage):
@@ -48,8 +67,13 @@ class FaissVectorDBStorage(BaseVectorStorage):
         **Cross-file atomicity is not guaranteed**: the two ``atomic_write``
         renames in ``_save_faiss_index`` are independent, so a crash
         between them can leave ``.index`` and ``.meta.json`` referring to
-        different snapshots. ``_load_faiss_index`` tolerates this by
-        skipping metadata rows whose ``fid`` exceeds ``self._index.ntotal``.
+        different snapshots. ``_load_faiss_index`` tolerates both
+        directions on load: ``meta > index`` rows are dropped silently;
+        ``index > meta`` (the more dangerous case) is logged as a warning
+        but **not** auto-repaired — orphan vectors remain in the loaded
+        index but are unreachable via custom-id lookups. Repair semantics
+        (truncate index vs rebuild meta) are deliberately left to a
+        follow-up PR.
 
     Concurrency invariants (the code here is correct *only* while all
     three hold):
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3152, opened 2026-05-28
refact(postgres): defer PGVector embedding to index_done_callback

danielaskdd:
## Summary

`PGVectorStorage` now follows the same deferred-embedding contract as `OpenSearchVectorDBStorage`, `NanoVectorDBStorage`, and `FaissVectorDBStorage`: `upsert()` buffers raw documents process-locally, and embedding + persistence runs in one PG transaction during `index_done_callback()` / `finalize()`.

## Motivation

PG was the last vector backend still embedding eagerly inside `upsert()`. That blocked:

- Deduplication of repeated upserts of the same id within a working batch.
- Coalescing many small `upsert()` calls into one larger embedding batch.
- A uniform lifecycle contract across all vector backends.

### Production (`lightrag/kg/postgres_impl.py`)

- New module-level dataclass `_PendingPGVectorDoc(item, created_at, vector=None)`.
- `__post_init__` initializes `_pending_vector_docs`, `_pending_vector_deletes`, `_flush_lock`.
- `initialize()` creates the namespace lock via `get_namespace_lock(self.namespace, workspace=self.workspace)` after workspace resolution.
- `upsert()` buffers only — no embedding, no DB call. Same-id upsert overwrites the prior pending doc (and discards any cached stale vector). A pending upsert wins over an earlier pending delete for the same id.
- New `_flush_pending_vector_ops()`:
  - Runs entirely under `_flush_lock`.
  - Embeds pending docs whose vector is `None`, batched by `embedding_batch_num`, with explicit count-mismatch check.
  - Builds tuples via the existing `_upsert_chunks` / `_upsert_entities` / `_upsert_relationships` helpers (no signature change). Converts cached `list[float]` to `np.float32` array right before `executemany` for the pgvector codec.
  - Persists upserts + deletes in **one asyncpg transaction** wrapped in `_run_with_retry`. **All-or-nothing**: any failure leaves buffers and cached vectors intact for the next flush.
- `index_done_callback()` flushes.
- `finalize()` loud-failure contract: captures `Exception` from flush, releases the client in `finally`, then raises `RuntimeError` naming pending upsert/delete counts. `BaseException` (cancellation/shutdown) still propagates.
- `delete(ids)` buffers and cancels matching pending upserts.
- `delete_entity(name)` runs the SQL `WHERE entity_name=$2` predicate immediately under the lock (kept as a safety net for legacy rows whose `id` may not equal `compute_mdhash_id(name, "ent-")`), and also prunes matching pending docs.
- `delete_entity_relation(name)` prunes pending relationship docs by `src_id`/`tgt_id` and runs the server-side predicate delete under the lock.
- `get_by_id` / `get_by_ids` / `get_vectors_by_ids` serve pending state under the lock and release the lock before SQL fallback. SQL fallback now uses parameterized `id = ANY($2)` instead of the previous string-built `IN (...)`. `get_vectors_by_ids` lazily embeds buffered docs and caches the vector on the pending record so the next flush reuses it.
- `drop()` clears both buffers and runs the existing workspace-scoped table delete under the lock.

### What's Broken / Behaviour Changes

- `upsert()` is no longer durable on return. Callers depending on cross-worker visibility must call `index_done_callback()` (the LightRAG pipeline already does). Documented as the "correctness premise" in `upsert()`'s docstring.
- Flush is all-or-nothing per transaction: a single bad row in `executemany` rolls back the whole batch, including the same-flush deletes. The buffer remains intact for the next attempt. This trade-off was deliberately chosen over per-doc partial-failure handling (which PG cannot expose like OpenSearch's bulk API).
- `finalize()` now raises `RuntimeError` instead of silently dropping unflushed writes.

### Tests

- New `tests/kg/postgres_impl/test_postgres_vector_deferred.py` (18 tests) covering: buffer-only `upsert()`, deferred batching across many small calls, same-id overwrite, lazy vector cache + flush reuse, upsert after lazy cache discards the cached vector, embedding failure leaves buffer intact, `_run_with_retry` failure preserves buffer and cached vector, delete cancels pending upsert and vice versa, `delete_entity` prunes pending + runs SQL, `delete_entity_relation` predicate, `drop` clears buffers, read-your-writes for all three read methods, `id = ANY($2)` SQL fallback shape, and `finalize()` raise-with-counts on flush failure.
- `tests/kg/postgres_impl/test_postgres_upsert.py` updated to seed `_flush_lock` and to assert the new `flush[...]` timing label after `index_done_callback()`.

### How It Works

1. `await rag.ainsert(...)` → pipeline → `PGVectorStorage.upsert()` → buffer.
2. (...repeat with many small upserts; same-id deduped, deletes cancel upserts, etc.)
3. `await rag._insert_done()` → `index_done_callback()` → `_flush_pending_vector_ops()` →
   1. embed all `vector is None` pending docs in `embedding_batch_num` batches via `asyncio.gather`.
   2. `executemany` upserts + `execute` delete inside one `connection.transaction()`.
   3. clear buffers on success; keep them on any failure.
4. `await rag.finalize_storages()` → `PGVectorStorage.finalize()` flushes, releases the client, raises if anything is still pending.

## Test plan

- [x] `ruff check lightrag/kg/postgres_impl.py tests/kg/postgres_impl/` — clean.
- [x] `./scripts/test.sh tests/kg/postgres_impl/` — all passed.
- [x] `./scripts/test.sh tests/kg/opensearch_impl/` — regression sanity, all passed.
- [x] `./scripts/test.sh tests/kg/postgres_impl/ tests/kg/opensearch_impl/ tests/kg/nano_impl/ tests/kg/faiss_impl/` — 352 passed.
- [ ] Live PG harness with pgvector enabled (not available in this environment):
  - Insert a small corpus end-to-end and confirm `index_done_callback()` writes all chunks/entities/relations.
  - Run a query and compare results with the pre-refactor baseline.
  - Exercise `delete_entity()` and `delete_entity_relation()` followed by `index_done_callback()`.
  - Confirm vectors round-trip through the pgvector asyncpg codec (already converted to `np.float32` ndarray immediately before `executemany`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01T115bMNJLmh3v5jpfTt4kS)_

### DIFF
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -3796,9 +3796,7 @@ class PGVectorStorage(BaseVectorStorage):
             if self.db is None:
                 return
 
-            timing_label = (
-                f"{self.workspace} PGVectorStorage.flush[{self.namespace}]"
-            )
+            timing_label = f"{self.workspace} PGVectorStorage.flush[{self.namespace}]"
             total_start = time.perf_counter()
             performance_timing_log(
                 "[%s] start upserts=%s deletes=%s max_batch_size=%s",
@@ -3816,8 +3814,7 @@ class PGVectorStorage(BaseVectorStorage):
             ]
             if docs_to_embed:
                 contents = [
-                    pending_doc.item["content"]
-                    for _, pending_doc in docs_to_embed
+                    pending_doc.item["content"] for _, pending_doc in docs_to_embed
                 ]
                 batches = [
                     contents[i : i + self._max_batch_size]
@@ -3856,9 +3853,7 @@ class PGVectorStorage(BaseVectorStorage):
                 build_tuple = self._upsert_chunks
             elif is_namespace(self.namespace, NameSpace.VECTOR_STORE_ENTITIES):
                 build_tuple = self._upsert_entities
-            elif is_namespace(
-                self.namespace, NameSpace.VECTOR_STORE_RELATIONSHIPS
-            ):
+            elif is_namespace(self.namespace, NameSpace.VECTOR_STORE_RELATIONSHIPS):
                 build_tuple = self._upsert_relationships
             else:
                 raise ValueError(f"{self.namespace} is not supported")
@@ -3911,9 +3906,7 @@ class PGVectorStorage(BaseVectorStorage):
                         )
 
             try:
-                await self.db._run_with_retry(
-                    _flush_batch, timing_label=timing_label
-                )
+                await self.db._run_with_retry(_flush_batch, timing_label=timing_label)
             except Exception as e:
                 logger.warning(
                     f"[{self.workspace}] PGVectorStorage flush failed; "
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

PR #3154, opened 2026-05-28
🐛 fix(kg): post-prune pending buffer in vector delete_entity_relation

danielaskdd:
## Summary

Align all vector storage backends with the post-prune contract already used by `PGVectorStorage.delete_entity_relation`. Fixes a silent data-loss path in `Qdrant`, `Mongo`, `Milvus`, `OpenSearch`, `Faiss`, and `Nano`, plus a compound silent-failure bug in `OpenSearch` and `Nano`.

## Motivation

While reviewing `PGVectorStorage` and `QdrantVectorDBStorage` for equivalence, we found the two backends disagreed on the ordering between buffer pruning and the destructive server-side / materialized delete in `delete_entity_relation`:

- **pre-prune (Qdrant, Mongo, Milvus, OpenSearch, Faiss, Nano):** matching pending upserts are removed from `_pending_vector_docs` *before* the destructive call. If the destructive call raises, the buffered upserts are gone forever even though the higher-level operation (e.g. `adelete_by_entity` in `utils_graph.py`) catches the exception, short-circuits, and leaves both the graph and the persisted vector rows untouched. The user is told the delete failed, but writes they never asked to discard are silently lost.
- **post-prune (Postgres):** the destructive call runs first; pending upserts are only popped after success. On failure the buffer survives, the exception re-raises, the caller short-circuits past `_persist_graph_updates`, and the next retry can converge.

`OpenSearch` and `Nano` made this worse by swallowing the destructive failure into `logger.error` instead of re-raising. That defeats the caller short-circuit and lets `adelete_by_entity` continue through `delete_node` + `_persist_graph_updates`, so the graph node and the pending buffer get cleaned up while the persisted relation rows remain — permanent graph / vector-store inconsistency, reported back as `status="success"`.

The single in-tree caller, `adelete_by_entity` in [lightrag/utils_graph.py](../lightrag/utils_graph.py), already short-circuits correctly when `delete_entity_relation` raises (its `except` skips both `delete_node` and `_persist_graph_updates`), so the post-prune contract is the strictly safer pattern.

## What's Changed

Unified contract now followed by all 7 vector backends (`PG`, `Qdrant`, `Mongo`, `Milvus`, `OpenSearch`, `Faiss`, `Nano`):

1. Destructive side runs first — server-side `delete_by_query` / `delete_many` / `client.delete` / materialized index rebuild.
2. Matching pending upserts in `_pending_vector_docs` (or `_pending_upserts`) are pruned **only after** that succeeds.
3. Failures are re-raised so `adelete_by_entity`'s `except` clause skips `_persist_graph_updates` and the buffer state stays consistent with the unchanged graph.
4. `db is None` / `client is None` / `_index_ready is False` early-return paths still prune the buffer to record the delete intent (post-finalize / pre-init handling).
5. Empty-result branches also prune so a buffered upsert cannot re-create a just-deleted relation on the next flush.

Per-file changes:

| File | Method | Change |
|---|---|---|
| `lightrag/kg/qdrant_impl.py` | `delete_entity_relation` | post-prune (was pre-prune); pre-init / post-finalize path still prunes |
| `lightrag/kg/mongo_impl.py` | `delete_entity_relation` | post-prune; empty-result branch also prunes |
| `lightrag/kg/milvus_impl.py` | `delete_entity_relation` | post-prune; empty-result branch also prunes |
| `lightrag/kg/opensearch_impl.py` | `delete_entity_relation` | post-prune; non-missing-index `OpenSearchException` now re-raises (was logged and swallowed); missing-index treated as "already deleted" and still prunes |
| `lightrag/kg/faiss_impl.py` | `delete_entity_relation` | post-prune (materialized rebuild first) |
| `lightrag/kg/nano_vector_db_impl.py` | `delete_entity_relation`, `delete_entity` | post-prune; outer `except` re-raises (was logged and swallowed) |

PG was already post-prune (#3152) — no change needed in this PR.

## What's Broken and How It Works

- **Public API:** unchanged. `delete_entity_relation` and `delete_entity` keep the same signature and intent (`delete all relation/entity vectors for this name`).
- **Failure semantics, OpenSearch / Nano:** these now raise on destructive failure instead of silently logging. This is a behavior change for callers that depended on the old "always returns successfully" behavior. The only in-tree caller, `adelete_by_entity`, already wraps in `try / except Exception` and converts to `DeletionResult(status="fail")`, so the externally-visible behavior is "operations that previously reported success while leaving inconsistent state now correctly report failure".
- **Failure semantics, Mongo / Milvus / Qdrant / Faiss:** unchanged — these already re-raised destructive failures; only the prune ordering moved.
- **Empty-result handling:** Mongo and Milvus previously skipped pruning when the server-side query returned no rows. They now prune in this case too so a buffered upsert cannot re-create the relation. This is a behavior fix (previously a buffered relation upsert against an entity with no persisted relations would survive a `delete_entity_relation` call).
- **`db/client/index is None` paths:** continue to behave as buffer-only prune.
- **Pre-existing semantic note on Milvus** (deferred-buffer ↔ persisted divergence when an upsert is rewriting `src_id` / `tgt_id`) is preserved in the docstring.

## Test Plan

- [ ] `ruff format --check` clean on all 6 touched files
- [ ] `ruff check --ignore=E402` clean on all 6 touched files
- [ ] `ast.parse` succeeds on all 6 touched files
- [ ] Manual: with each vector backend, simulate a destructive failure during `delete_entity_relation` (force the server call to raise) and verify:
  - the exception propagates out of `delete_entity_relation`
  - `adelete_by_entity` returns `DeletionResult(status="fail")`
  - `_pending_vector_docs` still contains the matching pending upserts
  - retrying `adelete_by_entity` after the failure clears the persisted rows and prunes the buffer
- [ ] Manual: OpenSearch missing-index path still short-circuits (prunes buffer, marks index missing, returns) without raising
- [ ] Manual: empty-result branch on Mongo / Milvus now prunes a pre-buffered relation upsert against an entity with no persisted rows

### DIFF
--- a/lightrag/kg/faiss_impl.py
+++ b/lightrag/kg/faiss_impl.py
@@ -541,6 +541,14 @@ class FaissVectorDBStorage(BaseVectorStorage):
 
         Errors propagate (same rationale as ``delete``).
 
+        Buffer semantics — post-prune with caller short-circuit contract:
+            The materialized index rebuild runs first; matching pending
+            upserts are pruned **only after** it succeeds. If the
+            rebuild raises, the pending buffer stays intact so the
+            caller (``adelete_by_entity`` in ``utils_graph.py``) can
+            short-circuit before ``_persist_graph_updates`` flushes a
+            half-cleaned buffer.
+
         **Not pipeline-gated** — see class docstring
         *Non-pipeline write paths*. The caller is responsible for
         ensuring single-writer serialization.
@@ -548,19 +556,10 @@ class FaissVectorDBStorage(BaseVectorStorage):
         async with self._storage_lock:
             self._reload_index_from_disk_locked(for_write=True)
 
-            # Prune matching buffered upserts (their records carry src_id /
-            # tgt_id from the relationships vdb meta_fields)...
-            pending_ids = [
-                doc_id
-                for doc_id, pdoc in self._pending_upserts.items()
-                if pdoc.record.get("src_id") == entity_name
-                or pdoc.record.get("tgt_id") == entity_name
-            ]
-            for doc_id in pending_ids:
-                del self._pending_upserts[doc_id]
-
-            # ...then scan the materialized rows. .get() so rows from
-            # foreign namespaces (no src_id / tgt_id) silently don't match.
+            # Materialized side first so a failure leaves the pending
+            # buffer intact for the caller's retry path. .get() so rows
+            # from foreign namespaces (no src_id / tgt_id) silently
+            # don't match.
             relations = [
                 fid
                 for fid, meta in self._id_to_meta.items()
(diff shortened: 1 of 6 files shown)


danielaskdd:
@codex review

PR #3164, opened 2026-05-30
🐛 fix(milvus): batch flush upserts/deletes under gRPC message limit

danielaskdd:
## Description

`MilvusVectorDBStorage._flush_pending_vector_ops` embedded pending vectors in batches but then merged **all** of them into a single `client.upsert()` call. Once enough vectors accumulated, that one gRPC request exceeded Milvus' server-side message limit (`grpc.serverMaxRecvSize`, default 64MB) and failed:

```
StatusCode.RESOURCE_EXHAUSTED
grpc: received message larger than max (71331392 vs. 67108864)
```

That ceiling lives on the Milvus server/proxy and cannot be raised from the client library, so the fix is to split the write **client-side** — mirroring the batching pattern the Qdrant backend already uses.

## Related Issues

n/a

## Changes Made

- **`lightrag/kg/milvus_impl.py`**
  - Add `_build_upsert_batches`: splits records by a JSON-estimated **payload byte budget** (primary limiter) and a **record-count cap** (secondary). JSON serialization overestimates the actual protobuf size (a JSON float string is far longer than the 4 protobuf bytes it encodes), so batches stay conservatively below the server limit and never underestimate.
  - A single record larger than the byte budget is emitted as its **own batch** with a warning, rather than raising — its real protobuf size is often still under 64MB, so the server is the final arbiter.
  - `upsert` and `delete` (by record count) are both batched. Buffers are cleared **only after every batch succeeds**, preserving the existing fail-fast / whole-buffer-retry contract of `index_done_callback`.
  - New env knobs (mirroring `QDRANT_UPSERT_MAX_*`): `MILVUS_UPSERT_MAX_PAYLOAD_BYTES` (default 32MB), `MILVUS_UPSERT_MAX_RECORDS_PER_BATCH` (1000), `MILVUS_DELETE_MAX_RECORDS_PER_BATCH` (1000); a non-positive value disables that splitting dimension.
- **`env.example` / `env.docker-compose-full`**: document the three new variables in the Milvus section.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

- Verified `_build_upsert_batches` against six cases (empty input; single batch; byte-budget split; oversized-record-gets-own-batch; record-count split; splitting disabled when limits ≤ 0) — all pass.
- Existing Milvus suite (`tests/kg/milvus_impl/`, incl. deferred-embedding flush path) passes: **85 passed**.
- To exercise the multi-batch path manually: `export MILVUS_UPSERT_MAX_PAYLOAD_BYTES=1048576` (1MB) and watch for the `upsert split into N batches` log line.
- No unit test added for the new batching helper in this PR; happy to add one under `tests/kg/milvus_impl/` if preferred.

### DIFF
--- a/env.docker-compose-full
+++ b/env.docker-compose-full
@@ -886,6 +886,14 @@ MINIO_SECRET_ACCESS_KEY=minioadmin
 ### DB specific workspace should not be set, keep for compatible only
 # MILVUS_WORKSPACE=forced_workspace_name
 
+### Milvus upsert/delete batching (enabled by default)
+### Split large flushes by estimated JSON payload size and record count to stay
+### under the server-side 64MB gRPC message limit. A single record larger than the
+### byte budget is sent as its own batch instead of failing.
+# MILVUS_UPSERT_MAX_PAYLOAD_BYTES=33554432
+# MILVUS_UPSERT_MAX_RECORDS_PER_BATCH=1000
+# MILVUS_DELETE_MAX_RECORDS_PER_BATCH=1000
+
 ### Milvus Vector Index Configuration
 ### Index type: AUTOINDEX (default), HNSW, HNSW_SQ, HNSW_PQ, IVF_FLAT, IVF_SQ8, DISKANN
 # MILVUS_INDEX_TYPE=AUTOINDEX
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

PR #3165, opened 2026-05-30
🐛 fix(qdrant): resilient flush — own-batch oversized points, batch deletes

danielaskdd:
## Description

While adding flush batching to the Milvus backend (#3164), the same review surfaced two robustness gaps in `QdrantVectorDBStorage._flush_pending_vector_ops`. This PR brings Qdrant's flush in line with the more resilient Milvus behavior.

1. **A single oversized point poisoned the entire flush.** `_build_upsert_batches` raised `ValueError` when one point exceeded the upsert payload budget. That call sits inside the flush's `try` block, so the error was re-raised and the buffer retained — meaning **every** subsequent `index_done_callback` retry hit the same point and failed again, permanently blocking all healthy points buffered alongside it from ever committing. The conservative default budget (16MB, deliberately below the ~32MB server/gateway limit) made this reachable even for points the server would have accepted.
2. **Deletes were not batched.** All delete ids were sent in a single `client.delete()` call, so a very large delete set could exceed the request limit — the same class of failure the upsert batching already guards against.

## Related Issues

#XXXX (follow-up to #3164)

## Changes Made

- **`lightrag/kg/qdrant_impl.py`**
  - `_build_upsert_batches`: a point larger than the byte budget is now emitted as its **own single-point batch** instead of raising. The JSON size estimate is conservative, so such a request may still succeed; the server is left as the final arbiter rather than failing the whole flush.
  - Flush emits a `warning` when a single point is sent over-budget on its own (mirroring the Milvus backend), so oversized records stay visible in logs.
  - Deletes are chunked by point count via the new `QDRANT_DELETE_MAX_POINTS_PER_BATCH` env var (default 1000; a non-positive value disables splitting), mirroring the existing `QDRANT_UPSERT_MAX_*` knobs.
  - Buffers are still cleared only after **every** batch succeeds — the fail-fast / whole-buffer-retry contract is unchanged.
- **`env.example` / `env.docker-compose-full`**: document `QDRANT_DELETE_MAX_POINTS_PER_BATCH` and the new own-batch behavior in the Qdrant section.
- **Tests**: replaced the unit test that asserted the old `raise` behavior with one asserting the oversized point gets its own batch; added coverage for isolating a mid-stream oversized point (neighbors not polluted) and for delete chunking.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

- `tests/kg/qdrant_impl/` passes: **32 passed**.
- Note on the Qdrant transport: the client uses REST/HTTP (JSON), so the JSON byte estimate closely matches the real request size. A point above the *server's actual* limit will still be rejected by the server — but (a) the conservative 16MB default leaves a band where the own-batch point now succeeds, and (b) more importantly, one oversized point no longer holds the entire buffer hostage.

### DIFF
--- a/env.docker-compose-full
+++ b/env.docker-compose-full
@@ -921,11 +921,14 @@ MINIO_SECRET_ACCESS_KEY=minioadmin
 QDRANT_URL=http://localhost:6333
 # QDRANT_DEVICE=cpu
 # QDRANT_API_KEY=your-api-key
-### Qdrant upsert batching (enabled by default)
-### Split large upserts by estimated JSON payload size and point count
-### Default 16MB keeps safe headroom below common 32MB gateway/request limits
+### Qdrant upsert/delete batching (enabled by default)
+### Split large upserts by estimated JSON payload size and point count, and
+### large deletes by point count, to stay under the server/gateway request limit.
+### Default 16MB keeps safe headroom below common 32MB gateway/request limits.
+### A single point larger than the byte budget is sent as its own batch instead of failing.
 # QDRANT_UPSERT_MAX_PAYLOAD_BYTES=16777216
 # QDRANT_UPSERT_MAX_POINTS_PER_BATCH=128
+# QDRANT_DELETE_MAX_POINTS_PER_BATCH=1000
 ### DB specific workspace should not be set, keep for compatible only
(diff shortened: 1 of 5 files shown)


danielaskdd:
@codex review

PR #3167, opened 2026-05-30
feat(mongo): batch bulk_write upserts/deletes across all MongoDB storage paths

danielaskdd:
## Description

A single MongoDB flush could previously accumulate thousands of vector/KV/graph documents and ship them as **one** `bulk_write` (or one `delete_many`). That keeps the entire op list — each carrying a full embedding or document body — in memory at once and offers no payload control, no observability, and no tuning knob. `MilvusVectorDBStorage` already solved the equivalent problem by splitting upserts by **payload-byte budget (primary) + record count (secondary)** and deletes by record count.

This PR brings that batching to **every MongoDB write path** (vector, KV, graph) behind shared helpers and shared, env-configurable limits, and aligns the vector-flush log output with the now-standardized vector-DB batch logging.

## Related Issues

N/A (maintenance / robustness improvement).

## Changes Made

**MongoDB batched writes (`lightrag/kg/mongo_impl.py`)**
- New module-level helpers shared by all upsert paths:
  - `_estimate_doc_bytes()` — compact-JSON byte estimate (deliberately overestimates real BSON, so splits stay conservatively under server limits). It is a splitting *heuristic* over the dominant payload field (`$set` body / `update_doc`), not the exact wire size; the under-count of the filter/`$setOnInsert`/`$or` wrapper is immaterial next to an embedding/document body and against the 48MB bulk-command limit.
  - `_chunk_by_budget()` — generic payload-size-primary / record-count-secondary splitter (non-positive limit disables that dimension; an oversized single record is emitted as its own batch).
  - `_resolve_upsert_batch_limits()` — reads `MONGO_UPSERT_MAX_*` env vars with module defaults + non-positive warnings, shared across all three storage classes.
  - `_run_batched_bulk_write()` — executes `UpdateOne` ops as size/count-bounded `bulk_write` batches (used by KV & graph).
- `MongoVectorDBStorage._flush_pending_vector_ops`: upserts split into batches; deletes switched to **chunked `delete_many({"_id": {"$in": …}})`** (1:1 equivalent of Milvus `client.delete(pks=…)`). Fail-fast: any batch failure raises and leaves both buffers intact for the next `index_done_callback` retry (idempotent).
- `MongoKVStorage.upsert` and `MongoGraphStorage.upsert_nodes_batch` / `upsert_edges_batch`: now batched via the shared helper.
  - **Graph** paths preserve their original `ordered` semantics (nodes/edges `ordered=True`, source-node placeholders `ordered=False`).
  - **KV** intentionally switches from the old single `bulk_write(...)` (which used pymongo's implicit `ordered=True` default) to `ordered=False`: every op targets a distinct flattened `_id`, so the writes are order-independent, and `ordered=False` lets the server apply idempotent upserts in parallel. This is a deliberate change, documented at the call site.

**Logging standardization (`milvus_impl.py`, `qdrant_impl.py`)**
- Vector-DB batch logs standardized to include `max_payload` / `batch` context; Qdrant wording aligned from "points" to "records".
- Milvus: the split-INFO log gains the `max_payload`/`batch` context suffix (net-new on this line; it uses Milvus's real attribute `_max_upsert_records_per_batch`). _No pre-existing bug is being fixed here — `main` never referenced an undefined attribute on this line; it simply lacked the context suffix._
- `MongoVectorDBStorage` flush logs aligned 1:1 with Milvus (with backend-appropriate field swaps: `_id` vs `id`, `MongoDB` vs `Milvus`).

**Config & tests**
- `env.example`: documented `MONGO_UPSERT_MAX_PAYLOAD_BYTES` (16 MiB), `MONGO_UPSERT_MAX_RECORDS_PER_BATCH` (128), `MONGO_DELETE_MAX_RECORDS_PER_BATCH` (1000).
- Updated graph/vector unit tests for the shared batching attributes.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

- Default limits chosen for MongoDB's ceilings: 16 MiB payload budget stays well under the 48 MB bulk-command message limit and bounds peak memory; the JSON estimate overestimates real BSON, so the split never underestimates.
- `pymongo` already auto-splits `bulk_write` at the wire level, so the practical wins here are **bounded peak memory, observable/tunable batches, and consistent behavior** across all MongoDB write paths — not avoiding a hard wire error.
- Verification: `tests/kg/` offline suite passes (429 passed); `ruff` clean; helper unit checks cover count/byte caps, disabled dimensions, empty input, oversized-single passthrough, and `ordered` passthrough.
- Out of scope (noted for follow-up): very large `$in` read/delete queries still pass the full id list in one query document — only a concern at ~hundreds of thousands of ids.

### DIFF
--- a/env.example
+++ b/env.example
@@ -828,7 +828,8 @@ MONGO_URI=mongodb://localhost:27017/
 MONGO_DATABASE=LightRAG
 ### DB specific workspace should not be set, keep for compatible only
 # MONGODB_WORKSPACE=forced_workspace_name
-# Flush-time batching limits for MongoVectorDBStorage (non-positive disables that dimension)
+# Flush-time bulk_write batching limits for MongoDB upsert paths (KV, vector, graph).
+# (non-positive disables that dimension; DELETE cap applies to MongoVectorDBStorage)
 # MONGO_UPSERT_MAX_PAYLOAD_BYTES=16777216
 # MONGO_UPSERT_MAX_RECORDS_PER_BATCH=128
 # MONGO_DELETE_MAX_RECORDS_PER_BATCH=1000
--- a/env.example
+++ b/env.example
@@ -828,6 +828,10 @@ MONGO_URI=mongodb://localhost:27017/
 MONGO_DATABASE=LightRAG
 ### DB specific workspace should not be set, keep for compatible only
 # MONGODB_WORKSPACE=forced_workspace_name
+# Flush-time batching limits for MongoVectorDBStorage (non-positive disables that dimension)
+# MONGO_UPSERT_MAX_PAYLOAD_BYTES=16777216
+# MONGO_UPSERT_MAX_RECORDS_PER_BATCH=128
+# MONGO_DELETE_MAX_RECORDS_PER_BATCH=1000
 
 # Community/local Docker MongoDB example for KV, graph, or doc-status storage only:
 # MONGO_URI=mongodb://localhost:27017/
--- a/lightrag/kg/milvus_impl.py
+++ b/lightrag/kg/milvus_impl.py
@@ -1783,6 +1783,7 @@ class MilvusVectorDBStorage(BaseVectorStorage):
                         logger.info(
                             f"[{self.workspace}] {self.namespace} flush: upsert split into "
                             f"{len(upsert_batches)} batches for {len(list_data)} records "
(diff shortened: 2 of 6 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3168, opened 2026-05-30
✨ feat(opensearch): bound bulk write payload/record size

danielaskdd:
## Summary

Ports PR #3167's MongoDB batching contract to the OpenSearch backend: every bulk write is now bounded by **payload bytes (primary) + record count (secondary)**, env-tunable, with visible split logging and fail-fast retry semantics — keeping write behaviour consistent across backends.

## Background

opensearch-py's `helpers.async_bulk()` already splits a request via `async_streaming_bulk → _chunk_actions → _ActionChunker` using `max_chunk_bytes` (byte budget, primary) and `chunk_size` (record count, secondary), with an oversized single action emitted as its own chunk — **semantically identical** to Mongo's `_chunk_by_budget`. So the "batching" here is **not** a re-implementation of the split: we expose the two limiter dimensions as env vars and pass them through as `chunk_size` / `max_chunk_bytes`.

The only path without auto-chunking is graph `remove_edges`'s raw `client.bulk(body=...)`, which is split manually — mirroring Mongo's manual-split delete path.

## Changes

- **`_resolve_bulk_batch_limits()`** — reads `OPENSEARCH_UPSERT_MAX_PAYLOAD_BYTES` / `OPENSEARCH_UPSERT_MAX_RECORDS_PER_BATCH` / `OPENSEARCH_DELETE_MAX_RECORDS_PER_BATCH`, with module defaults and non-positive-disables-dimension + warning.
- **`_run_chunked_async_bulk()`** — thin wrapper around `helpers.async_bulk`; empty-actions short-circuit, forwards `chunk_size`/`max_chunk_bytes`, INFO log on split. Non-positive caps map to "whole list in one chunk" / a large finite byte sentinel (`1 << 62`, since `async_bulk` needs a positive int).
- Every storage class (`KV`, `DocStatus`, `Graph`, `VectorDB`) reads & stores the three limits in `__post_init__`.
- All bulk write paths wired through the wrapper: **KV / Vector flush split into separate delete/upsert phases** (so the delete cap can differ from the upsert cap, mirroring mongo_impl), DocStatus upsert/delete, graph node/edge upsert, `remove_nodes`.
- `remove_edges` raw bulk split manually by the delete record-count cap.
- `env.example`: three `OPENSEARCH_*` knobs aligned with `MONGO_*`.

## Defaults

| Knob | Default | Rationale |
|---|---|---|
| `OPENSEARCH_UPSERT_MAX_PAYLOAD_BYTES` | `104857600` (100 MiB) | sits at the typical `http.max_content_length` ceiling |
| `OPENSEARCH_UPSERT_MAX_RECORDS_PER_BATCH` | `128` | matches Mongo |
| `OPENSEARCH_DELETE_MAX_RECORDS_PER_BATCH` | `1000` | matches Mongo |

## Test plan

- [x] `ruff check lightrag/kg/opensearch_impl.py` — clean
- [x] `pytest tests/kg/opensearch_impl/ -q` — **205 passed**
- [x] New helper tests: env defaults / override / non-positive-disable+warning, empty short-circuit, limit forwarding, sentinel mapping
- [x] New flush test: mixed flush issues two `async_bulk` calls each with its own cap

## Out of scope

- `remove_edges`'s current swallow-and-log error behaviour (pre-existing silent-failure; preserved unchanged here).
- Oversized `$in` / query id-list bounds (a Mongo-PR follow-up too).

### DIFF
--- a/env.example
+++ b/env.example
@@ -858,6 +858,10 @@ OPENSEARCH_VERIFY_CERTS=false
 # OPENSEARCH_KNN_EF_SEARCH=100
 ### PPL graphlookup for server-side graph traversal (auto-detected if not set)
 # OPENSEARCH_USE_PPL_GRAPHLOOKUP=true
+### Bulk batching limits (split a single async_bulk request; non-positive disables that dimension)
+# OPENSEARCH_UPSERT_MAX_PAYLOAD_BYTES=104857600
+# OPENSEARCH_UPSERT_MAX_RECORDS_PER_BATCH=128
+# OPENSEARCH_DELETE_MAX_RECORDS_PER_BATCH=1000
 ### DB specific workspace should not be set, keep for compatible only
 # OPENSEARCH_WORKSPACE=forced_workspace_name
 
--- a/lightrag/kg/opensearch_impl.py
+++ b/lightrag/kg/opensearch_impl.py
@@ -264,10 +264,14 @@ async def _run_chunked_async_bulk(
         else _OPENSEARCH_UNBOUNDED_PAYLOAD_BYTES
     )
     if len(actions) > chunk_size:
+        # Log format aligned with mongo_impl's flush split log
+        # (max_payload=/batch= field names, raw configured values). Unlike
+        # Mongo we cannot report the final batch count up front: async_bulk's
+        # _ActionChunker decides it at stream time by byte budget, so this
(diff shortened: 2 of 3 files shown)


danielaskdd:
@codex review

PR #3170, opened 2026-05-30
⚡ feat(postgres): chunk-level transaction fallback for PGGraph batch writes

danielaskdd:
## Summary

**PR 2** of the PostgreSQL batching plan (after the merged KV/Vector/DocStatus PR). Apache AGE has no parameterized `UNWIND` bulk upsert, so `PGGraphStorage`'s batch methods kept per-row Cypher but ran **each row** in its own connection-acquire → AGE-configure → transaction cycle — high overhead under large ingests. This groups rows into payload/record-bounded **chunks**, each run in a **single transaction** on one shared connection. It does not reach single-`UNWIND` performance, but removes the per-row transaction/connection overhead with low risk.

## Changes

- **Shared SQL builders**: extract `_build_upsert_node_sql` / `_build_upsert_edge_sql` so the single-row (`upsert_node`/`upsert_edge`) and batch paths emit identical Cypher and cannot drift.
- **Limits**: `PGGraphStorage.__post_init__` resolves the shared `POSTGRES_UPSERT_MAX_*` / `POSTGRES_DELETE_MAX_RECORDS_PER_BATCH` knobs.
- **`upsert_nodes_batch` / `upsert_edges_batch`**: dedup (last-write-wins) → `_chunk_by_budget` → each chunk run in one transaction via a `@retry`-wrapped `_upsert_node_chunk` / `_upsert_edge_chunk`. Each chunk keeps the **two-layer retry** of the single-row path: `_run_with_retry` for connection-level transients, the tenacity `@retry` + `_is_transient_graph_write_error` for query-level deadlock/serialization errors wrapped as `PGGraphQueryException`.
- **Edge serialization (advisory locks)** — the edge upsert's `OPTIONAL MATCH + DELETE + CREATE` is not atomic against a concurrent writer of the same pair (both could observe no edge and both `CREATE`, leaving duplicate `DIRECTED` rows). The graph-edit endpoints (`/graph/relation/edit` etc.) only best-effort-check `pipeline_status` and don't reserve the writer slot, so the storage-layer advisory lock is the documented last line of defense. Locks are arranged so the single and batch paths share a lock domain while preserving pipeline concurrency:
  - **single `upsert_edge`** takes a per-edge **EXCLUSIVE** lock keyed on `(graph_name, ordered (src,tgt))` — serialises same-edge single-vs-single, lets different edges run concurrently — **plus** a graph-wide **SHARED** lock (`pg_advisory_xact_lock_shared`, keyed on `graph_name`).
  - **batch `upsert_edges_batch`** takes one graph-wide **EXCLUSIVE** lock per chunk (`_GRAPH_ADVISORY_LOCK_SQL`) — a single advisory lock regardless of edge count, no per-edge pile-up.
  - shared/shared is compatible (concurrent single writers on different edges don't serialise); shared/exclusive conflicts, so a bulk batch and any single edge write on the same graph cannot interleave their `OPTIONAL MATCH/DELETE/CREATE` → no cross-path duplicate rows.
- **`remove_nodes`**: chunk the inlined `IN [...]` list (bounds Cypher text) but run **all chunks in ONE transaction**, preserving the original single-statement all-or-nothing semantics.
- **`remove_edges`**: chunk and run **one transaction per chunk** (the old path opened one transaction *per edge*), bounding Cypher text and transaction duration per chunk.

## Test plan

- [x] `ruff check` — clean
- [x] New `tests/kg/postgres_impl/test_postgres_graph_batch.py`: chunk counts, per-chunk vs single-transaction (`remove_nodes` = 1 tx, `remove_edges` = 1 tx/chunk), dedup last-write-wins, byte-budget split, missing-`entity_id` raises before any tx, empty no-ops, and that the edge batch takes exactly **one graph-keyed exclusive lock per chunk** (not per-edge).
- [x] `test_postgres_upsert_edge_cypher.py` / `test_postgres_cypher_injection.py`: single `upsert_edge` issues per-edge exclusive lock + graph-wide shared lock + cypher (asserted against the 3-statement sequence); cross-backend `test_batch_graph_operations.py` ordering tests updated to the chunk flow.
- [x] `pytest tests/kg/ -q` — **680 passed, 17 skipped** (live-PG cases skip)

## Out of scope (follow-up)

- **PR 3**: explore AGE true `UNWIND` bulk upsert to replace the per-row-in-chunk execution.

### DIFF
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -5814,21 +5814,6 @@ def _is_transient_graph_write_error(exc: BaseException) -> bool:
     )
 
 
-# Transaction-scoped advisory lock that serialises concurrent upserts of the
-# same logical edge. Keyed on (graph_name, ordered (src, tgt)) so {A,B}/{B,A}
-# collide while the same pair in a different graph/workspace does not. See
-# PGGraphStorage.upsert_edge for the full rationale.
-_EDGE_ADVISORY_LOCK_SQL = (
-    "SELECT pg_advisory_xact_lock("
-    "  hashtextextended("
-    "    $1::text || E'\\x01' ||"
-    "    LEAST($2::text, $3::text) || E'\\x01' || GREATEST($2::text, $3::text),"
-    "    0"
-    "  )"
-    ")"
-)
-
-
 @final
 @dataclass
 class PGGraphStorage(BaseGraphStorage):
@@ -6346,8 +6331,7 @@ class PGGraphStorage(BaseGraphStorage):
         Shared by ``upsert_edge`` and ``upsert_edges_batch``. The endpoint ids
         are parameterized; edge properties are inlined in the CREATE clause (the
         only reliable way to persist edge properties in AGE -- see
-        ``upsert_edge`` for the full rationale). The advisory lock is issued
-        separately via ``_EDGE_ADVISORY_LOCK_SQL``.
+        ``upsert_edge`` for the full rationale).
         """
         props_literal = self._format_properties(edge_data) if edge_data else "{}"
         cypher_query = f"""MATCH (source:base {{entity_id: $src_id}})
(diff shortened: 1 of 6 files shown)


danielaskdd:
@codex  review

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3171, opened 2026-05-31
🔒 fix(opensearch): canonical edge ids to close reciprocal upsert race

danielaskdd:
## Background

PR #3170 closed the AGE/PostgreSQL read-then-write edge-write race. The same review (`GraphDB-consistency.md`) flagged a pre-existing **same-edge concurrency gap in the OpenSearch graph backend** that today relies solely on the application-level keyed lock.

`upsert_edge` keyed each edge by a **directed** id `hash("src-tgt")` and then did an `exists(reverse)` **read-then-write** probe to reuse the reverse-direction doc. Two concurrent reciprocal writers `(A,B)` / `(B,A)` can each miss the other and `index()` into two different `_id`s → **duplicate edge documents** (which then inflate `node_degree` / `get_node_edges`).

## What this PR does

**Canonical edge id — idempotent by construction, no lock**
- New `_canonical_edge_id(src, tgt) = hash("-".join(sorted((src, tgt))))`. Both directions collapse onto the **same `_id`**, so concurrent reciprocal writers overwrite one doc (last-write-wins) instead of racing into two.
- `upsert_edge`: drop the `exists(reverse)` read-then-write probe; write straight to the canonical id.
- `upsert_edges_batch`: drop the reverse-id `mget`; key by canonical id with in-batch last-write-wins dedup.
- **Reads unchanged & compatible**: the canonical id is always one of the two directed ids, so `has_edge`/`get_edge` (`mget` both orientations) and `remove_edges` (delete both candidates) keep working untouched; `node_degree`/`get_node_edges`/`get_nodes_edges_batch`/`delete_node` are field-term queries, id-format-agnostic.

**One-time reindex migration for legacy data**
- `_migrate_edges_to_canonical_id_if_needed()` runs in `initialize()` inside `get_data_init_lock()`, guarded by an index `_meta.edge_id_canonical_v1` flag so it scans **at most once** per index.
- Scrolls existing edges and re-keys non-canonical docs onto their canonical id (`index(canonical)` + `delete(old)`), collapsing legacy reciprocal duplicates (last-write-wins, edges are undirected).
- **Operator-friendly for large indexes**: flushes in bounded batches (no full in-memory buffer), with `count`-based start log, periodic progress (`scanned X/total, migrated Y`), completion summary, and an explicit "already on canonical ids; skipping" line on subsequent boots.
- **Multi-node safe**: bulk uses `raise_on_error=False`, tolerates benign stale-doc `404` deletes, and leaves the index **unflagged on real errors** so the next startup retries.

## Testing

`pytest tests/kg/opensearch_impl/ -q` → **211 passed**, `ruff` clean.

New/updated coverage: canonical-id reciprocal write folding, batch reciprocal collapse (last-write-wins), migration reindex + idempotent skip-when-flagged, large-scan progress logging, and 404-tolerance vs real-error bail.

## Notes
- Pre-existing gap; **not** introduced by #3170. This is the OpenSearch half of the consistency follow-ups; a separate PR will cover Mongo (`edge_key` + partial unique index + dedup/backfill migration).
- No API or schema-mapping change; edges index gains only a `_meta` flag.

### DIFF
--- a/lightrag/kg/opensearch_impl.py
+++ b/lightrag/kg/opensearch_impl.py
@@ -293,30 +293,19 @@ _EDGE_ID_CANONICAL_META_FLAG = "edge_id_canonical_v1"
 _EDGE_MIGRATION_PROGRESS_INTERVAL = 50_000
 
 
-def _reciprocal_edge_ids(source_node_id: str, target_node_id: str) -> tuple[str, str]:
-    """Return ``(canonical_id, other_orientation_id)`` for an edge.
+def _canonical_edge_id(source_node_id: str, target_node_id: str) -> str:
+    """Direction-independent edge document ``_id``.
 
-    ``canonical_id = hash(sorted(src, tgt))`` collapses an edge and its reverse
-    onto the same ``_id``, so concurrent ``(A,B)``/``(B,A)`` writes overwrite one
-    document (last-write-wins) instead of racing into two — idempotent by
-    construction, no ``exists(reverse)`` read-then-write and no lock needed. The
-    canonical id is always one of the two directed ids
-    ``hash("src-tgt")``/``hash("tgt-src")``, so the bidirectional ``mget`` in
-    ``has_edge``/``get_edge`` keeps finding it.
-
-    ``other_orientation_id`` is the reverse-direction directed id. The two are
-    equal only for a self-loop (``src == tgt``).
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3172, opened 2026-05-31
🔒 fix(mongo): canonical edge_key + unique index to close upsert duplicate race

danielaskdd:
## Background

Follow-up to the OpenSearch consistency fix (#3171). `GraphDB-consistency.md` flagged the **MongoDB graph backend's same-edge upsert race**: `upsert_edge` matched on a bidirectional `$or` filter with **no unique index**, so two concurrent writers — or two processes bypassing the application keyed lock — both miss and both insert, producing **duplicate edge documents** (which inflate `node_degree` / `get_node_edges`). Today this relies solely on the keyed lock.

## What this PR does

**Canonical endpoint fields + compound partial unique index (DB-level backstop)**
- New `_canonical_edge_endpoints(src, tgt) = tuple(sorted((src, tgt)))` → two separate fields `edge_lo` / `edge_hi`, written into `$set`. Direction-independent, so `(A,B)` and `(B,A)` map to the same pair.
- Two **separate** fields (not a single delimiter-joined `edge_key` string) so distinct pairs never collide even if an id contains the would-be delimiter (e.g. custom-KG ids) — no input sanitisation needed.
- `upsert_edge` / `upsert_edges_batch` match on `{edge_lo, edge_hi}` instead of `$or`, so a unique index can reject the racing second insert.
- Partial **compound** unique index `edge_endpoints_unique` on `(edge_lo, edge_hi)` (`partialFilterExpression` requires both fields exist and are strings).

**Duplicate-key handling (the unique index makes the race observable)**
- `upsert_edge`: retry once on `DuplicateKeyError` — the retry matches the now-existing doc and updates it.
- `upsert_edges_batch`: dedupe by `(edge_lo, edge_hi)` within the batch (collapsing reciprocal directions, also avoiding intra-batch 11000), then retry once on a **duplicate-key-only** `BulkWriteError`. The bulk keeps `ordered=True` (intra-batch last-write-wins is already guaranteed by the endpoint-pair dedupe, so ordering is not load-bearing; kept for continuity). A `writeConcernError` (durability failure, empty `writeErrors`) or any non-11000 error is **re-raised, never masked**. Note: under `ordered=True` the bulk aborts at the first failing op, so a non-11000 error hidden behind a leading 11000 race surfaces one retry later rather than being masked.

**Fail-fast one-time migration** (`create_edge_indexes_and_migrate_if_not_exists`, in `initialize` under `get_data_init_lock`)
1. dedupe legacy reciprocal duplicates, **merging the full relation payload** into the survivor — `source_id`/`source_ids`/`file_path`/`description` unioned, `keywords` set-unioned, `weight` summed (like `_merge_edges_then_upsert`) — so **no provenance or relation evidence is lost**. The merge is **idempotent across fail-fast retries** (weight is re-added only for duplicates whose `source_ids` are not yet folded into the survivor);
2. backfill `edge_lo` / `edge_hi` on every remaining doc (one doc per canonical pair ⇒ the pairs are unique);
3. build the partial compound unique index.
- Gated by the index's existence (the completion flag). **Any error raises**, so startup aborts rather than serving a half-migrated collection — the `{edge_lo, edge_hi}` filter relies on every doc having the endpoints. Mirrors the OpenSearch migration's fail-fast design.

**Reads unchanged & compatible**: `has_edge`/`get_edge`/`node_degree`/`get_node_edges`/`get_nodes_edges_batch`/`remove_edges`/`delete_node` keep the bidirectional `$or` (correct both before and after migration). `edge_lo`/`edge_hi` are excluded from `KnowledgeGraphEdge.properties`.

## Lessons applied from #3171
- **Fail-fast migration** (no per-write cleanup overhead; gate startup on a complete migration).
- **No data loss on dedup**: union provenance and merge the full relation payload instead of dropping a side.
- **Don't mask real errors**: retry only pure duplicate-key races; surface writeConcern/other failures.
- **Idempotent guard** via the unique index; **don't mutate the caller's `edge_data`**.

## Testing
`pytest tests/kg/mongo_impl/ -q` → **46 passed**, `ruff` check + format clean.

New coverage: canonical-pair direction independence, delimiter non-collision, `{edge_lo, edge_hi}` filter + `$set`, single-edge dup retry / persistent-dup re-raise, batch reciprocal dedup + last-write-wins, batch dup-only retry, non-dup and **writeConcern** re-raise, **non-dup error hidden behind a leading dup surfaces on retry**, migration skip-when-indexed, dedupe-merge-provenance (incl. string-weight coercion + retry idempotency) + backfill + compound-unique-index build.

## Notes
- Pre-existing gap; not introduced by #3170/#3171. This is the Mongo half of the consistency follow-ups.
- The migration's `edge_lo`/`edge_hi` backfill uses an **aggregation-pipeline update** (`update_many(filter, [pipeline])`), which requires **MongoDB 4.2+**.
- Reads keep the bidirectional `$or` and so do not use the new compound index; switching the read path to `{edge_lo, edge_hi}` (which the fail-fast migration guarantees on every served doc) to leverage the unique index is a possible follow-up, left out here to keep this PR scoped to closing the write race.
- Assumes stop-the-world / single-deployment restarts (`get_data_init_lock` only serialises one deployment's workers). A true rolling deploy with mixed code versions writing one collection could leave a straggler duplicate; remedy is to drop the `edge_endpoints_unique` index and let the next startup re-migrate.

### DIFF
--- a/lightrag/kg/mongo_impl.py
+++ b/lightrag/kg/mongo_impl.py
@@ -1797,6 +1797,14 @@ class MongoGraphStorage(BaseGraphStorage):
             # Retry ONLY when every failure is a duplicate-key race; a
             # writeConcern failure (durability problem, empty writeErrors) or any
             # other write error must surface, not be masked by a blind retry.
+            #
+            # NOTE: under ordered=True the bulk aborts at the FIRST failing op, so
+            # writeErrors holds at most one entry — the all(...) check therefore
+            # only inspects that first error, not the whole batch. Ops after it
+            # never ran; they re-run when we retry the entire op list below. So a
+            # non-11000 error hidden behind a leading 11000 is not masked — it
+            # simply surfaces one retry later (the retry hits it and re-raises,
+            # since by then the leading dup has resolved to a plain update).
             dup_only = (
                 bool(write_errors)
                 and all(we.get("code") == _DUPLICATE_KEY_CODE for we in write_errors)
--- a/lightrag/kg/mongo_impl.py
+++ b/lightrag/kg/mongo_impl.py
@@ -1192,12 +1192,18 @@ class MongoGraphStorage(BaseGraphStorage):
 
         Groups by the canonical (sorted) endpoint pair; for each group with more
         than one doc, keeps the newest by ``created_at`` and **merges the
-        non-survivors' relation payload into it before deleting them**, mirroring
-        ``_merge_edges_then_upsert`` so no relation evidence is lost:
-        ``source_ids``/``source_id``/``file_path`` are unioned, ``keywords`` are
-        comma-set-unioned, distinct ``description`` values are ``GRAPH_FIELD_SEP``
-        joined (the no-LLM merge fallback), and ``weight`` is summed. Returns the
-        number of docs removed.
+        non-survivors' relation payload into it before deleting them** so no
+        relation evidence is lost: ``source_ids``/``source_id``/``file_path`` and
+        ``description`` are unioned over their ``GRAPH_FIELD_SEP`` components,
+        ``keywords`` are comma-set-unioned, and ``weight`` takes the max.
+
+        The merge is **idempotent across retries**: if a transient error aborts
+        startup after the survivor update but before the delete, the next run
+        re-processes the same group and must produce the same survivor. All
+        fields union their split components (so re-merging an already-merged
+        survivor is a no-op), and ``weight`` uses max rather than sum (summing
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3173, opened 2026-05-31
🔒 fix(opensearch): merge reverse payload into canonical on duplicate migration

danielaskdd:
## Background

Follow-up to #3171 (merged), which migrated OpenSearch edges onto canonical sorted-pair `_id`s. That migration handled a reciprocal **duplicate** (a `create` 409 — a forward canonical doc already exists) by simply deleting the reverse doc, which **silently discarded the reverse doc's `description` / `keywords` / `weight` / `source_ids`**. Those duplicates are exactly the upsert-race artifacts the migration repairs, and sequential `_merge_edges_then_upsert` normally preserves that relation evidence — so the migration could degrade query context.

This also brings OpenSearch in line with the Mongo consistency fix (#3172), which merges the full relation payload when consolidating duplicates.

## What this PR does

When the migration hits a reciprocal duplicate (`create` 409), it now **merges the reverse doc's relation payload into the existing canonical doc before deleting the reverse**, instead of dropping it.

- New `_merge_edge_payloads(docs)` — shared field semantics with `mongo_impl` / `operate.py._merge_edges_then_upsert` (minus LLM description summarisation):
  - `source_id` / `source_ids` / `file_path` / `description` union their `GRAPH_FIELD_SEP` components,
  - `keywords` comma-set-unioned,
  - `weight` is **summed** (duplicate docs carry separate accumulated weight — matches the Mongo dedupe-merge in #3172).
  - **Idempotent** (every field unions split components; the weight sum counts each duplicate only while its `source_ids` are not already folded into the base) so re-merging an already-merged doc is a no-op across fail-fast retries.
  - **String-weight safe**: legacy weights are coerced to float, non-numeric skipped (the graph API is `dict[str, str]`), so a bad value can't crash startup.
- New `_merge_into_canonical_edge()` — `GET` the canonical doc, merge, write back with **optimistic concurrency** (`if_seq_no` / `if_primary_term`). A concurrent live write during a rolling deploy is never clobbered: on a version conflict it re-reads (incorporating that write) and re-merges, up to a bounded retry count, then **fails fast before any delete** (startup aborts, flag stays unset, next startup rescans). This keeps #3171's "create-only never clobbers a live write" guarantee through the merge path. If the canonical vanished between the 409 and the merge `GET`, it is recreated from the (merged) reverse sources.
- `_flush_pending` captures the 409 canonical ids and merges each before the delete phase. **All** pending sources mapping to one canonical are folded in together (a node pair with 3+ legacy docs can yield >1 reverse doc per canonical — only one wins the insert, the rest 409), so no reverse payload is dropped. Any non-409 create error still fails fast.

## Testing

`pytest tests/kg/opensearch_impl/ -q` → **221 passed**, `ruff` check + format clean.

New coverage:
- merge-on-conflict (asserts unioned `description`/`source_ids`/`file_path`, set-unioned `keywords`, summed `weight`, `if_seq_no`/`if_primary_term`, direction fields preserved),
- version-conflict retry (re-read + retry, never clobber),
- **multi reverse-doc → one canonical** (every 409'd source folded in, not just the last — `source_ids` + `weight` summed across all),
- **NotFoundError recreate** (canonical vanished → recreated from merged reverse sources, no optimistic-concurrency clobber),
- **persistent-conflict abort** (bounded retries → `RuntimeError` raised before any delete).

## Notes
- Builds on #3171 (merged); pairs with the Mongo dedupe-merge in #3172 — both backends now consolidate duplicates without losing relation evidence, idempotently and type-safely.

### DIFF
--- a/lightrag/kg/opensearch_impl.py
+++ b/lightrag/kg/opensearch_impl.py
@@ -27,7 +27,7 @@ from ..base import (
     DocStatus,
     DocStatusStorage,
 )
-from ..utils import logger, compute_mdhash_id, _cooperative_yield
+from ..utils import logger, compute_mdhash_id, _cooperative_yield, merge_source_ids
 from ..types import KnowledgeGraph, KnowledgeGraphNode, KnowledgeGraphEdge
 from ..constants import GRAPH_FIELD_SEP
 from ..kg.shared_storage import get_data_init_lock, get_namespace_lock
@@ -38,7 +38,12 @@ if not pm.is_installed("opensearch-py"):
     pm.install("opensearch-py")
 
 from opensearchpy import AsyncOpenSearch, helpers  # type: ignore
-from opensearchpy.exceptions import OpenSearchException, NotFoundError, RequestError  # type: ignore
+from opensearchpy.exceptions import (  # type: ignore
+    OpenSearchException,
+    NotFoundError,
+    RequestError,
+    ConflictError,
+)
 
 config = configparser.ConfigParser()
 config.read("config.ini", "utf-8")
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3174, opened 2026-05-31
⚡️ perf(mongo): exact single-edge reads use canonical (edge_lo, edge_hi) index

danielaskdd:
## Background

Stacked on #3172 (canonical `(edge_lo, edge_hi)` + compound unique index). That PR closes the same-edge upsert race and adds a partial **compound unique index** `edge_endpoints_unique` on `(edge_lo, edge_hi)`, plus a **fail-fast migration** that guarantees every served edge doc carries the endpoints before reads begin. It deliberately left the read path on the bidirectional `$or` to keep that PR scoped to the write race. This is the follow-up.

> **Base:** `feat/mongo-canonical-edge-key` (depends on #3172). Once #3172 merges, GitHub will retarget this to `main`.

## What this PR does

Switch the **exact single-edge** lookups from the bidirectional `source_node_id`/`target_node_id` `$or` to the canonical `{edge_lo, edge_hi}` filter, so they are served by the compound unique index instead of a scan/secondary index:

- **`has_edge`** — direction-independent point lookup, one index hit.
- **`get_edge`** — same canonical point lookup.
- **`remove_edges`** — one canonical clause per edge (vs. the old two bidirectional clauses, halving the `$or`), with reciprocal/duplicate inputs collapsed.

**Why it's safe:** the fail-fast migration in `initialize` (from #3172) aborts startup unless every doc has `edge_lo`/`edge_hi`, and every new write sets them — so once reads are being served, the canonical filter matches exactly what the old `$or` matched.

## What this PR does NOT change

Per-node scans keep the bidirectional `$or` on purpose: `node_degree`, `get_node_edges`, `get_nodes_edges_batch`, `node_degrees_batch`, `delete_node`, and the KG-traversal aggregations all fan out over *every edge touching a node*. `edge_hi` is not a prefix of the `(edge_lo, edge_hi)` compound index, so an `{$or: [{edge_lo: X}, {edge_hi: X}]}` rewrite would gain nothing (and the `edge_hi` half would be unindexed). Those queries want their own `source_node_id`/`target_node_id` indexes, which is a separate concern.

## Testing

`pytest tests/kg/mongo_impl/ -q` → **51 passed**, `ruff` check + format clean.

New `TestMongoEdgeReadCanonicalFilter`: `has_edge`/`get_edge` use the canonical filter and are direction-independent; `remove_edges` emits one canonical clause per edge and collapses reciprocals; empty `remove_edges` is a no-op.

### DIFF
--- a/lightrag/kg/mongo_impl.py
+++ b/lightrag/kg/mongo_impl.py
@@ -1424,20 +1424,15 @@ class MongoGraphStorage(BaseGraphStorage):
     async def has_edge(self, source_node_id: str, target_node_id: str) -> bool:
         """
         Check if there's a direct single-hop edge between source_node_id and target_node_id.
+
+        Matches on the canonical ``(edge_lo, edge_hi)`` pair (direction-independent)
+        instead of the bidirectional ``$or``, so this point lookup is served by the
+        compound unique index. Safe because the fail-fast migration in
+        ``initialize`` guarantees every served doc carries the endpoints.
         """
+        edge_lo, edge_hi = _canonical_edge_endpoints(source_node_id, target_node_id)
         doc = await self.edge_collection.find_one(
-            {
-                "$or": [
-                    {
-                        "source_node_id": source_node_id,
-                        "target_node_id": target_node_id,
-                    },
-                    {
-                        "source_node_id": target_node_id,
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3175, opened 2026-05-31
⚡ perf(opensearch): collapse edge point-lookups onto canonical _id

danielaskdd:
## Background

`OpenSearchGraphStorage` keys every edge document under `_canonical_edge_id(src, tgt)` — the two endpoints sorted then hashed, so `(A,B)` and `(B,A)` collapse onto the same document. The startup `_migrate_edges_to_canonical_id_if_needed()` is **fail-fast**: the service does not finish initializing until migration completes, after which the edge index holds canonical `_id`s only.

Given that invariant, point-lookup/delete paths for a known endpoint pair no longer need to build forward/reverse candidate IDs — they can address the single canonical `_id` directly.

## Changes

- **`has_edge()`**: point-checks the canonical `_id` with `client.exists()`, mirroring `has_node()` — real-time (translog-backed), independent of the index refresh cycle, no candidate-id fan-out.
- **`get_edge()`**: reads the single canonical document via `_mget_optional_doc()`, identical to `get_node()`; return behavior unchanged (still writes `_id` back into the result dict).
- **`remove_edges()`**: emits one delete op per canonical `_id`, deduped via a set so reciprocal inputs `(A,B)`/`(B,A)` collapse to a single op.

Traversal-style paths that rely on the `source_node_id`/`target_node_id` field indexes (`node_degree`, `get_node_edges`, BFS, subgraph aggregations, etc.) are untouched — the canonical `_id` is a hash and cannot answer "all edges of node X", so those must still query the field indexes.

## Compatibility note

The canonical-only delete in `remove_edges()` depends on the migration having fully succeeded. If there is a rolling-upgrade window where non-canonical edges are still being written, a canonical delete cannot reap those stragglers — in that case, clear the `_meta` flag and re-run the migration as documented in the migration comments.

## Tests

Updated and extended `tests/kg/opensearch_impl/test_opensearch_storage.py`:
- `test_has_edge` now mocks `client.exists`, asserts it is called with the canonical `_id` and that `mget` is not called; added `test_has_edge_false`.
- `test_get_edge` uses a single canonical document and asserts the mget body and the returned `_id`.
- `test_remove_edges` asserts one op per distinct canonical edge; added `test_remove_edges_dedups_reciprocal` to verify dedup.

```
./scripts/test.sh tests/kg/opensearch_impl/test_opensearch_storage.py
# 208 passed
```

### DIFF
--- a/lightrag/kg/opensearch_impl.py
+++ b/lightrag/kg/opensearch_impl.py
@@ -2376,25 +2376,18 @@ class OpenSearchGraphStorage(BaseGraphStorage):
             return False
 
     async def has_edge(self, source_node_id: str, target_node_id: str) -> bool:
-        """Check whether an edge exists between two nodes (bidirectional).
+        """Check whether an edge exists between two nodes.
 
-        Uses mget with the two candidate edge IDs so the check is real-time
-        (translog-backed), consistent with has_node() and independent of the
-        index refresh cycle.
+        Startup migration is fail-fast, so after initialize() every edge is keyed
+        by its canonical (sorted-pair) ``_id``. Point-check that single id with
+        exists(), mirroring has_node() — real-time (translog-backed), independent
+        of the index refresh cycle, no candidate-id fan-out.
         """
         if not self._indices_ready:
             return False
         try:
-            forward_id = compute_mdhash_id(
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3176, opened 2026-05-31
🐛 fix(postgres): rank get_knowledge_graph("*") nodes by undirected degree

danielaskdd:
## Problem

`PGGraphStorage.get_knowledge_graph("*")` selects its top-N nodes (when the graph exceeds `max_nodes` and must be truncated) by node degree. The previous query used:

```cypher
OPTIONAL MATCH (n)-[r]->()
RETURN id(n) as node_id, count(r) as degree
```

This counts **only outgoing edges**. A node that is mostly an edge *target* (high in-degree, low out-degree) gets a low degree and can be dropped on truncation — so the `*` view loses genuinely central nodes. This violates LightRAG's undirected edge semantics (any direction is the same connection).

The sibling `get_popular_labels()` already computes degree the right way (native SQL, `start_id` + `end_id`); this PR brings the `*` node selection in line.

## Change

Replace the degree query with native SQL over AGE's underlying tables:

- **Undirected degree** — `UNION ALL` of `start_id` and `end_id` from `_ag_label_edge`, mirroring `get_popular_labels()`.
- **Isolated nodes preserved** — `LEFT JOIN {graph}.base` + `COALESCE(degree, 0)` keeps degree-0 nodes when the graph is *not* truncated (the old `OPTIONAL MATCH` did this; a plain `JOIN` would silently drop them).
- **Deterministic ordering** — `ORDER BY degree DESC, v.id ASC` tie-break.
- **Parameterized `LIMIT`** — `$1` via `params={"limit": max_nodes}`.

### Intentionally unchanged

The directed subgraph edge read `(a)-[r]->(b)` is **left as-is**: `a` iterates over every selected node, so it already captures every edge whose endpoints are both selected. Switching it to `-[r]-` would only produce duplicate rows handled by the existing `edge_id` dedup — no behavioral gain. Other edge paths (`has_edge`, `get_edge(s)`, `get_node_edges`, `node_degrees_batch`, `remove_edges`, …) are already undirected and untouched.

No public API or wire-shape change.

## Tests

New `tests/kg/postgres_impl/test_postgres_get_knowledge_graph_all.py` (mocks `_query`):

1. Degree SQL is undirected (`start_id`/`end_id`/`UNION ALL`), preserves isolated nodes (`LEFT JOIN`/`COALESCE`), stable tie-break, and no longer contains the out-only `-[r]->()` fragment.
2. `LIMIT` is parameterized (`$1` + `params={"limit": ...}`), not interpolated.
3. An isolated (degree-0) node survives end-to-end into the returned `KnowledgeGraph`.
4. Regression guard: subgraph read stays `(a)-[r]->(b)` and dedupes duplicate edge rows by `edge_id`.

```
./scripts/test.sh tests/kg/postgres_impl/test_postgres_get_knowledge_graph_all.py \
  tests/kg/postgres_impl/test_postgres_upsert_edge_cypher.py \
  tests/kg/postgres_impl/test_postgres_age_quote_fix.py \
  tests/kg/postgres_impl/test_postgres_graph_batch.py
# 41 passed
```

Integration tests against a live PostgreSQL/AGE backend (`LIGHTRAG_GRAPH_STORAGE=PGGraphStorage`) were not run in this environment.

### DIFF
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -7589,9 +7589,7 @@ class PGGraphStorage(BaseGraphStorage):
                 LEFT JOIN node_degrees d ON d.node_id = v.id
                 ORDER BY degree DESC, v.id ASC
                 LIMIT $1"""
-            node_results = await self._query(
-                query_nodes, params={"limit": max_nodes}
-            )
+            node_results = await self._query(query_nodes, params={"limit": max_nodes})
 
             node_ids = [str(result["node_id"]) for result in node_results]
 
--- a/lightrag/kg/postgres_impl.py
+++ b/lightrag/kg/postgres_impl.py
@@ -7567,15 +7567,31 @@ class PGGraphStorage(BaseGraphStorage):
             total_nodes = count_result[0]["total_nodes"] if count_result else 0
             is_truncated = total_nodes > max_nodes
 
-            # Get max_nodes with highest degrees
-            query_nodes = f"""SELECT * FROM cypher('{self.graph_name}', $$
-                    MATCH (n:base)
-                    OPTIONAL MATCH (n)-[r]->()
-                    RETURN id(n) as node_id, count(r) as degree
-                $$) AS (node_id BIGINT, degree BIGINT)
-                ORDER BY degree DESC
-                LIMIT {max_nodes}"""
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3177, opened 2026-05-31
🔒 fix(mongo): rebuild FAILED Atlas vector index instead of wedging queries

danielaskdd:
## Description

A MongoDB-backed deployment started failing every query with:

```
Query failed: ... cannot query vector index vector_knn_index_space2_entities ... while in state FAILED
```

Atlas/`mongot` leaves a vector search index in the **terminal `FAILED`** state after a build error and never retries it on its own. `MongoVectorDBStorage.create_vector_index_if_not_exists` matched the existing index **only by name** and returned as if healthy, so a `FAILED` index wedged all `$vectorSearch` queries permanently. Worse, `drop()` / `clear_documents` could not recover either — their "recreate" step goes through the same function and hit the same early `return`, so a `FAILED` index was unrecoverable through the normal API.

Root-cause note: in the affected database the stored vectors were all valid (1481 docs, all 1024-dim, no NaN/Inf/zero-norm), so the `FAILED` state was a stale leftover from an earlier one-off build failure, not corrupt data. Dropping and recreating the index rebuilt it to `READY` immediately.

## Changes Made

- `create_vector_index_if_not_exists`: on a name match, detect `status == "FAILED"`, drop the dead index, wait for the drop to clear, then fall through to recreate. Healthy/transitional states (`PENDING`/`BUILDING`) and the dimension-mismatch guard are unchanged.
- Add `_wait_for_search_index_absent`: bounded (120s) best-effort poll so the recreate doesn't collide with a name still in the `DELETING` state; on timeout it logs and proceeds rather than blocking `initialize()` indefinitely.
- This also restores `drop()` / `clear_documents` as a recovery path for a `FAILED` index.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Unit tests added (if applicable) — existing `tests/kg/mongo_impl/test_mongo_deferred_embedding.py` (23) pass; verified idempotency against a live healthy index (READY index untouched) and the drop→wait→recreate path against a live `FAILED` index (rebuilt to `READY`, `$vectorSearch` returns).

## Additional Notes

The dimension-mismatch case still raises `ValueError` (intentional guard against silently rebuilding under a changed embedding model); only the terminal `FAILED` state triggers an automatic rebuild.

### DIFF
--- a/lightrag/kg/mongo_impl.py
+++ b/lightrag/kg/mongo_impl.py
@@ -2992,14 +2992,19 @@ class MongoVectorDBStorage(BaseVectorStorage):
         """Create the Atlas Vector Search index, repairing a FAILED one.
 
         Atlas/mongot leaves a vector index in the terminal ``FAILED`` state
-        after a build error and never retries it on its own; every
-        subsequent ``$vectorSearch`` then raises ``cannot query vector index
-        ... while in state FAILED``. Matching the index only by name would
-        treat that dead index as healthy and wedge all queries permanently,
-        so a same-dimension FAILED index is dropped and rebuilt here. The
-        dimension-mismatch guard runs *before* the rebuild: a FAILED index
-        built under a different embedding model raises rather than being
-        auto-rebuilt against incompatible stored vectors. Transitional states
+        after a build error and never retries it on its own; when that index
+        is also non-queryable every subsequent ``$vectorSearch`` raises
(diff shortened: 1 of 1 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3180, opened 2026-05-31
🔧 refactor(constants): centralize LLM call priorities and align analysis stage with processing

danielaskdd:
## Description

Multimodal **analysis-stage** LLM calls (drawing VLM analysis, table/equation EXTRACT analysis) used `DEFAULT_MM_ANALYSIS_PRIORITY = 12`, while **processing-stage** entity/relation extraction uses the wrapper's default priority `10`. Since lower numbers run first (`asyncio.PriorityQueue` semantics) and the table/equation analysis calls go through the **same `extract` role queue** as processing-stage entity extraction, a busy ingestion queue (priority 10) continually preempted analysis tasks (priority 12). Analysis nodes were starved and stalled, dragging down overall throughput once the processing queue drained but still waited on analysis output.

This PR sets the analysis priority equal to the processing priority so both advance evenly (fair, FIFO-style interleaving within the shared queue), and centralizes every previously inlined `_priority` literal into named constants so the priority scheme has a single source of truth.

## Related Issues

N/A

## Changes Made

- Added named priority constants in `lightrag/constants.py`:
  - `DEFAULT_QUERY_PRIORITY = 5` — interactive query / keyword LLM calls and query-time embeddings
  - `DEFAULT_SUMMARY_PRIORITY = 8` — entity/relation description summary generation
  - `DEFAULT_PROCESSING_PRIORITY = 10` — ingestion extraction; also the `priority_limit_async_func_call` wrapper default
- `DEFAULT_MM_ANALYSIS_PRIORITY` now references `DEFAULT_PROCESSING_PRIORITY` (effectively `12 → 10`), making the "analysis == processing" intent explicit in code.
- Replaced all hardcoded `_priority` literals with the constants across:
  - `lightrag/utils.py` (wrapper default), `lightrag/operate.py`, `lightrag/lightrag.py`
  - All vector-DB backends: `qdrant`, `mongo`, `faiss`, `postgres`, `nano_vector_db`, `opensearch`, `milvus`, and deprecated `chroma`
- No public interfaces added/removed; existing priority ordering (5 < 8 < 10) is preserved — only the analysis stage moves from 12 to 10.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

- Each LLM role (`extract` / `keyword` / `query` / `vlm`) has its own priority queue; priority only orders calls *within* a queue. The cross-stage contention this PR fixes is specifically in the shared **`extract`** queue (analysis table/equation extraction vs. processing entity extraction). The VLM call has its own analysis-only queue, so its value change is a no-op in practice but kept consistent via the shared constant.
- `grep -rn "_priority=[0-9]" lightrag/` returns no remaining hardcoded literals.
- `tests/pipeline/test_pipeline_analyze_multimodal.py` passes (15 passed). That suite asserts `_priority` is consumed by the wrapper (not observable on the raw mock) and references the constant by name, so it remains accurate.

### DIFF
--- a/lightrag/constants.py
+++ b/lightrag/constants.py
@@ -289,6 +289,28 @@ DEFAULT_MAX_PARALLEL_PARSE_DOCLING = 2
 DEFAULT_QUEUE_SIZE_DEFAULT = 100
 DEFAULT_QUEUE_SIZE_INSERT = 4
 
+# LLM / embedding call priority levels.  Lower values run first
+# (asyncio.PriorityQueue semantics); priority only orders calls *within* a
+# single role queue (extract / keyword / query / vlm).  These name the values
+# passed as the ``_priority`` argument to the priority_limit_async_func_call
+# wrapper, centralizing the magic numbers that were previously inlined at each
+# call site.
+#
+# Query stage (interactive: query/keyword LLM calls and query-time embeddings)
+# gets the highest priority so user requests stay responsive.
+DEFAULT_QUERY_PRIORITY = 5
+# Entity/relation description summary generation — ahead of raw extraction but
+# behind interactive query work.
+DEFAULT_SUMMARY_PRIORITY = 8
+# Processing stage entity/relation extraction (ingestion).  Also the wrapper's
+# baseline default for any call that does not pass ``_priority``.
+DEFAULT_PROCESSING_PRIORITY = 10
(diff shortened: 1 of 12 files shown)


danielaskdd:
@codex review

PR #3188, opened 2026-06-02
🐛 fix(pipeline): clean stale per-attempt metadata earlier for PENDING docs

danielaskdd:
## Description

When `apipeline_process_enqueue_documents` resumes work, interrupted documents are reset back to `PENDING` by `_validate_and_fix_document_consistency`. That reset reused the full carry-over list (`doc_status_transition_metadata`), so the **previous attempt's** per-attempt metadata — `parse_*`/`analyzing_*` timings, `*_stage_skipped` flags, `parse_warnings`, `chunk_opts` — was dragged into the new `PENDING` record.

The only cleanup happened later, in `_parse_worker` right before `PARSING` (and it only scrubbed 5 of those keys, missing `parse_warnings`/`chunk_opts`). As a result, while a document waited in the `PENDING` queue the WebUI rendered stale parse/analyze timestamps that did not reflect "waiting to be reprocessed".

The fix keeps the **classification** correct (`process_options`/`source_file` are long-lived directives that must survive a reset; everything else is a per-attempt product) but fixes the **mechanism**: cleanup is moved to a single point during the consistency check, so a document is clean the moment it enters `PENDING`.

## Related Issues

N/A

## Changes Made

- **`lightrag/utils_pipeline.py`**
  - Add `read_source_file_basename()` — the backward-compatible `source_file_name` → `source_file` reader now lives here so reset helpers can reuse it without a reverse import from `pipeline.py`.
  - Add `_DOC_STATUS_METADATA_DIRECTIVE_KEYS` and `doc_status_reset_metadata()` — a directive-only rebuild used on the reset path. It **iterates** the directive whitelist (so future directive additions are carried automatically) and normalises `source_file` with legacy tolerance.
  - Add `_DOC_STATUS_METADATA_ATTEMPT_KEYS` and `doc_status_metadata_has_attempt_fields()` — a precise trigger (timing/skip pairs, `parse_warnings`, `chunk_opts`, and the `PROCESSING` `extraction_meta` fields) for normalising an already-`PENDING` doc.
- **`lightrag/pipeline.py`**
  - `_read_source_file` now delegates to the shared helper.
  - `_validate_and_fix_document_consistency` is the single cleanup point: interrupted docs are reset to a clean `PENDING`, and already-`PENDING` docs that still carry a stale per-attempt field are normalised in place. In both cases the cleaned metadata is mirrored onto the in-memory `status_doc` so later transitions (which rebuild metadata from it) don't reintroduce stale fields.
  - Remove the now-redundant stale-field `pop` loop in `_parse_worker` (the `parse_start_time` stamp and the `source_file` normalisation block are kept).
  - The forward carry-over (`_DOC_STATUS_METADATA_CARRY_OVER_KEYS` / `doc_status_transition_metadata`) is unchanged — forward transitions still carry timing fields through to the terminal record.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

New tests in `tests/pipeline/test_doc_status_reset_metadata.py` cover: `doc_status_reset_metadata` (keep directives / drop attempt fields / legacy `source_file_name` tolerance / blank handling), the `has_attempt_fields` trigger, the interrupted reset clearing both storage and in-memory metadata, stale-`PENDING` normalisation vs. clean-`PENDING` left untouched, and a **semantic lock**: a `PENDING` doc holding only non-attempt custom metadata is left untouched, while one holding a stale key + custom field is collapsed to directives-only.

Verification: `tests/pipeline/` 202 passed, `tests/api/routes/test_document_routes_docx_archive.py` 46 passed, `ruff check` clean.

### DIFF
--- a/lightrag/pipeline.py
+++ b/lightrag/pipeline.py
@@ -83,7 +83,10 @@ from lightrag.utils_pipeline import (
     load_lightrag_document_content,
     make_lightrag_doc_content,
     normalize_document_file_path,
+    doc_status_metadata_has_attempt_fields,
+    doc_status_reset_metadata,
     parsed_artifact_dir_for,
+    read_source_file_basename,
     resolve_doc_file_path,
     sidecar_blocks_path,
     sidecar_uri_for,
@@ -126,18 +129,11 @@ def _call_source_file_resolver(
     return resolver(source_file or file_path)
 
 
-def _read_source_file(data: Any) -> str | None:
-    """Read the source-file basename with backward compatibility.
-
-    The ``source_file_name`` → ``source_file`` rename means documents enqueued
-    before the change persisted the old key in full_docs content_data /
-    doc_status.metadata.  Resumed/legacy docs are read through here so they
-    still resolve their original source basename; write sites always use the
-    new ``source_file`` key.
-    """
-    if not isinstance(data, dict):
-        return None
-    return data.get("source_file") or data.get("source_file_name")
+# Backward-compatible source-file reader.  Implementation lives in
+# utils_pipeline so reset/normalisation helpers there can reuse it without a
+# reverse import into this module; kept as a module-level alias for the
+# existing call sites below.
+_read_source_file = read_source_file_basename
 
 
 # Map ``process_options.chunking`` selector → ``extraction_meta.chunk_method``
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

PR #3190, opened 2026-06-02
🐛 fix(parser): split over-long outline headings at soft line break

danielaskdd:
## Description

The native docx engine could abort the parse of an entire document by misclassifying a normal body paragraph as an over-long heading.

A `<w:p>` paragraph that carries an outline level (`w:outlineLvl`) but whose text exceeds `MAX_HEADING_LENGTH` (200) was passed to `validate_heading_length()`, which raises `DocxContentError` and tears down the whole document parse.

The common real-world cause is a WPS/Word authoring pattern: the author sets an outline level on a paragraph (intending a short sub-heading) but types the body with **soft line breaks** (Shift+Enter → `<w:br/>` → `'\n'`) instead of starting a new paragraph. As a result the heading text and a long body live in a *single* `<w:p>`. `extract_text_from_run` turns the soft break into `'\n'`, so the paragraph's `full_text` becomes e.g. `"A、处置程序\n施工现场一旦发生事故时,..."` (413 chars) and trips the length check.

## Related Issues

#XXXX

## Changes Made

In `extract_docx_blocks` ([lightrag/parser/docx/parse_document.py](../blob/main/lightrag/parser/docx/parse_document.py)), at the heading-detection call site:

- When a paragraph is detected as a heading but its text exceeds `MAX_HEADING_LENGTH`:
  - **Contains a soft break** → split at the first `'\n'`: the first line stays the heading, the remainder is carried as a regular body paragraph in the same block. The author's intended sub-heading is preserved and the body content is kept in full.
  - **No usable soft break** (a genuine single-line over-long heading) → demote the whole paragraph to body text.
- Either path avoids the `DocxContentError` crash and never drops content.
- Within-limit paragraphs are untouched, so golden outputs are unchanged.
- `validate_heading_length()` is left intact as a defense-in-depth guard on the anchor re-split path; the split heading's first line is now well under the limit.
- New non-fatal counters surfaced via `parse_warnings`: `heading_softbreak_split_count`, `demoted_oversize_heading_count`.

Added regression tests in `tests/parser/docx/test_native_docx_extract.py`:
- over-long outline heading **with** soft break → splits (heading kept, body preserved);
- over-long outline heading **without** soft break → demoted to body;
- short outline paragraph → still recognized as a heading (no regression).

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Verified end to end on a real document that previously failed with `ERROR: Heading too long (413 characters, max 200)`:

- `python -m pytest tests/parser/docx/test_native_docx_extract.py tests/parser/docx/test_native_docx_golden.py tests/parser/docx/test_validation_raises.py -q` → 27 passed (golden unchanged).
- Re-running the failing CLI command now completes; the `A、处置程序` sub-heading is preserved as the block heading and the full ~1100-char body is retained as block content, with a single soft-break-split warning on stderr.

### DIFF
--- a/lightrag/parser/docx/parse_document.py
+++ b/lightrag/parser/docx/parse_document.py
@@ -1608,6 +1608,44 @@ def extract_docx_blocks(
             # Check if this is a heading using the new function
             outline_level = get_heading_level(element, styles_outline)
 
+            # A "heading" longer than MAX_HEADING_LENGTH is not a real heading.
+            # The common cause (WPS/Word): the author set an outline level on a
+            # paragraph but typed the body with soft line breaks (Shift+Enter →
+            # <w:br/> → '\n') instead of starting a new paragraph, so heading
+            # text + body live in one <w:p>. Split at the first soft break: the
+            # first line stays the heading, the remainder becomes body text. If
+            # there is no usable soft break (a genuine single-line over-long
+            # heading), demote the whole paragraph to body text. Either way we
+            # avoid crashing via validate_heading_length() and never drop content.
+            demoted_body_text = None
+            if outline_level is not None and len(full_text) > MAX_HEADING_LENGTH:
+                head, sep, rest = full_text.partition("\n")
+                if sep and len(head) <= MAX_HEADING_LENGTH:
+                    full_text = head
+                    demoted_body_text = rest.strip() or None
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3191, opened 2026-06-02
🐛 fix(pipeline): guarantee worker teardown + isolate enqueue get_by_id on backend outage

danielaskdd:
## Description

Fixes a `busy`-flag desync observed after a Redis (KV/doc_status) backend outage during active pipeline processing: the WebUI's flashing red "pipeline busy" indicator went off (looked stopped), yet the `/documents/pipeline_status` dialog kept showing `history_messages` updating and the backend kept processing remaining files.

This is **not** a `/health` bug. With a single workspace, `/health` and `/documents/pipeline_status` read the same shared `pipeline_status` Manager dict, and the Manager IPC was healthy (the status-card lock view kept updating). `/health` faithfully reported a `busy` value that had been cleared **too early**.

## Root Cause

The pipeline's storage-exception handling only covered the **per-document worker path** (parse/merge failures → `_finalize_doc_failure` marks the doc FAILED and swallows, `busy` stays True). The **orchestrator path** was unguarded:

1. `_run_pipeline_batch` creates the parse/analyze/process worker tasks, then calls `await self.full_docs.get_by_id(doc_id)` (engine routing — a KV/Redis read) inside the enqueue loop, and only cancelled the workers **after** all queue joins succeed. There was **no `try/finally`** around the worker lifecycle.
2. During the outage, that `get_by_id` raised `ConnectionError/RetryError`. The exception escaped `_run_pipeline_batch` (workers never cancelled → **orphaned**), propagated past the un-wrapped call site in `apipeline_process_enqueue_documents`, and hit the outer `finally`, which unconditionally set `pipeline_status["busy"] = False`.
3. The orphaned workers kept draining the already-enqueued docs and appending to `history_messages` — and, after Redis recovered, finished the work — while `busy=False`. Secondary risk: a later auto-scan/enqueue could see `busy=False` and start a **second** pass alongside the orphans.

## Changes Made

`lightrag/pipeline.py` — `_run_pipeline_batch`:
- **Wrap the enqueue loop + the three queue `join()`s in `try`, with `worker.cancel()` + `await gather(..., return_exceptions=True)` in `finally`.** Workers are always reclaimed, so no orphans and no `busy` desync.
- **Per-document isolation of the enqueue-time `get_by_id`.** A transient/corrupt single doc is marked FAILED (via `_finalize_doc_failure`) and the batch **continues to completion**, so in-flight documents reach terminal states instead of being stranded in PARSING/ANALYZING by an abrupt whole-batch abort. (A whole-batch-abort variant was tried and reverted — see Notes — because in production it stranded in-flight docs and perturbed teardown.)
- The internal-error abort still discards not-yet-flushed shared index buffers (`_discard_pending_index_ops`) after workers are stopped, so stale KG/vector records are not carried into the next batch.

`lightrag/pipeline.py` — `_finalize_doc_failure`:
- **Label internal-error cancels correctly in `doc_status`.** In-flight merge workers raise a hardcoded `"User cancelled during <stage>"` on a cooperative cancel; the doc's `error_msg` recorded that verbatim even when the batch was aborted by an **internal error**. Swap the generic `"User cancelled"` prefix for the reason-aware `_cancellation_label`, so `doc_status` records `"Cancelled by internal error: <detail> during <stage>"` while a genuine user cancel still reads `"User cancelled during <stage>"`.

`tests/pipeline/test_pipeline_orchestrator_get_by_id_failure.py` (new):
- single-doc enqueue `get_by_id` failure is isolated (bad → FAILED, good → PROCESSED), `busy` released, **no orphaned worker tasks**;
- full-outage escape still cancels all workers and releases `busy`.

`tests/pipeline/test_pipeline_internal_abort.py`:
- two `_finalize_doc_failure` cases asserting the `doc_status` `error_msg` label for internal-error vs user-cancel.

## Notes

A review-bot suggestion to (a) drop per-doc isolation in favor of a clean whole-batch abort and (b) discard buffers from inside the `finally` on any exception escape was implemented and then **reverted after a production retest**: removing isolation stranded in-flight documents in PARSING/ANALYZING, and discarding inside the teardown `finally` made the abort path itself hit Redis (LLM-cache flush → tenacity retry) during the very outage that triggered it. The bot's underlying concerns (stale buffers on a hard escape; the same unguarded read in `_validate_and_fix_document_consistency` running first) are valid but are deferred as separate hardening.

Out of scope (deliberate): `get_docs_by_statuses` at the loop top runs *between* batches with no live workers, so an escape there is a clean stop. This repo already has the `_atomic_release_busy_or_consume_pending` / `busy_released_in_loop` loop-exit-race fix (commit 8d46b0c3).

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Additional Notes

Verification: `tests/pipeline/` → **207 passed**; `ruff check` clean. Production retest on commit `6fbebf47` (the state this PR is now restored to) confirmed clean teardown — all workers cancelled, no stranded docs.

### DIFF
--- a/lightrag/pipeline.py
+++ b/lightrag/pipeline.py
@@ -1275,38 +1275,18 @@ class _PipelineMixin:
         # ``busy`` — leaving busy=False while processing visibly continues.
         batch_aborted_by_exception = False
         try:
-            # Add pending files to the correct parsing queue
-            for current_file_number, (doc_id, status_doc) in enumerate(
-                to_process_docs.items(), start=1
-            ):
-                file_path = getattr(status_doc, "file_path", "unknown_source")
-                # Per-document isolation: the engine-routing get_by_id is the only
-                # orchestrator-level storage read in this loop. A transient/corrupt
-                # single-doc failure must FAIL just that document and continue with
-                # the rest of the batch — not escape and abort the whole batch.
-                # During a full outage _finalize_doc_failure's own doc_status write
-                # also raises; that escape is caught by the finally below (workers
-                # are cleanly cancelled) and the batch aborts as a whole.
-                try:
-                    content_data = await self.full_docs.get_by_id(doc_id) or {}
-                except Exception as e:
-                    await self._finalize_doc_failure(
-                        doc_id=doc_id,
-                        status_doc=status_doc,
-                        file_path=file_path,
-                        error=e,
-                        stage_label="parse",
-                        current_file_number=current_file_number,
-                        total_files=total_files,
-                        failed_chunks_snapshot=([], 0),
-                        pending_tasks=[],
-                        metadata_extra={},
-                        pipeline_status=pipeline_status,
-                        pipeline_status_lock=pipeline_status_lock,
-                    )
-                    continue
+            # Add pending files to the correct parsing queue. An orchestrator-
+            # level storage read here (e.g. full_docs.get_by_id during a backend
+            # outage) that raises aborts the whole batch — the finally below
+            # cancels the workers cleanly and the caller releases ``busy``, with
+            # affected documents left queued (PENDING) for the next run. We do
+            # NOT per-doc isolate this read: the same unguarded read also runs in
+            # _validate_and_fix_document_consistency before this batch, so a clean
+            # whole-batch abort is the consistent behavior across both sites.
+            for doc_id, status_doc in to_process_docs.items():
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex revew

danielaskdd:
@codex review

PR #3206, opened 2026-06-05
fix: honor Postgres search path in table checks

he-yufeng:
## Summary
- make PostgreSQL table existence checks follow the active `search_path` with `to_regclass`
- reuse that helper when creating the full entity/relation tables instead of checking only `public`
- add focused coverage for the search-path-aware table check

## To verify
- `python -m pytest tests\kg\postgres_impl\test_postgres_search_path_table_exists.py -q`
- `python -m pytest tests\kg\postgres_impl\test_postgres_client_manager.py tests\kg\postgres_impl\test_postgres_migration.py -q`
- `python -m py_compile lightrag\kg\postgres_impl.py tests\kg\postgres_impl\test_postgres_search_path_table_exists.py`
- `git diff --check`


danielaskdd:
Thanks for sharing.

PR #3212, opened 2026-06-06
✨ feat(chunker): glue body-less heading blocks into deeper child (P strategy)

danielaskdd:
## Description

Adds a pre-Stage-D pass to the **Paragraph Semantic (P) chunking strategy** that handles **body-less heading blocks** — section headings that carry no body of their own (e.g. `## 2.4` with no text under it before its first subsection).

Previously such a heading became an independent same-level small block and was absorbed **backward into the tail of an unrelated same-level sibling** (`## 2.3`) by Stage D's peer merging / tail absorption — separating the parent heading from its real content (`### 2.4.1`).

The new pass glues a body-less heading **forward into its first strictly-deeper child** before Stage D runs, so the heading always travels with its content.

## Related Issues

N/A

## Changes Made

- `lightrag/chunker/paragraph_semantic.py`
  - `_is_heading_only(block)` — detects a block whose `content` consists solely of markdown heading lines (`^#{1,6} +`), guarded by a non-empty `heading`.
  - `_glue_heading_only_blocks(blocks, ...)` — runs between Stage B/C and Stage D:
    - **Forward-only**: when the next block is **strictly deeper** (greater `level`) and not a protected table slice, merge the heading forward into that child with `keep="left"` (keeping the shallower parent identity / level). Re-evaluated in place so a chain of bare ancestor headings (`# 2` → `## 2.4` → `### 2.4.1`) collapses into one block keeping the shallowest identity.
    - A heading whose next block is **not** deeper is left untouched for Stage D — it is deliberately **not** pulled backward into a deeper previous block (that would invert the hierarchy / demote the heading's level).
    - **Hard cap preserved**: if prepending the heading line(s) tips the bonded block past `chunk_token_size`, it is re-split via Stage C's `_split_long_block` (nothing downstream re-splits an oversized chunk).
    - Backfill into the previous chunk is left to Stage D's existing `target_ideal` / tail-absorption rules.
- `tests/chunker/test_paragraph_semantic_merge_and_fallback.py` — regression tests: detection, forward glue, chain collapse, cap-respecting re-split, no-glue when next not deeper, no backward glue into deeper/same-level previous, table-role protection.
- `docs/ParagraphSemanticChunking.md` / `docs/ParagraphSemanticChunking-zh.md` — new §8.5 plus pipeline-diagram updates (en + zh).

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (if necessary)
- [x] Unit tests added (if applicable)

## Additional Notes

[...]

PR #3214, opened 2026-06-06
✨ feat(chunker): glue body-less heading blocks into their deeper child (P strategy)

danielaskdd:
## Description

Adds a pre-Stage-D pass to the **Paragraph Semantic (P) chunking strategy** that handles **body-less heading blocks** — section headings with no body of their own (e.g. `## 2.4` with no text under it before its first subsection).

Previously such a heading became an independent same-level small block and was absorbed **backward into the tail of an unrelated same-level sibling** (`## 2.3`) by Stage D's peer merging / tail absorption — separating the parent heading from its real content (`### 2.4.1`).

The new pass glues a body-less heading **forward into its first strictly-deeper child** before Stage D runs, so the heading always travels with its content.

> Note: this is a clean single-commit reimplementation that supersedes #3212 (which carried a long iterative review history). Final behaviour is identical to where #3212 converged, with a simpler, fully-tested cap-split path.

## Related Issues

N/A

## Changes Made

- `lightrag/chunker/paragraph_semantic.py`
  - `_is_heading_only(block)` — detects a block whose `content` is solely markdown heading lines (`^#{1,6} +`), guarded by a non-empty `heading`.
  - `_glue_heading_only_blocks(blocks, ...)` — runs between Stage B/C and Stage D:
    - **Forward-only glue**: when the next block is **strictly deeper** (greater `level`) and its `table_chunk_role` is `none` or `first`, merge the heading forward into that child with `keep="left"` (shallower parent identity / level). Gluing into a `first` table slice keeps the `first` role so the table boundary stays protected from backward absorption. A chain of bare ancestor headings (`# 2` → `## 2.4` → `### 2.4.1`) collapses to the shallowest identity.
    - A heading whose next block is **not** deeper is left for Stage D — never pulled backward into a deeper previous block (that would invert the hierarchy / demote the heading's level).
    - **Hard cap** (`_split_to_cap`): if prepending the heading tips the bonded block past `chunk_token_size`, peel the heading prefix, split **only the body** at the full cap (so later body pieces keep the full budget), and glue the prefix back onto the first piece — the heading is never handed to the splitter, so it can't be sliced off as a heading-only orphan that Stage D would re-absorb. When the prefix alone exceeds the cap (very long title / tiny chunk size), the whole block is split directly and the oversized heading line is character-split (cap wins over heading-intactness).
    - Backfill into the previous chunk is left to Stage D's existing `target_ideal` / tail-absorption rules.
- `tests/chunker/test_paragraph_semantic_merge_and_fallback.py` — regression tests: detection, forward glue, chain collapse, no-glue when next not deeper / into deeper or same-level previous / into middle-or-last table slice, glue into a `first` table slice (role preserved), and three cap-split cases (no-anchor orphan avoidance, no over-fragmentation of later body chunks, prefix-larger-than-cap).
- `docs/ParagraphSemanticChunking.md` / `docs/ParagraphSemanticChunking-zh.md` — new §8.5 plus pipeline-diagram updates (en + zh).

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

- Offline tests: `pytest tests/chunker/ -m offline -q` → **163 passed, 19 deselected**; `ruff check` clean.
- Scoped to the P strategy only; no Stage D thresholds were changed.
- Edge note: a body line that genuinely begins with `#` + space is treated as a heading line — the same accepted heuristic ambiguity already documented in `lightrag/parser/_markdown.py`.


danielaskdd:
@codex review

PR #3215, opened 2026-06-06
✨ feat(chunker): land parent-path gate + bridge overlap for P strategy

danielaskdd:
## Description

Lands three Paragraph-Semantic (P) chunking features that `docs/ParagraphSemanticChunking-zh.md` documented but `lightrag/chunker/paragraph_semantic.py` never implemented, and corrects the doc passages that had drifted from the actual implementation. Originated from a doc-vs-code audit of the P strategy.

## Related Issues

N/A

## Changes Made

**Code (`lightrag/chunker/paragraph_semantic.py`)**

- **Parent-path consistency constraint (§9.1 #4)** — `_merge_small_blocks` previously merged purely on `level` + `table_chunk_role` + size, so two same-level sections under different parents could be fused (cross-topic pollution). Added `_same_parent_path` / `_is_descendant` helpers and gated all five merge points:
  - Phase A (forward/backward) and tail-absorption run: require **identical** `parent_headings` (true siblings).
  - Phase B (shallow-absorbs-deep, forward/backward): require the deep block to be a **descendant** of the shallow one (shallow's full heading path is a prefix of the deep block's `parent_headings`).
  - Empty parent paths (preamble / non-hierarchical) compare as compatible.
- **Bridge single-side budget ≤ `chunk_token_size/2`** — `_context_capacity` now also clamps each side to `target_max // 2` so duplicated bridge text can't dominate a table boundary block.
- **Bridge middle-block overlap** — the standalone middle text block now keeps R-style overlap with the prefix/suffix copied into the neighbouring table blocks. Overlap indices stay inside the bridge tokens, so `<table>` markup is never dragged into the middle block.

**Docs (`docs/ParagraphSemanticChunking-zh.md`)**

- §3.3 output schema → nested `heading` dict + optional `sidecar`; internal-field list corrected (`blockids`, not `uuid/uuid_end/type`).
- §3.2 documents the `blockid` input field.
- §2 pipeline diagram moves `[part n]` numbering before Stage D (Stage C.1), matching its per-row timing.
- §9.2 merge-condition wording tightened; §9.1 #4 refined to the implemented strict semantics; §7 documents the middle-block overlap.

**Tests (`tests/chunker/`)**

- 4 parent-path gate cases (same-level equal/unequal, cross-level descendant/non-descendant) + 1 tail-absorption-stops-at-divergent-parent case.
- 1 single-side `/2` cap case.
- Updated the long-bridge middle-text case to assert the middle block carries the full bridge and excludes `<table>` markup.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

`tests/chunker/` passes (169 passed, 19 deselected) with `-m offline`; no real docx/embedding needed. The middle-block-overlap behavior is intentionally maximal: because each side's prefix/suffix is itself ≤ the overlap budget, the middle block ends up carrying the **complete** bridge (head/tail also duplicated into the adjacent tables), so the bridge text is never fragmented across blocks.


danielaskdd:
@codex review

PR #3225, opened 2026-06-08
feat(extract): inject section heading breadcrumb into extraction prompts

danielaskdd:
## What

Entity/relation extraction now receives the **section heading hierarchy** of each text chunk as background context.

Each chunk already stores its heading info (`heading: {level, heading, parent_headings}`), but `_process_single_content` only ever passed the cleaned body text as `{input_text}`. For papers, the section path (e.g. `Methods → Data Collection`) helps the LLM disambiguate entities and ground relationship descriptions in the right context.

## How

- **New helper** `format_heading_context(dp)` (`chunk_schema.py`): reuses `normalize_chunk_heading` / `_clean_heading_text`, returns the **full breadcrumb** (parent chain **+** the chunk's own heading), or `""` when absent.
- **New prompt template** `entity_extraction_section_context`: the `---Section Context---` marker lives **only here**. Extraction code never hardcodes the marker — it just produces the breadcrumb and decides whether to inject the block. Changing the section title/wording is now a prompt-only edit.
- **User prompts** (text + JSON) gain a `{heading_context_block}` placeholder right above `---Input Text---`.
- **System prompts** (text + JSON) gain a static, conditionally-worded instruction: use the Section Context *only as background*, never extract entities from the heading text itself. Static wording keeps the system prompt cache-friendly.
- Continue-extraction prompts are untouched — they run with conversation history that already carries the first user prompt.

## Key guarantee: zero noise when a chunk has no heading

When `format_heading_context` returns `""`, the wrapper template is never invoked, `heading_context_block = ""`, and the rendered user prompt is **byte-identical** to the prior no-context form. The `---Section Context---` line never appears.

## Length bounding (addresses Codex review)

The breadcrumb is metadata layered on top of the already-chunk-sized input text, so an unbounded heading chain could push an otherwise-valid chunk past the provider context window. Two bounds, with shared constants in `constants.py`:

- **Per-level char cap** — `format_heading_context` truncates each heading level at `DEFAULT_HEADING_LEVEL_MAX_CHARS` (80) with an ellipsis, so one runaway title can't dominate the prompt.
- **Overall token budget** — `_process_single_content` token-budgets the joined breadcrumb via `_truncate_section_context` against `DEFAULT_MAX_SECTION_CONTEXT_TOKENS` (256). Over budget it collapses to `first → … → leaf` (top-level document location + nearest/most-specific section), eliding the middle.
- **Invariant** — the per-level cap is kept below 1/3 of the token budget so the collapsed two-level form always fits (guarded by a test). The gleaning guard already counts the block through `history`.

## Tests

15 offline tests in `tests/extraction/test_entity_extraction_stability.py`:
- `format_heading_context` full-path vs empty; per-level cap (and disable).
- No-heading → marker absent **and** rendered prompt byte-identical to the placeholder-removed (pre-change) template — text & JSON.
- Heading present → block rendered above the input text — text & JSON.
- Both system prompts carry the instruction.
- `_truncate_section_context`: no-op within budget; over-budget keeps first + leaf and drops the middle; 2-level path returned intact; disabled / no-tokenizer no-op; the 1/3 budget invariant.
- End-to-end via `extract_entities`: chunk with heading injects the breadcrumb; heading-free chunk does not; a pathological 500-char heading never reaches the prompt verbatim.

`tests/extraction` + `tests/sidecar`: **133 passed**. `pre-commit run ruff-format` / `ruff`: clean.

### DIFF
--- a/lightrag/chunk_schema.py
+++ b/lightrag/chunk_schema.py
@@ -26,8 +26,15 @@ from __future__ import annotations
 import re
 from typing import Any
 
+from lightrag.constants import DEFAULT_HEADING_LEVEL_MAX_CHARS
+
 
 _SIDECAR_TYPES = frozenset({"block", "drawing", "table", "equation"})
+
+# Separator joining heading levels into a single breadcrumb line. Shared so the
+# extraction-side token budgeter (see ``operate._truncate_section_context``) can
+# split the rendered breadcrumb back into levels without a drifting magic string.
+HEADING_BREADCRUMB_SEP = " → "
 
 
 def normalize_chunk_heading(dp: dict[str, Any]) -> dict[str, Any] | None:
@@ -118,14 +125,7 @@ def format_parent_headings(dp: dict[str, Any]) -> str:
     cleaned = [
         c for c in (_clean_heading_text(h) for h in normalized["parent_headings"]) if c
     ]
-    return " → ".join(cleaned)
-
-
-# Per-level character cap for headings rendered into the extraction
-# `---Section Context---` block. Bounds a single pathologically long section
-# title so it cannot dominate the prompt. The overall breadcrumb is
-# additionally token-budgeted by the caller (see ``operate.py``).
-DEFAULT_HEADING_LEVEL_MAX_CHARS = 120
+    return HEADING_BREADCRUMB_SEP.join(cleaned)
 
 
 def _truncate_heading_level(text: str, max_chars: int) -> str:
@@ -166,7 +166,7 @@ def format_heading_context(
         for c in (_clean_heading_text(h) for h in chain)
         if c
     ]
-    return " → ".join(cleaned)
+    return HEADING_BREADCRUMB_SEP.join(cleaned)
 
 
 def normalize_chunk_sidecar(dp: dict[str, Any]) -> dict[str, Any] | None:
@@ -348,6 +348,7 @@ __all__ = [
     "normalize_chunk_heading",
     "format_parent_headings",
     "format_heading_context",
+    "HEADING_BREADCRUMB_SEP",
     "normalize_chunk_sidecar",
     "strip_internal_multimodal_markup_for_extraction",
 ]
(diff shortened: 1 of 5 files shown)


danielaskdd:
@codex review

danielaskdd:
Addressed in 286d2b33 — the heading breadcrumb is now bounded on two levels:

- **Per-level char cap**: `format_heading_context` truncates each heading level at `DEFAULT_HEADING_LEVEL_MAX_CHARS` (120) with an ellipsis, so one runaway section title can't dominate the prompt.
- **Overall token budget**: `_process_single_content` token-budgets the joined breadcrumb via `_truncate_section_context` against `DEFAULT_MAX_SECTION_CONTEXT_TOKENS` (256), keeping the nearest (tail) section on truncation. The existing gleaning guard already counts the block through `history`.

So heading metadata can no longer push an otherwise-valid chunk past the provider context window. Added unit tests for both caps + an e2e test asserting a pathological 500-char heading never reaches the prompt verbatim.

danielaskdd:
@codex review

danielaskdd:
Both new findings addressed in cf5bfd64.

**Prompt injection (prompt.py)** — the breadcrumb is collapsed to a single line upstream (`_clean_heading_text` folds all whitespace incl. newlines), and the template now renders it *after* a descriptive label on the same line:

```
---Section Context---
The input text is an excerpt from this document section path (reference only): <breadcrumb>
```

Structural prompt markers (`---X---` sections, ``` fences) are line-start constructs, so a heading such as `---Output---` lands mid-line as inert data and can no longer forge a prompt section outside the input fence. Marker/wrapping still lives only in the prompt template. Added `test_section_context_breadcrumb_is_not_at_line_start`.

**Token cap bypass (operate.py)** — `_truncate_section_context` no longer returns 1–2 level paths unchanged. After the optional `first → … → leaf` collapse it always applies a hard token-truncation backstop, so token-dense / byte-level tokenizers can't push the breadcrumb past `DEFAULT_MAX_SECTION_CONTEXT_TOKENS`. Added tests asserting the result token count is always `<= max_tokens` for dense short paths and for a still-over-budget collapsed form.

`tests/extraction` + `tests/sidecar`: 135 passed; ruff clean.

danielaskdd:
@codex review

PR #3245, opened 2026-06-11
fix: prevent deadlock in sync wrappers and clean up zombie chunks on extraction failure

### LINKED ISSUE #1968: [Bug]:Critical Issues in Concurrent Insertion and Persistent State Management

### **Bug Report: Critical Issues in Concurrent Insertion and Persistent State Management**

**Project**: LightRAG
**Version**: lightrag-hku 1.4.6

**Dear LightRAG Development Team,**

First, thank you for creating and sharing the LightRAG project. The concept of building a knowledge graph-based RAG system is highly valuable, and I've been exploring its capabilities for a research project.

During the integration process, I encountered several critical issues related to the document insertion pipeline (`ainsert` method) that currently prevent robust and reliable use, especially in scenarios involving batch processing or error recovery. This report details these issues with the hope that it can help improve the framework.

---

#### **1. Core Issue: Race Condition in Concurrent `ainsert` Leading to `Document content not found`**

When attempting to insert multiple documents (or text chunks) by passing a list to the `ainsert` method, the internal concurrent processing mechanism appears to have a race condition. This consistently leads to a `Document content not found` exception, followed by an `UnboundLocalError` due to secondary issues in exception handling.

**How to Reproduce:**
The issue can be reliably reproduced in a clean environment with the following minimal setup:

**Setup Code:**
```python
import asyncio
import shutil
from lightrag import LightRAG
from lightrag.kg.shared_storage import initialize_pipeline_status

# Assume llm_model_func and embedding_func are configured
# ...

async def run_minimal_test():
    """A minimal test case to demonstrate the concurrent insertion bug."""
    
    WORKING_DIR = "./lightrag_test_dir"
    if os.path.exists(WORKING_DIR):
        shutil.rmtree(WORKING_DIR) # Ensure a clean environment

    rag = LightRAG(
        working_dir=WORKING_DIR,
        embedding_func=embedding_func,
        # ... other necessary configs
    )

    # Test Case 1: Single insertion (This usually works)
    try:
        await rag.ainsert("This is a single document.")
        print("[SUCCESS] Single document insertion completed.")
    except Exception as e:
        print(f"[FAIL] Single document insertion failed: {e}")

    # Test Case 2: Concurrent insertion (This consistently fails)
    try:
        documents = [
            "This is the first document in a batch.",
            "This is the second document in a batch.",
            "This is the third document in a batch."
        ]
        await rag.ainsert(documents)
        print("[SUCCESS] Concurrent document insertion completed.")
    except Exception as e:
        print(f"[FAIL] Concurrent document insertion failed: {e}")

# asyncio.run(run_minimal_test())
```

**Observed Behavior & Error Log:**
The single insertion succeeds, but the concurrent insertion fails with the following traceback:

```
[FAIL] Concurrent document insertion failed: local variable 'first_stage_tasks' referenced before assignment
```
And the logs show the preceding exceptions:
```
Exception: Document content not found in full_docs for doc_id: ...
...
UnboundLocalError: local variable 'first_stage_tasks' referenced before assignment
```

**Diagnosis:**
This strongly suggests that when multiple `process_document` tasks are launched concurrently via `asyncio.gather`, the tasks that are supposed to store the document content in the internal `full_docs` storage do not complete before the worker tasks attempt to retrieve that content.

---

#### **2. Compounding Issue: Persistent State Pollution from Failed Tasks**

The above issue is severely compounded by a state management problem. When an insertion task fails, it leaves a "zombie" entry in the persistent state files within the `working_dir` (likely in `doc_status_storage.json` and potentially other files).

This "polluted" state is reloaded upon the next initialization of the `LightRAG` instance. Consequently, any subsequent `ainsert` call—even for a single, valid document—will be contaminated by these zombie tasks. The system will attempt to re-process the failed tasks alongside the new ones, leading to an immediate and unavoidable repeat of the `Document content not found` error.

**How to Reproduce:**
1. Run the minimal test from Issue #1. This will create a `working_dir` with polluted state.
2. **Do not delete** the `working_dir`.
3. Rerun the script, but this time **only** execute the single document insertion test case.

**Observed Behavior:**
Even the previously successful single document insertion will now fail. The logs will show that `LightRAG` is attempting to process multiple documents (e.g., `Failed to extract document 1/2`), one of which is the new document, and the other(s) are the zombie tasks from the previous failed run.

**Diagnosis:**
This indicates that the framework lacks a robust mechanism for cleaning up failed tasks from its persistent queue. This makes the system highly brittle, as a single transient failure (e.g., a network error during an LLM call) can render the entire `working_dir` permanently corrupted, preventing any further documents from being inserted.

---

#### **3. Architectural Issue: `insert()` Method Deadlocks in Async Environments**

In an attempt to work around the concurrency bug (Issue #1), I tried to force sequential processing by calling the synchronous `insert()` method for each document within an async loop, using `loop.run_in_executor`.

However, an analysis of the traceback revealed that the `insert()` method is a "fake sync" wrapper that internally calls `loop.run_until_complete(self.ainsert(...))`.

**Code Snippet of the Flawed Workaround:**
```python
# This code causes a deadlock
loop = asyncio.get_running_loop()
for chunk in text_chunks:
    await loop.run_in_executor(None, rag.insert, chunk)
```

**Observed Behavior:**
The program deadlocks.

**Diagnosis:**
This design leads to a classic deadlock in an async environment. The main event loop is waiting for the executor thread to finish, while the executor thread is trying to block and take over the main event loop. The synchronous `insert` method, in its current form, cannot be safely used from within an already running asyncio event loop.

---

### **Summary & Suggested Solutions**

The combination of these three issues makes the current document insertion pipeline very fragile. To build a robust application, a workaround is needed. My current stable solution involves:
1.  **Forcibly deleting the entire `working_dir` before every batch insertion process.** This is the only reliable way to clear the state pollution. This, however, comes at the significant cost of losing all LLM and embedding caches.
2.  **Submitting documents one by one using `await rag.ainsert(single_document)` in a `for` loop.** This manually enforces sequential processing and bypasses the faulty concurrency logic.

It would be a great improvement for the framework if:
*   The race condition in the concurrent `ainsert` pipeline could be resolved, perhaps by ensuring document content is fully stored before worker tasks are dispatched.
*   A mechanism to safely purge or ignore failed tasks from the persistent state could be introduced, preventing the "zombie task" pollution.
*   The synchronous `insert()` method could be re-implemented to be truly synchronous (i.e., using synchronous I/O) or removed to avoid the deadlock trap.

Thank you for your time and consideration. I believe addressing these stability issues would greatly benefit the LightRAG community. I am happy to provide any further information or testing assistance.

Best regards,

mytiome

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### LightRAG Config Used

# Paste your config here

### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version:1.4.6
- Operating System:
- Python Version:3.10.14
- Related Issues:

returnSGD:
## Summary

Fixes #1968 — three related issues causing concurrent insert unreliability:

### 1. Sync deadlock protection (main fix)

All 12 sync wrapper methods (`insert`, `query`, `query_data`, `query_llm`, `delete_by_entity`, `delete_by_relation`, `edit_entity`, `edit_relation`, `create_entity`, `create_relation`, `merge_entities`, `insert_custom_kg`, `insert_custom_chunks`) now call `_check_sync_in_async_context()` before `loop.run_until_complete()`. 

When a sync wrapper is called from within a running asyncio event loop (e.g. inside a FastAPI handler, or any `async def` function), it now raises a clear `RuntimeError`:

```
RuntimeError: insert() cannot be called from within a running async event loop
(this would deadlock). Use `await ainsert(...)` instead.
```

This replaces the silent deadlock that previously occurred when `loop.run_until_complete()` was called on an already-running loop.

### 2. Zombie chunk cleanup on extraction failure

When entity/relation extraction fails after chunks have already been upserted to `chunks_vdb` and `text_chunks` (the `asyncio.gather` at the start of the extraction stage succeeds, but the subsequent extraction fails), orphaned chunks are now cleaned up:
- Deletes chunks from `chunks_vdb` and `text_chunks` (best-effort, logged on failure)
- Clears `chunks_list` / `chunks_count` on the status_doc so re-processing starts clean
- Chunk IDs are doc-scoped (`doc_id-` prefix) so this does not affect other documents

### 3. Defensive first_stage_tasks initialization

The `first_stage_tasks` variable in `process_single_document` is already initialized at function scope before the `async with ctx.semaphore:` block. This prevents `UnboundLocalError` in the exception handler when an error occurs before task creation — a race condition that could previously occur under concurrent inserts.

## Test plan

- [x] `_check_sync_in_async_context` raises correctly when called from async context
- [x] `_check_sync_in_async_context` does NOT raise when no loop is running
- [x] Syntax check passes on both modified files
- [x] Integration test: concurrent `ainsert` calls with extraction failures
- [x] Manual test: call `insert()` from inside a FastAPI endpoint

## Files changed

- `lightrag/lightrag.py` — +54 lines (helper function + 12 guard calls)
- `lightrag/pipeline.py` — +35 lines (chunk cleanup on extraction failure)

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -29,6 +29,35 @@ from typing import (
     Dict,
     Union,
 )
+def _check_sync_in_async_context(
+    sync_name: str, async_name: str
+) -> None:
+    """Raise a clear error when a sync wrapper is called from within
+    a running asyncio event loop, which would otherwise deadlock.
+
+    All sync wrappers (``insert``, ``query``, ``delete_by_entity``,
+    etc.) call ``loop.run_until_complete()`` internally.  When the
+    caller is already inside an async function — e.g. a FastAPI
+    handler — this blocks the running loop and causes a deadlock that
+    is silent until the request times out.
+
+    ``asyncio.get_running_loop()`` (Python ≥ 3.7) raises
+    ``RuntimeError`` when *no* loop is running, so the presence of a
+    running loop means the caller is in an async context that would
+    deadlock.  The error message points directly to the correct async
+    alternative.
+    """
+    try:
+        asyncio.get_running_loop()
+    except RuntimeError:
+        return  # No running loop — safe to call sync wrapper
+    raise RuntimeError(
+        f"{sync_name}() cannot be called from within a running "
+        f"async event loop (this would deadlock). "
+        f"Use `await {async_name}(...)` instead."
+    )
+
+
 from lightrag.prompt import (
     PROMPTS,
     get_default_entity_extraction_prompt_profile,
@@ -1295,7 +1324,12 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
 
         Returns:
             str: tracking ID for monitoring processing status
+
+        Raises:
+            RuntimeError: if called from within a running async event loop.
+                Use ``ainsert()`` instead.
         """
+        _check_sync_in_async_context("insert", "ainsert")
         loop = always_get_an_event_loop()
         return loop.run_until_complete(
             self.ainsert(
@@ -1384,6 +1418,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
         text_chunks: list[str],
         doc_id: str | list[str] | None = None,
     ) -> None:
+        _check_sync_in_async_context("insert_custom_chunks", "ainsert_custom_chunks")
         loop = always_get_an_event_loop()
         loop.run_until_complete(
             self.ainsert_custom_chunks(full_text, text_chunks, doc_id)
@@ -1657,6 +1692,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
     def insert_custom_kg(
         self, custom_kg: dict[str, Any], full_doc_id: str = None
     ) -> None:
+        _check_sync_in_async_context("insert_custom_kg", "ainsert_custom_kg")
         loop = always_get_an_event_loop()
         loop.run_until_complete(self.ainsert_custom_kg(custom_kg, full_doc_id))
 
@@ -1944,6 +1980,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
         Returns:
             str: The result of the query execution.
         """
+        _check_sync_in_async_context("query", "aquery")
         loop = always_get_an_event_loop()
 
         return loop.run_until_complete(self.aquery(query, param, system_prompt))  # type: ignore
@@ -1999,6 +2036,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
         Returns:
             dict[str, Any]: Same structured data result as aquery_data.
         """
+        _check_sync_in_async_context("query_data", "aquery_data")
         loop = always_get_an_event_loop()
         return loop.run_until_complete(self.aquery_data(query, param))
 
@@ -2369,6 +2407,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
         Returns:
             dict[str, Any]: Same complete response format as aquery_llm.
         """
+        _check_sync_in_async_context("query_llm", "aquery_llm")
         loop = always_get_an_event_loop()
         return loop.run_until_complete(self.aquery_llm(query, param, system_prompt))
 
@@ -3908,7 +3947,12 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
 
         Returns:
             DeletionResult: An object containing the outcome of the deletion process.
+
+        Raises:
+            RuntimeError: if called from within a running async event loop.
+                Use ``adelete_by_entity()`` instead.
         """
+        _check_sync_in_async_context("delete_by_entity", "adelete_by_entity")
         loop = always_get_an_event_loop()
         return loop.run_until_complete(self.adelete_by_entity(entity_name))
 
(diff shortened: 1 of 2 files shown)


danielaskdd:
Thanks for digging into #1968 and putting this together — the instinct to make the synchronous wrappers fail clearly instead of mysteriously is a good one. After reviewing it against `main`, though, we're going to pass on merging this PR as-is, for a few reasons:

**1. The "deadlock" framing is misleading.** Calling a sync wrapper from within a running loop on the same thread does **not** silently deadlock — `loop.run_until_complete()` raises `RuntimeError: This event loop is already running` immediately. The genuinely problematic case (`run_in_executor`) is different: the coroutine runs on a *different* loop than the storages were initialized on, and the shared `asyncio.Lock` objects in `lightrag.kg.shared_storage` bind to the first loop that touches them, so the second loop raises "bound to a different event loop" / stalls. Both cases share the same fix (use the `a*` coroutine), but neither is a deadlock — so the guard message points users in a slightly wrong direction.

**2. The zombie-chunk cleanup is redundant.** Orphaned chunks from a failed extraction are already cleaned up on both relevant paths, and more completely (they also purge KG entries / source-ids):
- on retry: `_purge_stale_extraction_if_resuming` (`pipeline.py`)
- on delete: `adelete_by_doc_id` reads `chunks_list` and deletes from `chunks_vdb` / `text_chunks` (`lightrag.py`)

Also, the added `status_doc.chunks_list = []` only mutates the in-memory object, which is discarded — `_finalize_doc_failure` has already persisted the FAILED row *with* those chunk IDs, and the next run reloads from storage, so the clear has no effect on re-processing.

**3. The `first_stage_tasks` part is a no-op.** It's already initialized at function scope in `process_single_document` (`pipeline.py`), so there's no code change there — the `UnboundLocalError` from #1968 is already fixed on `main`.

**Minor:** the `_check_sync_in_async_context` helper is defined *between* import statements, which trips ruff `E402` (module-level import not at top of file).

For the part that does have value — the sync-interface guard — we opened #3271 as an alternative: it centralizes the loop handling into one `_run_sync()` helper (instead of 12 scattered guards), covers `clear_cache` and `export_data` too, keeps the wording accurate (immediate `RuntimeError`, not a deadlock), avoids the E402 issue, and adds regression tests.

Thanks again for the report and the effort here — it directly informed #3271.

---

PR #3253, opened 2026-06-12
feat(concurrency): make MAX_ASYNC a true cross-worker limit under gunicorn + aggregated queue stats

danielaskdd:
# Summary

Under gunicorn multi-worker, `priority_limit_async_func_call` enforces concurrency **per process**: with `MAX_ASYNC_LLM=4` and `--workers 4` the provider actually sees ~16 concurrent calls, and `/health` only reports whichever worker happened to answer (1/N view). This PR fixes both, reusing the existing `shared_storage` namespace + keyed-lock machinery (no Redis, no new Manager registries):

1. **MAX_ASYNC becomes a true total limit.** Under `lightrag-gunicorn --workers N`, every existing `MAX_ASYNC_*` setting (per LLM role / embedding / rerank) now acts BOTH as each worker's local limit and as a cross-worker global cap — the provider never sees more than MAX_ASYNC concurrent calls in total, exactly like single-process mode. **No new env vars**: the gunicorn master derives the per-group caps (`llm:extract` / `llm:keyword` / `llm:query` / `llm:vlm` / `embedding` / `rerank`) from the already-parsed MAX_ASYNC values before fork. Crash self-healing (lease + heartbeat + reaper) and longest-waiter-first soft fairness across workers included.
2. **Near-realtime cross-worker queue stats**: each worker publishes debounced snapshots; `/health` aggregates all workers. Existing flat field names are preserved — **zero webui changes**.

Unchanged paths: single-worker mode (`workers=1`, uvicorn or gunicorn) passes no limits — the per-process max_async already IS the total there; standalone decorator usage (no `concurrency_group`) never touches shared storage; embedded library users are unaffected.

> ⚠️ **Behavior change for existing multi-worker deployments**: effective provider concurrency drops from ~`MAX_ASYNC × workers` to `MAX_ASYNC`. This restores MAX_ASYNC's documented meaning; deployments that relied on the multiplication should raise `MAX_ASYNC_*` accordingly. Runtime caveat (**gunicorn multi-worker only**, documented in env.example): changing a role's max_async through the API rebuilds only the receiving worker's local wrapper — the cross-worker cap keeps the value read at startup, so raising it cannot push the total above the startup cap, and lowering it only shrinks that one worker's share. In single-process mode (uvicorn or `--workers 1`) there is no separate cap: the wrapper's max_size IS the global limit, and an API update rebuilds the wrapper and takes full effect immediately.

# How the global gate works

- **One slot pump per process**: a single coroutine per process acquires cross-process slots and hands `(lease, task)` pairs to executor workers — not `max_size` polling workers — dividing the cross-process poll/IPC rate by `max_size`. A slot is requested only when there is BOTH physically queued work and an idle local worker to run it immediately, so the worker herd can never grab slots it cannot use (no `global_in_use` inflation, no waiter-seniority churn from no-op acquires, and scarce global capacity is never parked behind a busy process).
- **Slot-first, drain-second**: the pump acquires a slot *before* consuming the local queue. While slots are saturated, tasks stay queued and cancellable — user timeouts/cancellations never reach the provider, local priority order is preserved (query can still overtake ingestion within a process), and execution timing starts only at dispatch, so slot waits are never misjudged as stuck tasks.
- **Single-key gate state, O(1) IPC per attempt**: each group's whole gate state (leases + waiters) lives under ONE key of the Manager-dict namespace with whole-value replacement. An acquire attempt under the group's keyed lock costs exactly one proxy read plus at most one write — a failed attempt against a healthy, saturated gate writes nothing (covered by a write-counting test). Reaping, capacity counting and waiter ranking all run on the local copy.
- **Event loop never blocks on cross-process locks**: `UnifiedLock` now acquires the Manager lock proxy through the default executor, so waiting for another process's critical section no longer freezes this worker's event loop (HTTP serving, heartbeats and other queues keep running). Cancellation while parked in the executor is safe: an orphaned acquisition is released by a done-callback (otherwise every process would deadlock), and the per-process async gate is rolled back (previously a `CancelledError` could leak it).
- **Leases with heartbeat self-healing**: each held slot is a lease (`pid`, `updated_at`) renewed by the existing 5s health check. Dead owners (kill -9 / OOM) are reclaimed immediately via PID probe; heartbeat-expired-but-alive owners get a suspect grace before reclamation, so a momentarily stalled event loop or transient renewal failure never falsely reclaims a live long-running task.
- **Longest-waiter-first adaptive polling (soft FIFO across workers)**: on every failed attempt the pump registers/refreshes its process's waiter record (`wait_start`, `last_poll`) inside the same critical section and learns whether it is the longest-waiting *live* poller: the favored process keeps the fastest poll interval (0.05s), everyone else backs off exponentially up to a small bounded cap (0.4s). There is **no hard gate** — any poller that finds a free slot takes it — so no head-of-line blocking, and a freed slot idles at most ~one deferred period. A successful acquire resets the winner's seniority, yielding approximate round-robin under sustained contention. Ghost-waiter guardrails: waiter records are reaped **in the same pass as leases**, stale-poll records are ignored for ranking and reaped, wrapper shutdown clears its record, and the first acquisition attempt is always immediate.
- **Fail-closed**: any shared-storage error during acquisition returns "no slot" (rate-limited warning) — capacity is never exceeded due to infrastructure failures. Failed releases are parked for health-check retry; heartbeat expiry guarantees eventual reclamation either way.
- **Logical admission instead of a bounded queue** (limited groups only): the physical queue becomes unbounded and `max_queue_size` is enforced against `live_queued` — live queued tasks only, exactly matching the bounded-queue semantics of unlimited groups (running tasks and cancelled zombies don't consume capacity, `max_queue_size<=0` = unlimited). Reservations are released exactly once (`worker_started` transfer rule). Zombie tuples are drained in bounded batches while holding a slot, and a batched health-check compaction bounds the physical queue even during prolonged fail-closed periods — with exact `task_done()` bookkeeping so `queue.join()`-based graceful shutdown never wedges (shutdown also drains undelivered dispatch entries, returning their leases). Shutdown wakes admission sleepers so no request hangs forever.

# Observability

- Counter-update points publish debounced snapshots (0.1s min interval); the 5s health check force-flushes (also keeping snapshots ahead of the 15s stale TTL).
- New `get_aggregated_queue_stats()` on wrappers: publishes a fresh local snapshot, sums all live workers' flat counters, and adds `reporting_workers` / `per_worker` / `global_limit` / `global_in_use` / `global_slot_waits` / `global_waiting_workers` / `global_longest_wait`. Any failure falls back to the local snapshot — `available` keeps meaning "wrapper exists", an idle queue can never turn unavailable.
- In limited mode `queued` = live queued count (zombies excluded); `physical_queued` exposes the raw queue length for debugging.

# Configuration

None added. Under gunicorn multi-worker the existing settings double as the cross-worker totals:

```
EXTRACT_MAX_ASYNC_LLM / KEYWORD_… / QUERY_… / VLM_…  (fallback: MAX_ASYNC_LLM)   # llm:{role}
EMBEDDING_FUNC_MAX_ASYNC                                                          # embedding
MAX_ASYNC_RERANK                                                                  # rerank
```

`lightrag-gunicorn` assembles these into `initialize_share_data(workers, global_concurrency_limits=...)` in the master before fork (read-only module config inherited by workers; the library-internal no-arg `initialize_share_data()` call hits the already-initialized guard and never overwrites it). env.example documents the new semantics, the self-healing behavior, the admission-counting rule, and the runtime-update caveat.

# What's changed

- `lightrag/kg/shared_storage.py`: single-key per-group gate state (`{"leases": ..., "waiters": ...}`, whole-value replacement; write-only-when-changed); `try_acquire_global_slot` / `try_acquire_global_slot_tracked` / `release_global_slot` / `renew_global_slots` / `reconcile_global_slots` / `global_concurrency_in_use` / `clear_slot_waiter` / `global_slot_waiters`; `publish/unpublish/aggregate_queue_stats` (stale reaping re-checks `updated_at` under the internal lock); sync `is_global_concurrency_limited` (no IPC). `UnifiedLock`: executor-offloaded mp acquire with cancellation safety (orphaned-acquire release + async-gate rollback on `CancelledError`).
- `lightrag/utils.py`: `concurrency_group` parameter; mode resolved once at first `ensure_workers()` with lazy queue creation; `slot_pump` (single per-process acquirer: slot-first + bounded zombie drain + waiter-rank-driven adaptive backoff + idle-worker gating) dispatching to `limited_worker` executors; reservation-based admission; per-tick maintenance (renew / pending-release retry / reconcile / compaction / stats flush, each failure-isolated, exposed as `run_maintenance`); `get_aggregated_queue_stats`.
- `lightrag/llm_roles.py`: roles pass `concurrency_group="llm:{role}"`; queue status methods prefer the aggregated view with local fallback.
- `lightrag/lightrag.py`: embedding/rerank wrappers pass their group names.
- `lightrag/api/run_with_gunicorn.py`: derives the global limits from MAX_ASYNC (multi-worker only).
- `lightrag/constants.py`, `env.example`: gate tuning defaults + docs.

# Tests

45 offline tests (single-process shared storage exercises the same logical code paths as multi-worker):

- `tests/kg/test_global_concurrency_slots.py` (23): limit config first-init/no-overwrite/finalize-reset; acquire-to-capacity/release/idempotency/group isolation; dead-PID immediate reclaim; heartbeat-expiry suspect grace (live PID never falsely reclaimed; renewal clears suspect; no resurrection); fail-closed acquire; **zero-write contract for failed attempts on an unchanged gate**; waiter tracking (register on failure / clear on success, longest-live-waiter ranking, stale-poll records ignored and reaped, waiter records cleaned in the same reap pass as their process's timed-out lease, plain acquire never registers); stats aggregation sums/schema, dead-PID + stale reaping, unpublish.
- `tests/llm/test_global_concurrency_queue.py` (22): in-flight peak ≤ global limit (incl. two wrappers sharing one gate); unconfigured group keeps original path with acquire-never-called mock assertion; standalone (`group=None`) never touches shared storage; **pump acquires exactly once for a single task regardless of worker count**; **no slot hoarding while all local workers are busy**; user timeout during slot wait never calls the provider; in-process priority overtake; admission counts only live queued tasks; zombies don't consume admission capacity nor reach the provider; shutdown wakes admission sleepers and clears the waiter record; fail-closed acquire holds tasks until recovery; release failure preserves the original result/exception and retries via maintenance; stats failures never break calls; bounded zombie drain; compaction bounds the physical queue and `queue.join()` shutdown completes promptly; aggregated stats schema + global fields; long tasks survive past heartbeat TTL via renewal.

Multiprocess-path verification (real Manager, which single-process tests cannot exercise): executor-offloaded lock stays responsive under contention, 10 consecutive cancellations never poison the lock, and an end-to-end pump run over Manager-backed storage holds peak ≤ limit with all leases returned.

Full suite: **1640 passed**; the remaining failures/collection errors are pre-existing environment issues (tiktoken BPE download blocked by the sandbox network policy, webui asset 307s, missing optional backend deps) and are **byte-for-byte identical to the failure set on the base commit**. `ruff check` and `ruff format` are clean.

# Out of scope (matching the design plan)

- Redis / multi-machine coordination (single host only).
- Strict cross-process FIFO/priority ordering (the longest-waiter bias is soft — bounded unfairness by design; priority applies strictly within each process's queue).
- Strong consistency for aggregated stats (debounced snapshots, ≤ ~0.1s + heartbeat lag; the responding worker's own entry is always fresh).
- A global-limit entry point for embedded (non-API) library users.
- Propagating runtime max_async updates to the cross-worker cap under gunicorn multi-worker (startup value wins; documented — single-process mode is fully dynamic).

https://claude.ai/code/session_017rbHanRTbEVv3Rt2X6gTMz
https://claude.ai/code/session_01Tdaqf3H4T6d9JqFytDveor

### DIFF
--- a/env.example
+++ b/env.example
@@ -495,7 +495,12 @@ LLM_MODEL=gpt-5.4-mini
 ###       cap: the provider never sees more than MAX_ASYNC concurrent calls
 ###       in total, exactly like single-process mode. Slots held by crashed
 ###       workers (kill -9 / OOM) are reclaimed automatically via lease
-###       heartbeats. For globally capped groups, MAX_QUEUE_SIZE-style
+###       heartbeats. This cross-worker cap is best-effort, not a hard
+###       guarantee: if a worker's event loop stalls for ~40s while running
+###       a call, its slot may be reclaimed and the true total briefly reach
+###       the cap + 1; under normal operation (every worker renews its 5s
+###       heartbeat) the cap holds exactly. For globally capped groups,
+###       MAX_QUEUE_SIZE-style
 ###       admission counts only live QUEUED tasks (running tasks and
 ###       cancelled requests are excluded), matching the bounded-queue
 ###       semantics of single-process mode.
--- a/env.example
+++ b/env.example
@@ -490,30 +490,19 @@ LLM_MODEL=gpt-5.4-mini
 ### Max concurrency requests of LLM
 ### MAX_ASYNC is still accepted as a deprecated alias
 ### NOTE: with gunicorn multi-worker (lightrag-gunicorn --workers N) every
-###       MAX_ASYNC_* / *_MAX_ASYNC_* setting is a PER-WORKER-PROCESS limit,
-###       so the real total concurrency is ~ MAX_ASYNC x WORKERS.
+###       MAX_ASYNC_* / *_MAX_ASYNC_* setting (LLM roles, embedding, rerank)
+###       is enforced BOTH per worker process AND as a cross-worker global
+###       cap: the provider never sees more than MAX_ASYNC concurrent calls
+###       in total, exactly like single-process mode. Slots held by crashed
+###       workers (kill -9 / OOM) are reclaimed automatically via lease
+###       heartbeats. For globally capped groups, MAX_QUEUE_SIZE-style
+###       admission counts only live QUEUED tasks (running tasks and
+###       cancelled requests are excluded), matching the bounded-queue
+###       semantics of single-process mode.
+###       Runtime caveat: changing a role's max_async through the API
+###       updates only that worker's local limit — the cross-worker cap
+###       keeps the value read at startup.
 MAX_ASYNC_LLM=4
-
-###########################################################################
-### Cross-worker GLOBAL concurrency limits (gunicorn multi-worker only)
-### *_GLOBAL_* variables cap the TOTAL in-flight calls across ALL gunicorn
-### workers for one call group (per LLM role / embedding / rerank), on top
-### of the per-worker MAX_ASYNC limits. Set them to
-###     <= WORKERS x per-worker max_async
-### Unset (default) means no global gate: behavior is unchanged.
-### Self-healing: slots held by crashed workers (kill -9 / OOM) are
-### reclaimed automatically via lease heartbeats.
-### Queue capacity note: for globally limited groups MAX_QUEUE_SIZE-style
-### admission counts only live QUEUED tasks (running tasks and cancelled
-### requests are excluded), matching the bounded-queue semantics of
-### non-limited groups.
-###########################################################################
-# EXTRACT_GLOBAL_MAX_ASYNC_LLM=8
-# KEYWORD_GLOBAL_MAX_ASYNC_LLM=8
-# QUERY_GLOBAL_MAX_ASYNC_LLM=8
-# VLM_GLOBAL_MAX_ASYNC_LLM=8
-# EMBEDDING_FUNC_GLOBAL_MAX_ASYNC=16
-# GLOBAL_MAX_ASYNC_RERANK=8
 
 ###########################################################################
 ### Role-specific LLM/VLM overrides
--- a/env.example
+++ b/env.example
@@ -492,18 +492,13 @@ LLM_MODEL=gpt-5.4-mini
 ### NOTE: with gunicorn multi-worker (lightrag-gunicorn --workers N) every
 ###       MAX_ASYNC_* / *_MAX_ASYNC_* setting (LLM roles, embedding, rerank)
 ###       is enforced BOTH per worker process AND as a cross-worker global
-###       cap: the provider never sees more than MAX_ASYNC concurrent calls
-###       in total, exactly like single-process mode. Slots held by crashed
-###       workers (kill -9 / OOM) are reclaimed automatically via lease
-###       heartbeats. This cross-worker cap is best-effort, not a hard
-###       guarantee: if a worker's event loop stalls for ~40s while running
-###       a call, its slot may be reclaimed and the true total briefly reach
-###       the cap + 1; under normal operation (every worker renews its 5s
-###       heartbeat) the cap holds exactly. For globally capped groups,
-###       MAX_QUEUE_SIZE-style
-###       admission counts only live QUEUED tasks (running tasks and
-###       cancelled requests are excluded), matching the bounded-queue
-###       semantics of single-process mode.
+###       cap. Under normal operation this keeps total in-process provider
+###       calls clamped to MAX_ASYNC, similar to single-process mode. Slots
+###       held by crashed workers (kill -9 / OOM) are reclaimed automatically
+###       via lease heartbeats; if a worker is terminated externally while its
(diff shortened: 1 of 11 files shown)


danielaskdd:
@codex review

PR #3255, opened 2026-06-12
fix(neo4j): escape single quotes in workspace label to prevent Cypher injection

skymacro:
## Summary

`_get_workspace_label()` in `lightrag/kg/neo4j_impl.py` only escaped backticks for backtick-quoted identifier context, but the label is also interpolated into **single-quoted string literals** inside APOC `labelFilter` configuration (lines 1359 and 1413). A workspace name containing a single quote (`'`) could break out of the string and inject arbitrary Cypher.

## Motivation

Defense-in-depth: even though `validate_workspace` rejects workspace names containing path separators (`/`, `\`), the escaping layer should be correct on its own to prevent Cypher injection from any input that passes validation.

## What Changed

- **`lightrag/kg/neo4j_impl.py`** — Added `.replace("'", "''")` to `_get_workspace_label()` so single quotes are escaped (doubled) in addition to backticks. This is harmless in backtick identifier context and prevents injection in the `labelFilter` string-literal context.
- **`tests/kg/neo4j_impl/test_workspace_label_injection.py`** — New test file with 17 cases covering normal names, whitespace fallback, backtick escaping, single-quote escaping, mixed escaping, and injection payloads. Two structural assertions verify escaped values cannot break out of either context.

## What is Broken

Nothing. The change is additive — escaping `'` → `''` inside backtick-quoted identifiers is a no-op semantically.

## How it Works

Cypher escapes single quotes inside string literals by doubling them (`'` → `''`), same as SQL. By applying this escaping in `_get_workspace_label()`, the returned value is safe for both:
- Backtick-quoted identifiers: `MATCH (n:\`{label}\`)`
- Single-quoted string literals: `labelFilter: '{label}'`

### DIFF
--- a/lightrag/kg/neo4j_impl.py
+++ b/lightrag/kg/neo4j_impl.py
@@ -104,18 +104,18 @@ class Neo4JStorage(BaseGraphStorage):
         self._driver = None
 
     def _get_workspace_label(self) -> str:
-        """Return sanitized workspace label safe for use as a backtick-quoted identifier in Cypher queries.
+        """Return sanitized workspace label safe for use in Cypher queries.
 
-        Escapes backticks by doubling them to prevent Cypher injection
-        via the LIGHTRAG-WORKSPACE header, while preserving a 1-to-1 mapping
-        for all other characters. The returned value is intended to be used
-        inside backticks (for example, MATCH (n:`{label}`)) and is not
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
PR #3257 supersedes this PR

PR #3257, opened 2026-06-12
fix(neo4j): bind APOC labelFilter as parameter to prevent Cypher injection

danielaskdd:
## Summary

The workspace label is interpolated into the APOC `labelFilter` **single-quoted string literal** at two `get_knowledge_graph` sites (`lightrag/kg/neo4j_impl.py`, formerly lines 1359 and 1413):

```cypher
labelFilter: '{workspace_label}',
```

A workspace name containing a single quote (e.g. `team'a`, which passes `validate_workspace` since it has no path separators) could close the string literal and inject arbitrary Cypher via the `LIGHTRAG-WORKSPACE` header.

This PR supersedes #3255, which attempted to fix the same injection point by doubling single quotes (`'` → `''`) inside the shared `_get_workspace_label()`.

## Motivation

The approach in #3255 is incorrect:

- **Cypher does not escape string literals by doubling quotes** (that is SQL). Cypher uses backslash escaping (`\'`), so `'team''a'` is not the string `team'a` — it parses as two adjacent string literals (a syntax error).
- The same value is used in **two different syntactic contexts** — backtick-quoted identifiers (`` MATCH (n:`{label}`) ``) and single-quoted string literals (`labelFilter: '{label}'`). In backtick context, single quotes are literal and must **not** be escaped; doubling them corrupts the label. A single shared escape value therefore cannot be correct in both contexts.

This matches the automated review comment on #3255 (P2), which is technically correct.

## What Changed

- **`lightrag/kg/neo4j_impl.py`**
  - Added `_get_raw_workspace_label()` returning the un-escaped label (with `strip()` + `"base"` fallback). This is the actual Neo4j label name, safe to bind as a query parameter.
  - `_get_workspace_label()` now derives its backtick-escaped form from the raw label (behaviour unchanged for backtick contexts).
  - Both `apoc.path.subgraphAll` sites now bind `labelFilter: $label_filter` instead of interpolating the label into a string literal. The same config map already binds `$max_depth`/`$max_nodes`, so parameter binding here is valid and behaviour-preserving for normal labels — and injection-proof by construction.
- **`tests/kg/neo4j_impl/test_workspace_label_injection.py`** — New tests:
  - `_get_workspace_label`: backtick escaping; single quotes preserved 1-to-1 (not doubled).
  - `_get_raw_workspace_label`: no escaping; `strip`/`base` fallback.
  - A functional test mocking the Neo4j driver to assert that for a malicious workspace the label is bound as the `label_filter` parameter and the query contains no interpolated `labelFilter: '...'` literal.

## What is Broken

Nothing. Parameter binding produces the same APOC `labelFilter` value as before for normal workspace names; only the injection vector is removed.

## How it Works

Cypher evaluates the config map (including bound parameters) before passing it to the APOC procedure, so a parameterized `labelFilter` value can never be parsed as query text. There is no string literal for a single quote to break out of, regardless of the workspace name.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ep21KfAQXmMiFtLp6HBdaz)_

### DIFF
--- a/lightrag/kg/neo4j_impl.py
+++ b/lightrag/kg/neo4j_impl.py
@@ -103,6 +103,18 @@ class Neo4JStorage(BaseGraphStorage):
 
         self._driver = None
 
+    def _get_raw_workspace_label(self) -> str:
+        """Return the actual Neo4j label name for this workspace (no escaping).
+
+        This is the un-escaped label as it is stored on nodes. It is safe to
+        bind as a query parameter (for example, APOC ``labelFilter``), where
+        Neo4j handles the value without string interpolation and therefore
+        without any risk of Cypher injection. It must NOT be interpolated
+        directly into a query string.
+        """
+        workspace = self.workspace.strip()
+        return workspace if workspace else "base"
+
     def _get_workspace_label(self) -> str:
         """Return sanitized workspace label safe for use as a backtick-quoted identifier in Cypher queries.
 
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

PR #3246, opened 2026-06-11
feat: enhance streaming response for WebUI (#1286)

returnSGD:
## Summary

- **Backend**: `/query` endpoint now supports `stream=true`, returning NDJSON (`application/x-ndjson`) via a shared `_build_stream_generator` helper — eliminating ~40 lines of duplicated streaming logic between `/query` and `/query/stream`
- **Frontend**: Refactored `queryTextStream` by extracting `_readNdjsonStream`, `_buildStreamHeaders`, and `_classifyStreamError` helpers, removing DRY duplication between the normal path and the guest-token retry path (~190 lines net reduction)

### Backend (`query_routes.py`)
- The `/query` endpoint now respects the `stream` parameter (previously it was ignored)
- When `stream=True`, it returns the same NDJSON streaming response as `/query/stream`
- Both endpoints share `_build_stream_generator()` for consistent NDJSON formatting and error handling

### Frontend (`lightrag.ts`)
- `_readNdjsonStream()`: shared NDJSON parser using `ReadableStream.getReader()` with proper `releaseLock()` in `finally`
- `_buildStreamHeaders()`: centralized auth header construction
- `_classifyStreamError()`: user-friendly error classification (auth errors, HTTP status codes, network errors)
- Guest-token 401 retry path now reuses `_readNdjsonStream` instead of duplicating the entire parser

## Test plan
- [x] Python syntax verified with `ast.parse()`
- [x] Manual test: send `POST /query` with `stream=true` and verify NDJSON response
- [x] Manual test: verify WebUI streaming chat with stream toggle enabled
- [x] Manual test: verify guest mode silent token refresh during streaming

PR #3269, opened 2026-06-13
fix: DRY refactor of WebUI NDJSON stream parsing

returnSGD:
## Summary
DRY refactor of the WebUI NDJSON stream parsing in `lightrag_webui/src/api/lightrag.ts`, with a small backend readability extract and full test coverage for both ends.

### Frontend (`lightrag.ts`)
- `_readNdjsonStream` — unified stream reader, shared by the normal path and the guest-token retry path (eliminates the duplicated `reader.read()` + buffer-split + `JSON.parse` loop)
- `_buildStreamHeaders` — single header construction helper
- `_classifyStreamError` — centralized error classification for consistent user-facing messages
- **Retry path no longer mislabels stream errors**: token refresh + retry `fetch` stay inside the guarded block, but the resulting response is assigned to `activeResponse` and read **after** the auth block. A mid-stream failure on the *refreshed* stream is now classified uniformly by the outer handler instead of being reported as "Failed to refresh authentication" + a spurious login redirect.

### Backend (`query_routes.py`)
- `_build_stream_generator` extracts the inline NDJSON formatting closure from `/query/stream` for readability
- `stream` field default changed `True → None`; `/query/stream` treats `None` as `True`, `/query` forces `False`
- `/query` remains JSON-only — no streaming, no NDJSON, backward-compatible

### Behavior note
Previously, JSON parse failures inside the stream called `onError` (in three places with inconsistent messages). `_readNdjsonStream` now uniformly `console.warn`s and skips malformed mid-stream lines, but if the stream is cut off and the final buffer holds an unparseable partial JSON object, a single `onError` alerts the user that the response may be truncated. Server-sent `{"error": ...}` lines are still forwarded to `onError` unchanged.

## Tests
- `tests/api/routes/test_query_stream_routes.py` — verifies `/query` declares `application/json` and `/query/stream` declares/serves `application/x-ndjson` (OpenAPI spec + live content-type)
- `lightrag_webui/src/api/lightrag-stream.test.ts` — covers normal NDJSON parsing, error-line forwarding, malformed-line skipping, multi-byte split across chunks, truncated-stream `onError`, user abort (silent), HTTP 403/404/429/5xx messages, network errors, guest-token 401 retry, and auth-header construction
- Auth-header test now **asserts the actual `Authorization: Bearer …` / `Accept` / `Content-Type` headers** (replaced the previous `expect(true).toBe(true)` placeholder) and adds an "omits Bearer token when none stored" case
- All fetch mocks go through a typed `installFetchMock` helper so the new test file is clean under `tsc -b` (no `fetch.preconnect` type errors) and ESLint

## Related
- Related to [#1286](https://github.com/HKUDS/LightRAG/issues/1286)
- Replaces closed PR [#3246](https://github.com/HKUDS/LightRAG/pull/3246) per reviewer feedback — backend streaming on `/query` has been removed

## Test plan
- [x] `queryTextStream` works with normal auth
- [x] `queryTextStream` works with guest-token auto-refresh (401 retry path)
- [x] Mid-stream failure on the *refreshed* retry stream is classified uniformly (no false login redirect)
- [x] Stop button (user abort) exits silently without surfacing an error
- [x] HTTP errors (403, 404, 429, 5xx) show user-friendly messages
- [x] Truncated stream (unparseable final buffer) alerts user via `onError`
- [x] Auth headers (`Authorization` / `Accept` / `Content-Type`) verified by tests
- [x] `/query` continues to return `application/json` only
- [x] `/query/stream` continues to return `application/x-ndjson`
- [x] `bun test` green, `tsc -b` clean, ESLint clean, `pytest` green


returnSGD:
## Test Plan Results

### Backend (`/query` & `/query/stream` endpoints)

```
tests/api/routes/test_query_stream_routes.py::TestQueryRouteJsonOnly::test_openapi_spec_declares_json_response PASSED
tests/api/routes/test_query_stream_routes.py::TestQueryRouteJsonOnly::test_query_route_exists_and_accepts_post PASSED
tests/api/routes/test_query_stream_routes.py::TestQueryStreamRoute::test_openapi_spec_declares_ndjson_response PASSED
tests/api/routes/test_query_stream_routes.py::TestQueryStreamRoute::test_stream_route_exists_and_accepts_post PASSED
tests/api/routes/test_query_stream_routes.py::TestQueryStreamResponseContentType::test_stream_response_has_ndjson_content_type PASSED
```

| # | Test | Result |
|---|------|--------|
| 1 | `/query` OpenAPI declares `application/json` (not ndjson) | ✅ |
| 2 | `/query` route exists and accepts POST | ✅ |
| 3 | `/query/stream` OpenAPI declares `application/x-ndjson` | ✅ |
| 4 | `/query/stream` route exists and accepts POST | ✅ |
| 5 | `/query/stream` response Content-Type header is `application/x-ndjson` | ✅ |

---

### Frontend (`queryTextStream` — NDJSON stream parsing)

```
lightrag_webui/src/api/lightrag-stream.test.ts — 17 pass, 0 fail
```

| # | Scenario | Result |
|---|----------|--------|
| 1 | Parses NDJSON response chunks and calls `onChunk` | ✅ |
| 2 | Forwards error lines to `onError` | ✅ |
| 3 | Skips malformed JSON lines without breaking the stream | ✅ |
| 4 | Handles multi-byte characters split across chunks | ✅ |
| 5 | User abort (Stop button, `signal.aborted`) exits silently | ✅ |
| 6 | `AbortError` without a signal exits silently | ✅ |
| 7 | HTTP 403 → "You do not have permission to access this resource (403 Forbidden)" | ✅ |
| 8 | HTTP 404 → "The requested resource does not exist (404 Not Found)" | ✅ |
| 9 | HTTP 429 → "Too many requests, please try again later (429 Too Many Requests)" | ✅ |
| 10 | HTTP 500 → "Server error, please try again later (500)" | ✅ |
| 11 | HTTP 502 → "Server error, please try again later (502)" | ✅ |
| 12 | HTTP 503 → "Server error, please try again later (503)" | ✅ |
| 13 | `Failed to fetch` → network error message | ✅ |
| 14 | `NetworkError` → network error message | ✅ |
| 15 | Request is sent to `/query/stream` endpoint | ✅ |
| 16 | Guest-token 401 auto-refresh retry succeeds | ✅ |
| 17 | Includes Bearer token header when present (implicit) | ✅ |

---

### Test files added
- `tests/api/routes/test_query_stream_routes.py` — backend OpenAPI + Content-Type verification
- `lightrag_webui/src/api/lightrag-stream.test.ts` — frontend stream parsing unit tests

**Run commands:**
```bash
# Backend
/condaa/python -m pytest tests/api/routes/test_query_stream_routes.py -v

# Frontend
cd lightrag_webui && bun test src/api/lightrag-stream.test.ts --timeout 10000
```

danielaskdd (commented):
Overall this is a reasonable cleanup. On the frontend, extracting the shared NDJSON read loop into `_readNdjsonStream` removes genuine duplication between the normal path and the guest-token 401 retry path, and moving `reader.releaseLock()` into a `finally` is more robust than before. I verified the backend `stream` default change (`True → None`) against all three endpoints — `/query` forces `False`, `/query/stream` already falls back via `request.stream if request.stream is not None else True`, and `/query/data` passes `False` — so it's functionally equivalent and safe.

A few things I'd like to confirm / adjust:

1. **One user-visible `onError` behavior change worth calling out in the description.** Previously a JSON parse failure called `onError` (in three places — normal-path per-line, normal-path final buffer, and retry-path per-line, with inconsistent messages). The new `_readNdjsonStream` uniformly `console.warn`s and skips, and no longer invokes `onError`. For a single truncated line this is an improvement, but the side effect is: **if the stream is cut off mid-line and the final buffer holds a partial JSON object, the old code surfaced a parse error, whereas the new code silently drops it and behaves as if the stream ended normally** — the user gets no indication. If we want truncated/corrupt responses to still notify the user, consider keeping a single soft `onError` for an unparseable leftover buffer after the stream ends. (Note: server-sent `{"error": ...}` lines are still forwarded to `onError`, so that path is unchanged — good.)

2. **The title says "(frontend only)" but the PR also changes the backend** `query_routes.py` (extracts `_build_stream_generator`, changes the `stream` default, removes docstring lines). Worth correcting the title to avoid misleading reviewers.

3. **The link to #1286 is weak.** #1286 is closed as completed, and its only outstanding point was "store session history in the backend," which this PR doesn't touch. If it's just borrowing context, consider changing "Addresses #1286" to "Related to" (or dropping it) so it isn't read as advancing that issue's substance.

4. **The Test Plan references `tests/api/routes/test_query_stream_routes.py` and `lightrag_webui/src/api/lightrag-stream.test.ts`, but neither file appears in this PR's diff** (only the 2 source files are changed). Could you confirm whether these were meant to be committed? If they're new tests, it'd be good to include them here.

5. (minor) `_build_stream_generator` currently has a single caller (`/query/stream`), so this part is really "move an inline closure out for readability" rather than deduplication — the "deduplicates" wording in the description could be softened.

---

danielaskdd:
@codex review

PR #3271, opened 2026-06-13
fix: centralize sync-wrapper event-loop guard with accurate errors

danielaskdd:
## Summary

Centralizes the event-loop handling shared by all of `LightRAG`'s synchronous wrappers into a single `_run_sync()` helper, and replaces the obscure failures that occur when a sync wrapper is called against the wrong event loop with one clear, actionable error.

This is an alternative to #3245, focused only on the sync-interface improvement (the part of #3245 that had real value), implemented more completely and without its drawbacks.

## Motivation

Closes the sync-interface portion of #1968. The 15 sync wrappers (`insert`, `query`, `query_data`, `query_llm`, `insert_custom_chunks`, `insert_custom_kg`, `clear_cache`, `delete_by_entity`, `delete_by_relation`, `edit_entity`, `edit_relation`, `create_entity`, `create_relation`, `merge_entities`, `export_data`) each acquired a loop and called `loop.run_until_complete()` inline. That is only valid when the current thread has no running loop **and** the loop it drives is the one the storages were initialized on. Two misuse modes break this, with obscure failures:

- **Same thread** (directly inside an `async def` / FastAPI handler): `run_until_complete()` raises `RuntimeError: This event loop is already running`. This is a fail-fast error, **not** a deadlock, but the message gives no hint about the fix.
- **Wrong loop while the original loop is still alive** (`loop.run_in_executor(None, rag.insert, ...)` spins up a fresh loop on a pool thread while the app's loop keeps running, or an asyncio loop runs on another thread): the coroutine runs on a *different* loop than the one the storages were initialized on. The shared `asyncio.Lock` objects in `lightrag.kg.shared_storage` bind to the first loop that acquires them, so a second loop raises `RuntimeError: <Lock> is bound to a different event loop` or stalls on a callback scheduled on an idle loop.

Both have the same correct fix: use the `a*` coroutine directly from async code.

## What's changed

- New module-level `_run_sync(coro_factory, *, sync_name, async_name, owning_loop)` helper in `lightrag/lightrag.py`. It guards **both** misuse modes:
  - detects a running loop on the current thread via `asyncio.get_running_loop()` up front (same-thread case);
  - compares the loop it is about to drive against `owning_loop` and fails fast on a mismatch, but **only while `owning_loop` is still open** (wrong-loop case, covers `run_in_executor` and loops running on another thread without relying on thread-position heuristics).
  Each raises a single clear `RuntimeError` naming the correct `await a*(...)` alternative. The coroutine is built lazily through a factory, so a rejected call never leaves an un-awaited coroutine behind.
- `LightRAG._owning_loop` is captured in `initialize_storages()` as the loop the storages (and their `shared_storage` locks) bind to. It is a plain `__post_init__` instance attribute, **not** a dataclass field, so `_build_global_config()`'s `asdict(self)` never tries to (deep)copy a live event loop (which raises `TypeError: cannot pickle '_contextvars.Context'` on Python 3.14). The 15 sync wrappers pass it through to `_run_sync(...)`.
- The cross-loop check is deliberately scoped to a **still-alive** `owning_loop`. A *closed* one, such as the common `rag = asyncio.run(initialize_rag())` pattern where `asyncio.run` closes the loop after `initialize_storages()`, is let through to run on a fresh loop, preserving the long-standing behavior (any lock still bound to the closed loop is handled by `asyncio.Lock` itself, exactly as before).
- The `_run_sync()` docstring now explicitly documents the sync wrapper boundary: these wrappers are compatibility conveniences for simple, single-threaded scripts. SDK integrations, async apps, and concurrent workloads should prefer the `a*` coroutine APIs. The sync wrappers are not cross-thread-safe entry points for concurrent calls against the same `LightRAG` instance; callers that must use them from multiple threads need external serialization or a single-loop worker.
- All 15 sync wrappers now delegate to `_run_sync(...)`, removing the duplicated `always_get_an_event_loop()` / `run_until_complete()` boilerplate.
- `always_get_an_event_loop()` no longer triggers the `asyncio.get_event_loop()` auto-create `DeprecationWarning` on Python 3.12+, while preserving loop reuse (which the owning-loop guard relies on).
- Regression tests in `tests/test_sync_wrapper_guard.py` (same-thread guard, lazy-factory no-leak, matching-loop runs, live mismatched-loop fails fast, closed owning-loop runs).

## Sync wrapper scope

The synchronous API remains supported for simple single-threaded scripts and existing examples, including `rag = asyncio.run(initialize_rag()); rag.insert(...)`. It should not be treated as the preferred SDK integration surface or as a cross-thread-safe concurrency API.

For application code, especially FastAPI/Jupyter/task-queue integrations or any workload with an existing event loop, prefer `await rag.ainsert(...)`, `await rag.aquery(...)`, and the other async methods. If synchronous wrappers must be called from multiple threads, serialize those calls externally or route them through one event loop.

## How this compares to #3245

- **DRY**: one helper instead of 12 copy-pasted guard calls layered on top of the existing per-wrapper loop boilerplate.
- **More complete**: also covers `clear_cache` and `export_data` (which #3245 did not) **and** the off-thread / wrong-loop case (which neither #3245 nor the running-loop-only check caught).
- **No lint regression**: the helper is placed after all imports. #3245 inserted its function *between* import statements, which triggers ruff `E402`.
- **Accurate wording**: the same-thread condition is an immediate `RuntimeError`, not a "silent deadlock" as #3245's message stated. The new messages say so honestly and point at the async API.

This PR intentionally does **not** include #3245's "zombie chunk cleanup" change: orphaned chunks from a failed extraction are already cleaned up by the existing resume path (`_purge_stale_extraction_if_resuming`) and by `adelete_by_doc_id`, so that part was redundant.

## What's broken / how it works

No behavioral change for correct sync usage, including the `rag = asyncio.run(initialize_rag())` + sync-wrapper pattern used throughout the examples (the init loop is closed, so the guard stays out of the way). The only new behavior is a clear fail-fast `RuntimeError` (instead of an obscure one or a stall) when a sync wrapper is driven from within a running loop, or from a *different, still-alive* loop than the one the instance was initialized on (the `run_in_executor` hazard).

The sync wrappers remain a compatibility/convenience layer, not the recommended API for async or concurrent SDK integrations. The async `a*` methods are the safe default for those environments.

## Test plan

- [x] `_run_sync` runs the coroutine and returns its result when no loop is running
- [x] `_run_sync` runs the coroutine when the driven loop is the instance's owning loop
- [x] `_run_sync` raises `RuntimeError` naming the sync + async methods when called inside a running loop, with no misleading "deadlock" wording
- [x] `_run_sync` raises `RuntimeError` when driven from a different, still-alive loop (off-thread `run_in_executor` pattern), without creating the coroutine
- [x] `_run_sync` runs the coroutine when `owning_loop` is closed (the `asyncio.run(initialize_rag())` example pattern), no regression
- [x] `_run_sync` does not create (leak) a coroutine when a guard rejects
- [x] `_owning_loop` is not a dataclass field, so `asdict(self)` excludes it (Python 3.14 `asdict` no longer chokes on a live loop)
- [x] `pre-commit run ruff / ruff-format` on `lightrag/lightrag.py` and `tests/test_sync_wrapper_guard.py` pass
- [x] `pytest tests/test_sync_wrapper_guard.py -W error::DeprecationWarning` — 6 passed, no warnings

Refs #1968. Alternative to #3245.

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -4379,9 +4379,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
             include_vector_data: Whether to include data from the vector database.
         """
         _run_sync(
-            lambda: self.aexport_data(
-                output_path, file_format, include_vector_data
-            ),
+            lambda: self.aexport_data(output_path, file_format, include_vector_data),
             sync_name="export_data",
             async_name="aexport_data",
         )
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -166,43 +166,48 @@ def _run_sync(
     *,
     sync_name: str,
     async_name: str,
+    owning_loop: Optional[asyncio.AbstractEventLoop] = None,
 ) -> _SyncResultT:
     """Drive an async coroutine to completion from a synchronous wrapper.
 
     The synchronous wrappers (``insert``, ``query``, ``delete_by_entity``,
     …) all share the same shape: acquire an event loop and call
-    ``loop.run_until_complete()``.  That call is only valid when **no** event
-    loop is already running on the current thread:
+    ``loop.run_until_complete()``.  That call is only valid when the current
+    thread has **no** running loop *and* the loop it ends up driving is the
+    same one the instance's storages were initialized on.  Two misuse modes
+    break this, and both have the same fix — use the ``a*`` coroutine from
+    async code:
 
-    * Called from within a running loop on the same thread (e.g. directly
-      inside an ``async def`` / FastAPI handler), ``run_until_complete()``
-      raises ``RuntimeError: This event loop is already running``.  This is a
-      fail-fast error, not a deadlock — but the message gives no hint about
-      the fix.
-    * Pushed onto another thread's loop (e.g. ``loop.run_in_executor(None,
-      rag.insert, …)``), it runs on a *different* loop than the one the
-      storage backends were initialized on.  The shared ``asyncio.Lock``
-      objects in ``lightrag.kg.shared_storage`` bind to the first loop that
-      acquires them, so a second loop raises ``RuntimeError: <Lock> is bound
-      to a different event loop`` or stalls waiting on a callback scheduled
-      on an idle loop.
+    * **Same thread, inside a running loop** (e.g. directly inside an
+      ``async def`` / FastAPI handler): ``run_until_complete()`` would raise
+      ``RuntimeError: This event loop is already running``.  Detected up front
+      via :func:`asyncio.get_running_loop`.
+    * **A different loop than the storages bound to** (e.g.
+      ``loop.run_in_executor(None, rag.insert, …)`` runs the wrapper on a pool
+      thread with no loop, or an asyncio loop runs on another thread): the
+      shared ``asyncio.Lock`` objects in ``lightrag.kg.shared_storage`` are
+      bound to ``owning_loop``, so acquiring them from a second loop raises
+      ``RuntimeError: <Lock> is bound to a different event loop`` or stalls on
+      a callback scheduled on an idle loop.  Detected by comparing the loop we
+      are about to drive against ``owning_loop``.
 
-    Both failure modes have the same correct fix: use the ``a*`` coroutine
-    directly from async code.  This helper detects a running loop up front and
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

PR #3272, opened 2026-06-13
fix(parser): strip C0 control separators from parsed body before persist

danielaskdd:
## Description

The document **parser** path did not clean illegal control characters from extracted body text before it became chunk content and was persisted to storage.

Two of the three ingest branches already sanitize at enqueue time (`ainsert` with explicit ids, and the plain-text branch both call `sanitize_text_for_encoding`), and the entity/relation extraction layer cleans every field via `sanitize_and_normalize_extracted_text`. But the **file-upload `PENDING_PARSE` path** — which runs a real parser (legacy PDF/txt, native docx, MinerU, Docling) — had no equivalent cleaning. As a result the C0 separators `\x1c`–`\x1f` (FS/GS/RS/US) and other control chars could survive from the parsed document all the way into `text_chunks` and the vector DB.

This is also a latent hazard because `\x1f` is used as `KEY_SEP` in `lightrag/kg/shared_storage.py`.

This PR cleans the parsed body **before it is persisted to disk**, making the parser path consistent with the other ingest entry points.

## Related Issues

N/A

## Changes Made

- **New helper `strip_control_characters()`** in `lightrag/utils.py`: removes surrogates + C0 control/separator chars (reusing the existing `_SURROGATE_PATTERN` / `_CONTROL_CHAR_PATTERN_ALL` regexes, which already cover `\x1c`–`\x1f` while preserving `\t`/`\n`/`\r`). Unlike `sanitize_text_for_encoding` it deliberately does **not** `html.unescape` or `.strip()`, so it is safe for text carrying intentional markup (`<table>`/`<drawing>`/`<equation>`) or significant whitespace. It is a no-op for control-char-free input.
- **Injection A — `lightrag/sidecar/writer.py`**: clean each block's rendered text before it feeds `blockid`, the `blocks.jsonl` `content` (the chunk source for docx/MinerU/Docling), `merged_text` (the full_docs body) and `document_hash`. A block that is only control chars + whitespace now collapses to empty and is dropped. Because the helper is a no-op on clean input, existing blockids/hashes and golden/byte-equivalence snapshots are unchanged.
- **Injection B — `lightrag/pipeline.py` `_persist_parsed_full_docs`**: clean `record["content"]` before `content_hash` is computed. This is the single convergence point for every engine's persist, so it covers the RAW (legacy) chunk source, keeps the dedup hash stable (computed on the sanitized body), and is an idempotent backstop for sidecar engines and any future non-sidecar engine.

Not touched: the enqueue branches, the entity/relation extraction layer (already cleaned independently), and `_clean_heading_text`.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

New tests:
- `tests/test_strip_control_characters.py` — helper unit tests (separator removal, whitespace preserved, no-op on clean input, no spurious spaces inserted between CJK).
- `tests/sidecar/test_writer.py` — control chars stripped from `blocks.jsonl` content & `document_hash`; **blockid/document_hash unchanged for clean input** (snapshot guard); all-control-char block is dropped.
- `tests/pipeline/test_persist_parsed_control_chars.py` — exercises the real `_persist_parsed_full_docs`: RAW content cleaned, `content_hash` derived from the cleaned body, no-op on clean content.

Verification:
- New tests + `tests/parser/test_legacy_parser.py`: **38 passed**.
- `tests/sidecar/ tests/parser/ tests/chunker/` (incl. golden / byte-equivalence): **517 passed** — no regression.
- `pre-commit run ruff-format` / `ruff` clean.

### DIFF
--- a/lightrag/pipeline.py
+++ b/lightrag/pipeline.py
@@ -1649,6 +1649,20 @@ class _PipelineMixin:
                     )
                 ).to_dict()
 
+                # Align the in-memory body with the sanitized copy that
+                # _persist_parsed_full_docs wrote to full_docs: a parser may
+                # return ParseResult(content=...) carrying the pre-clean text
+                # (e.g. legacy returns the raw extraction verbatim). Downstream
+                # this body feeds content_summary / content_length on doc_status
+                # and the duplicate-check length, so leaving C0 control chars
+                # (incl. NUL, which breaks PostgreSQL text writes) here would let
+                # them reach doc_status. No-op for sidecar engines (already
+                # cleaned at write_sidecar) and for already-clean content.
+                if isinstance(parsed_data_w.get("content"), str):
+                    parsed_data_w["content"] = strip_control_characters(
+                        parsed_data_w["content"]
+                    )
+
                 # Mirror non-fatal parser warnings (e.g. legacy docx tables
                 # missing w14:paraId) onto the in-memory status_doc so the
                 # ANALYZING / PROCESSING / PROCESSED / FAILED upserts carry
--- a/lightrag/pipeline.py
+++ b/lightrag/pipeline.py
@@ -72,6 +72,7 @@ from lightrag.utils import (
     sanitize_text_for_encoding,
     save_to_cache,
     serialize_llm_cache_identity,
+    strip_control_characters,
 )
 from lightrag.utils_pipeline import (
     archive_docx_source_after_full_docs_sync,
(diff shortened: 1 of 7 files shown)


danielaskdd:
Addressed the [P2] finding in a5840471.

You're right — `_persist_parsed_full_docs` only cleaned the persisted copy, while the parse worker still consumed the parser's returned `ParseResult.content` (legacy returns the raw extraction verbatim) for `content_summary` / `content_length` and the post-parse duplicate-length check, so control chars (incl. NUL → PostgreSQL text-write failures) could still reach `doc_status`.

Fix: took the second option you suggested — normalize `parsed_data_w["content"]` with `strip_control_characters` immediately after `parser.parse(...).to_dict()`, before the duplicate/status logic. This is the single convergence point for every engine (no-op for sidecar engines already cleaned at `write_sidecar`, and for already-clean content), and it makes `content_length` agree with the persisted body.

Added a parse-worker E2E test (`tests/pipeline/test_parse_worker_content_sanitized.py`): a PENDING_PARSE `.txt` whose body carries `\x1c`-`\x1f` + NUL → asserts `doc_status.content_summary`/`content_length` and `full_docs` are all cleaned and agree on the body. Full `tests/pipeline tests/parser tests/sidecar` + the helper unit tests: 558 passed.

PR #3304, opened 2026-06-20
perf(webui): made graph rendering more efficient -> large knowledge graphs now usable

Xaverrrrr:
## Description

The graph viewer in the webui completely failed for large knowledge graphs and was basically unusable.

This PR makes graph rendering efficient enough for large graphs, fixes a series of interaction/layout regressions found during review, and adds an **edge-count-adaptive** mode so small graphs keep the nicer experience while large graphs stay fast.

**Result => a 73k-node / 60k-edge knowledge graph that was unusable before now renders within seconds.**

### Performance
- Initial FA2 layout now runs in a Web Worker with Barnes-Hut (`inferSettings`); was synchronous O(V²) on the main thread and froze the tab for minutes.
- Replaced `seedrandom`, `setTimeout(0)` batching, and per-bad-edge try/catch with cheaper equivalents in graph build.
- Hover/highlight reducers are now O(1) (precomputed neighbor `Set`) and set to `null` when nothing is focused, so no per-node/edge callback runs on idle refreshes.
- Expansion + prune made O(n) (Maps/Sets instead of `.find()`/`.indexOf()` in loops).

### Edge-count-adaptive rendering (new)
A single shared threshold (`EDGE_PERF_LIMIT = 5000`) trades small-graph experience against large-graph cost, driven by a reactive node/edge count signal in the graph store (expand/prune mutate the graph in place, so `sigmaGraph.order/.size` are not reactive on their own):
- **≤ 5000 edges**: edges render as **curves**; edge events (hover/click to select an edge) follow the user setting.
- **> 5000 edges**: edges render **straight**; edge events are **fully disabled** (no WebGL picking buffer allocated at construction, no whole-graph edge reducer pass), and the "Edge Events" menu item is **greyed out** while preserving its checked value.
- The behavior recomputes after UI **expand/prune**, and **auto-recovers** (curves + edge events) when the edge count drops back to ≤ 5000, because the user setting is never overwritten.
- Curve/straight is switched at runtime via `defaultEdgeType` + a full `refresh()` (edges carry no per-edge type), and re-applied after a sigma rebuild (a rebuild resets the default back to straight).

### Bug fixes (review-driven)
- Worker layouts (Noverlap / Force / FA2) actually run now; a single layout supervisor owns the running worker so two layouts can't fight over node coordinates (jitter).
- Stopped an unbounded zero-backoff retry loop that hammered the backend on fetch errors; malformed graph data no longer triggers retries.
- Preserve zero edge weight instead of coercing it to a default.
- Resolve system-dark colors correctly when reducers are disabled (idle graph rendered light-theme defaults under `theme === 'system'` + OS dark).
- Preserve camera rotation/zoom when a worker layout finishes; don't re-run the initial layout on theme switch (run once per graph, not per sigma instance).
- Reliably restore edges after the camera stops moving with `hideEdgesOnMove` (edges no longer stay hidden after a small pan with inertia).
- Restore edge click/hover selection via the "Edge Events" setting (the picking buffer must be baked in at construction).
- Tuned worker layout params (Force Directed jitter, Noverlap margin).
- Removed the dead `maxLayoutIterations` setting and backfilled i18n.

### UX
- "Computing Layout…" overlay shown during any layout run.
- Bottom-left status line now shows the live **`node:` / `edge:`** counts of the rendered graph (replaces the static `Max` value), updating on build/expand/prune.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

These edits were largely made using Claude Code (Fable & Opus). All changes were reviewed and tested locally.


danielaskdd:
@codex review

lightrag_webui/src/components/graph/LayoutsControl.tsx
```diff
@@ -26,165 +27,197 @@ type LayoutName =
...
-        }, autoRunFor)
+    // Show the overlay synchronously, but DEFER supervisor.start() so the
+    // browser actually paints first. Force's runFrame() runs sync on the
+    // main thread (O(V^2) at 70k+ nodes = multi-second block); rAF won't
+    // work here because rAF callbacks fire BEFORE the next paint, so the
+    // block lands before the overlay ever shows. setTimeout with a small
+    // delay guarantees at least one paint happens first. FA2/Noverlap
+    // don't strictly need this (their start() just posts to a real worker
+    // and returns), but the small delay is irrelevant for them too.
+    useGraphStore.getState().setIsLayoutComputing(true)
```
danielaskdd: **The "Computing Layout…" overlay blocks interaction for the entire worker budget, which defeats the point of running these layouts in a worker.**

`isLayoutComputing` is set `true` here and only cleared in `stop()`, which fires after `workerBudgetMs(graph.order)` — i.e. `min(1500 + order/10, 10000)` ≈ **8.8s for a 73k-node graph**. The overlay in `GraphViewer.tsx` is `absolute inset-0 z-10 bg-background/80` with **no `pointer-events-none`**, so it covers the whole `SigmaContainer` (including this control's own play/pause button) for that whole window.

For FA2/Noverlap (real Web Workers) the UI is supposed to stay interactive while the layout settles in the background — that's the entire reason for the worker. The opaque, click-blocking overlay negates that and even prevents the user from pausing. Note the **initial** FA2 layout in `GraphControl.tsx` deliberately does *not* set this flag, so the two paths are inconsistent.

Suggestion: don't use the full-screen blocking overlay for worker layouts (either skip `setIsLayoutComputing` like the initial layout does, or use a small non-blocking indicator). Reserve the blocking overlay for the synchronous layout path.

---

lightrag_webui/src/components/graph/GraphControl.tsx
```diff
@@ -59,56 +59,95 @@ const GraphControl = ({ disableHoverEffect }: { disableHoverEffect?: boolean })
...
+      layout.start()
+      console.log(`FA2 worker layout started (${sigmaGraph.order} nodes)`)
+    } catch (error) {
+      console.error('Error starting FA2 worker layout:', error)
+      return
+    }
+
+    // Time budget instead of an iteration count: the UI stays interactive
+    // while the layout settles in the background, then we stop it.
+    const budgetMs = Math.min(1500 + sigmaGraph.order / 10, 10000)
```
danielaskdd: **The initial FA2 worker and a manually-selected worker layout can run concurrently and fight over the same graph coordinates.**

This initial FA2 supervisor runs for a time budget of up to 10s and is only `kill()`-ed on unmount or when `sigmaGraph` changes. If the user opens the Layouts menu and selects Force / Noverlap / Force Atlas within that window, `WorkerLayoutControl` builds a **second** supervisor bound to the same store graph (`buildSupervisor(...)` → `useGraphStore.getState().sigmaGraph`). Two supervisors then mutate the same nodes' `x/y` simultaneously, producing jitter / a tug-of-war until this budget timer fires.

Suggestion: give the two paths a single layout owner — e.g. store the initial supervisor in the graph store and have `WorkerLayoutControl.start()` (and manual layout selection) kill it before starting, so only one layout ever runs at a time.

---

lightrag_webui/src/hooks/useLightragGraph.tsx
```diff
@@ -344,174 +405,200 @@ const useLightrangeGraph = () => {
...
+          state.setIsFetching(false)
+          dataLoadedRef.current = false
+          fetchInProgressRef.current = false
+          // NOTE: graphDataFetchAttempted is reset, but lastFetchSignatureRef
+          // is deliberately KEPT. Clearing the flag alone used to create an
+          // unbounded zero-backoff retry loop: error -> flag cleared ->
+          // effect refires -> identical request -> error -> ... A failed
+          // query can now only be retried via an explicit refresh
+          // (graphDataVersion bump) or a parameter change.
+          state.setGraphDataFetchAttempted(false)
```
danielaskdd: **Killing the retry loop is right, but as written a *transient* fetch failure wedges the query permanently with no auto-recovery.**

On error `graphDataFetchAttempted` is reset but `lastFetchSignatureRef` is deliberately kept, so the next effect run computes the identical `label|depth|maxNodes|version` signature, hits the suppression guard, and returns. After a transient backend hiccup (restart / network blip) the same query can never be re-issued until the user changes a parameter or bumps `graphDataVersion` via refresh — a first-load failure just stays blank.

The reason recording-on-issue is needed is that `isFetching` is in the effect deps, so it re-fires after every attempt; recording only on *success* alone would reintroduce the loop on a persistent failure. So the suggestion is a bounded retry rather than permanent suppression: record the signature on **success** to guard the completed-fetch loop, and on **failure** allow a capped number of retries with backoff (e.g. 3 attempts, exponential delay) before giving up. That keeps the loop fixed while letting transient errors recover on their own.

Worth confirming the intended behavior — the PR description frames "no retry" as a feature, so flagging the trade-off rather than asserting a hard bug.

---

danielaskdd (commented):
Reviewed the graph-rendering perf work — the core rewrites (worker FA2, O(1) reducers, Map/Set expansion/prune, retry-loop fix) look solid. Three follow-up items left as inline comments below.

---

lightrag_webui/src/components/graph/LayoutsControl.tsx
Xaverrrrr: You're right, the layout was unnecessary - I removed it again.

lightrag_webui/src/components/graph/GraphControl.tsx
Xaverrrrr: Your suggestion is fair, I implemented it.

lightrag_webui/src/hooks/useLightragGraph.tsx
Xaverrrrr: Signature is now recorded on success. Transient failures now trigger retry with exponential backoff (max 3 iterations) just to be save.

danielaskdd:
@codex review

danielaskdd:
Thanks for the great performance work @Xaverrrrr! I pushed a series of follow-up commits on top to fix some interaction/layout regressions surfaced during review and to add an edge-count-adaptive mode. Quick summary of what I changed:

**Review-driven fixes**
- Resolve system-dark colors correctly when reducers are disabled (idle graph was rendering light-theme defaults under `theme === 'system'` + OS dark).
- Preserve zero edge weight, and stop retrying on malformed graph data.
- Layout lifecycle / camera UX: a single layout supervisor owns the running worker so two layouts can't fight over node coordinates (jitter), pause/refresh handled cleanly.
- Restore edge click/hover selection via the "Edge Events" setting (the WebGL picking buffer has to be baked in at construction, so it's keyed off the setting and rebuilds the instance).
- Preserve camera rotation/zoom when a worker layout finishes; don't re-run the initial layout on theme switch (run once per graph, not per sigma instance).
- Reliably restore edges after the camera stops with `hideEdgesOnMove` (they no longer stay hidden after a small pan with inertia).
- Tuned worker layout params (Force Directed jitter, Noverlap margin); removed the dead `maxLayoutIterations` setting and backfilled i18n.

**Edge-count-adaptive rendering (new)**
A single shared threshold (`EDGE_PERF_LIMIT = 5000`), driven by a new reactive node/edge count signal in the graph store (expand/prune mutate the graph in place, so `order`/`size` aren't reactive on their own):
- ≤ 5000 edges → curved edges, edge events follow the user setting.
- \> 5000 edges → straight edges, edge events fully disabled (no picking buffer, no whole-graph edge reducer pass), and the "Edge Events" menu item greyed out while keeping its checked value.
- Recomputes after expand/prune and auto-recovers when the count drops back, since the user setting is never overwritten.

**UX**
- Bottom-left status line now shows live `node:` / `edge:` counts instead of the static `Max` value.

All of this passes `tsc` / `eslint` / `bun run build` and was tested locally.

PR #3314, opened 2026-06-24
fix: strip null bytes from extracted text to prevent PostgreSQL encoding errors

### LINKED ISSUE #3308: [Bug]: PDF parsing fails with "invalid byte sequence for encoding UTF8: 0x00"

### Describe the bug

When ingesting/parsing certain PDF files, the parse worker fails and crashes. The issue occurs because the extracted text from the PDF contains null bytes (`0x00`), which causes an encoding error when attempting to insert the text into the database (likely PostgreSQL).

**Error Message**
```text
ERROR: Parse worker failed (native): invalid byte sequence for encoding "UTF8": 0x00
```

**Additional context / Suggested Fix**
This is a known limitation in PostgreSQL, which does not support null bytes (`\x00`) in `TEXT` or `VARCHAR` fields. 

A simple fix would be to sanitize the extracted text in the document parser before it is sent to the storage layer. For example, adding a simple replace function:
```python
# Inside the text extraction/parsing logic
cleaned_text = raw_extracted_text.replace('\x00', '')
```

Thank you for looking into this!

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### LightRAG Config Used

# Paste your config here
i use PostgreSQL for all storage:
```
LIGHTRAG_KV_STORAGE=PGKVStorage
LIGHTRAG_DOC_STATUS_STORAGE=PGDocStatusStorage
LIGHTRAG_GRAPH_STORAGE=PGGraphStorage
LIGHTRAG_VECTOR_STORAGE=PGVectorStorage
```

### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

michaelxer:
## Problem

When ingesting certain PDF files, the parse worker crashes with:

```
ERROR: Parse worker failed (native): invalid byte sequence for encoding "UTF8": 0x00
```

PDF text extractors (pypdf) can produce strings containing NUL characters (`\x00`). PostgreSQL does not support null bytes in `TEXT` or `VARCHAR` fields, so inserting this text fails.

Fixes #3308

## Solution

Added a `_strip_null_bytes()` sanitizer in the legacy `extract_text()` entry point. All extraction paths — binary formats (PDF, DOCX, PPTX, XLSX) and the text decode path — pass through this sanitizer before returning.

The fix is minimal (1 function + 3 call-site changes) and has zero performance impact on clean text.

## Testing

- Added unit tests for `_strip_null_bytes()` covering: embedded nulls, all-null input, clean text, and empty string.
- All 5 existing tests in `tests/parser/test_legacy_parser.py` continue to pass.
- The sanitizer is a pure `str.replace("\x00", "")` — no regex, no external deps.

## Scope

- `lightrag/parser/legacy/extractors.py`: +`_strip_null_bytes()` helper, wrap all 3 return paths in `extract_text()`.
- `tests/parser/test_legacy_parser.py`: +1 test function for the sanitizer.

### DIFF
--- a/lightrag/parser/legacy/extractors.py
+++ b/lightrag/parser/legacy/extractors.py
@@ -167,6 +167,17 @@ def _decode_text(file_bytes: bytes) -> str:
     return content
 
 
+def _strip_null_bytes(text: str) -> str:
+    """Remove NUL characters that cause PostgreSQL encoding errors.
+
+    PDF extractors (pypdf) can produce text containing ``\\x00`` bytes,
+    which PostgreSQL rejects with ``invalid byte sequence for encoding
+    "UTF8": 0x00``.  Stripping them is safe — NUL carries no semantic
+    content in natural-language text.
+    """
+    return text.replace("\x00", "")
+
+
 def extract_text(
     file_bytes: bytes, suffix: str, *, pdf_password: str | None = None
 ) -> str:
@@ -179,7 +190,7 @@ def extract_text(
     suffix = suffix.lower().lstrip(".")
     extractor = _BINARY_EXTRACTORS.get(suffix)
     if extractor is _extract_pdf_pypdf:
-        return _extract_pdf_pypdf(file_bytes, pdf_password)
+        return _strip_null_bytes(_extract_pdf_pypdf(file_bytes, pdf_password))
     if extractor is not None:
-        return extractor(file_bytes)
-    return _decode_text(file_bytes)
+        return _strip_null_bytes(extractor(file_bytes))
+    return _strip_null_bytes(_decode_text(file_bytes))
--- a/tests/parser/test_legacy_parser.py
+++ b/tests/parser/test_legacy_parser.py
@@ -119,3 +119,14 @@ async def test_legacy_parse_passes_pdf_password_from_env(
 
     assert seen == {"suffix": "pdf", "pdf_password": "s3cret"}
     assert result.content == "decrypted text"
+
+
+async def test_extract_text_strips_null_bytes(tmp_path):
+    """Null bytes from PDF extractors must be stripped to avoid PostgreSQL
+    ``invalid byte sequence for encoding "UTF8": 0x00`` errors (#3308)."""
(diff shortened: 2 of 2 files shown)


danielaskdd:
Thanks a lot for the contribution and for taking the time to write tests, @michaelxer! 🙏

After review, this fix is already covered on `main` (and in the released **v1.5.3**) by **PR #3272**, so we're going to close this one as redundant.

The existing fix sanitizes the parsed body at the **single convergence point for every parser engine** — `LightRAG._persist_parsed_full_docs` (`lightrag/pipeline.py`) — via the `strip_control_characters()` helper in `lightrag/utils.py`:

```python
_CONTROL_CHAR_PATTERN_ALL = re.compile(r"[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]")
```

A few reasons it already supersedes this PR:

- **Same path is covered.** The legacy/PDF engine (`lightrag/parser/legacy/parser.py`) persists through `_persist_parsed_full_docs`, so its output is already stripped before it reaches `full_docs`.
- **Broader coverage.** It removes the NUL byte from this issue (`\x00`, within the `\x00-\x08` range) *and* the other C0 control characters that trigger the same PostgreSQL `0x00`-class errors, while preserving `\t`/`\n`/`\r` and normal text.
- **Both writes are sanitized.** It is applied to the in-memory body feeding `doc_status` as well, so both the `full_docs` and `doc_status` PostgreSQL writes are clean — not just the legacy `extract_text()` return path.

Adding `_strip_null_bytes()` inside `extract_text()` would duplicate this work on the legacy path (harmless but redundant) and only handle a subset (`\x00` only). Centralizing at the persist layer is the intended design so every engine — legacy, native, MinerU, Docling — benefits without per-extractor changes.

Please upgrade to v1.5.3 (or the latest `main`) and the original crash in #3308 should be resolved. If you still hit it after upgrading, please comment on #3308 with your version and the full stack trace and we'll dig in.

Thanks again for helping make LightRAG more robust! 🚀

---

PR #3317, opened 2026-06-24
fix(api): don't pair wildcard CORS origin with credentials

danielaskdd:
## Summary

The API server's CORS middleware hardcoded `allow_credentials=True` while `CORS_ORIGINS` defaults to `"*"`. Per the [Fetch spec](https://fetch.spec.whatwg.org/#http-cors-protocol), the wildcard origin `*` and credentialed responses are mutually exclusive — a server must not pair `Access-Control-Allow-Origin: *` with `Access-Control-Allow-Credentials: true`. With this combination, Starlette's `CORSMiddleware` reflects **any** cookie-bearing origin back with `Access-Control-Allow-Credentials: true`, which is the classic permissive-CORS pattern flagged by security scanners (and the subject of an internal advisory review).

## Motivation

LightRAG Server authenticates exclusively via **request headers** — `Authorization: Bearer <JWT>` (`OAuth2PasswordBearer`) and `X-API-Key` (`APIKeyHeader`). It never issues or reads cookies (no `set_cookie`/`request.cookies` anywhere in the backend), and the WebUI stores its token in `localStorage` and sets the `Authorization` header manually (no `withCredentials`).

Because CORS credentials only govern **ambient** credentials (cookies, HTTP Basic/Digest, TLS client certs) — none of which LightRAG uses — the wildcard+credentials combo is not directly exploitable in a stock deployment. However it is:
- **spec-invalid** configuration that static analyzers correctly flag, and
- a **latent risk**: if an operator fronts the server with a reverse proxy that enforces cookie/Basic auth, the browser *will* attach those ambient credentials and Starlette *will* reflect the attacker's origin with `Allow-Credentials: true`, enabling cross-origin reads.

## What's changed

- `lightrag/api/lightrag_server.py`: derive `allow_credentials = cors_origins != ["*"]`. Credentials are enabled only for an **explicit origin allowlist**; the `"*"` wildcard now yields a clean `Access-Control-Allow-Origin: *` with no credentials header. Behavior for the default config is otherwise unchanged.
- `tests/api/test_cors.py`: new regression tests pinning both paths — wildcard ⇒ no `Access-Control-Allow-Credentials`; explicit allowlist ⇒ reflected origin **with** credentials; disallowed origins are never echoed back.
- `env.example` / `env.docker-compose-full`: document `CORS_ORIGINS` (default `*`, bundled WebUI is same-origin and needs no CORS, credentials only for an explicit allowlist).

## What's broken / compatibility

No behavioral change for the common case. Header-based auth (Bearer / API-Key) is unaffected, since those headers are not "credentials" in the CORS sense. The only difference: a hypothetical cross-origin **cookie**-based client under a wildcard origin would no longer receive `Allow-Credentials: true` — but LightRAG issues no cookies, so nothing in the project relies on that. Such deployments should set an explicit `CORS_ORIGINS` allowlist (which now correctly enables credentials).

## Testing

```
./scripts/test.sh tests/api/test_cors.py    # 4 passed
ruff check lightrag/api/lightrag_server.py tests/api/test_cors.py    # All checks passed
```

---
_Generated by [Claude Code](https://claude.ai/code/session_017ZwddG56G4QwLcbEsYLxmd)_

### DIFF
--- a/env.docker-compose-full
+++ b/env.docker-compose-full
@@ -14,6 +14,11 @@ WEBUI_DESCRIPTION='Simple and Fast Graph Based RAG System'
 # WORKERS=2
 ### gunicorn worker timeout(as default LLM request timeout if LLM_TIMEOUT is not set)
 # TIMEOUT=150
+### CORS allowed origins for browser cross-origin requests. Defaults to "*"
+### (any origin). The bundled WebUI is served same-origin and does not need
+### this; set an explicit allowlist only when a different-origin web app calls
+### the API from a browser. Credentialed (cookie) cross-origin requests are
+### only enabled for an explicit allowlist, never for the "*" wildcard.
 # CORS_ORIGINS=http://localhost:3000,http://localhost:8080
 
 ### Path Prefix Configuration (Optional)
--- a/env.example
+++ b/env.example
@@ -14,6 +14,11 @@ WEBUI_DESCRIPTION='Simple and Fast Graph Based RAG System'
 # WORKERS=2
 ### gunicorn worker timeout(as default LLM request timeout if LLM_TIMEOUT is not set)
 # TIMEOUT=150
+### CORS allowed origins for browser cross-origin requests. Defaults to "*"
+### (any origin). The bundled WebUI is served same-origin and does not need
+### this; set an explicit allowlist only when a different-origin web app calls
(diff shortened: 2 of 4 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3323, opened 2026-06-24
fix: Add workspace isolation to cache hash computation to prevent cross-user cache pollution

ggi2003:
## Summary
This PR fixes a critical cache pollution issue where cache hash computation did not include workspace information, leading to cross-user/session cache contamination in multi-user scenarios.

## Motivation
In multi-user environments, different users or sessions sharing the same LightRAG instance could experience cache pollution when workspace information was not included in cache hash computation. This caused:
- Users receiving cached responses from other users' queries
- Incorrect entity extraction results due to cross-workspace cache hits
- Data isolation violations in multi-tenant deployments

## What's Changed
Modified cache hash computation to include workspace parameter across all cache-related operations:

1. **utils.py**: 
   - Added `workspace` parameter to `compute_args_hash()` function
   - Updated `use_llm_func_with_cache()` to accept and pass workspace parameter

2. **operate.py**:
   - Updated cache hash calls in `kg_query()` (query cache)
   - Updated cache hash calls in `extract_keywords_only()` (keyword extraction cache)
   - Updated cache hash calls in `naive_query()` (query cache)
   - Updated three `use_llm_func_with_cache()` calls for entity extraction and summary generation

3. **pipeline.py**:
   - Updated cache hash calls in VLM image analysis
   - Updated cache hash calls in entity extraction

## How It Works
- Workspace parameter is prepended to the hash computation string
- Different workspaces now generate different cache hashes even for identical queries
- Workspace is obtained from `self.workspace` in pipeline.py and `global_config.get("workspace")` in operate.py
- Backward compatible: None workspace values are converted to empty string for consistent hashing

## Testing
- All modifications passed compilation verification
- Existing cache-related tests continue to pass
- No breaking changes to existing functionality

## Impact
- ✅ Fixes cache pollution in multi-user/multi-session scenarios
- ✅ Maintains backward compatibility with single-user deployments
- ✅ No performance impact on cache operations
- ✅ Enables proper data isolation in multi-tenant environments

Users can now implement workspace-based cache isolation by setting `workspace=f"{user_id}_{session_id}"` for complete user/session cache separation.

## Before the modification, the multi - user retrieval cache was contaminated.
修改前Snipaste_2026-06-24_11-36-11
修改前Snipaste_2026-06-24_11-44-40

## After the modification, the multi - user retrieval cache is correct.
修改后Snipaste_2026-06-24_11-53-39

预编译通过Snipaste_2026-06-24_13-56-42


danielaskdd:
@codex review

danielaskdd:
Thanks for the detailed write-up and the repro screenshots! I dug into this, but I don't think there is actually a cross-workspace cache pollution problem to fix here — the LLM response cache is **already physically isolated by workspace at the storage layer**, before the hash ever comes into play.

### Why the cache is already workspace-isolated

`handle_cache()` reads through `hashing_kv.get_by_id(flattened_key)`, and `hashing_kv` is the `llm_response_cache` storage instance, which is constructed with the instance's workspace:

```python
self.llm_response_cache = self.key_string_value_json_storage_cls(
    namespace=NameSpace.KV_STORE_LLM_RESPONSE_CACHE,
    workspace=self.workspace,   # <-- workspace bound here
    ...
)
```

Every KV backend then partitions data by that workspace dimension, so the `args_hash` only ever acts as the `id` *within* a workspace:

- **JSON KV**: stored under a per-workspace directory `working_dir/<workspace>/kv_store_<namespace>.json`
- **Redis**: keys are prefixed — `final_namespace = f"{workspace}_{namespace}"`
- **PostgreSQL**: `LIGHTRAG_LLM_CACHE` has `PRIMARY KEY (workspace, id)` and reads are scoped by `workspace`

Two different workspaces therefore land in two physically separate cache stores already. The hash never needs to carry the workspace.

### Why this PR doesn't change the behavior

`global_config["workspace"]` comes from `asdict(self)["workspace"]`, which is exactly the same `self.workspace` used to build the cache store. Within a single LightRAG instance the workspace is a constant equal to the storage workspace, so prepending it to the hash is a no-op for isolation.

The "multiple users sharing one instance" scenario also isn't enabled by this change: a single instance binds exactly one workspace and one workspace-scoped cache store; there is no per-query workspace override. The only way two identities collide is if they deliberately share the **same** workspace (e.g. the default `""`) — and in that case `workspace_str == ""`, so the computed hash is identical to before and the PR provides zero additional isolation.

### Side effects worth noting

- For any deployment that *does* set a non-empty workspace, this change alters every entity-extraction/query cache hash → full cache invalidation on upgrade with no isolation benefit in return.
- The concatenation `workspace_str + "".join(args)` has no delimiter, so `"user1" + "2…"` and `"user12" + "…"` hash to the same value — which would actually *re-introduce* cross-workspace collisions, the opposite of the goal.

### Could you share a minimal repro?

If you observed real pollution, my strong suspicion is that the two "users" were running under the **same** workspace rather than separate ones. If you can share a minimal repro where two distinct workspaces leak into each other, I'll happily reopen and dig in — that would point to a storage-backend bug (a backend not honoring its workspace dimension), which is where the real fix would belong, rather than the hash layer.

Given the above, I'm planning to close this PR. Thanks again for taking the time to investigate and report — much appreciated! 🙏

PR #3330, opened 2026-06-24
🔒 fix(docker): patch dependency CVEs, run as non-root, pin bookworm base (#3296)

### LINKED ISSUE #3296: [Bug]: [Docker] Vulnerable dependencies and root user in ghcr.io/hkuds/lightrag:latest block enterprise deployment

- [ ] I have searched the existing issues and this bug is not already filed.

## Summary

  The official Docker image `ghcr.io/hkuds/lightrag:latest` fails enterprise security
  scans (Prisma/Twistlock), blocking deployment in corporate environments. Three fixes
  are needed.

  ## 1. Upgrade Python dependencies (fixes available now)

  | Package | Current | Fix version | CVE |
  |---|---|---|---|
  | starlette | 1.0.0 | 1.3.1 | CVE-2026-54283, CVE-2026-48818 |
  | cryptography | 48.0.0 | 48.0.1 | GHSA-537c-gmf6-5ccf |
  | python-multipart | 0.0.28 | 0.0.30 | CVE-2026-53539 |

  **Important**: a wrapper Dockerfile with `pip install --upgrade` does NOT fix this
  because LightRAG uses a `uv` virtualenv internally. Security scanners inspect the
  venv packages, not system Python. The fix must be in the upstream image.

  ## 2. Add a non-root user to the Dockerfile (CIS Docker 4.1)

  The image runs as root, which fails CIS Docker benchmark 4.1 — a hard block in
  most enterprise environments.

  Suggested fix:
  ```dockerfile
  RUN groupadd -r lightrag && useradd -r -g lightrag lightrag && \
      chown -R lightrag:lightrag /app
  USER lightrag

  3. Remove perl from the base image (no fix in Debian yet)

  The Debian trixie base carries perl 5.40.x with CVE-2026-12087 (critical, CVSS 9.1).
  No Debian patch is available yet. Consider switching the base to:
  - python:3.11-slim-bookworm (older perl, lower exposure), or
  - Alpine-based image (no perl at all)
  
  Impact

  Without these fixes, teams in regulated industries cannot deploy LightRAG in
  production without a manual security exception per deployment — which creates
  significant friction for adoption.

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### LightRAG Config Used

# Paste your config here


### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version: latest (ghcr.io/hkuds/lightrag:latest)
- Operating System: Debian trixie (base image)
- Python Version: 3.11
- Related Issues: N/A

danielaskdd:
## Description

Hardens the official `ghcr.io/hkuds/lightrag` images (`latest` + `lite`) so they pass enterprise security scanners (Prisma/Twistlock), which currently block deployment in regulated environments. Fixes all three issues raised in #3296:

1. **Vulnerable dependencies** — the venv shipped vulnerable `starlette`/`cryptography`/`python-multipart`. Because the image installs via `uv sync --frozen`, the fix has to live in `uv.lock` (a downstream `pip install --upgrade` can't patch the uv venv that scanners inspect).
2. **Runs as root** — violates CIS Docker 4.1.
3. **trixie perl CVE** — the floating `python:3.12-slim` base resolved to Debian trixie (perl 5.40.x, CVE-2026-12087, no patch yet).

The non-root change is designed to **not break existing deployments**: containers still start as root, fix the ownership of the mounted data dir, then drop privileges via `gosu`, so previously root-owned bind-mount/PVC data is adopted on upgrade and stays readable/writable.

## Related Issues

Fixes #3296

## Changes Made

- **Dependency CVEs** (`pyproject.toml`, `uv.lock`): add security floors in the `api` extra and re-lock to surgical fix versions (no major jumps):
  - starlette `1.0.0 → 1.3.1` (CVE-2026-54283, CVE-2026-48818)
  - cryptography `48.0.0 → 48.0.1` (GHSA-537c-gmf6-5ccf)
  - python-multipart `0.0.28 → 0.0.30` (CVE-2026-53539)
- **Non-root** (`Dockerfile`, `Dockerfile.lite`, new `docker-entrypoint.sh`): install `gosu` (bookworm main, 1.14), create a fixed uid/gid 1000 `lightrag` user with a home dir, set `HOME` + pip/uv cache dirs (so `pipmaster`'s runtime installs stay writable), and split `ENTRYPOINT`/`CMD`. The entrypoint chowns the data dirs then `exec gosu lightrag …`; when started non-root (compose `user:` / k8s `runAsUser`) it skips the chown and execs directly. Two backward-compat guards:
  - Splitting `ENTRYPOINT`/`CMD` would otherwise break `docker run <image> --port 9622` (flags replace `CMD`, so `gosu lightrag --port …` would try to exec `--port`). The entrypoint detects a leading-`-` first arg and prepends the server command, preserving the documented flag-append usage.
  - It chowns not just `/app/data` but also the env-configured `WORKING_DIR`/`INPUT_DIR`/`PROMPT_DIR`/`TIKTOKEN_CACHE_DIR` (guarded against unset/system-root values), so deployments that point data outside `/app/data` (e.g. the multi-site example) keep working under the non-root user.
- **Base image** (`Dockerfile`, `Dockerfile.lite`): pin the final stage to `python:3.12-slim-bookworm` (keeps Python 3.12 for venv compatibility, lowers perl exposure, aligns the Debian release with the builder stage).
- **Docs** (`docs/DockerDeployment.md`): add a "Container security (non-root)" section documenting uid 1000, automatic ownership adoption (incl. custom `WORKING_DIR`), the `.env` readability requirement, and the start-as-root/drop-privilege trade-off.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

**Verified locally** (both images built for amd64):

- Process runs as `uid=1000(lightrag)`; base `VERSION_CODENAME=bookworm`; venv carries 1.3.1 / 48.0.1 / 0.0.30.
- **Seamless upgrade**: a volume pre-populated as root (simulating pre-upgrade data) is read and written fine after upgrade — the entrypoint re-owns it to uid 1000.
- tiktoken cache (`/app/data/tiktoken`) is owned by uid 1000 and writable (cache-miss writes won't raise, since `TIKTOKEN_CACHE_DIR` makes tiktoken re-raise on write failure).
- 347 API tests pass (`pytest -k "api or health or auth or route or upload or server"`); only a non-blocking starlette `TestClient`/httpx deprecation warning.

**Known trade-offs:**

- The image starts as root and drops privileges at runtime (postgres/redis official-image pattern), so the running process is non-root but the image `config.User` is still root. Scanners that judge CIS 4.1 purely by the `USER` directive may still flag it. This trade-off was chosen deliberately to keep the bundled `docker-compose.yml` a seamless upgrade for existing root-owned data; a strict `USER 1000` would require users to manually migrate data ownership.
- `.env` handling: the entrypoint **does not** chown `/app/.env` (only `/app/data`). Chowning a bind-mounted `.env` would change the *host* file's owner to uid 1000 and force the host user to `sudo` just to edit their config. `.env` only needs to be *readable* by uid 1000, which the default `0644` already satisfies — so the host keeps ownership and edits it freely. A `0600` `.env` owned by another uid must be made readable (`chmod`/`chown` on the host) or supplied via environment variables (`env_file:`/`environment:`, k8s `env`/`envFrom`); otherwise the server exits at startup with a clear `PermissionError`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,7 +66,10 @@ RUN mkdir -p /app/data/tiktoken \
     if [ -n "${status:-}" ] && [ "$status" -ne 0 ] && [ "$status" -ne 2 ]; then exit "$status"; fi
 
 # Final stage
-FROM python:3.12-slim
+# Pin to bookworm: keeps Python 3.12 (venv compat with the builder stage) while
+# avoiding Debian trixie's perl 5.40.x exposure (CVE-2026-12087, no patch yet),
+# and aligns the final Debian release with the builder (also bookworm).
+FROM python:3.12-slim-bookworm
 
 WORKDIR /app
 
@@ -104,7 +107,30 @@ ENV WORKING_DIR=/app/data/rag_storage
 ENV INPUT_DIR=/app/data/inputs
 ENV PROMPT_DIR=/app/data/prompts
 
+# Create a non-root user (CIS Docker 4.1) and install gosu for privilege drop.
+# Fixed UID/GID 1000 gives predictable ownership for bind-mounts / PVCs.
+# chown -R /app MUST run after every data COPY above so the venv (pipmaster
+# installs packages at runtime), data dirs, and the tiktoken cache are writable.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends gosu \
+    && rm -rf /var/lib/apt/lists/* \
+    && groupadd -g 1000 lightrag \
+    && useradd -u 1000 -g lightrag -m -d /home/lightrag -s /usr/sbin/nologin lightrag \
+    && chown -R lightrag:lightrag /app /home/lightrag
+
+# HOME and cache dirs for the non-root user so pipmaster's runtime pip installs
+# never fall back to an unwritable /root or a missing HOME.
+ENV HOME=/home/lightrag \
+    XDG_CACHE_HOME=/home/lightrag/.cache \
+    PIP_CACHE_DIR=/home/lightrag/.cache/pip \
+    UV_CACHE_DIR=/home/lightrag/.cache/uv
+
+# Entrypoint starts as root, fixes mount ownership, then drops to lightrag.
+COPY docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+RUN chmod +x /usr/local/bin/docker-entrypoint.sh
+
 # Expose API port
 EXPOSE 9621
 
-ENTRYPOINT ["python", "-m", "lightrag.api.lightrag_server"]
+ENTRYPOINT ["docker-entrypoint.sh"]
+CMD ["python", "-m", "lightrag.api.lightrag_server"]
--- a/Dockerfile.lite
+++ b/Dockerfile.lite
@@ -66,7 +66,10 @@ RUN mkdir -p /app/data/tiktoken \
     if [ -n "${status:-}" ] && [ "$status" -ne 0 ] && [ "$status" -ne 2 ]; then exit "$status"; fi
 
 # Final stage
-FROM python:3.12-slim
+# Pin to bookworm: keeps Python 3.12 (venv compat with the builder stage) while
+# avoiding Debian trixie's perl 5.40.x exposure (CVE-2026-12087, no patch yet),
+# and aligns the final Debian release with the builder (also bookworm).
+FROM python:3.12-slim-bookworm
 
 WORKDIR /app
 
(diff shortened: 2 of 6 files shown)


RedEye1605 (commented):
Good security-focused fix. Addressing multiple CVEs (CVE-2026-12087, CVE-2026-53539, CVE-2026-54283/CVE-2026-48818) in one PR is efficient, and pinning the Debian base to bookworm in both Dockerfile stages is the right call for consistency.

The CIS Docker 4.1 non-root user implementation is well thought out. Specifically:
- Using fixed UID/GID 1000 for predictable bind-mount ownership is cleaner than random IDs
- The `|| true` on `chown` handles read-only mounts gracefully without silent failures
- The deliberate decision not to chown `/app/.env` is the right call — changing host file ownership via a bind-mount would be surprising and disruptive

One small suggestion: the data dir path `/app/data` is hardcoded in `docker-entrypoint.sh`. If this ever diverges from `ENV WORKING_DIR=/app/data/rag_storage` in the Dockerfile, the entrypoint will miss fixing ownership. Consider using `$WORKING_DIR` instead:

```sh
chown -R lightrag:lightrag "\${WORKING_DIR:-/app/data}" 2>/dev/null || true
```

This would make the entrypoint resilient to such drift. Otherwise, this is a clean and responsible security hardening PR — good work.

danielaskdd:
@RedEye1605 Thanks for the careful review — and great catch on the hardcoded `/app/data`. That drift concern is exactly right, and it's now resolved in the latest revision.

The entrypoint no longer chowns a single hardcoded path. It iterates over `/app/data` **plus** every data location configurable via env — `WORKING_DIR`, `INPUT_DIR`, `PROMPT_DIR`, and `TIKTOKEN_CACHE_DIR` — so it stays correct even if those drift from the Dockerfile defaults or point outside `/app/data` (e.g. the multi-site deployment example). Unset values and system roots (`/`, `/etc`, …) are skipped so a stray `WORKING_DIR=/` can't trigger a catastrophic `chown -R /`:

```sh
for _d in /app/data "$WORKING_DIR" "$INPUT_DIR" "$PROMPT_DIR" "$TIKTOKEN_CACHE_DIR"; do
    case "$_d" in
        ""|/|/bin|/boot|/dev|/etc|/home|/lib|/lib64|/proc|/root|/run|/sbin|/sys|/usr|/var) continue ;;
    esac
    [ -d "$_d" ] && chown -R lightrag:lightrag "$_d" 2>/dev/null || true
done
```

I went a bit beyond the single-`$WORKING_DIR` suggestion to also cover the input/prompt/tiktoken dirs, since any of them can be relocated independently. Verified locally: a custom `WORKING_DIR=/data/site01/storage` on a root-owned volume is adopted and writable after the drop to uid 1000, while `WORKING_DIR=/` is correctly skipped.

Thanks again for the thorough and constructive feedback!

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

Issue #3315, opened 2026-06-24
[Feature Request]:Support remote image storage (S3/OSS) for multimodal document parsing


forever-Li13:
### Do you need to file a feature request?

- [x] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

### Feature Request Description

##Summary
When parsing multimodal documents (e.g., PDF with images) via MinerU / Docling / Native adapters, extracted images are currently stored in the local *.blocks.assets/ directory, and local relative paths are written into drawings.json and blocks.jsonl. There is no built-in mechanism to upload images to remote object storage (S3, OSS, etc.) and store the corresponding remote URLs in the knowledge base.

##Why this matters
1. Frontend consumption: The LightRAG WebUI and any external consumers need to display images referenced in the knowledge graph. Local paths like foo.blocks.assets/bar.png only resolve on the machine that performed the parsing; they cannot be served to remote clients.
2. Storage efficiency: For deployments processing large volumes of PDFs, image assets can consume significant disk space. Object storage is purpose-built for this workload and offers lifecycle management, CDN acceleration, etc.
3. Consistency with the storage abstraction philosophy: LightRAG already abstracts KV / Vector / Graph / DocStatus storage behind pluggable backends (Neo4j, Milvus, PostgreSQL, etc.). Image asset storage is the only remaining piece hardcoded to the local filesystem.

PR #3331, opened 2026-06-25
feat: Add support for S3 asset uploads

forever-Li13:
## Description

Add optional S3-compatible object storage upload for sidecar image assets. When enabled via environment variables, materialized image assets (charts, figures, etc.) are automatically uploaded to S3 after the sidecar writer creates them, and remote URLs are used in the output files for front-end display.

## Related Issues

[#3315 ](https://github.com/HKUDS/LightRAG/issues/3315)

## Changes Made

| File | Change |
|------|--------|
| `lightrag/sidecar/s3_uploader.py` | **New** — S3 upload module, reads config from env vars, uploads assets via boto3, returns `{ref: remote_url}` mapping |
| `lightrag/sidecar/writer.py` | Integrate S3 upload into the sidecar write pipeline (Stage 1b); remote URLs used in `blocks.jsonl` placeholders, `drawings.json` gets a `remote_url` field while preserving local `path` |
| `pyproject.toml` | Add `s3-upload` optional dependency group (`boto3>=1.26.0,<2.0.0`) |

### How it works

1. After `_materialize_assets` (Stage 1), `upload_assets_to_s3` is called (Stage 1b)
2. If `LIGHTRAG_ASSET_UPLOAD_ENABLED` is not set, the function returns `None` — **zero overhead when disabled**
3. When enabled, assets are uploaded to the configured S3-compatible endpoint
4. Remote URLs flow into two places:
   - **`blocks.jsonl`**: placeholders use the remote URL (for front-end display)
   - **`drawings.json`**: a new `remote_url` field is added; the local `path` is always preserved so VLM analysis reads from disk

### Configuration (environment variables)

| Variable | Required | Description |
|----------|----------|-------------|
| `LIGHTRAG_ASSET_UPLOAD_ENABLED` | Yes | `true`/`1`/`yes` to enable |
| `LIGHTRAG_ASSET_UPLOAD_ENDPOINT` | Yes | S3-compatible endpoint URL |
| `LIGHTRAG_ASSET_UPLOAD_BUCKET` | Yes | Target bucket name |
| `LIGHTRAG_ASSET_UPLOAD_ACCESS_KEY` | Yes | Access key (falls back to `AWS_ACCESS_KEY_ID`) |
| `LIGHTRAG_ASSET_UPLOAD_SECRET_KEY` | Yes | Secret key (falls back to `AWS_SECRET_ACCESS_KEY`) |
| `LIGHTRAG_ASSET_UPLOAD_PREFIX` | No | Key prefix (default: `lightrag/`) |
| `LIGHTRAG_ASSET_UPLOAD_REGION` | No | AWS region (default: `us-east-1`) |
| `LIGHTRAG_ASSET_UPLOAD_PUBLIC_URL_PREFIX` | No | Custom public URL prefix (e.g. CDN) |

## Checklist

- [x] Changes tested locally
- [x] Code reviewed


danielaskdd:
Thanks for the contribution, @forever-Li13 — and for tackling #3315, which is a real pain point. I'd like to share some architectural feedback before we move forward, because I think this PR is solving the right problem in a way that's narrower than where LightRAG needs to go.

## The concern: uploading only the image assets is a point solution

This PR uploads the *materialized sidecar image assets* to S3 as a post-step (Stage 1b) and writes remote URLs into `blocks.jsonl` / `drawings.json`, while everything else stays on local disk. The trouble is that `INPUT_DIR` is not just images — it holds three different things:

1. **User-uploaded source files** (`/documents/upload` streams bytes to `input_dir/<file>`)
2. **Parser intermediates / caches** (`*.mineru_raw/`, `*.docling_raw/`, `*.native_raw/`)
3. **Sidecar artifacts + archived sources** (`__parsed__/<name>.parsed/` with `*.blocks.jsonl`, `*.drawings.json`, `*.tables.json`, `*.blocks.assets/`, plus the moved-aside source)

With this PR, the sources, the parser intermediates, and the `blocks.jsonl` itself still live only on the local filesystem, and VLM analysis still hard-depends on reading images from local disk (as the PR notes, "local path is always preserved so VLM analysis reads from disk"). Scan/dedup/delete/archive are all still local `iterdir`/`glob`/`unlink`/`rmtree`.

So it doesn't actually move `INPUT_DIR` off the local disk. That matters because the direction we want LightRAG to go is **distributed / multi-replica deployment**, where:
- any replica must be able to see uploads and sidecars written by any other replica;
- dedup, scan, deletion and **query-time traceability of the original file/assets** must be consistent across replicas;
- data lifecycle should be decoupled from the container/PVC lifecycle.

A per-asset upload can't get us there.

## The direction: back the *entire* `INPUT_DIR` with pluggable object storage

The idea is to turn `INPUT_DIR` from "a local `Path`" into "a logical namespace + a pluggable backend" — a `FileStore` abstraction analogous to the existing `kg/` storage backends:

- **`LocalFileStore`** (default, zero-config, byte-for-byte identical to today's behavior)
- **`S3FileStore`** (S3-compatible; reuses the `boto3` dependency you already added here)

Core interface: `put_stream / get_bytes / list / delete_prefix / copy / move / public_url`, plus two bridge context managers — `materialize(key)` (download an object to a local temp file for tools that need a real path, e.g. the external mineru/docling adapters and VLM) and `staging_dir()` (write a batch of artifacts locally, then flush back to the store).

A few things that make this tractable:

- **The seam already exists.** `full_docs.sidecar_location` is already a scheme-based URI, and `resolve_sidecar_uri()` already reserves `s3://` (it returns `None` for non-`file://` today — see the comment block in `utils_pipeline.py`). We just need to route by scheme to the right store. Historical `file://` URIs keep resolving, so it's backward compatible.
- **Client traceability (#3315)** becomes `store.public_url(key)`: the S3 backend returns a **presigned GET URL** (302 redirect, offloads bandwidth, supports expiry + optional CDN prefix); the local backend serves bytes via an endpoint. This naturally supersedes the `remote_url` field — when the store *is* S3, the asset is already there, no extra upload needed.
- **VLM** reads the local copy produced by `staging_dir` in the same parse batch (zero extra download), and only `materialize`s from the store when re-analyzing on a different replica — which removes the "must keep a local path forever" constraint.
- Rollout can be **phased and fully backward-compatible** (abstraction first with the local backend → migrate touchpoints → S3 backend → traceability endpoint → migration tool + drop the inputs PVC in k8s), with `INPUT_STORAGE=local` always the default.

Two smaller notes for whichever direction we take: the env vars would be worth converging onto a single `S3_*` namespace, and `verify=False` on the boto3 client (hardcoded here) should default to TLS verification with an opt-out.

## Ask

Would you be interested in implementing this broader version? If so, the ideal next step is a **detailed design proposal** (the `FileStore` interface, the touchpoint mapping, and the phased plan) that we can review and reach consensus on *before* implementation — that way the work lands cleanly and we avoid churn. I've done a fair bit of the touchpoint inventory already and am happy to share it and collaborate on the design.

Either way, really appreciate you pushing on this area. Let me know what you think.

---

forever-Li13:
Thanks for the detailed and thoughtful review — I really appreciate the architectural perspective and the time you put into laying this out.
You're right that the current PR is a point solution, and I agree that the FileStore abstraction is the right direction. 
I'd be happy to take on the broader version. The phased, backward-compatible rollout you described (abstraction first → local backend → S3 backend → traceability endpoint → migration tool) sounds like the right plan.
A couple of quick follow-ups as I think about the design:
• Could you share the touchpoint inventory you mentioned? That would save time and make sure I don't miss any call sites.
• For the FileStore interface — do you have a rough sketch of the method signatures in mind, or do you want me to draft one based on the interface you outlined (put_stream / get_bytes / list / delete_prefix / copy / move / public_url + materialize() / staging_dir())?
• Regarding the two smaller notes: agreed on consolidating env vars under a single S3_* namespace, and the verify=False should definitely be opt-out, not hardcoded — I'll fix that in the S3 backend regardless.
Happy to write up a detailed design proposal as the next step. Let me know how you'd like to collaborate on it.
Either way, thanks again for the guidance on this. Really appreciate it.

danielaskdd:
One more thing, and I want to be transparent about it: the broader design I sketched above was drafted with AI assistance, and I haven't yet carefully evaluated it for soundness myself. Please treat it as a **starting point for discussion**, not a vetted spec — some of the specifics (the `FileStore` interface, the touchpoint mapping, the `materialize`/`staging_dir` caching for VLM) may well be off or need rethinking once we dig in.

So none of it is meant as a hard requirement on you. I'd genuinely value your judgment here — you know this PR's code path best. If anything in there looks wrong or over-engineered, push back.

And thank you again for the contribution and for engaging on this — really appreciate the effort. Looking forward to your thoughts.

---

danielaskdd:
For reference, I've written up the full design as a document so it's easier to read and comment on than these inline summaries:

- English: https://github.com/HKUDS/LightRAG/blob/claude/determined-turing-0pi8qs/docs/S3InputStorageDesign.md
- 中文: https://github.com/HKUDS/LightRAG/blob/claude/determined-turing-0pi8qs/docs/S3InputStorageDesign-zh.md

Same caveat as above — it's an AI-assisted draft I haven't fully vetted, so treat it as a discussion starting point.

---

danielaskdd:
One suggestion on logistics: since this has grown into an architectural discussion (a pluggable storage backend for the whole `INPUT_DIR`) that's bigger than this single PR, would you be up for opening a thread under [Discussions](https://github.com/HKUDS/LightRAG/discussions) and putting the consolidated design proposal in the opening post?

A Discussion is a better home than a PR comment thread for this kind of design conversation — it keeps the proposal editable as we converge, supports threaded replies, and is easier for others to find and weigh in on. We can keep this PR focused on the concrete image-upload change and link it from there.

Happy to help seed the post with the design doc content if that's useful. What do you think?

---

PR #3357, opened 2026-07-05
fixed an issue where js protoype names would lead to the graph not being rendered

Xaverrrrr:
Graphology doesn't sanitize edge names apparently, which lead to issues when graphs contained edges with JS prototype names, such as hasOwnProperty. 

To fix this, we implement "hasEdgeSafe" instead of "hasEdge"

Changes were made using Claude code, reviewed manually and tested locally


danielaskdd:
@codex review

danielaskdd:
Pushed two commits (`48a7f34f`, `142ea17d`) addressing the Codex review finding, which I verified against graphology 0.26.0 as a **real, newly-introduced regression** (the original `hasEdge` returned `false` correctly in the same scenario).

**Root cause.** Switching from `hasEdge(source, target)` to `edge(source, target)` fixed the `hasOwnProperty`-method-shadowing crash, but `edge()` still reaches adjacency by prototype-walking property lookup. When a node gains a neighbor literally named `__proto__`, graphology's `adjacency['__proto__'] = edgeData` reassigns that adjacency object's **prototype** to the `EdgeData` instance. Afterwards `edge(A, 'source')` / `edge(A, 'target')` read the inherited `EdgeData.source`/`.target` node objects and return a phantom key, so `hasEdgeSafe` falsely reports a duplicate and the build loop **silently drops a real edge** to an entity named `source`/`target`.

**Fix.** Validate `edge()`'s key with the Map-backed single-arg `hasEdge(edgeKey)`, which does no prototype walk, so any phantom node key collapses to `false`:

```ts
export const hasEdgeSafe = (graph, source, target) => {
  if (!graph.hasNode(source) || !graph.hasNode(target)) return false
  const edgeKey = graph.edge(source, target)
  return edgeKey !== undefined && graph.hasEdge(edgeKey)
}
```

I also extracted the two prototype-safety helpers into `src/utils/graphEdgeSafety.ts` (only depends on graphology) and corrected the stale comment that claimed "inherited junk can never masquerade as a real edge" — the Codex scenario disproved exactly that line.

**Tests** (`graphEdgeSafety.test.ts`, 10 cases): `hasOwnProperty` neighbor no longer throws; `__proto__` neighbor no longer yields phantom `source`/`target` edges; the build loop keeps real edges while still deduping genuine duplicates. The regression case fails against the old `edge() !== undefined` implementation, so it genuinely locks the fix.

`tsc --noEmit` clean, `eslint` clean, `bun test` 51/51 pass.

danielaskdd:
@codex review

PR #3359, opened 2026-07-06
perf(operate): tokenize each description once in entity/relation summary map-reduce

hata33:
## Why

`_handle_entity_relation_summary` runs the map-reduce description merge for **every entity and every edge** during ingestion (it is invoked once per entity from `_merge_nodes_then_upsert` and once per edge from `_merge_edges_then_upsert`, both fanned out as N concurrent tasks under a semaphore in `extract_entities`). On each map-reduce iteration it tokenized every description **twice** — once to compute `total_tokens`, then again while building chunks — doubling the tiktoken BPE cost on the hottest CPU op in the merge stage. For a 1000-entity merge with a handful of descriptions each, that is thousands of redundant encodes per ingestion batch.

(Self-discovered while profiling the ingestion path; no tracking issue.)

## Root cause

`lightrag/operate.py`, function `_handle_entity_relation_summary`. Two passes walk the **same** `current_list` in the same `while` iteration:

```python
# Pass 1 — sum total tokens (operate.py:316-318)
for i, desc in enumerate(current_list, start=1):
    total_tokens += len(tokenizer.encode(desc))   # encode #1 per desc

# ... (no mutation of current_list between the passes) ...

# Pass 2 — build chunks (operate.py:356-357)
for i, desc in enumerate(current_list, start=1):
    desc_tokens = len(tokenizer.encode(desc))      # encode #2, same desc
```

`current_list` is unchanged between the two passes (Pass 2 is only reached when the list does not fit the context window), so the per-description counts computed in Pass 1 are simply discarded and recomputed in Pass 2.

## What changed

- Pass 1 now records each `len(tokenizer.encode(desc))` into a `desc_token_counts` list while summing `total_tokens`.
- Pass 2 iterates `zip(current_list, desc_token_counts)` and reuses the counts instead of re-encoding.

Externally observable behavior is **unchanged** — same input list, same iteration order, identical chunk boundaries and `current_tokens` arithmetic. No core runtime (pipeline / shared_storage) semantics touched; the `_cooperative_yield` cadence is preserved in both passes.

## Side effect analysis

- Default value: not changed
- Backward compatibility: output is byte-identical (verified by the regression test)
- Security boundary: none (pure compute refactor)

## Validation

- `python -m pytest tests/extraction/test_summary_map_reduce_perf.py -v` → **2 passed** (one asserts the chunking output is preserved; one asserts each description is encoded exactly once per map-reduce iteration)
- `python -m pytest tests/extraction/test_summary_sanitization.py tests/extraction/test_summary_map_reduce_perf.py -v` → **6 passed** (no regression in the existing summary tests)
- `ruff check lightrag/operate.py` → All checks passed
- `ruff format --check lightrag/operate.py` → clean

## AI assistance

**Tool(s) used:** Claude Code
**How:** AI profiled the ingestion path, located the double-encode, implemented the memoization, and wrote the regression test. I reviewed every line and verified the diff is behavior-preserving.

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -313,8 +313,13 @@ async def _handle_entity_relation_summary(
         # Calculate total tokens in current list while periodically yielding so
         # a large merge does not monopolize the event loop in single-worker mode.
         total_tokens = 0
+        # Tokenize each description once; the chunk-building pass below
+        # reuses these counts instead of re-encoding the same strings.
+        desc_token_counts = []
         for i, desc in enumerate(current_list, start=1):
-            total_tokens += len(tokenizer.encode(desc))
+            tokens = len(tokenizer.encode(desc))
+            desc_token_counts.append(tokens)
+            total_tokens += tokens
             await _cooperative_yield(i, every=32)
 
         # If total length is within limits, perform final summarization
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3251, opened 2026-06-12
🐛 fix(mongo): close MongoClient on ClientManager release

skymacro:
## Problem

`ClientManager.get_client()` creates an `AsyncMongoClient` but only stores the database reference (`client.get_database()`). The original client object is lost after `get_client()` returns, so `release_client()` can never close the underlying connection — it only sets the db reference to `None`.

`AsyncMongoClient.__del__` does **not** close connections — it only emits a warning (`Unclosed AsyncMongoClient`). Connections in the pool remain open until Python GC happens, which is non-deterministic in long-running server processes.

Under repeated init/finalize cycles (common in web servers), MongoDB connections accumulate and can eventually hit the server connection limit.

## Fix

- Store the `AsyncMongoClient` alongside the db in `_instances`
- In `release_client()`, call `client.close()` when ref_count reaches 0
- Clear both client and db references after closing

## Changes

- `lightrag/kg/mongo_impl.py` — 5 lines added
- `tests/kg/mongo_impl/test_client_manager_lifecycle.py` — 4 new unit tests (all mock-based, no MongoDB instance required)

## Test Results

```
tests/kg/mongo_impl/test_client_manager_lifecycle.py::TestClientManagerLifecycle::test_release_client_closes_connection_when_ref_count_zero PASSED
tests/kg/mongo_impl/test_client_manager_lifecycle.py::TestClientManagerLifecycle::test_release_client_keeps_connection_with_multiple_refs PASSED
tests/kg/mongo_impl/test_client_manager_lifecycle.py::TestClientManagerLifecycle::test_release_client_noop_for_wrong_db PASSED
tests/kg/mongo_impl/test_client_manager_lifecycle.py::TestClientManagerLifecycle::test_release_client_noop_for_none PASSED
4 passed
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3360, opened 2026-07-06
perf(kg): close Milvus and Qdrant clients in finalize() to release connections

hata33:
## Why

`MilvusVectorDBStorage.finalize()` and `QdrantVectorDBStorage.finalize()` flush pending writes but **never release the underlying client**, leaking one gRPC/HTTP transport per vector storage per process. A LightRAG instance creates 3 vector storages (`entities_vdb`, `relationships_vdb`, `chunks_vdb`), so a 4-worker Gunicorn deployment leaks 12 long-lived transports to the vector server with no close path.

Every other server-backed storage closes its client in `finalize()`: Neo4j driver (`neo4j_impl.py`), Postgres pool, Mongo client (#3251), OpenSearch `ClientManager`. This continues that pattern (#3251 Mongo, #3261 Anthropic).

## Root cause

Both `finalize()` methods flushed and surfaced unflushed-buffer errors but stopped there. Their docstrings claimed the client was safe to leave to GC:

- `lightrag/kg/milvus_impl.py:2872` — *"Milvus has no client connection to release (the MilvusClient is stateless …)"*
- `lightrag/kg/qdrant_impl.py:1290` — *"Qdrant has no client connection that needs explicit release here (… torn down on GC)"*

That claim is contradicted inside the same Milvus file: `_rebuild_milvus_client()` (milvus_impl.py:513) explicitly calls `old_client.close()` to release the gRPC channel, with its own docstring noting *"close() is best-effort — on a dead channel it is a no-op release"* — i.e. on a live channel it is a real release. So `close()` is meaningful and was simply not wired into `finalize()`.

## What changed

- After the flush (so the flush can still use the client), each `finalize()` calls `self._client.close()` best-effort, wrapped in `try/except` that logs on failure — identical to the pattern already used by `_rebuild_milvus_client()`.
- Updated both `finalize()` docstrings to reflect that the client is now released.
- `self._client` is intentionally **not** set to `None`: `close()` already releases the transport, and keeping the reference preserves post-finalize inspection (and the existing deferred-embedding tests that assert on `s._client.upsert` / `s._client.delete` after a clean flush).

Externally observable behavior during normal operation is unchanged — `finalize()` is a shutdown/teardown path; the only new effect is that the transport is released eagerly instead of waiting for GC.

## Side effect analysis

- Default value: not changed
- Backward compatibility: no API or runtime behavior change; `finalize()` still flushes first and surfaces the same `RuntimeError` on unflushed buffers (the close runs after the flush and before those checks)
- Security boundary: none opened (closes a connection; does not open any path)
- Double-finalize safety: `close()` on an already-closed Milvus channel is a documented no-op; Qdrant failures are caught by the `try/except`

## Validation

- `python -m pytest tests/kg/milvus_impl/test_milvus_deferred_embedding.py tests/kg/qdrant_impl/test_qdrant_deferred_embedding.py -q` → **56 passed** (2 new regression tests asserting `finalize()` calls `client.close()`, plus all existing flush/upsert/drop tests unchanged)
- `ruff check` / `ruff format --check` on the 4 changed files → clean

## AI assistance

**Tool(s) used:** Claude Code
**How:** AI audited the storage backends for resource leaks, located the two missing `close()` calls, implemented the best-effort release + docstring updates, and wrote the regression tests. I reviewed every line and ran the full deferred-embedding suites to confirm no regression.

### DIFF
--- a/lightrag/kg/milvus_impl.py
+++ b/lightrag/kg/milvus_impl.py
@@ -2867,18 +2867,34 @@ class MilvusVectorDBStorage(BaseVectorStorage):
             return result
 
     async def finalize(self):
-        """Flush pending vector ops; surface unflushed data as RuntimeError.
+        """Flush pending vector ops, then release the Milvus gRPC channel.
 
-        Milvus has no client connection to release (the MilvusClient is
-        stateless from the storage layer's perspective), but we still need
-        to fail loudly when a transient bulk error left writes buffered —
-        the caller must not believe storage finalized cleanly.
+        Every other server-backed storage releases its client here (Neo4j
+        driver, Postgres pool, Mongo client, OpenSearch ClientManager); the
+        MilvusClient owns a gRPC channel that should be closed explicitly
+        too — ``close()`` is already used for that purpose by
+        ``_rebuild_milvus_client``. We still fail loudly when a transient
+        bulk error left writes buffered — the caller must not believe
+        storage finalized cleanly.
         """
         flush_error: Exception | None = None
         try:
             await self._flush_pending_vector_ops()
         except Exception as e:
             flush_error = e
+
+        # Release the gRPC channel after the flush so the flush can still use
+        # the client. Best-effort: close() on a dead channel is a no-op (see
+        # _rebuild_milvus_client). The connection is freed on every exit path,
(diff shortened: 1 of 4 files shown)


danielaskdd:
@codex review

PR #3365, opened 2026-07-07
fix(rerank): support flat payload format for Aliyun qwen3-rerank models

### LINKED ISSUE #3084: 阿里云百炼平台reranker模型接口有些乱，使用aliyun接口同时用qwen3-rerank模型会有问题
https://github.com/HKUDS/LightRAG/blob/405525a5e5b2c7d4385a0a3d4726accd285f9934/lightrag/rerank.py#L245-L254
# api接口问题
参考https://help.aliyun.com/zh/model-studio/text-rerank-api#d46aed9b6cmg3文档，qwen3-vl-rerank和gte-rerank-v2的paload是个格式的
```json
{
    "model": "gte-rerank-v2",
    "input":{
         "query": "什么是文本排序模型",
         "documents": [
         "文本排序模型广泛用于搜索引擎和推荐系统中，它们根据文本相关性对候选文本进行排序",
         "量子计算是计算科学的一个前沿领域",
         "预训练语言模型的发展给文本排序模型带来了新的进展"
         ]
    },
    "parameters": {
        "return_documents": true,
        "top_n": 2
    }
}
```
qwen3-rerank模型的payload是这个格式的
```json
{
        "model": "qwen3-rerank",
        "documents": [
                "文本排序模型广泛用于搜索引擎和推荐系统中，它们根据文本相关性对候选文本进行排序",
                "量子计算是计算科学的一个前沿领域",
                "预训练语言模型的发展给文本排序模型带来了新的进展"
        ],
        "query": "什么是文本排序模型",
        "top_n": 2,
        "instruct": "Given a web search query, retrieve relevant passages that answer the query."
}
```
# 解决
在不修改源代码的情况下使用aliyun百炼网站提供qwen3-rerank模型，接口可以使用RERANK_BINDING=cohere

# 额外
文档中写gte-rerank-v2模型快要下架了，后面建议使用qwen3-rerank作为替代

freya0926:
## Summary

Fixes #3084

The Aliyun DashScope rerank endpoint serves two incompatible API formats depending on the model (see the [Text Rerank API docs](https://help.aliyun.com/zh/model-studio/text-rerank-api)):

- **gte-rerank-\*** and **qwen3-vl-rerank** use a nested request body `{"model", "input": {"query", "documents"}, "parameters": {"top_n", ...}}` and return results under `output.results`.
- The **qwen3-rerank** series uses a flat, Cohere-style request body `{"model", "query", "documents", "top_n", "instruct"}` and returns top-level `results`.

`generic_rerank_api` hard-wired `request_format="aliyun"` to the nested shape, so using `RERANK_BINDING=aliyun` with `qwen3-rerank` failed (the request body is rejected / results come back empty because the flat response has no `output` key). The issue (reported in Chinese) also notes that Aliyun plans to retire gte-rerank-v2 and recommends qwen3-rerank as its replacement, and that the only workaround was misusing `RERANK_BINDING=cohere`.

## What's changed

- Added `is_aliyun_flat_rerank_model()`: detects the qwen3-rerank family by model name (case-insensitive, matches `qwen*rerank*` but excludes `qwen3-vl-rerank`, which keeps the nested format per the docs).
- In `generic_rerank_api`, when `request_format == "aliyun"` and the model is flat-format, fall back to the standard request builder and response parser, and drop the unsupported `return_documents` parameter.
- `env.example`: documented that qwen3-rerank uses the flat format (auto-detected) and needs the corresponding flat-format endpoint as `RERANK_BINDING_HOST`.
- Added `tests/rerank/test_aliyun_rerank_format.py`: 12 mocked-HTTP regression tests covering both model families (request payload shape assertions and response parsing).

## How it works

`RERANK_BINDING=aliyun` keeps working unchanged for gte-rerank-v2 and qwen3-vl-rerank; with `RERANK_MODEL=qwen3-rerank` the flat payload/response format is now selected automatically. Note the qwen3-rerank endpoint path differs from the gte one, so `RERANK_BINDING_HOST` must point at the flat (compatible) rerank endpoint.

## Tests

`python -m pytest tests/rerank/test_aliyun_rerank_format.py tests/chunker/test_rerank_chunking.py -v` — 31 passed. The 3 flat-format tests fail without the fix, reproducing the issue.

### DIFF
--- a/env.example
+++ b/env.example
@@ -189,6 +189,8 @@ RERANK_BINDING=null
 # RERANK_MAX_TOKENS_PER_DOC=480
 
 ### Aliyun Dashscope
+### qwen3-rerank series models use a flat payload format (auto-detected from model name)
+### and require the corresponding flat-format endpoint as RERANK_BINDING_HOST
 # # RERANK_MODEL=gte-rerank-v2
 # # RERANK_BINDING_HOST=https://dashscope.aliyuncs.com/api/v1/services/rerank/text-rerank/text-rerank
 # # RERANK_BINDING_API_KEY=your_rerank_api_key_here
--- a/lightrag/rerank.py
+++ b/lightrag/rerank.py
@@ -171,6 +171,32 @@ def aggregate_chunk_scores(
     return aggregated_results
 
 
+def is_aliyun_flat_rerank_model(model: str) -> bool:
+    """
+    Check whether an Aliyun DashScope rerank model uses the flat (standard)
+    request/response format instead of the nested input/parameters format.
+
+    Aliyun's qwen3-rerank series expects a flat payload
+    {"model", "query", "documents", "top_n", ...} and returns top-level
+    {"results": [...]}, while gte-rerank-* and qwen3-vl-rerank use the nested
+    {"model", "input": {...}, "parameters": {...}} payload and return
+    {"output": {"results": [...]}}.
+    Reference: https://help.aliyun.com/zh/model-studio/text-rerank-api
+
+    Args:
+        model: Rerank model name
+
+    Returns:
+        True if the model uses the flat (standard) format
+    """
+    model_lower = model.lower()
+    return (
+        model_lower.startswith("qwen")
+        and "rerank" in model_lower
+        and "-vl-" not in model_lower
+    )
+
+
 @retry(
     stop=stop_after_attempt(3),
     wait=wait_exponential(multiplier=1, min=4, max=60),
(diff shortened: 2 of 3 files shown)


danielaskdd:
Thanks a lot for digging into this and for the thorough tests, @freya0926 — the diagnosis here is correct.

After reviewing, we're going to resolve #3084 with a **docs-only** change (#3368) instead of adding the flat-format branch, for two reasons:

1. **`qwen3-rerank` uses a genuinely separate, Cohere-compatible endpoint** (`.../compatible-api/v1/reranks`), not just a different payload on the same URL. Its request/response is byte-for-byte the standard Cohere format that `cohere_rerank`/`generic_rerank_api` already produce, so `RERANK_BINDING=cohere` handles it correctly and semantically (pointing the Cohere binding at a Cohere-compatible endpoint is exactly its purpose).

2. Because the endpoint differs, this PR's `aliyun` path still requires the user to manually override `RERANK_BINDING_HOST` (the `aliyun` default points at the `text-rerank` endpoint). So the model-name auto-detection doesn't remove any configuration step, while the `qwen*rerank* && !-vl-` heuristic adds a name-matching maintenance surface that can misfire on future model names.

So the underlying issue was really a documentation gap. #3368 documents the two endpoints/formats and the correct `cohere` binding config in `env.example` and both API-Server guides.

Closing this in favor of #3368 — but your investigation and the reference links directly informed that fix, so thank you! 🙏

---

PR #3370, opened 2026-07-07
fix(kg): release Redis client and pool reference in finalize()

hata33:
## Problem

`RedisKVStorage` and `RedisDocStatusStorage` own a `redis.asyncio` client backed by a **shared, reference-counted** connection pool (`RedisConnectionManager`). They define a `close()` that releases the per-instance client and decrements the pool reference, but both classes inherit the base class's **no-op `finalize()`** (`lightrag/base.py:170`).

The standard shutdown path calls `finalize()` on every storage (`LightRAG.finalize_storages`), so on a normal shutdown **neither `close()` nor `release_pool()` ever runs**: the per-instance client is not released and the shared pool's reference count only ever grows. The pool is reclaimed only at process exit.

While fixing this, three further defects in the teardown path surfaced and are addressed here (see **Changes**).

## Root cause

- **Missing `finalize()`** — `RedisKVStorage` / `RedisDocStatusStorage` override `close()` but not `finalize()`, so they fall through to the no-op base. Every other server-backed storage closes its client in `finalize()` (Neo4j, Postgres, Mongo #3251, OpenSearch, Milvus/Qdrant #3360); `MongoKVStorage.finalize()` is the direct KV analogue.
- **Pool never actually disconnected** — in `redis.asyncio`, `ConnectionPool.disconnect()` is a **coroutine**. The old `release_pool` called it synchronously (never awaited), so even when the refcount reached 0 the sockets were never closed and an unawaited-coroutine warning was emitted.
- **`close()` not idempotent** — `close()` never cleared `self._redis_url`, so a second `close()` (finalize-after-`__aexit__`, or an explicit close + finalize) called `release_pool()` again. With a **shared** pool (Redis KV backend runs ~7 `RedisKVStorage` + 1 `RedisDocStatusStorage` on one pool), the second release stole a sibling's reference and could drop the refcount to 0, disconnecting the pool out from under a still-live sibling.
- **Init-error path masked the real exception** — the synchronous `__post_init__` error path used `asyncio.ensure_future(pool.aclose())`; with no running loop this raises `RuntimeError`, replacing the original initialization exception.

## Changes

`lightrag/kg/redis_impl.py`:

- **`finalize()` override** on both storages, delegating to `close()`.
- **Idempotent, re-entrancy- and cancellation-safe `close()`**: detaches `_redis` / `_redis_url` / `_pool` into locals and nulls them **before the first `await`** (a concurrent/second `close()` is then a complete no-op and cannot release the shared pool twice); closes the client, then releases the pool reference in a **`finally`** so the reference is released even if `aclose()` is cancelled (`CancelledError` is a `BaseException` and escapes `except Exception`); `CancelledError` is re-raised.
- **Real async pool teardown**: `RedisConnectionManager.release_pool()` and `close_all_pools()` are now `async` and actually `await pool.aclose()`. New sync `release_pool_ref()` decrements the refcount and returns the pool to disconnect (or `None`).
- **Safe background close from the sync init-error path**: new `schedule_pool_close()` / `_close_pool_safely()` — schedules the disconnect as a managed task (strong-referenced in `_cleanup_tasks` until done, errors swallowed), and is a **no-op when no loop is running** (never creates a coroutine, so it cannot mask the original init exception or emit an unawaited-coroutine warning).
- **Client teardown uses `aclose()`** (was the deprecated `Redis.close()`); with an externally-provided pool `auto_close_connection_pool = False`, so closing the client does not disconnect the shared pool.

`pyproject.toml` + `uv.lock`:

- **`redis>=5.0.0` → `redis>=5.0.1`**. `Redis.aclose()` / `ConnectionPool.aclose()` only exist from redis-py **5.0.1** (`Redis.close` is `@deprecated_function(version="5.0.1", reason="Use aclose() instead")`). The lock's resolved version is unchanged (8.0.0); only the declared floor moves, dropping a single patch release without narrowing Redis Server compatibility.

## Notes on API / behavior

- `RedisConnectionManager.release_pool()` and `close_all_pools()` changed from sync to **async**; `release_pool_ref()` / `schedule_pool_close()` are new. All in-repo callers are updated; `close_all_pools()` has no in-repo callers.
- Pool disconnect is now genuinely awaited before `finalize_storages()` returns (previously it was never awaited).
- Pool close remains **best-effort**: `aclose()` errors are logged, not propagated — they do not necessarily fail `finalize_storages()`.
- Shared-pool safety is preserved: sibling storages on the same URL are unaffected; the pool disconnects only when the last reference is released (refcount → 0). After the idempotency fix, double-finalize and finalize-after-`__aexit__` are safe (a second `close()` is a no-op).
- Security boundary: none opened (closes a connection; opens no path).

## Validation

- `./scripts/test.sh tests/kg/redis_impl/ -q` → **24 passed** (16 in the new `test_redis_finalize.py`, covering: finalize→close wiring; real-refcount double-close **not** stealing a sibling reference; manager last-reference disconnect / `close_all_pools`; init-error path with no loop / with loop / failing background `aclose()`; cancellation safety via mock and real refcount).
- **Counter-proof**: temporarily reverting the `close()` idempotency fix makes the shared-refcount test fail as expected — the test genuinely locks the fix.
- **Minimum version**: `uv run --with 'redis==5.0.1' python -c 'import redis; assert redis.__version__ == "5.0.1"'` then the suite on redis-py 5.0.1 → **24 passed** (confirms `aclose()` is available at the declared floor).
- `ruff check` / `ruff format --check` on the changed files → clean.

## AI assistance

**Tool(s) used:** Claude Code
**How:** AI audited the Redis storage teardown path, found the missing `finalize()` plus the three secondary defects (un-awaited pool disconnect, non-idempotent `close()`, init-error exception masking) and the redis-py 5.0.1 API-floor mismatch, implemented the fixes, and wrote the regression suite including a counter-proof and a minimum-version run. Reviewed line by line.


danielaskdd:
@codex review

hata33:
Thanks for the catch! Fixed in the latest push (`c42e615`).

`redis.asyncio.ConnectionPool.disconnect()` / `aclose()` are coroutines, so the previous sync `release_pool` never actually closed the sockets. The fix splits the concern:

- **`release_pool_ref`** (sync `@classmethod`) — decrements the reference count under the lock and returns the pool to disconnect (or `None`). Used in the sync `__post_init__` error path, where the returned pool is scheduled via `asyncio.ensure_future(pool.aclose())`.
- **`release_pool`** (now `async`) — calls `release_pool_ref`, then `await pool.aclose()` so sockets are closed before `finalize_storages()` returns. Awaited from `close()` / `finalize()`.
- **`close_all_pools`** (now `async`) — awaits `pool.aclose()` for each pool.

danielaskdd:
Pushed `7434f76`, which hardens the Redis teardown path beyond the original `finalize()` fix:

- **`close()` is now idempotent + cancellation-safe** — all instance state is detached before the first `await`, and the pool reference is released in a `finally`. This fixes a double-close (finalize-after-`__aexit__`) stealing a **sibling's** shared-pool reference and disconnecting the pool from under a live sibling.
- **Pool disconnect is now actually awaited** — `release_pool()` / `close_all_pools()` are async and `await pool.aclose()`; the previous sync `disconnect()` on an async pool was a never-awaited coroutine, so the pool was never really closed even at refcount 0.
- **Sync init-error path no longer masks the original exception** — background close is a managed task, and a no-op when no event loop is running.
- **`redis` floor bumped to `>=5.0.1`** — `aclose()` only exists from 5.0.1 (lock's resolved version unchanged at 8.0.0).

Tests expanded to **24 passing** (real-refcount double-close, init-error paths, cancellation), plus a counter-proof and a minimum-version run on redis-py 5.0.1. PR description updated accordingly.

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3371, opened 2026-07-07
fix(llm): close AsyncOpenAI client in nvidia_openai_embed

hata33:
## Problem

`nvidia_openai_embed` creates an `AsyncOpenAI` client per call but never closes it (`lightrag/llm/nvidia_openai.py:59-68`):

```python
openai_async_client = AsyncOpenAI(base_url=base_url)
response = await openai_async_client.embeddings.create(...)
return np.array([dp.embedding for dp in response.data])
```

Each embedding call therefore leaks the client's `httpx` connection pool until garbage collection. Over a long-running process (e.g. a busy ingest worker calling `embed` thousands of times) this accumulates one leaked pool per call.

## Root cause

The function constructs the client inline and uses it without an `async with` / `finally close`. The sibling embedding function already does it correctly — `openai_embed` holds its client in `async with openai_async_client:` (`lightrag/llm/openai.py:1006`) — and PR #3261 fixed the same class of leak for the Anthropic backend. `nvidia_openai_embed` was simply never updated to match.

## Changes

- Wrap the call body in `async with openai_async_client:` so the client is released on every exit path: success, error, **and each `@retry` attempt** (the client is created inside the function body, so tenacity re-creates and closes it on each retry). +4 / -3 lines in `nvidia_openai.py`; the returned embedding array is unchanged.

## Side effect analysis

- Default value: not changed
- Backward compatibility: no API or runtime behavior change; the embedding result is byte-identical. The only new effect is that the httpx pool is released eagerly instead of waiting for GC.
- Security boundary: none opened (closes a connection; opens no path)

## Validation

- `python -m pytest tests/llm/nvidia_impl/test_nvidia_client_cleanup.py -v` → **2 passed** (client closed on success and on error, mirroring `test_anthropic_client_cleanup.py`)
- `ruff check` / `ruff format --check` on the changed files → clean

## AI assistance

**Tool(s) used:** Claude Code
**How:** AI audited LLM backends for the same per-call client leak fixed in #3261, identified the missing close in `nvidia_openai_embed`, applied the `async with` pattern already used by `openai_embed`, and wrote regression tests mirroring the Anthropic cleanup suite. I reviewed every line and ran the tests to confirm.

### DIFF
--- a/lightrag/llm/nvidia_openai.py
+++ b/lightrag/llm/nvidia_openai.py
@@ -59,10 +59,14 @@ async def nvidia_openai_embed(
     openai_async_client = (
         AsyncOpenAI() if base_url is None else AsyncOpenAI(base_url=base_url)
     )
-    response = await openai_async_client.embeddings.create(
-        model=model,
-        input=texts,
-        encoding_format=encode,
-        extra_body={"input_type": input_type, "truncate": trunc},
-    )
-    return np.array([dp.embedding for dp in response.data])
+    # Hold the client in an async-with so its httpx connection pool is
+    # released on every exit path (success, error, and each @retry attempt),
+    # instead of leaking one pool per call until GC. Mirrors ``openai_embed``.
+    async with openai_async_client:
+        response = await openai_async_client.embeddings.create(
+            model=model,
+            input=texts,
+            encoding_format=encode,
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3372, opened 2026-07-07
fix: cross-dedup entity/relation descriptions against stored ones on re-merge

### LINKED ISSUE #3367: [Bug]: merge_nodes_and_edges accumulates entity/relation descriptions on any reprocess/resume (file pipeline + ainsert, not just custom chunks)

### Describe the bug

`_merge_nodes_then_upsert` merges an existing node's stored description with the newly extracted ones via `already_description + sorted_descriptions` (`lightrag/operate.py:~2167`), deduping only WITHIN the new batch, not between stored and new. The same is true for relations in `_merge_edges_then_upsert` (`operate.py:~2525`). Both feed `_handle_entity_relation_summary`, which — below the force-summary threshold — simply `separator.join(...)` without cross-dedup (`operate.py:~335`).

So re-running merge for a doc whose entities/relations already exist (any reprocess or resume) re-appends the same descriptions. Each reprocess reads the full stored description back into `already_description` and adds one more copy, so the fragment count grows by one per reprocess (N -> N+1): a single failed-then-retried entity goes 1 -> 2 (looks like a doubling), and repeated retries accumulate linearly.

Both reprocess paths do purge first (pipeline resume via `_purge_stale_extraction_if_resuming`; `ainsert_custom_chunks` via `_purge_doc_chunks_and_kg`), but purge discovers targets via the `full_entities`/`full_relations` indexes, which merge writes only in its final phase (`operate.py:~3253`). A merge that wrote partial graph state then failed leaves no index row -> purge finds nothing -> the retry accumulates.

Scope: `source_id`/`file_path` are deduped and safe; only descriptions accumulate. It is bounded (later re-summarization collapses it) but silently inflates entity/relation descriptions by one duplicate copy per failed-then-retried reprocess across all ingestion paths, and each reprocess re-reads the whole stored description so later merges also do more token work. Affects the file pipeline, `ainsert`, and `ainsert_custom_chunks` alike (all share `merge_nodes_and_edges`).

### Steps to reproduce

Minimal, no DB/LLM — drives the real merge round-trip (`get_node` -> merge -> `upsert_node`) against an in-memory graph. Verified on the latest `main`.

```python
import asyncio
import lightrag.operate as operate
from lightrag.operate import _merge_nodes_then_upsert, _handle_entity_relation_summary
from lightrag.constants import GRAPH_FIELD_SEP

class FakeTokenizer:
    def encode(self, s):  # token count == char count; thresholds kept slack
        return list(range(len(s)))

class MemGraph:  # minimal in-memory graph: real read-back + write round-trip
    def __init__(self):
        self.nodes = {}
    async def get_node(self, name):
        return self.nodes.get(name)
    async def upsert_node(self, name, node_data):
        self.nodes[name] = dict(node_data)

cfg = {
    "tokenizer": FakeTokenizer(),
    "summary_context_size": 1_000_000,
    "summary_max_tokens": 1_000_000,
    "force_llm_summary_on_merge": 6,
    "source_ids_limit_method": operate.SOURCE_IDS_LIMIT_METHOD_KEEP,
    "max_source_ids_per_entity": 10_000,
    "max_file_paths": 100,
    "file_path_more_placeholder": "...",
}
D = "Alice is a software engineer at Acme."
batch = {"entity_name": "ALICE", "entity_type": "person",
         "description": D, "source_id": "chunk-1", "file_path": "doc1.txt", "timestamp": 1}

async def main():
    g = MemGraph()
    for _ in range(3):  # 3 reprocesses of the SAME single description, no purge between
        await _merge_nodes_then_upsert("ALICE", [dict(batch)], g, None, cfg)
        frags = g.nodes["ALICE"]["description"].split(GRAPH_FIELD_SEP)
        print(f"stored fragments: {len(frags)}")
    joined, _ = await _handle_entity_relation_summary(
        "Relation", "ALICE~ACME", [D, D], GRAPH_FIELD_SEP, cfg, None)
    print("summary([D, D]) fragments:", len(joined.split(GRAPH_FIELD_SEP)))

asyncio.run(main())
```

### Expected Behavior

Re-merging a doc whose entities/relations already exist should be idempotent w.r.t. descriptions: a description already stored on a node/edge should not be appended again. Reprocessing the same content any number of times should leave the fragment count at 1 (for one unique description), not grow by one per reprocess.

### LightRAG Config Used

The accumulation is independent of user config; it reproduces with default merge settings. The knobs that matter (kept generous above so the no-LLM join branch is exercised):

```
force_llm_summary_on_merge: 6
summary_max_tokens: 1000000
summary_context_size: 1000000
source_ids_limit_method: KEEP
max_source_ids_per_entity: 10000
```

### Logs and screenshots

```
stored fragments: 1
stored fragments: 2
stored fragments: 3
summary([D, D]) fragments: 2
```

Node description after 3 reprocesses (real `upsert_node` value):

```
'Alice is a software engineer at Acme.<SEP>Alice is a software engineer at Acme.<SEP>Alice is a software engineer at Acme.'
```

### Additional Information

- LightRAG Version: 1.5.5 (latest `main`)
- Operating System: macOS (Darwin 25.5.0)
- Python Version: 3.12
- Related Issues: #3352 / #3353 (surfaced there; this is the shared core cause)
- Proposed fix: cross-dedup `already_description` against the new descriptions in `_merge_nodes_then_upsert` / `_merge_edges_then_upsert`, making re-merge idempotent regardless of whether purge found the prior index rows.

freya0926:
## Summary

Closes #3367.

`_merge_nodes_then_upsert` (`lightrag/operate.py:~2167`) and `_merge_edges_then_upsert` (`lightrag/operate.py:~2525`) combine `already_description` (read back from the graph) with the newly extracted descriptions via plain concatenation:

```python
description_list = already_description + sorted_descriptions
```

The new batch is deduped against itself (`unique_nodes`/`unique_edges`) but never against what's already stored. So any reprocess or resume of a document whose entities/relations already exist re-appends the same description every time — a single failed-then-retried entity goes 1 -> 2 -> 3 stored fragments per reprocess instead of staying idempotent at 1, and each reprocess re-reads the whole stored description, so later merges also do more token work.

Both reprocess paths purge first, but purge discovers targets via the `full_entities`/`full_relations` indexes, which merge only writes in its final phase — so a merge that wrote partial graph state then failed leaves no index row, purge finds nothing, and the retry accumulates. (Full analysis in the issue, including a from-scratch repro.)

## Fix

Dedup the new batch's sorted descriptions against `already_description` before concatenating, in both functions:

```python
already_description_set = set(already_description)
new_unique_descriptions = [
    desc for desc in sorted_descriptions if desc not in already_description_set
]
description_list = already_description + new_unique_descriptions
```

`source_id`/`file_path` were already deduped correctly and are untouched. `_handle_entity_relation_summary`'s own no-LLM join (`operate.py:~335`) needs no separate change — it now receives an already-deduped `description_list` from both call sites.

## Test plan

- [x] Added `tests/extraction/test_merge_description_dedup.py`:
  - `test_reprocessing_same_entity_description_is_idempotent` — merges the same single entity description 3 times, asserts stored fragment count stays at 1 (was 1/2/3 before the fix — verified by reverting the fix locally and confirming all 3 new tests fail with the exact 1/2/3 growth pattern from the issue)
  - `test_reprocessing_adds_only_genuinely_new_descriptions` — a genuinely new description is still appended once; only exact repeats of stored content are skipped
  - `test_reprocessing_same_relation_description_is_idempotent` — same guarantee for `_merge_edges_then_upsert`
- [x] `pytest tests/extraction/` — 106 passed (1 pre-existing, unrelated failure in `test_entity_extraction_stability.py` about a prompt-file error message string, reproduces identically on unmodified `main`)
- [x] `ruff check` / `ruff format --check` on changed files — clean

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -2163,8 +2163,15 @@ async def _merge_nodes_then_upsert(
         )
         sorted_descriptions = [dp["description"] for dp in sorted_nodes]
 
-        # Combine already_description with sorted new sorted descriptions
-        description_list = already_description + sorted_descriptions
+        # Combine already_description with new sorted descriptions, deduped
+        # against what's already stored so re-running merge for an entity that
+        # already exists (any reprocess/resume) doesn't re-append descriptions
+        # already present.
+        already_description_set = set(already_description)
+        new_unique_descriptions = [
+            desc for desc in sorted_descriptions if desc not in already_description_set
+        ]
+        description_list = already_description + new_unique_descriptions
         if not description_list:
             fallback_description = f"Entity {entity_name}"
             logger.warning(
@@ -2521,8 +2528,15 @@ async def _merge_edges_then_upsert(
         )
         sorted_descriptions = [dp["description"] for dp in sorted_edges]
 
-        # Combine already_description with sorted new descriptions
-        description_list = already_description + sorted_descriptions
+        # Combine already_description with new sorted descriptions, deduped
+        # against what's already stored so re-running merge for a relation that
+        # already exists (any reprocess/resume) doesn't re-append descriptions
+        # already present.
+        already_description_set = set(already_description)
+        new_unique_descriptions = [
+            desc for desc in sorted_descriptions if desc not in already_description_set
+        ]
+        description_list = already_description + new_unique_descriptions
         if not description_list:
             logger.error(f"Relation {src_id}~{tgt_id} has no description")
             raise ValueError(f"Relation {src_id}~{tgt_id} has no description")
--- a/tests/extraction/test_merge_description_dedup.py
+++ b/tests/extraction/test_merge_description_dedup.py
@@ -1,0 +1,175 @@
+"""Regression tests: re-merging an entity/relation that already exists must
+not re-append descriptions already stored on it.
+
+``_merge_nodes_then_upsert``/``_merge_edges_then_upsert`` combined
+``already_description`` (read back from the graph) with the newly extracted
+descriptions via plain concatenation, deduping only *within* the new batch.
+Any reprocess or resume of a document whose entities/relations already exist
+would therefore re-append the same description every time: a single
+failed-then-retried entity went 1 -> 2 -> 3 fragments instead of staying
+idempotent at 1.
+
+See https://github.com/HKUDS/LightRAG/issues/3367.
+"""
+
+import pytest
+
+from lightrag.constants import GRAPH_FIELD_SEP
+from lightrag.operate import _merge_edges_then_upsert, _merge_nodes_then_upsert
+from lightrag.utils import Tokenizer, TokenizerInterface
+
+
+class DummyTokenizer(TokenizerInterface):
+    """1:1 character-to-token mapping, matching other operate.py tests."""
+
+    def encode(self, content: str):
+        return [ord(ch) for ch in content]
+
+    def decode(self, tokens):
+        return "".join(chr(token) for token in tokens)
+
+
+def _make_global_config() -> dict:
+    tokenizer = Tokenizer("dummy", DummyTokenizer())
+    return {
+        "tokenizer": tokenizer,
+        "summary_context_size": 1_000_000,
+        "summary_max_tokens": 1_000_000,
+        "force_llm_summary_on_merge": 6,
+        "source_ids_limit_method": "KEEP",
+        "max_source_ids_per_entity": 10_000,
+        "max_source_ids_per_relation": 10_000,
+        "max_file_paths": 100,
+        "file_path_more_placeholder": "...",
+    }
+
+
+class MemGraph:
+    """Minimal in-memory graph storage exercising the real get/upsert round-trip."""
+
+    def __init__(self):
+        self.nodes = {}
+        self.edges = {}
+
+    async def get_node(self, name):
+        return self.nodes.get(name)
+
+    async def upsert_node(self, name, node_data):
+        self.nodes[name] = dict(node_data)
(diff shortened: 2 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
_(English version of the note above.)_

Thanks for this PR and its clear analysis and regression tests 🙏

This PR and #3395 address the same issue (#3367) with the same core idea (cross-dedup of stored ↔ new descriptions). After comparing both, we've decided to merge **#3395**, so this PR is being closed. Reasons for the choice:

1. **Self-heals existing corrupted data** — `already_description` comes from `existing_desc.split(GRAPH_FIELD_SEP)`, so graphs already hit by this bug currently store duplicate fragments; #3395 collapses these on the next merge, whereas this PR only stops future growth and keeps the historical duplicates.
2. **More accurate accounting/logging** — #3395 recomputes `already_fragment` as the surviving deduped count, keeping `Merged/LLMmrg | a+b` and `dd N` correct after collapsing.
3. **DRY** — #3395 extracts a single `_combine_descriptions_dedup` helper reused at both the node and edge sites, avoiding the divergence risk of this PR's two copies.

This is not a knock on the quality of this PR — it's equally correct, concise, and already passed Codex review. One of its test ideas (`test_reprocessing_adds_only_genuinely_new_descriptions`, mixing an old chunk + a new one in the same batch, closer to a real reprocess) has been suggested for adoption in the review of #3395. Thanks again for your contribution! 🙌

---

PR #3374, opened 2026-07-07
fix: respect LIGHTRAG-WORKSPACE header in /query endpoints (#2904)

### LINKED ISSUE #2904: [Bug]: LIGHTRAG-WORKSPACE header ignored in /query — context assembly uses default workspace
### Do you need to file a bug?

- [x] I have searched the existing issues and this bug is not already reported.
- [x] I am using the latest version of LightRAG.

### Bug Description

When using the `LIGHTRAG-WORKSPACE` header to route queries to a specific workspace, the knowledge graph entity search partially respects the header, but the **context assembly step falls back to the default workspace**. This causes queries to return context from the wrong workspace in multi-tenant setups.

### Steps to Reproduce

1. Start LightRAG Server v1.4.13 with a default workspace
2. Ingest documents into workspace `tenant_a` using the `LIGHTRAG-WORKSPACE: tenant_a` header
3. Ingest different documents into workspace `tenant_b`
4. Query with `LIGHTRAG-WORKSPACE: tenant_a` header

**Expected:** Only context from `tenant_a` documents is returned
**Actual:** Context assembly pulls from the default workspace, mixing or missing tenant-specific data

### Root Cause Analysis

Looking at the source code:

- **`/health` endpoint** correctly calls `get_workspace_from_request(request)` to extract the header
- **`/query` route** (`query_routes.py`) does **not** extract the workspace from the request header — it passes the global RAG instance directly to `aquery`/`aquery_llm`
- The global RAG instance was initialized at server startup with the default workspace
- KG search partially works because the storage backend has some workspace awareness, but context assembly in `aquery_llm()` uses the global instance's working directory

### Environment

- LightRAG Server version: v1.4.13
- LLM: Ollama (qwen3:30b-instruct)
- Embedding: Ollama (bge-m3)
- Storage: NetworkX + nano-vectordb (default)
- Deployment: Docker

### Workaround

We route queries through a custom FastAPI service that maintains **per-workspace LightRAG instances** (one `LightRAG()` object per workspace with its own `working_dir`). This gives correct isolation but requires extra memory and doesn't use the Server's API.

### Related Issues

- #2527 — Multi-workspace support question (confirmed for v1.6.x)
- #2373 — Workspace-level document isolation feature request
- #1289 — Multiple workspace support feature request

### Suggested Fix

The `/query` route handler in `query_routes.py` should:
1. Extract workspace from the `LIGHTRAG-WORKSPACE` header (like `/health` does)
2. Resolve or create the appropriate workspace-scoped RAG instance
3. Pass that instance to `aquery`/`aquery_llm` instead of the global default

This would make the existing header actually work end-to-end, ahead of the full multi-workspace redesign in v1.6.x.

### Additional Context

This is a **data isolation bug** — in multi-tenant deployments, queries leak context across workspaces. The header exists and is documented, but only partially implemented for queries.

BigStartByXuyb:
## Summary

Fixes #2904

The three query endpoints /query, /query/stream and /query/data ignored the LIGHTRAG-WORKSPACE header entirely. All queries were routed to the default workspace regardless of what the client specified, which breaks multi-tenant deployments.

## Root Cause

create_query_routes received a single rag instance bound to the default workspace at server startup. All query handlers used this instance directly without ever reading the request header, so workspace-scoped storage backends were never consulted.

## What Changed

Extracted parse_workspace_header as a shared utility in utils_api.py to eliminate the duplicated header parsing logic in the health endpoint.

In lightrag_server.py the RAG construction kwargs are now stored for reuse. A workspace-scoped instance pool with async double-checked locking creates initializes and caches RAG instances on first access per workspace.

In query_routes.py create_query_routes accepts an optional resolve_workspace_rag callable. All three query endpoints resolve the workspace-scoped RAG instance from the request header before querying. Fully backward compatible when no resolver is provided the default rag from the closure is used as before.

## Checklist

- [x] 本地测试更改
- [x] 代码已审核
- [x] 添加单元测试（如适用）

## Test Plan

- 6 new regression tests in tests/api/routes/test_query_workspace.py
- No header uses default workspace
- Header present routes to correct workspace
- Covers all three query endpoints
- Header sanitization for invalid characters
- Backward compatibility when no resolver is passed
- All existing query route tests pass with zero regressions

### DIFF
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -9,6 +9,7 @@ from fastapi.openapi.docs import (
     get_swagger_ui_html,
     get_swagger_ui_oauth2_redirect_html,
 )
+import asyncio
 import json
 import os
 import re
@@ -29,6 +30,7 @@ from dotenv import load_dotenv
 from lightrag.api.utils_api import (
     get_combined_auth_dependency,
     get_auth_status_dependency,
+    parse_workspace_header,
     display_splash_screen,
     check_env_file,
 )
@@ -1418,34 +1420,12 @@ def create_app(args):
     auth_status = get_auth_status_dependency(api_key)
 
     def get_workspace_from_request(request: Request) -> str | None:
+        """Extract workspace from the LIGHTRAG-WORKSPACE header.
+
+        Delegates to the shared ``parse_workspace_header`` utility so the
+        sanitisation logic is not duplicated.
         """
-        Extract workspace from HTTP request header or use default.
-
-        This enables multi-workspace API support by checking the custom
-        'LIGHTRAG-WORKSPACE' header. If not present, falls back to the
-        server's default workspace configuration.
-
-        Args:
-            request: FastAPI Request object
-
-        Returns:
-            Workspace identifier (may be empty string for global namespace)
-        """
-        # Check custom header first
-        workspace = request.headers.get("LIGHTRAG-WORKSPACE", "").strip()
-
-        if not workspace:
-            workspace = None
-        else:
-            sanitized = re.sub(r"[^a-zA-Z0-9_]", "_", workspace)
-            if sanitized != workspace:
-                logger.warning(
-                    f"Workspace header '{workspace}' contains invalid characters. "
-                    f"Sanitized to '{sanitized}'."
-                )
-                workspace = sanitized
-
-        return workspace
+        return parse_workspace_header(request)
 
     # Create working directory if it doesn't exist
     Path(args.working_dir).mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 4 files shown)


danielaskdd:
Hi @PR-author, thank you for submitting this fix. We reviewed the issue and the implementation approach, and we agree that this PR points out a real problem: the current query endpoints do not correctly handle the `LIGHTRAG-WORKSPACE` header, which may cause `/query`, `/query/stream`, and `/query/data` to keep querying the default workspace in multi-workspace deployments.

However, after further review, we found that this PR introduces several risks and incomplete behaviors. For that reason, we do not plan to merge it at this time.

## 1. The fix is incomplete and may lead to inconsistent API behavior

This PR mainly updates the query endpoints so that query requests can dynamically select a workspace-scoped `LightRAG` instance based on the `LIGHTRAG-WORKSPACE` header.

However, workspace semantics are not limited to query endpoints. Document upload, scan, delete, clear, document list, pipeline status, and other document / pipeline-related APIs are also strongly tied to workspace behavior.

At the moment, many of those endpoints are still bound to the default `rag` instance and the default `DocumentManager` created at server startup. This can lead to inconsistent behavior:

- query endpoints may route to workspace A based on the request header;
- upload / scan / delete / document-list endpoints may still operate on the default workspace;
- users may upload documents into the default workspace while querying another workspace.

As a result, this PR fixes the query entry points, but it does not provide a complete request-level workspace routing solution.

## 2. The WebUI does not currently send the `LIGHTRAG-WORKSPACE` header

The current WebUI does not have a workspace selector, nor does it add `LIGHTRAG-WORKSPACE` to axios or fetch requests.

This means that even if this PR were merged, the default WebUI usage path would not trigger the new workspace routing logic. Users would not be able to explicitly switch workspaces from the UI, and normal requests and streaming requests would not be guaranteed to use the same workspace.

This creates a mismatch between backend capability and frontend behavior: the backend would partially support workspace headers, but the frontend would not make use of them.

## 3. Dynamically creating `LightRAG` instances has significant resource risks

This PR introduces a workspace RAG instance pool and dynamically creates and caches a new `LightRAG` instance whenever a request references a previously unseen workspace.

This design has several resource-related risks:

- the instance pool has no size limit;
- there is no workspace allowlist;
- there is no TTL / LRU eviction;
- arbitrary header values can trigger creation of new instances;
- each instance may hold its own storage clients, connection pools, queues, locks, caches, and LLM / embedding / rerank state;
- with many workspaces, this may multiply DB connections, memory usage, file descriptors, and LLM concurrency.

If the service is exposed externally, a malicious or misconfigured client could repeatedly send different `LIGHTRAG-WORKSPACE` values and cause the server to continuously create new instances until resources are exhausted.

We do not want to merge this kind of dynamic instance creation behavior without resource limits and access control.

## 4. The global initialization lock may introduce cold-start performance issues

The PR uses a single global async lock to protect workspace RAG instance creation, and `initialize_storages()` is executed while holding that lock.

This avoids duplicate initialization for the same workspace, but it also serializes first-time initialization across different workspaces:

- if workspace A is slow to initialize;
- the first request for workspace B, even though unrelated, must wait for workspace A to finish;
- concurrent cold starts across multiple workspaces will queue up behind the same lock.

A more appropriate approach would be a per-workspace initialization lock or initialization task, combined with a global concurrency limit, rather than running full storage initialization under one global lock.

## 5. This may change the global LLM / embedding / rerank concurrency semantics

The current implementation creates a full `LightRAG` instance for each workspace. If each instance owns its own async queues or semaphores, then settings such as `max_async`, embedding concurrency, and rerank concurrency may change from service-level limits to per-workspace limits.

For example:

```bash
max_async = 8
active workspaces = 20
possible effective concurrency ≈ 160
```

This could overload local Ollama, external LLM providers, embedding services, rerank services, or backend storage systems.

This behavior needs to be explicitly designed and documented. It should not be introduced as a side effect of a query bug fix.

## 6. Workspace sanitization may cause name collisions

The PR replaces characters outside [a-zA-Z0-9_] with _ in the workspace header. This can cause collisions, for example:

* `tenant-a` → `tenant_a`
* `tenant_a` → `tenant_a`
* `tenant/a` → `tenant_a`
* `tenant.a` → `tenant_a`

Different original workspace names may map to the same effective workspace. In a multi-tenant isolation context, this is a potential security and data-isolation risk.

A safer approach would be to reject invalid workspace names with a 400 response, or define an explicit workspace encoding strategy, rather than silently sanitizing the value.

## 7. Test coverage is insufficient

The new tests cover the router-level resolver behavior, which is useful. However, they mostly rely on mock RAG instances and do not cover the full server lifecycle or real workspace routing behavior.

The following areas are not covered:

* workspace RAG pool behavior at the create_app() level;
* whether initialize_storages() is correctly called for a newly created workspace;
* whether concurrent requests for the same new workspace create only one instance;
* initialization behavior for concurrent requests across different workspaces;
* whether shutdown correctly finalizes all workspace instances;
* whether upload, scan, delete, and query behave consistently across workspaces;
* whether storage isolation actually works;
* workspace sanitization collision behavior;
* whether the WebUI sends the workspace header.

Since this PR touches multi-tenant isolation and instance lifecycle management, router-level mock tests are not sufficient for merging.

## 8. This needs a more complete workspace routing design

We agree that workspace is crucial feature for production for LightRAG, but we would prefer to solve it with a more complete design rather than a query-only patch.

A more robust direction would be:

1. define valid workspace naming rules and reject invalid names;
2. bind workspace access to authentication / authorization instead of allowing arbitrary headers;
3. introduce a unified request-level workspace resolver;
4. make query, document, graph, pipeline, and status APIs use the same workspace resolver;
5. make DocumentManager workspace-scoped or request-scoped as well;
6. add size limits, TTL/LRU eviction, or an allowlist to the workspace RAG instance pool;
7. use per-workspace initialization locks / tasks instead of a global initialization lock;
8. clearly define whether LLM / embedding / rerank concurrency limits are service-level or workspace-level;
9. add a workspace selector to the WebUI and ensure both normal requests and streaming fetch requests send the same workspace header;
10. add end-to-end tests covering upload, scan, query, delete, and status behavior across different workspaces.

## Decision

For these reasons, we do not plan to merge this PR and will close it.

Thank you again for identifying the issue where query endpoints ignore the workspace header. The issue itself is valid, but the current implementation introduces too many risks around resource usage, isolation, consistency, and test coverage. We would prefer to address this through a complete multi-workspace routing design that unifies workspace behavior across query, document, pipeline, and WebUI flows.

danielaskdd:
Based on the design suggestions from #3511 and #3563, and my understanding of the project, I have drafted a PRD for multi-workspace support. Please continue the discussion in this issue: #3631

PR #3375, opened 2026-07-07
fix: preserve cached xlsx formula values

eteriall:
## Description

Improve XLSX extraction so formula cells preserve their indexable content.

Previously, formula cells could be indexed as the formula text itself (e.g. `=SUM(A1:A10)`) instead of the computed value when parsing spreadsheets. This meant documents containing formulas sometimes lost meaningful searchable content if the workbook values weren't handled correctly.

This change makes the extractor prefer cached calculated values when they are available, while falling back to the formula text when a workbook has not been recalculated and no cached value exists.

## Related Issues

None.

## Changes Made

- Updated `_extract_xlsx` to read cached workbook values and formula text together.
- Added regression coverage for cached-formula and no-cache XLSX cases.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

On same files - graph before fix
Screenshot 2026-07-07 at 22 29 10

Chat before fix
image

Graph after fix

Screenshot 2026-07-07 at 22 18 52

Chat after fix

image

### DIFF
--- a/lightrag/parser/legacy/extractors.py
+++ b/lightrag/parser/legacy/extractors.py
@@ -98,8 +98,8 @@ def _extract_xlsx(file_bytes: bytes) -> str:
     """Extract XLSX content in tab-delimited format with sheet separators."""
     from openpyxl import load_workbook  # type: ignore
 
-    xlsx_file = BytesIO(file_bytes)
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3377, opened 2026-07-08
feat: optional BM25+vector hybrid entity seeding with RRF fusion

### LINKED ISSUE #3198: RFC: hybrid BM25 + vector retrieval with graph traversal seeding for jargon-heavy domains
## Problem

Vector-only retrieval silently fails on domain jargon — product names, acronyms, technical terms, and proper nouns that are underrepresented in embedding training data. In LightRAG's `local` mode, entities are seeded via vector search; if the embedding of a rare term is poor, the seed is never found and the graph traversal never starts. The knowledge is in the graph, but the query can't reach it.

This is a known production gap in RAG systems, particularly in specialized domains (medical, legal, engineering, financial).

## Evidence

Benchmark across two datasets with different characteristics:

| Dataset | Domain | BM25 NDCG@10 | Dense NDCG@10 |
|---------|--------|:---:|:---:|
| MIRACL (Wikipedia) | General | 0.64 | **0.79** |
| EZIS (Oracle DB manual) | **Technical jargon** | **0.92** | 0.80 |

For jargon-heavy domains, BM25 outperforms dense by 12 points; a hybrid RRF of the two is both fast and domain-robust. The point is not "BM25 beats dense" but that the winner flips by domain, so a query that can reach the graph through *either* seed path is the robust choice.

---

> **Design update (consolidating the discussion below).** This RFC originally proposed a dedicated keyword-storage role with several BM25 backends (preserved at the bottom for the thread that references it). Following the direction on #3377 — settle the design first, back the lexical index with the engine rather than an in-process index, and cover entities **and** chunks under one abstraction — the shape below reframes hybrid as an **optional capability of the existing vector store**. Everything here is author-proposed until the maintainers confirm it.

## Proposed shape: hybrid as a vector-store capability

Hybrid becomes an optional capability of the existing vector store (`SupportsHybridQuery` + a per-query `enable_hybrid`, mirroring `enable_rerank`), **not** a 5th storage role. A new `@abstractmethod` would break all 7 existing backends; a capability is duck-typed and negotiated at the call site, so backends that don't implement it simply keep dense behavior.

### What the first PR contains (minimal)

| Piece | State |
|---|---|
| `SupportsHybridQuery` contract in `base.py` + `enable_hybrid` on `QueryParam` | proposed |
| PG **entity** reference: generated `tsvector` + GIN + RRF-in-SQL, opt-in per deployment | prototyped (mechanism only; see status note) |
| `enable_hybrid` propagation through REST/OpenAPI, `QueryParam` copies, and the query **cache key** | proposed |
| Chunk retrieval | same capability shape; implementation and exact lexical policy **deferred** |
| Existing vector backends | unchanged; absence of the capability keeps dense behavior and emits an explicit fallback warning when hybrid was requested |

The PG lexical tier here is built-in `tsvector` only — no extension, fitting the no-AGE direction of #3103.

## Contract details that are easy to get subtly wrong

1. **Entity lexical leg is OR-of-terms, not AND.** Entity seeding passes the *extracted keywords* (e.g. `["CABG", "HTN"]`). Joining them and running `plainto_tsquery` would AND the terms and match neither entity. The contract preserves the keywords as a sequence and combines per-term queries with OR; chunk retrieval passes the raw query under its own policy. The lexical input is a typed, path-specific contract, not one shared string.
2. **Result keeps the dense keys.** The entity path reads `entity_name` (and `created_at`) off each row, so a fused row projects the same keys a dense row has — never a bare `(id, score)`.
3. **`enable_hybrid` is isolated in the answer cache.** It goes into the query cache key alongside `mode` / `enable_rerank`; otherwise a hybrid request reuses a dense-cached answer. This is why the change is not a single call site — it is the negotiation site plus REST / param / cache propagation.
4. **Capability means ready for this namespace, not merely implemented by the class.** PG advertises entity hybrid only after its generated column, matching text-search config, and index are ready. A backend that does not implement the capability may fall back to dense with an explicit one-time warning; schema, configuration, permission, and query failures propagate instead of masquerading as "no results."
5. **Tenant scope is part of retrieval correctness.** Each dense and lexical leg applies the workspace filter *before* its candidate limit; fusion and hydration retain `(workspace, id)`; tests include colliding IDs across workspaces. Corpus statistics for any future BM25 tier must also be workspace-scoped, or explicitly documented as a shared-statistics limitation.

Fused (RRF) scores are not comparable to `cosine_better_than_threshold`, so that threshold is never applied to the fused score. The dense candidate leg retains an explicitly defined dense-threshold policy; per-leg scores are preserved for diagnostics, and ties break on a stable `id`.

## Entity + chunk under one abstraction, and the benchmark

- **One capability, two hook points:** entity seeding (`_get_node_data`) and chunk retrieval (`_get_vector_context`). One scoping subtlety: chunk vectors are also re-scored on the KG `VECTOR` chunk-pick path (`get_vectors_by_ids` + Python cosine), which bypasses `query()` — see the open questions.
- **Recall benchmark:** to be upfront, the current PG check is a *mechanism proof on synthetic embeddings*, not a recall measurement. The real benchmark: a jargon/acronym query set over a real corpus, measuring entity-seed recall@k and chunk recall@k, dense-only vs hybrid, with `enable_rerank` held fixed. Exact-term domains beyond English are a natural second track — e.g. non-Latin scripts (CJK) the built-in parser cannot segment. That is a *tokenization* concern (a segmenting text-search config), orthogonal to the scoring tier: it works at the built-in `ts_rank_cd` tier and does not require a stronger BM25 tier.

## Open questions (for the maintainers)

- **Chunk-vector scope** — full-corpus chunk search only, or also the KG `VECTOR` chunk-pick path above?
- **Chunk lexical tier first** — is the built-in `tsvector` tier an acceptable first step for chunks, or would you want to gate chunk hybrid on a stronger BM25 tier from the start?
- **Rollout order** — PG-only first (extend the reference to chunks), then which backend second?
- **Forward-compatibility** — does the result contract + `enable_hybrid` cache-key shape look right, so a future stronger lexical tier (a true-BM25 / non-Latin-tokenizing option) can be added without a breaking change? The PG implementation is kept internally factored so that is cheap later — a private implementation detail with one implementation in this PR, **not** a strategy-selection config or a new public extension API. How, if ever, to register an *external* lexical provider is a separate question I would leave open.
- **CJK / non-Latin tokenization scope** — a segmenting text-search config would let CJK (and other non-space-segmented) deployments use hybrid at the built-in tier, landing the capability for a whole language family rather than one deployment. The cost: a chosen text-search config becomes *schema state* (baked into the generated `tsvector` column), so it needs a config-versioned migration / rebuild / rollback contract. Include that tokenization axis in the first PR, or defer it behind the same forward-compatibility door as the previous question?

## Out of scope for this PR

Stronger BM25 tiers, wiring the other four backends, strategy-selection configuration, and any external-provider / extension mechanism — deferred until the contract above is agreed and the benchmark exists. (CJK / non-Latin *tokenization* is not on this list — its inclusion is the open scoping question above, not a settled deferral.)

## Status note

A Postgres entity prototype exists from an earlier iteration, but it predates the contract points above (the typed lexical input, the result shape, and cache isolation), so this proposal is kept to the design; the reference implementation will be aligned once the shape is agreed — consensus before more code, as suggested on #3377. "Prototyped" means mechanism code exists, not that its behavior or rollout is validated.

---

<details>
<summary><b>Original proposal (superseded by the discussion above — preserved because the comments below reference its "Option A/B/C" and keyword-storage framing)</b></summary>

The original RFC proposed a dedicated `BaseKeywordStorage` role with three PostgreSQL BM25 backends — Option A (pgvector `sparsevec` with app-side BM25 weights), Option B (PL/pgSQL native BM25 via an inverted-index + triggers), and Option C (in-process `bm25s` for non-PG backends) — fused with the vector seed path via RRF before graph traversal, with a pluggable Python tokenizer. The discussion on #3377 and here moved the design toward a vector-store **capability** (engine-maintained lexical index, no new storage role), which is what the section above proposes. The original text is kept in the issue history.

</details>

BigStartByXuyb:
Sorry, my English isn't very good. The following introduction was translated with the help of Claude. The commits were merged into one.

## Summary

Adds an optional BM25 keyword index over entity names as a parallel seed path beside vector search in `_get_node_data`. Targets the exact failure mode in #3198: jargon/acronym queries where dense embeddings fail to seed the knowledge graph.

- Feature is **off by default** — zero behavior change unless `enable_bm25_seeding=True`
- Any BM25 path failure silently falls back to vector-only (a query never fails because of this)
- `bm25s` is an optional dependency (lazy import, graceful degradation)

## What's Changed

| Component | Description |
|-----------|-------------|
| `BaseKeywordStorage` | New pluggable storage interface (5th role) |
| `Bm25KeywordStorage` | Default impl: bm25s + TiktokenTokenizer (BPE, CJK-safe) |
| `fuse_seed_rankings` | RRF fusion of vector + BM25 ranked lists (k=60) |
| Query integration | `QueryParam.enable_bm25_seeding` flag, threaded through query chain |
| Lifecycle | Index rebuilt on insert + entity delete; workspace-isolated |
| Server config | `LIGHTRAG_KEYWORD_STORAGE` + `ENABLE_BM25_SEEDING` env vars |

## Design Decisions

- **Entity names only** (not chunks): smallest index that solves the seeding problem, avoids IDF drift
- **Full rebuild** (not incremental): milliseconds at entity-name scale, idempotent
- **Fusion before graph expansion**: one traversal on merged seeds, not two separate retrievals
- **Local path only** (MVP): global path is a follow-up

## Test Plan

- [x] 25 unit tests covering interface, BM25 build/search/persist/degrade, RRF fusion, flag-off regression, exception fallback, workspace isolation, CJK
- [x] `ruff check .` clean
- [x] `QueryParam().enable_bm25_seeding == False` confirmed

Closes #3198

### DIFF
--- a/docs/superpowers/plans/2026-07-08-bm25-entity-seeding.md
+++ b/docs/superpowers/plans/2026-07-08-bm25-entity-seeding.md
@@ -1,0 +1,1087 @@
+# BM25 Entity Seeding Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add an optional BM25 keyword index over entity names as a parallel seed path beside vector search in LightRAG's local retrieval, fused via Reciprocal Rank Fusion (upstream RFC HKUDS/LightRAG#3198, MVP scope).
+
+**Architecture:** A fifth pluggable storage role `KEYWORD_STORAGE` (interface `BaseKeywordStorage` in `lightrag/base.py`, default impl `Bm25KeywordStorage` in `lightrag/kg/bm25s_keyword_impl.py` backed by the optional `bm25s` library and the existing `TiktokenTokenizer`). The index is rebuilt from the graph's full entity-name set after each ingestion batch. At query time, `_get_node_data` optionally runs a BM25 search in parallel with the entity vector search and fuses both ranked lists with RRF before graph expansion. Feature is off by default; every failure path falls back to vector-only.
+
+**Tech Stack:** Python 3.10+, asyncio, dataclasses, `bm25s` (optional extra, lazy import), tiktoken (already a dependency), pytest (mock-based).
+
+**Spec:** `docs/superpowers/specs/2026-07-08-bm25-entity-seeding-design.md`
+
+## Global Constraints
+
+- Code, comments, and log messages in **English** (AGENTS.md).
+- `bm25s` must NOT become a hard dependency: lazy import, optional extra, graceful degradation when missing.
+- Default behavior with the feature disabled must be byte-for-byte identical to current behavior.
+- A query must never fail because of this feature (catch → `logger.warning` → vector-only fallback).
+- Tests are mock-based (no live services), live under `tests/` mirroring `lightrag/` layout, new dirs get an empty `__init__.py`.
+- Run tests via `./scripts/test.sh <path>`; lint via `ruff check .`.
+- Use `lightrag.utils.logger`, never `print`.
+- All new async APIs follow existing storage patterns (`@dataclass`, `StorageNameSpace` base).
+
+---
+
+### Task 1: `BaseKeywordStorage` interface + namespace + registry
+
+**Files:**
+- Modify: `lightrag/base.py` (append new ABC after `BaseGraphStorage`, i.e. before `class DocStatus` around line 793)
+- Modify: `lightrag/namespace.py` (add namespace constant)
+- Modify: `lightrag/kg/__init__.py` (registry entries)
+- Test: `tests/kg/test_keyword_storage_registry.py`
+
+**Interfaces:**
+- Consumes: `StorageNameSpace` (base.py:161) — fields `namespace, workspace, global_config`; abstract `index_done_callback`, `drop`.
+- Produces (later tasks rely on these exact names):
+  - `class BaseKeywordStorage(StorageNameSpace, ABC)` with
+    `async def index_entities(self, names: list[str]) -> None`,
+    `async def search(self, query: str, top_k: int) -> list[tuple[str, float]]`
+  - `NameSpace.KEYWORD_STORE_ENTITY_NAMES = "entity_keywords"`
+  - Registry key `"KEYWORD_STORAGE"`, impl name `"Bm25KeywordStorage"`, module `".kg.bm25s_keyword_impl"`.
+
+- [ ] **Step 1: Write the failing test**
+
+Create `tests/kg/test_keyword_storage_registry.py`:
+
+```python
+"""Registry and interface contract for the KEYWORD_STORAGE role."""
+
+import inspect
+
+from lightrag.kg import STORAGE_IMPLEMENTATIONS, STORAGE_ENV_REQUIREMENTS, STORAGES
+
+
+def test_keyword_storage_role_registered():
+    role = STORAGE_IMPLEMENTATIONS["KEYWORD_STORAGE"]
+    assert "Bm25KeywordStorage" in role["implementations"]
+    assert "index_entities" in role["required_methods"]
+    assert "search" in role["required_methods"]
+
+
+def test_bm25_keyword_storage_env_requirements_empty():
+    assert STORAGE_ENV_REQUIREMENTS["Bm25KeywordStorage"] == []
+
+
+def test_bm25_keyword_storage_module_mapped():
+    assert STORAGES["Bm25KeywordStorage"] == ".kg.bm25s_keyword_impl"
+
+
+def test_base_keyword_storage_interface():
+    from lightrag.base import BaseKeywordStorage
+
+    assert inspect.isabstract(BaseKeywordStorage)
+    sig = inspect.signature(BaseKeywordStorage.search)
+    assert list(sig.parameters) == ["self", "query", "top_k"]
+    sig = inspect.signature(BaseKeywordStorage.index_entities)
+    assert list(sig.parameters) == ["self", "names"]
+```
+
+- [ ] **Step 2: Run test to verify it fails**
+
+Run: `./scripts/test.sh tests/kg/test_keyword_storage_registry.py`
+Expected: FAIL — `KeyError: 'KEYWORD_STORAGE'` (and ImportError for `BaseKeywordStorage`).
+
+- [ ] **Step 3: Implement**
+
+In `lightrag/namespace.py`, inside `class NameSpace` after `DOC_STATUS = "doc_status"`:
+
+```python
+    KEYWORD_STORE_ENTITY_NAMES = "entity_keywords"
+```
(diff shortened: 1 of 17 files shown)


danielaskdd:
Hi @BigStartByXuyb — thank you for building this out, and please don't worry about the English at all.

First, the core result stands: BM25 seeding genuinely recovers the local-mode entities that dense retrieval misses on jargon/acronym queries. That gap is real and worth solving, so this is not a "no" — it's "let's settle the design first."

**We're going to hold off on merging this PR in its current form.** The main reason is the in-process / in-memory index. A few operational concerns come with it:

1. **Full rebuild on every insert batch.** `index_entities()` re-tokenizes and re-indexes the entire entity corpus on each `_insert_done`. That's O(N) in the number of entities and grows with the graph — it won't scale to large, frequently-updated graphs the way an incremental index would.
2. **The retriever lives in each process's memory.** Under a multi-worker API deployment, an insert in one worker rebuilds only that worker's index; the others stay stale until their next insert or a restart (reload from JSON). There is no shared source of truth.
3. **Persistence is only a JSON list of names**, re-indexed on load — effectively a per-process cache, not a shared, queryable index.
4. **Recall is unvalidated.** The tiktoken-token-id-as-BM25-term scheme is non-standard (BM25 over subword IDs), so it really needs a retrieval benchmark, not just unit tests, plus the extra synchronous search on the query path.

Our preference is to **drop the in-memory approach and back the keyword index with an external store that already does BM25/full-text well** — e.g. PostgreSQL (`tsvector`/GIN, or `pg_search` / `vchord_bm25` for true BM25) or MongoDB Atlas Search. That gives persistence, sharing across workers, and incremental updates for free, and lets the index scale with the corpus. This also lines up with @ysys143's `SupportsHybridQuery` discussion in #3198 — worth agreeing on that contract (a capability on the existing vector store vs. a standalone new role) before writing more code, since 5 of the 7 backends already ship native server-side hybrid.

**One more thing we'd like the design to cover:** this PR indexes entity names only, not text chunks. Improving **chunk recall** with BM25 is at least as important as entity seeding — it's the classic hybrid-retrieval win, and exact-term/jargon queries miss relevant chunks for the same reason they miss entities. We'd like the extended design to treat both: (a) BM25 seeding of graph entities, and (b) BM25 over text chunks for chunk retrieval, ideally under one coherent storage/abstraction rather than two separate mechanisms.

Could you put together a **more complete design proposal** along these lines — external BM25-capable storage, entity + chunk coverage, and alignment with the hybrid-query contract — before touching the code again? Let's reach consensus on the design here first, then implement. Thanks again for the strong start.

---

ysys143:
@danielaskdd @BigStartByXuyb — following your steer here, I posted an extended design on the RFC (#3198) so the discussion stays in one place: external BM25-capable store, **entity + chunk** coverage under the `SupportsHybridQuery` contract, an opt-in that keeps non-hybrid deployments zero-cost, and a recall-benchmark plan. Would welcome your thoughts there.

→ https://github.com/HKUDS/LightRAG/issues/3198#issuecomment-4937945791

PR #3386, opened 2026-07-10
fix(concurrency): propagate enqueuer's contextvars into priority queue workers

### LINKED ISSUE #3382: [Feature Request]: Tracability compatibility and enhancement

Propagate the caller's contextvars into priority_limit_async_func_call workers

## Summary
`priority_limit_async_func_call` executes every queued call in a persistent worker
task's creation-time `contextvars.Context`, not the enqueuer's context. This silently
drops all contextvars propagation across the queue. Capturing `copy_context()` at
enqueue and running each call under it makes the pool context-correct, which is a
prerequisite for OpenTelemetry/tracing span nesting and any request-scoped state.

## Problem
The pool spins up N persistent workers via `asyncio.create_task(worker())`. Because
asyncio snapshots contextvars at task-creation time, each worker runs `await func(...)`
in the context that happened to be active when that worker was first created — the
enqueuer's context is never carried across the queue. Every call funneled through a
given worker therefore shares one stale context snapshot.

Any `contextvars`-based propagation is lost across the pool. The most visible casualty
is OpenTelemetry / tracing context: a span opened inside `llm_model_func`,
`embedding_func`, a role LLM, or the rerank func attaches to whatever span was active
when the worker was created (or to nothing), instead of the caller's current span.
For a long-lived LightRAG instance reused across requests this yields:
- observations that detach into their own trace (sometimes an unnamed one), and
- cross-request pollution: request B's LLM/embedding spans nest under request A's
  span (the worker's snapshot), inflating A's trace duration.

Note: LightRAG's own optional Langfuse integration (`langfuse.openai` drop-in in
`lightrag/llm/openai.py`) does not rely on span nesting — each LLM call is a flat
generation — so it does not surface this. But it blocks any integration that wants
request-level tracing (a full pipeline → retrieval → generation waterfall) or that
uses contextvars to carry request-scoped state through the pool.

## Minimal reproduction (stdlib only, same pattern as the pool)
```python
import asyncio, contextvars
cv = contextvars.ContextVar("cv", default="<none>")

async def worker(q):
    while True:
        job = await q.get()
        await job()                 # runs in the worker's creation-time context
        q.task_done()

async def main():
    q = asyncio.Queue(); seen = []
    w = asyncio.create_task(worker(q))   # worker snapshots cv="<none>"
    await asyncio.sleep(0)
    cv.set("request-1")                  # caller's context
    await q.put(lambda: seen.append(cv.get()))
    await q.join(); w.cancel()
    print(seen[0])                       # prints "<none>", NOT "request-1"

asyncio.run(main())
```

## Root cause
worker() is created once and loops on queue.get(), executing
await func(*args, **kwargs) in its own (creation-time) context. The enqueue
site's context is never carried to the worker.

## Proposed fix
Capture contextvars.copy_context() at ENQUEUE and run each task under it.
```python
# enqueue
ctx = contextvars.copy_context()
await queue.put((priority, count, task_id, args, kwargs, ctx))
# worker
..., args, kwargs, ctx = await queue.get()
result = await asyncio.create_task(func(*args, **kwargs), context=ctx)
```

## Backward compatibility
No behavior change for callers that don't use contextvars (an empty copied
context behaves exactly like today).
Pure stdlib contextvars; no new dependencies.
Concurrency / priority / queue semantics unchanged — only each task's execution
context is corrected.
## Why it matters
Correct contextvars propagation is what lets any observability stack
(OpenTelemetry, Langfuse, OpenInference, ...) nest LLM/embedding/rerank spans under
the request that triggered them, and lets applications carry request-scoped state
(request IDs, auth, tenant) through the pool. Fixing it once in the pool removes the
need for every downstream to monkey-patch context restoration around the queue, and
lays the groundwork for a future request-level tracing story in LightRAG itself.

freya0926:
## Summary
`priority_limit_async_func_call`'s persistent worker tasks execute every queued call under the *worker's* creation-time `contextvars.Context`, not the enqueuer's context. This silently drops all contextvars propagation across the pool.

## Motivation
Fixes #3382.

The pool spins up N persistent workers via `asyncio.create_task(worker())`. Because asyncio snapshots contextvars at task-creation time, each worker runs `await func(...)` in whatever context was active when that worker was first created — never the calling context of the request that enqueued the work. For a long-lived `LightRAG` instance reused across requests, this breaks OpenTelemetry/tracing span nesting (spans opened inside `llm_model_func`, `embedding_func`, a role LLM, or the rerank func attach to the worker's stale snapshot instead of the caller's current span) and any other request-scoped state carried via contextvars.

## What's changed
- `lightrag/utils.py`: capture `contextvars.copy_context()` at **enqueue** time (both the bounded-queue default path and the unbounded global-limit path) and carry it through the queue tuple / `dispatch_queue` entry.
- At execution time (`worker()` and `limited_worker()`), the call now runs as `ctx.run(asyncio.ensure_future, func(*args, **kwargs))` — scheduling the call as a *fresh* Task whose own context is copied from the captured `ctx`. This is timeout/cancellation-compatible with the existing `asyncio.wait_for(...)` wrapping and works on Python 3.10+ (the project's floor), since `asyncio.create_task`'s `context=` kwarg only exists from 3.11.
- Shutdown drain code that unpacks the physical/dispatch queues was updated for the extra tuple element.
- No behavior change for callers that don't use contextvars — an empty copied context behaves exactly like before. Pure stdlib, no new dependencies, concurrency/priority/queue semantics unchanged.

Nothing is broken by this change; it only affects which `contextvars.Context` `func()` executes under. Added two regression tests in `tests/llm/test_global_concurrency_queue.py` (`test_context_propagates_to_default_path_worker`, `test_context_propagates_to_global_limit_path_worker`) that:
1. Prime the persistent worker before any contextvar is set (reproducing the exact moment the bug used to freeze the worker's context), then
2. Set a contextvar in the caller and assert the wrapped function observes the new value.

Both new tests fail on the pre-fix code (confirmed locally by reverting the `utils.py` change) and pass with the fix. Full `tests/llm/` suite (227 tests) and `tests/pipeline/test_pipeline_analyze_multimodal.py` pass locally.

🤖 Generated with [Claude Code](https://claude.ai/code)


danielaskdd:
@codex review

PR #3389, opened 2026-07-11
feat(documents): add crash-safe deferred batch deletion

morozsm:
## Summary

- persist deferred multi-document deletion intent in an atomic filesystem journal before destructive work
- perform one validated aggregate graph rebuild per batch, then verified metadata finalization
- support explicit crash/restart recovery with per-batch cross-process locking and read-only status endpoints
- preserve `delete_file` and LLM-cache cleanup intent across retries with at-most-once source-file cleanup
- reject unsafe workspace/batch path components on POSIX and Windows

This is a production-safety hardening/superset of the deferred-rebuild direction in #2819.

## Recovery API

- `GET /documents/delete_document/deferred`
- `GET /documents/delete_document/deferred/{batch_id}`
- `POST /documents/delete_document/{batch_id}/resume`

Operator behavior and lifecycle stages are documented in `docs/LightRAG-API-Server.md`.

## Verification

- focused deferred-delete regression suite: `53 passed`
- full CI-style offline suite: `2069 passed, 815 deselected`
- `ruff check`: passed
- `ruff format --check`: passed
- `git diff --check`: passed
- repeated independent adversarial review: final `PASS`

Full offline command:

```bash
PATH=/opt/homebrew/bin:$PATH \
  /tmp/lightrag-fulltest-venv/bin/python -m pytest tests/ -m offline -q
```

## Safety notes

- no automatic destructive resume at startup
- journals remain under persistent `working_dir/deletion_journals/<workspace>`
- `COMMITTED` requires successful rebuild and verification of every finalized metadata store
- uncertain document-delete outcomes and source-file cleanup are fail-closed / at-most-once

### DIFF
--- a/docs/LightRAG-API-Server.md
+++ b/docs/LightRAG-API-Server.md
@@ -1150,3 +1150,37 @@ This endpoint provides comprehensive status information including:
 * Content summary and metadata
 * Error messages if processing failed
 * Timestamps for creation and updates
+
+## Durable Deferred Batch Deletion Recovery
+
+`DELETE /documents/delete_document` records a durable journal before it begins a multi-document destructive delete. The journal is stored under `<working_dir>/deletion_journals/<workspace>/`; therefore `working_dir` must be persistent and shared with the server instance that will perform recovery. Do not use an ephemeral container filesystem if operators must be able to resume a failed deletion after restart.
+
+For a batch delete, LightRAG first prepares every document, performs **one** rebuild of the affected graph targets, verifies the rebuild, and only then removes source metadata. If any step fails, the journal remains unfinished and no failed batch is reported as a successful delete. The lifecycle stages are:
+
+- `PREPARED` — durable intent and source metadata snapshot exist; document preparation may still be resumed.
+- `REBUILD_PENDING` — all documents are prepared; the single aggregate rebuild is required.
+- `FINALIZATION_PENDING` — rebuild completed; source metadata finalization is required.
+- `FAILED_RETRYABLE` — an operator must explicitly resume the stored recovery stage.
+- `COMMITTED` — rebuild and metadata verification completed. Committed journals are not returned by the unfinished-batch list.
+
+All endpoints use the normal document API authentication (for example, `X-API-Key`). They expose no retained source-document metadata:
+
+```bash
+# List unfinished batches in the active workspace.
+curl -H "X-API-Key: $LIGHTRAG_API_KEY" \
+  http://localhost:9621/documents/delete_document/deferred
+
+# Inspect the stage, document IDs, error, and aggregate rebuild targets.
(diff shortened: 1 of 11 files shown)


morozsm:
Agent Review: PASS\n\nIndependent adversarial review completed against commit fb639a76 after all recovery, locking, Windows path-safety, finalization-verification, and at-most-once cleanup findings were resolved. Focused suite: 53 passed. Full offline suite: 2069 passed, 815 deselected.

morozsm:
Addressed both Codex review findings in 14a81586:\n\n- stale/not-found IDs are durably checkpointed as completed no-ops, so the batch proceeds to its single rebuild and finalization;\n- deferred deletion now checks the existing pipeline cancellation flag before every destructive document operation.\n\nVerification: 54 focused tests passed; full offline suite 2071 passed, 816 deselected; ruff, formatting, and diff checks passed. Independent follow-up review: PASS.

morozsm:
@codex review

danielaskdd:
Closing this PR after review. The two core ideas here are genuinely good and will be carried forward into the deferred-rebuild work (durably persisting rebuild targets so they survive a crash, and batch-aware ownership classification), but the implementation as a whole no longer fits the codebase:

1. **Competing journaling architecture.** Since this PR was written, main has landed a journaled, fail-closed, resumable per-document purge (#3400/#3416) whose journal lives in `doc_status.metadata` — the same storage backend as the data it protects, so it replicates and fails over with that data. A separate filesystem journal under `working_dir/deletion_journals/` introduces a second source of truth with a single-node assumption: `flock`/`msvcrt` locks are invisible to another API node handling the resume call, and unreliable on shared filesystems.

2. **Journals snapshot full document text and are never deleted.** No code path removes a COMMITTED journal, so deleting a document permanently leaves a plaintext copy of it on disk — a right-to-erasure problem and unbounded disk growth.

3. **INTENT checkpoint dead-end.** A crash inside `adelete_by_doc_id` leaves the per-doc checkpoint at INTENT, and every subsequent resume refuses to proceed ("refusing to repeat destructive deletion") with no override path — the batch is permanently wedged while the document sits half-deleted. `adelete_by_doc_id` is retry-safe by design, so the refusal is also unnecessary.

4. **Global fail-closed regression in `rebuild_knowledge_from_chunks`.** Turning missing extraction cache into a hard error breaks ordinary single-document deletion on any deployment that has cleared or evicted the LLM cache — those documents become undeletable.

5. Merge conflicts with current main in three files, plus a large new public surface (3 endpoints, 2 top-level modules) we don't want to maintain in this shape.

The deferred-rebuild direction itself is being pursued (see #2795 / #2819); a follow-up will rebuild it on top of the existing purge journal so rebuild intent survives crashes without a parallel journaling subsystem. Thanks for the detailed work — the failure-mode analysis here (at-most-once file cleanup, finalization verification) was useful input for that design.

Issue #2730, opened 2026-03-01
[Bug]:Payload error: JSON payload (46479219 bytes) is larger than allowed (limit: 33554432 bytes)


ndrewpj:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

Trying to index text and PDF file that has over 2 200 000 tokens and got an error after some time: Payload error: JSON payload (46479219 bytes) is larger than allowed (limit: 33554432 bytes)

### Steps to reproduce

1) Use qwen3 embeddinhg 4b via ollama latest
2) add text or PDF file with over 2 200 000 tokens
3) wait

### Expected Behavior

expecting no crush, no error and finalization of indexing

### LightRAG Config Used

# Paste your config here

```bash
### This is sample file of .env

###########################
### Server Configuration
###########################
HOST=0.0.0.0
PORT=9622
WEBUI_TITLE='My Graph KB Y'
WEBUI_DESCRIPTION="Simple and Fast Graph Based RAG System"
WORKERS=2
### gunicorn worker timeout(as default LLM request timeout if LLM_TIMEOUT is not set)
TIMEOUT=950
# CORS_ORIGINS=http://localhost:3000,http://localhost:8080

### Optional SSL Configuration
# SSL=true
# SSL_CERTFILE=/path/to/cert.pem
# SSL_KEYFILE=/path/to/key.pem

### Directory Configuration (defaults to current working directory)
### Default value is ./inputs and ./rag_storage
# INPUT_DIR=<absolute_path_for_doc_input_dir>
# WORKING_DIR=<absolute_path_for_working_dir>

### Tiktoken cache directory (Store cached files in this folder for offline deployment)
# TIKTOKEN_CACHE_DIR=/app/data/tiktoken

### Ollama Emulating Model and Tag
# OLLAMA_EMULATING_MODEL_NAME=lightrag
OLLAMA_EMULATING_MODEL_TAG=latest

### Max nodes for graph retrieval (Ensure WebUI local settings are also updated, which is limited to this value)
# MAX_GRAPH_NODES=1000

### Logging level
# LOG_LEVEL=INFO
# VERBOSE=False
# LOG_MAX_BYTES=10485760
# LOG_BACKUP_COUNT=5
### Logfile location (defaults to current working directory)
# LOG_DIR=/path/to/log/directory

#####################################
### Login and API-Key Configuration
[...]

Issue #2527, opened 2025-12-23
[Question]:Workspace isolation clarification: How to use multiple workspaces with a single LightRAG server / URL?


akash171198:
### Do you need to ask a question?

- [x] I have searched the existing question and discussions and this question is not already answered.
- [x] I believe this is a legitimate question, not just a bug or feature request.

### Your Question

Hi LightRAG team ,

I’m reviewing the workspace isolation changes mentioned in the v1.4.9.9 / v1.4.9.10 release notes, where it’s stated that multiple workspaces can be used and isolated.

However, in the current released server setup, I can only configure one workspace at startup (via env/config), and:

There is no option in the API to specify a workspace name while:
- uploading documents
- running a /query

- There is no option in the UI to select or switch a workspace
- All requests always go to the single default workspace that the server was started with

My question is:

 Is it currently possible to run one LightRAG server (one URL) and query different workspaces dynamically (per request or per document)?

If yes, could you please share:
- the supported API parameter / header / request format
- or an example of how workspace routing is expected to work

If not, is the intended design:

- one LightRAG instance per workspace (multiple services), or
- a future feature where workspace can be selected at request level?

### Additional Context

- I tested the /query endpoint via Postman and confirmed the server works correctly for a single workspace.
- At the moment, the workspace appears to be fixed at server startup, and there’s no documented way to override or select it at runtime.
- The release notes mention “workspace isolation” and “multiple workspaces”, but it’s not clear whether this refers to:
 internal instance-level isolation only, or
 actual API/UI-level workspace selection

My goal is to support multiple businesses / tenants using:

- one public URL
- workspace-level isolation for documents and queries

Any clarification on the current limitations and recommended architecture would be really helpful.
Thanks in advance!

danielaskdd:
Currently, a single LightRAG Server instance supports only one workspace. Multi-workspace support for a single server instance is already included in our development roadmap.

54chenxu:
What is the development roadmap for the feature of "single server instance supporting multiple workspaces"? When is it expected to be launched?

danielaskdd:
The multi-workspace feature is expected to be available starting from version 1.6.x.

leoterry-ulrica:
> The multi-workspace feature is expected to be available starting from version 1.6.x.

Approximately when will the 1.6.x version be released?

danielaskdd:
The release is scheduled for approximately August.

Kevin-JiangYC:
[...]

PR #2445, opened 2025-12-01
feat: Finish implement workspace isolation in lightrag_server

sillydong:
## Description

In this PR https://github.com/HKUDS/LightRAG/pull/2369, it proposed a way to get workspace from request header. It focused on pipeline status and in-mem storage. For the routers, it does not cover. All requests sent to server is still in the hardcoded workspace in env. In this new PR, they are fixed.
- There will be a dict of LightRAG instances and a dict of DocumentManager instances for different workspace kept in memory.  
- All routes (document/graph/ollama/query) supports get workspace from header. 
- WebUI supports switch workspace by manually input and update. User can easily change workspace in webpage.

## Related Issues

[Reference any related issues or tasks addressed by this pull request.]
Issue https://github.com/HKUDS/LightRAG/issues/1289
Issue https://github.com/HKUDS/LightRAG/issues/2373
Discussion https://github.com/HKUDS/LightRAG/discussions/1016

## Changes Made

[List the specific changes made in this pull request.]
- create `create_doc_manager` and `create_rag` in lightrag_server and pass to routers
- get workspace from request header, use `create_doc_manager` and `create_rag` to create instances for given workspace
- add input and button in webui to allow user change workspace
- send workspace with request in header `LIGHTRAG-WORKSPACE`

## Checklist

- [X] Changes tested locally
- [X] Code reviewed
- [ ] Documentation updated (if necessary)
- [ ] Unit tests added (if applicable)

## Additional Notes

[Add any additional notes or context for the reviewer(s).]

---

> [!NOTE]
> Route all API requests to per-workspace LightRAG/DocumentManager instances via LIGHTRAG-WORKSPACE header and add WebUI control to switch workspaces.
> 
> - **Backend (FastAPI)**:
>   - Implement per-workspace caching with `create_rag(request)` and `create_doc_manager(request)`; derive workspace from `LIGHTRAG-WORKSPACE` header (fallback to default).
>   - Update all routers (`documents`, `graph`, `query`, `ollama`) to accept `Request` and fetch workspace-scoped instances; change `include_router` usage to pass factory functions.
>   - Adjust lifespan to pre-create default instances and finalize all cached RAG storages on shutdown; expose current `workspace` in `/health`; set default via `set_default_workspace`.
> - **WebUI**:
>   - Add workspace input and update action in header; persist to `localStorage` and reload.
>   - Send `LIGHTRAG-WORKSPACE` with all requests (axios interceptor and streaming fetch).
> 
[...]

PR #3397, opened 2026-07-14
feat(api): server-side multi-workspace pools with HTTP routing and va…

Kevin-JiangYC:
## Description

Enable a single LightRAG server process to serve multiple workspaces, selected per-request via the existing LIGHTRAG-WORKSPACE HTTP header. Previously, a server could only serve the workspace specified at startup (--workspace). This PR introduces a lazy-initialized pooling layer (RagPool + DocManagerPool) with proper concurrency control, and wires it into all API route families.

⚠No changes have been made to the frontend or WebUI. Workspace selection is available exclusively through the HTTP API — clients must set the LIGHTRAG-WORKSPACE header on each request. WebUI still operates against the default workspace only. 

## Related Issues
Issues: #2373, #2527 . 
Built on PRs #2730 , #2904 , #3374 , #2445 .
 

## Changes Made

- RagPool: Lazily-initialized pool of LightRAG instances keyed by workspace. Uses a config factory pattern (config_factory(workspace) -> dict) instead of copying a shared base config dict — shared singletons (LLM functions, embedding functions, OllamaServerInfos) are captured by closure and reused by reference; per-workspace values (kwargs dicts, addon_params, role_llm_configs) are re-constructed fresh on every call. Two-layer asyncio.Lock (per-workspace) means different workspaces initialize in parallel without blocking each other. Error cleanup: if initialize_storages() or check_and_migrate_data() raises, finalize_storages() is called before the exception propagates — no leaked connections.

- DocManagerPool: Same two-layer lock pattern for DocumentManager instances, providing per-workspace file upload isolation via inputs/<workspace>/.

- DocumentManager class extracted from lightrag/api/routers/document_routes.py into its own module. This was necessary to avoid a circular import between workspace_pool.py (which imports DocumentManager) and document_routes.py (which would import DocManagerPool). The class itself remains unchanged.

- New validate_workspace_name() in lightrag/api/config.py — a rejection-based whitelist validator replacing the old re.sub(r"[^a-zA-Z0-9_]", "_", workspace) substitution approach. Allows [A-Za-z0-9_.-] plus CJK and Japanese kana, max 128 chars, with explicit path-traversal rejection (.., /, \). Invalid inputs get a clear HTTP 400 message instead of being silently mapped to a potentially conflicting sanitized name. 

- get_workspace_from_request() extracted from lightrag_server.py closure → lightrag/api/utils_api.py so all router modules can import it. Internalizes fallback logic with an explicit default_workspace parameter; when the header is present but empty, it rejects (HTTP 400) rather than silently falling back to the default — a malformed multi-workspace request should not accidentally route to the wrong data.

- In lightrag_server.py:
   - /status endpoint now resolves rag from rag_pool per-request.

   - build_rag_config(workspace) factory defined inside create_app(), capturing singletons via closure and re-constructing per-workspace containers.
   - Workspace names containing spaces, slashes, or backslashes are now rejected at request time with HTTP 400. Deployments using --workspace "my project" must rename their directories. The old re.sub approach silently mapped "A!" and "A?" both to "A_" while the new validator fails observably and tells the operator exactly what to fix.

- All endpoints of the 4 router familiers updated. 

- 60 new tests across 3 folders can be found in the following files:
   - tests/api/test_workspace_validation.py
   - tests/api/test_multi_workspace_routing.py
   - tests/api/test_workspace_pool.py

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes
Each workspace gets its own LightRAG instance with independent priority_limit_async_func_call limiters. Harmless for stateless cloud APIs but may overwhelm a local Ollama instance.

### DIFF
--- a/lightrag/api/config.py
+++ b/lightrag/api/config.py
@@ -238,6 +238,49 @@ def get_binding_env_value(env_key: str, default: str) -> str:
     return normalize_binding_name(get_env_value(env_key, default)) or default
 
 
+_WORKSPACE_NAME_RE = re.compile(
+    r"^[A-Za-z0-9_.\u4e00-\u9fff\u3040-\u309f\u30a0-\u30ff-]+$"
+)
+
+
+def validate_workspace_name(name: str) -> str:
+    """Validate and normalize a workspace name. Returns stripped name or raises.
+
+    Allowed characters: ``[A-Za-z0-9_.-]`` plus Chinese (\\u4e00-\\u9fff) and
+    Japanese kana.  Dots (``.``) are allowed (useful for versioned names like
+    ``v1.0``) because path-traversal is already blocked by the explicit ``..``
+    check below.
+
+    This is a rejection-based whitelist validator — unlike the old
+    substitution-based approach (``re.sub``), different invalid inputs are
+    NOT silently mapped to the same output.  The operator is told exactly
+    what to fix.
+
+    Args:
+        name: Raw workspace name from header or CLI.
+
+    Returns:
+        Validated, stripped workspace name.
+
+    Raises:
+        HTTPException(400): Invalid or empty workspace name.
+    """
+    from fastapi import HTTPException
+
+    name = name.strip()
+    if not name:
+        raise HTTPException(400, "LIGHTRAG-WORKSPACE header must not be empty")
+    if len(name) > 128:
+        raise HTTPException(400, "LIGHTRAG-WORKSPACE name too long (max 128 chars)")
+    if name in (".", ".."):
+        raise HTTPException(400, f"Invalid workspace name: {name}")
+    if "/" in name or "\\" in name:
+        raise HTTPException(400, f"Invalid workspace name: {name}")
+    if not _WORKSPACE_NAME_RE.match(name):
+        raise HTTPException(400, f"Workspace name contains invalid characters: {name}")
+    return name
+
+
 def parse_args() -> argparse.Namespace:
     """
     Parse command line arguments with environment variable fallback
@@ -736,16 +779,9 @@ def parse_args() -> argparse.Namespace:
     ollama_server_infos.LIGHTRAG_NAME = args.simulated_model_name
     ollama_server_infos.LIGHTRAG_TAG = args.simulated_model_tag
 
-    # Sanitize workspace: only alphanumeric characters and underscores are allowed
+    # Validate workspace name (rejection-based whitelist)
     if args.workspace:
-        sanitized = re.sub(r"[^a-zA-Z0-9_]", "_", args.workspace)
-        if sanitized != args.workspace:
-            logging.warning(
-                f"Workspace name '{args.workspace}' contains invalid characters. "
-                f"It has been sanitized to '{sanitized}'. "
-                "Only alphanumeric characters and underscores are allowed."
-            )
-            args.workspace = sanitized
+        args.workspace = validate_workspace_name(args.workspace)
 
     validate_auth_configuration(args)
     validate_bedrock_auth_configuration(args)
--- a/lightrag/api/document_manager.py
+++ b/lightrag/api/document_manager.py
@@ -1,0 +1,117 @@
+"""Document manager for LightRAG API — per-workspace input-directory management.
+
+Extracted from ``lightrag/api/routers/document_routes.py`` so that
+``workspace_pool.py`` can import ``DocumentManager`` without creating a
+circular import with the router module.
+"""
+
+from pathlib import Path
+from typing import List
+
+from lightrag.utils import logger, validate_workspace
+
+
+class DocumentManager:
+    def __init__(
+        self,
+        input_dir: str,
+        workspace: str = "",  # New parameter for workspace isolation
+    ):
+        # Reject path traversal before using workspace in the upload path
+        validate_workspace(workspace)
+        # Store the base input directory and workspace
+        self.base_input_dir = Path(input_dir)
+        self.workspace = workspace
+        self.indexed_files = set()
+
+        # Create workspace-specific input directory
+        # If workspace is provided, create a subdirectory for data isolation
+        if workspace:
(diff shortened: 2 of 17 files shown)


danielaskdd:
Hi @Kevin-JiangYC, thank you for the substantial work here — the engineering quality is clearly above average (the two-layer locking with double-check, the failure-path cleanup in `RagPool.get()`, the rejection-based name validator, and the 60 tests are all appreciated). However, the review surfaced several problems that are fundamentally **design questions**, and they shouldn't be settled implicitly inside an implementation PR. We are going to close this PR and would like to first reach agreement on the design in an RFC issue, then re-approach the implementation on top of that consensus.

## Problems found in this PR

1. **Backward-compatibility break for deployments without `--workspace`.** `WORKSPACE` defaults to `""` (`lightrag/api/config.py`), and the new `get_workspace_from_request()` returns HTTP 400 when there is no header and no default workspace — `test_no_header_no_default_returns_400` confirms this is intentional. On `main`, the empty workspace is a legitimate global namespace, so every existing deployment that never set `--workspace` (likely the majority) would fail on all requests after upgrading, and the WebUI (which never sends the header) would be completely broken on such deployments.

2. **Unknown workspaces are silently auto-created.** Any syntactically valid header value lazily instantiates a full LightRAG stack (storage namespaces, connections, data migration). There is no explicit create/list/delete management API, no workspace catalog, and no 404 semantics. A mistyped header silently creates and routes to a new empty workspace (a data hazard), and any client holding the single API key can grow the pool without bound (an abuse vector).

3. **`/health` has instance-creation side effects.** The handler calls `rag_pool.get(ws)`, so a liveness probe can instantiate a workspace — including storage initialization and migration. Observability endpoints must be side-effect free.

4. **No pool governance.** There is no capacity limit and no runtime eviction (`_evict()` is private and only reachable from `shutdown_all()`). Each instance holds database connections and in-memory state, so a long-running multi-tenant server accumulates resources monotonically.

5. **No global resource governance across workspaces.** Each workspace gets independent `llm_model_max_async` limiters, so N active workspaces mean N× the configured LLM/embedding concurrency — the PR's own notes acknowledge this may overwhelm a local Ollama. This needs a shared-limiter architecture, a cap on concurrently active pipelines, and basic fairness — not a per-instance `Semaphore(4)` band-aid inside one storage backend.

6. **Multi-worker (Gunicorn) semantics are unaddressed.** The pools are per-worker in-process dicts, and each worker independently runs `initialize_storages()` + `check_and_migrate_data()` on first access. Data migration now runs on the request path (first-request latency spike; cross-worker migration races rely on existing locks and are untested). There are no multi-worker tests.

7. **Restart recovery becomes lazy.** Unfinished documents in a non-default workspace are not resumed after a restart until the first request happens to touch that workspace.

8. **Unrelated changes are mixed in.** `trust_env=False` in `lightrag/llm/ollama.py` and the embedding-concurrency semaphore in `nano_vector_db_impl.py` are cross-cutting fixes that deserve their own PRs regardless of what happens here.

## Suggested next step: an RFC issue first

The docstrings in `workspace_pool.py` reference a numbered design document ("§14", "§14.4") — it sounds like you already have one. Please open a dedicated design issue (RFC) based on it, focused on **usage scenarios, constraints, and design requirements** rather than implementation details. It should cover at least:

1. **Instance/workspace binding lifecycle** — fixed binding with no mid-life switch (this PR's approach looks right), plus explicit semantics for the default/empty workspace: header absent vs. header empty, servers started without `--workspace`, and the upgrade path for existing single-workspace deployments (zero-migration is the requirement).
2. **Workspace lifecycle management** — explicit creation vs. auto-creation (we'd recommend an explicit create/list/delete API with 404 for unknown workspaces), name validation rules, deletion semantics (data destruction? in-flight jobs?), and a per-endpoint rule for which endpoints may instantiate a workspace (observability endpoints must not).
3. **Instance pool governance** — capacity limits, eviction policy and safe-eviction conditions (in-flight queries, running pipelines), initialization-failure handling.
4. **Pipeline/workspace binding** — explicit workspace context for all background jobs with no fallback to a global default; restart-recovery policy when multiple workspaces have unfinished documents (eager vs. lazy); when data migration runs (startup vs. first request).
5. **Cross-workspace parallelism and global resource governance** — shared vs. per-workspace LLM/embedding concurrency budgets, a cap on simultaneously active pipelines, and fairness so one tenant's bulk ingest cannot starve others.
6. **Multi-process mode and future clustering** — consistency between per-worker pools and Manager-shared state; memory/connection cost scaling as workers × active workspaces; a workspace catalog living in shared storage rather than process memory; keeping pipeline mutual exclusion abstract enough to move to external coordination later (no sticky-session assumptions).
7. **API contract** — header choice; how the Ollama-compatible API selects a workspace given that Ollama clients (e.g. Open WebUI) cannot set custom headers; WebUI plans; the full list of affected endpoints.
8. **Security boundary** — workspace isolation is not authorization: with a single API key, any client can reach any workspace. If per-workspace auth and quotas are out of scope for the first phase, the RFC should say so explicitly.
9. **Scope and phasing** — what phase one contains (this PR is roughly "logical isolation + instance pool + header routing") and what is deferred (physical isolation / storage profiles, WebUI support, quotas).

Note there is also a parallel proposal in #2527 with a different scope (storage profiles / physical isolation); a shared RFC would give both efforts one place to converge instead of two competing implementations.

Once the design is agreed, much of the code here — the pool concurrency model, the validator, the tests — should carry over directly. Thanks again for the effort; the goal is to make sure it lands on a stable foundation.

---

danielaskdd:
Please feel free to share your comments and suggestions under RFC #3520.

Kevin-JiangYC:
Hi @danielaskdd , thanks for the response! I'll check out #2527 & #3520 .

Kevin-JiangYC:
I've just opened an RFC at #3563 which addresses some of the issues brought up above. Any advice is appreciated.

Issue #3402, opened 2026-07-15
Cancellation safety of the pipeline busy-acquire pattern (busy set before the release-guaranteeing try/finally)


danielaskdd:
## Problem

Every acquirer of the pipeline `busy` slot follows the same shape:

```python
async with pipeline_status_lock:
    ...
    pipeline_status["busy"] = True        # slot taken (inside the lock)
    ...
# exiting the lock awaits __aexit__, which awaits asyncio.shield(...) in
# _KeyedLockContext.__aexit__ — a real cancellation delivery point
busy_released_in_loop = False              # sync
try:
    ...                                    # work
finally:
    pipeline_status["busy"] = False       # the only guaranteed release
```

If the task is **cancelled at the lock's `__aexit__`** — after `busy=True` was set but **before** entering the `try`/`finally` that releases it — nothing clears `busy`, so the workspace stays **permanently busy** and all later processing / clear / delete / scan / custom-chunks calls are rejected until restart.

This is a **pre-existing, codebase-wide property of the busy-acquire pattern**, not specific to any one caller. Affected sites include:
- `_PipelineMixin.apipeline_process_enqueue_documents` (the normal processing loop) — pipeline.py, acquire block then `try/finally`.
- `adelete_by_doc_id` (single-document deletion) and `background_delete_documents` / clear (destructive) — they set `busy`/`destructive_busy` then do work under separate try/finally.
- `ainsert_custom_chunks` handoff (added in #3401): when `request_pending` is set it keeps `busy=True` and relies on the awaited `apipeline_process_enqueue_documents(_holding_busy=True)` to release it; a cancel during that drain's startup awaits (`get_namespace_data` / lock acquire), before it reaches its own release path, leaks `busy`. (Codex flagged this instance on #3401; deferred here as it is the same general pattern.)

## Why a dedicated issue

Fixing it consistently means a **unified** cancellation-safety treatment of the acquire pattern, not a piecemeal patch in one caller (which would leave the identical window everywhere else and be inconsistent). Options to evaluate (no pre-supposed solution):
- Ensure `busy=True` and the release-guaranteeing `try/finally` are not separated by any cancellation point (e.g., set the flag and enter the guard within the same non-awaiting span, or move the flag-set to the first line inside the `try`).
- A context manager / helper that atomically takes the slot and guarantees release on any exit (including cancellation).
- Shielding just the acquire critical section.

## Scope / notes

[...]

PR #3407, opened 2026-07-16
fix(shared-storage): make keyed-lock release cancellation-safe

danielaskdd:
## Problem

Two cancellation-safety bugs in the shared keyed-lock layer (`lightrag/kg/shared_storage.py`) can strand a lock forever under `asyncio.CancelledError`:

- **`NamespaceLock.__aexit__`** cleared its `_ctx_var` only *after* `await ctx.__aexit__(...)`. That await internally awaits an `asyncio.shield`, a real cancellation-delivery point. A cancellation there skipped `_ctx_var.set(None)`, leaving a stale context — so the **same coroutine** re-entering the lock in a `finally` hit `RuntimeError("already acquired")` instead of a clean re-acquire. Any cancel-safe release built on top could never run.

- **`_KeyedLockContext.__aexit__`** released the underlying locks on a shielded coroutine that read `self._ul` **at run time**, while the `finally` set `self._ul = None`. Under (re-)cancellation the deferred release could observe `None` (and skip releasing the real lock), or `__aexit__` could return *before* the shielded release actually finished — either way the underlying keyed lock stays held forever and every future acquirer **deadlocks**. In multi-worker mode this wedges the whole workspace across processes.

## Fix

- Clear `_ctx_var` in a `try/finally` so it is cleared even when `ctx.__aexit__` is cancelled.
- Snapshot the acquired-lock entries and clear `self._ul` up-front; the release closure iterates the **snapshot**, not `self._ul`.
- Run the release on a **fixed task** and wait for that same task to complete via a shield loop that resists repeated cancellation — `__aexit__` never returns while a lock is still held, and a deferred cancellation is re-raised only after the release finishes. If the release task itself is directly cancelled, propagate rather than pretend it released.

## Tests

New `tests/kg/test_keyed_lock_cancel_safety.py` (4 cases, real `_KeyedLockContext`/`NamespaceLock`):
- `__aexit__` snapshots entries and keeps waiting for the release under cancellation, re-raising only after it completes.
- release task directly cancelled → `__aexit__` propagates.
- end-to-end: a task cancelled while holding the real keyed lock still releases it, a competing task re-acquires (no deadlock), refcount returns to 0.
- `NamespaceLock` is re-entrable in the same coroutine after a cancellation inside `ctx.__aexit__`.

`git stash`-ing the fix makes the two discriminating tests fail (old-fails / new-passes). `tests/kg` + `tests/pipeline` = 1182 passed, 20 skipped; ruff + ruff-format clean.

## Context

Prerequisite (PR-0) for the pipeline reservation cancellation-safety work tracked in #3402 — that work's cross-task release runs on a fresh task whose lock acquire would otherwise inherit the stale ContextVar, and relies on the underlying keyed lock actually releasing under cancellation.

### DIFF
--- a/lightrag/kg/shared_storage.py
+++ b/lightrag/kg/shared_storage.py
@@ -1062,6 +1062,16 @@ class _KeyedLockContext:
         if self._ul is None:
             return
 
+        # Snapshot the acquired-lock entries and clear ``self._ul`` BEFORE
+        # starting the release task. The release runs on a shielded task that
+        # may complete AFTER this coroutine is (re-)cancelled; if the closure
+        # below read ``self._ul`` at that later point it would find ``None``
+        # (cleared here) and skip releasing the underlying locks, deadlocking
+        # every future acquirer. Iterating a stable snapshot keeps the deferred
+        # release correct.
+        entries = list(self._ul)
+        self._ul = None
+
         async def release_all_locks():
             """Release all locks with comprehensive error handling, protected from cancellation"""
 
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3413, opened 2026-07-16
feat(shared-storage): dead-worker recovery for multiprocess keyed locks

danielaskdd:
## Problem

In multiprocess mode (`workers>1`) a keyed lock's underlying mutex is a `manager.Lock()` — an **ownerless, liveness-blind** mutex living in the Manager process. If the worker holding it is **SIGKILLed**, the Manager never notices: the lock stays locked forever and every other worker's `acquire()` **deadlocks**. This is not theoretical — KG entity/edge locks (`get_storage_keyed_lock([entity], namespace="{ws}:GraphDB")`) are held for an entire merge / LLM-extraction, so the kill window is wide.

Single-process mode uses `asyncio.Lock` (dies with the process — no cross-process orphan), so this is a **multiprocess-only** problem. There is no gunicorn per-worker-exit hook to lean on, so recovery must be intrinsic to the lock.

## Fix — holder-record lease (the record IS the lock)

Replace the `manager.Lock()` value with a holder record in a new `_keyed_lock_holders` manager.dict: `{owner_pid, process_start_id, lease_id}`.

- **acquire**: under `_registry_guard`, if unheld, write the record (a single atomic manager.dict write) and return; else snapshot and probe the owner's liveness **outside** the guard, and if the owner is *confirmed dead*, recheck the same `lease_id` and pop it, then retry. Live owner → backoff-poll.
- **release**: pop iff `lease_id` still matches (owner-checked — never clobber a reclaimer/new owner).
- **dead-only reclaim**: `_process_alive(pid, start_id)` returns `False` only when the PID is gone, or (Linux) its `/proc` start time differs from the recorded one (**PID reuse** = a different process). Any uncertainty — no permission, non-Linux, no start id — is treated as **ALIVE**. A live-but-slow owner is therefore never preempted, which is why **no fencing token is needed**: we only ever reclaim from the dead.

`UnifiedLock` gains an `mp_is_lease` path (async poll, no executor thread); the `manager.Lock()` executor path stays for `internal_lock`/`data_init_lock`. **Single-process mode is unchanged**; refcount + idle-cleanup bookkeeping is preserved (`_lock_registry` now stores a presence sentinel).

## Out of scope (unchanged, by design)

- SIGKILL of the holder of `_registry_guard` / `internal_lock` — µs-scale bookkeeping locks, already an unhardened base today; true hardening needs an OS robust mutex or external coordinator.
- Poll fairness under extreme contention (backoff + jitter for now).
- A live-but-**hung** holder of a data lock — stealing it would corrupt data; that's a separate liveness problem.

## Tests

`tests/kg/test_keyed_lock_dead_worker_recovery.py` drives a **real SIGKILLed subprocess PID** plus seeded holder records: reclaim-on-confirmed-dead (incl. the pipeline_status `NamespaceLock` full stack), **no reclaim of a live owner** (times out), owner-checked release, PID-reuse (Linux-only), and single-process-unchanged.

Full offline suite: **3699 passed, 35 skipped**; ruff + ruff-format clean.

## Relation to #3402

Orthogonal to the just-merged #3402 stack (#3407/#3408/#3409), which closed cooperative-cancellation wedges (in-process flags). This closes the cross-process orphaned-lock wedge (process killed). It is also the **mutex-layer prerequisite** for a follow-up pipeline-reservation dead-process recovery layer.

### DIFF
--- a/lightrag/kg/shared_storage.py
+++ b/lightrag/kg/shared_storage.py
@@ -1670,6 +1670,15 @@ def _process_alive(pid: Optional[int], start_id: Optional[str]) -> bool:
     if pid is None:
         return True  # no owner identity recorded → cannot declare dead
     if pid == os.getpid():
+        # Our own PID. Genuinely us only if the recorded start id matches ours:
+        # a record carrying our PID but a DIFFERENT start id was written by a
+        # dead predecessor whose PID the OS reused for us — that owner is dead,
+        # so we must NOT treat the lease as "still alive (me)" or it would never
+        # be reclaimed. If either start id is unknown (non-Linux / no /proc), we
+        # cannot confirm reuse and conservatively report alive.
+        mine = _my_start_id()
+        if start_id is not None and mine is not None and start_id != mine:
+            return False
         return True
     if not _pid_alive(pid):
         return False
--- a/lightrag/kg/shared_storage.py
+++ b/lightrag/kg/shared_storage.py
@@ -2,6 +2,7 @@ import os
 import sys
 import asyncio
 import multiprocessing as mp
+import random
 import uuid
 from multiprocessing.synchronize import Lock as ProcessLock
 from multiprocessing import Manager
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3418, opened 2026-07-17
Add Atlas Cloud LLM binding

binyangzhu000-sudo:
## Summary
- add an Atlas Cloud LLM binding that reuses LightRAG's existing OpenAI-compatible chat path
- default Atlas Cloud to https://api.atlascloud.ai/v1, ATLASCLOUD_API_KEY, and qwen/qwen3.5-flash when no LLM_MODEL is set
- add Atlas Cloud-specific provider options and focused config tests

## Validation
- /Users/zby/.local/bin/python3.11 -m py_compile lightrag/api/config.py lightrag/api/lightrag_server.py lightrag/llm/binding_options.py tests/api/config/test_api_config_atlascloud.py
- UV_PROJECT_ENVIRONMENT=.venv311 /Users/zby/.local/bin/uv run --python /Users/zby/.local/bin/python3.11 --with pytest pytest tests/api/config/test_api_config_atlascloud.py -q
- UV_PROJECT_ENVIRONMENT=.venv311 /Users/zby/.local/bin/uv run --python /Users/zby/.local/bin/python3.11 --with ruff ruff check lightrag/api/config.py lightrag/api/lightrag_server.py lightrag/llm/binding_options.py tests/api/config/test_api_config_atlascloud.py
- git diff --check
- Atlas Cloud live model catalog returned 118 models and confirmed qwen/qwen3.5-flash and deepseek-ai/deepseek-v4-pro

No README changes; env.example only documents the technical binding configuration.

### DIFF
--- a/env.example
+++ b/env.example
@@ -672,7 +672,7 @@ EMBEDDING_BATCH_NUM=32
 
 ###########################################################################
 ### Gloabal LLM Configuration
-###   LLM_BINDING type: openai, ollama, lollms, azure_openai, bedrock, gemini
+###   LLM_BINDING type: openai, atlascloud, ollama, lollms, azure_openai, bedrock, gemini
 ###   LLM_BINDING_HOST: Service endpoint (left empty if using the provider SDK default endpoint)
 ###   LLM_BINDING_API_KEY: api key
 ### If LightRAG deployed in Docker:
@@ -776,6 +776,14 @@ VLM_MIN_IMAGE_PIXEL=64
 
 ### Role-specific + Provider-sepecific LLM options
 # VLM_OPENAI_LLM_MAX_TOKENS=20000
+
+### Atlas Cloud OpenAI-compatible LLM:
+###     LLM_BINDING=atlascloud
+###     LLM_BINDING_HOST=https://api.atlascloud.ai/v1
+###     ATLASCLOUD_API_KEY=your_api_key
+###     LLM_MODEL=qwen/qwen3.5-flash
+### Atlas Cloud-specific LLM options use the ATLASCLOUD_LLM_* prefix.
+# ATLASCLOUD_LLM_TEMPERATURE=0.9
 
 ### Azure OpenAI Specific Parameters:
 ###     LLM_BINDING=azure_openai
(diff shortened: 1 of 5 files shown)


danielaskdd:
Thanks a lot for the contribution, and for the thorough validation steps — it's clear you put real care into this.

After reviewing the change, though, I don't think a dedicated `atlascloud` binding is the right fit for LightRAG, and I'd like to explain why so it's transparent.

Atlas Cloud exposes an **OpenAI-compatible** API, and LightRAG already supports any OpenAI-compatible provider through the existing `openai` binding. Concretely, this PR's `AtlasCloudLLMOptions` is an empty subclass of `OpenAILLMOptions`, and at runtime the `atlascloud` binding falls through to the exact same code path as `openai` (`create_optimized_openai_llm_func` → `openai_complete_if_cache`, with `base_url=LLM_BINDING_HOST`). So everything this PR enables can already be done today with:

```dotenv
LLM_BINDING=openai
LLM_BINDING_HOST=https://api.atlascloud.ai/v1
LLM_BINDING_API_KEY=your_api_key
LLM_MODEL=qwen/qwen3.5-flash
# provider-specific options via the existing OPENAI_LLM_* prefix, e.g.
# OPENAI_LLM_TEMPERATURE=0.9
```

This is actually the pattern we already document for other OpenAI-compatible providers — see `env.example`, where OpenRouter is shown exactly this way:

```
### OpenAI Specific Parameters (Openrouter of other OpenAI compatible API):
###     LLM_BINDING=openai
###     LLM_BINDING_HOST=https://openrouter.ai/api/v1
###     LLM_MODEL=google/gemini-2.5-flash
```

Adding a separate binding enum per OpenAI-compatible vendor doesn't scale (there are a great many of them), and the hardcoded default model (`qwen/qwen3.5-flash`) would drift out of date over time and become a maintenance liability — all for behavior the `openai` binding already provides.

If there's a genuine capability that Atlas Cloud needs but the `openai` binding can't express (a non-standard auth scheme, a request/response shape that isn't OpenAI-compatible, etc.), I'd be very happy to look at that specific gap — that would be a much stronger case. Short of that, I'm going to close this PR.

Thanks again for taking the time, and I hope you'll consider contributing again. 🙏

PR #3430, opened 2026-07-19
fix(utils): reject non-finite values in parse_optional_float

santhreal:
## Problem

`parse_optional_float` accepted env strings like `nan` / `inf` and passed them into semantic-chunker breakpoint thresholds.

## Repro

```python
from lightrag.utils import parse_optional_float

parse_optional_float("nan")  # previously returned nan
```

## Fix

Require a finite float after parsing so misconfigured knobs fail loudly at decode time.

### DIFF
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -12,6 +12,7 @@ import inspect
 import json
 import logging
 import logging.handlers
+import math
 import os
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3432, opened 2026-07-19
refactor(pipeline): operation-scoped lock-free logger for hot pipeline-status logging

danielaskdd:
## Description

Introduce an **operation-scoped `PipelineStatusLogger`** for lock-free pipeline-status logging, route the hot log sites in `operate.py` through it, and move per-item KG-rebuild detail out of pipeline history.

### Why this PR was closed and what reopened it

The original stages 1–2 (a one-shot lock-free helper replacing `async with pipeline_status_lock` at pure-log sites) were closed on 2026-07-19 with *"No significant positive impact"*: the stage-3 multi-worker benchmark showed the pipeline is LLM/IO dominated, so removing the lock alone is noise end-to-end.

This reopening follows the Manager-proxy RPC-reduction track instead: the logger now also **caches the `history_messages` handle**, so steady-state explicit proxy operations per log line drop from 3 to 2 (`latest_message` setitem + history `extend`), while cached nested-proxy reuse also avoids the repeated proxy reconstruction and reference-management (incref/decref) traffic that a per-message `DictProxy.get` incurs. Layer-1 microbenchmarks on the logging path showed the cached handle (avoiding ListProxy re-materialization) is the dominant win on that path; the layer-2 gate — **≥5% median end-to-end improvement on a real ingestion trace replay — has not been collected yet, so this PR stays Draft until it is.**

### Design: `PipelineStatusLogger` (lightrag/kg/shared_storage.py)

- One instance per pipeline operation (`extract_entities` / `merge_nodes_and_edges` call), shared by that operation's child tasks. Construction is proxy-free; the first history write fetches `pipeline_status["history_messages"]` once and caches the handle.
- **Never raises** — call sites are shaped `except ...: log(...); raise e`, so a raising status write would mask the real exception. `latest_message`, history, and the diagnostic logging are each guarded independently.
- **Cache recovery**: a failed or `None` fetch is not cached and is retried on the next write; a failed `extend` drops the cache so the next write re-fetches. Messages of the failed call are never retried (the write may have half-committed; a retry could duplicate).
- Lifecycle: must not be reused across `finalize_share_data()`; relies on the repo invariant that `history_messages` is only ever reset **in place** (`del h[:]` / `h[:] = [...]`), never replaced.
- **History trimming is deliberately not offered**: the `len`-then-delete is a read-modify-write and must stay under `pipeline_status_lock` (a locked trim API is deferred to stage 2).
- The earlier one-shot `append_pipeline_log` helper is **removed**: after the migration it had zero production callers (hot paths hold a logger; one-off sites can build one inline), and it is an internal, never-released symbol — no shim is kept. Its "why lock-free is safe" rationale moved into the class docstring, and its test coverage is subsumed by the logger tests.

### Coordination state stays locked

`busy` / `request_pending` / `cancellation_*` / reservation owner tokens / `cur_batch` / the history trim all remain inside `async with pipeline_status_lock` read-modify-write blocks. Cancellation-check KEEP blocks in `extract_entities` and `merge_nodes_and_edges` are unchanged (pinned by a raise-on-enter-lock test and an exact lock-acquisition count). The owner-lifecycle categories (`owner_stable_after_unlock`, `owner_transition`, `external_non_owner`, `history_reset`) are untouched — migrating `owner_stable` sites is a separate follow-up PR.

### Declared semantic deltas (each pinned by tests)

1. **Logging no longer requires `pipeline_status_lock` to be provided.** The old per-chunk extraction site did `async with pipeline_status_lock` while only checking `pipeline_status is not None` — with a status dict but no lock it raised `TypeError`. The logger writes regardless of the lock.
2. **A missing/None `history_messages` is silently skipped** instead of raising `KeyError` (best-effort logging must not depend on init timing).
3. **Per-item KG-rebuild completion detail leaves pipeline history** ("Rebuild `X` from N chunks" goes to the backend logger only — on large rebuilds one success line per item swamps the history), while per-item **failure** messages (cache parse failures, entity/relationship rebuild failures) **stay in history** so errors remain visible to pipeline-status consumers (the WebUI), alongside the operation-level and summary messages including failure counts. The entity-side completion write was dead code already (its only caller never passed `pipeline_status`); `_rebuild_single_entity` / `_rebuild_single_relationship` lose their now-dead `pipeline_status`/`pipeline_status_lock` parameters entirely (internal helpers, sole caller updated).

Note: the `append_pipeline_log` symbol (previously in `lightrag.kg.shared_storage`, re-exported through `lightrag.operate` imports) is gone entirely; any downstream deep import of it breaks — no shim is kept for internal symbols.

## Changes Made

- **`lightrag/kg/shared_storage.py`** — add `PipelineStatusLogger` (+ `_UNRESOLVED` sentinel), generalize `_debug_log_failure`, and remove the one-shot `append_pipeline_log` helper (zero callers after the migration).
- **`lightrag/operate.py`** — `extract_entities` and `merge_nodes_and_edges` build one logger per operation and share it with their child tasks (per-chunk extraction log; per-entity/per-relation merge logs via a trailing `status_logger` parameter on `_merge_nodes_then_upsert` / `_merge_edges_then_upsert` with a throwaway-logger fallback). `rebuild_knowledge_from_chunks` builds one for its per-item failure messages; KG-rebuild per-item completion logs are demoted to backend-only logging.
- **Tests** —
  - `tests/kg/test_pipeline_status_logger.py` (15): handle caching (fetch exactly once), cache recovery incl. no-resend-on-failure, in-place-reset identity (and the replacement-orphans-the-cache limitation), never-raise at every operation incl. a failing diagnostic logger, fork-inherited cached proxy visibility, multi-process concurrent cached-handle writers (no loss, no group tearing).
  - `tests/extraction/test_pipeline_log_delock.py` (3): raise-on-enter lock proves cancel checks still lock; a capture-logger injection proves the per-chunk log routes through the logger with the lock acquired exactly twice; a new test pins delta #1 (`pipeline_status_lock=None` still logs).
  - `tests/extraction/test_rebuild_history_scope.py` (5): parameterized parse/entity/relationship failure paths assert the failure detail reaches history while completion detail does not and the summaries (incl. failure counts) survive; a signature check pins that the single-item rebuild helpers cannot receive `pipeline_status` at all.
  - `tests/kg/test_append_pipeline_log.py`: removed together with the helper (its unique cases all have direct logger equivalents).

## Related Issues

N/A — internal performance / consistency work.

## Checklist

- [x] Changes tested locally (`tests/extraction tests/pipeline tests/kg tests/api`: 1821 passed, 36 skipped; ruff clean)
- [ ] Code reviewed
- [x] Documentation updated (if necessary) — helper/class are self-documenting; no user docs affected

## Additional Notes / Deferred

- **Stage 2 (separate PR, gated on the same benchmark):** migrate `owner_stable_after_unlock` sites (locked coordination `update`, then a lock-free history-only append — the logger deliberately ships with `log()` only until those call sites exist), and design a **locked** trim API (an unlocked trim is a concurrency hazard: two loggers observing len 10001 could each delete 5001 and empty the log).
- **Benchmark gate:** this PR remains **Draft** until a real ingestion-trace replay shows ≥5% median end-to-end improvement.


danielaskdd:
**No significant positive impact identified.

danielaskdd:
@codex review

PR #3433, opened 2026-07-19
perf(kg): snapshot JSON storage with DictProxy.copy() instead of dict(proxy)

danielaskdd:
## Description

Under multi-worker mode (`uvicorn/gunicorn --workers N`) the JSON storages' shared dict `self._data` is a `multiprocessing.Manager` **DictProxy**, where every proxy method call is one round-trip RPC to the Manager process. `index_done_callback` snapshotted the whole namespace with `dict(self._data)`, which walks the mapping protocol: `keys()` + one `__getitem__` **per key** = **N+1 RPCs** for an N-key namespace (`text_chunks`, `full_docs`, doc_status…). `DictProxy.copy()` marshals the entire mapping server-side in **a single RPC**.

This is the first of a small series of Manager-proxy RPC reductions; it is self-contained and touches only the persistence flush path.

## Changes Made

- `JsonKVStorage.index_done_callback` and `JsonDocStatusStorage.index_done_callback`: replace `dict(self._data) if hasattr(self._data, "_getvalue") else self._data` with `self._data.copy()`.
  - The `hasattr(_getvalue)` mode-branch is removed: a plain dict (single-process mode) copies cheaply and identically, and `write_json` only reads its argument (its docstring already documents "may be a shallow copy from shared memory"), so the shallow copy is safe in both modes and the on-disk output is byte-identical.
  - The reload-on-sanitize branch (`clear()` + `update()`) is unchanged.
- New `tests/kg/json_impl/test_json_storage_rpc_counts.py`:
  - counting DictProxy stand-in asserting `index_done_callback` performs **exactly one `copy()`** and **zero** per-key `keys()`/`__getitem__`;
  - real-Manager (`initialize_share_data(2)`) end-to-end checks that KV upsert+flush and DocStatus synchronous-flush persist correct on-disk content through a genuine proxy.

## Verification

- `./scripts/test.sh tests/kg/json_impl` — 39 passed.
- `pre-commit run ruff-format/ruff --files …` — passed.

## Scope notes (intentionally NOT changed)

- Per-key reads `get_by_id`/`get_by_ids` stay per-key — those are already one RPC per key; snapshotting a huge namespace to read one key would be a regression.
- `upsert`'s per-key `k in self._data` membership check is **left as-is**: a `set(self._data.keys())` snapshot would pull the whole namespace on every upsert, badly regressing the common single-key upsert against a large `text_chunks` — quantifying that trade-off is deferred to a benchmark.

## Checklist

- [x] Changes tested locally
- [x] Unit tests added

### DIFF
--- a/lightrag/kg/json_doc_status_impl.py
+++ b/lightrag/kg/json_doc_status_impl.py
@@ -235,9 +235,12 @@ class JsonDocStatusStorage(DocStatusStorage):
         """
         async with self._storage_lock:
             if self.storage_updated.value:
-                data_dict = (
-                    dict(self._data) if hasattr(self._data, "_getvalue") else self._data
-                )
+                # DictProxy.copy() is a single Manager RPC that marshals the
+                # whole mapping server-side; dict(proxy) would walk the mapping
+                # protocol and fetch every value with its own RPC. Plain dicts
+                # (single-process mode) copy cheaply and identically — write_json
+                # only reads its argument, so the shallow copy is safe there too.
+                data_dict = self._data.copy()
                 logger.debug(
                     f"[{self.workspace}] Process {os.getpid()} doc status writting {len(data_dict)} records to {self.namespace}"
                 )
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

PR #3434, opened 2026-07-19
perf(shared_storage): cut Manager RPCs in update flags, namespace lookup, status API

danielaskdd:
## Description

Three independent reductions of per-element `multiprocessing.Manager` proxy access. Under multi-worker mode (`--workers N`) every proxy method call is one round-trip RPC to the Manager process; single-process mode uses plain containers and is unaffected. Stacks conceptually with #3433 (JSON storage snapshot) but touches disjoint code.

The unifying principle: collapse "loop that re-fetches a proxy handle and indexes it per element" into a single slice/snapshot.

## Changes Made

**1. update flags — slice the ListProxy once**
- `set_all_update_flags` / `clear_all_update_flags` iterated `range(len(...))` and re-indexed `_update_flags[ns][i]` each step (DictProxy getitem + ListProxy getitem + ValueProxy set per flag). Now they slice the flag list once (`_update_flags[ns][:]`, one RPC) and set `.value` on each returned ValueProxy handle. For N worker flags: ~`(1 + 2 + 3N)` RPCs → `(2 + N)`. Writes still propagate (the slice returns the same referents), and the internal lock excludes concurrent appends.
- `get_all_update_flags_status` sliced each namespace's flag list once instead of iterating the ListProxy element-by-element (which costs one getitem RPC per element via the sequence protocol).

**2. get_namespace_data — per-process namespace cache**
- A namespace's shared dict is created once and never removed at runtime (the only `clear` is in `finalize_share_data`). Added `_namespace_data_cache` keyed by final namespace so a hot-path hit returns the cached reference **without the internal lock or the `__contains__`/`__getitem__` RPCs**.
- The cache only ever holds already-created namespaces, so the `PipelineNotInitializedError` guard (which fires on a *miss*) and workspace isolation are unchanged. Reset in `initialize_share_data` (`{}`) / `finalize_share_data` (`None`) alongside the existing `_lease_ns_cache` / `_queue_stats_ns_cache`. Verified `_shared_dicts` has no runtime writer outside `get_namespace_data`, so the cache is the sole creation path.

**3. pipeline_status endpoint — slice history, don't list() it**
- `list(status_dict["history_messages"])` walked the sequence protocol with one `__getitem__` RPC per element (history caps at ~10k lines). Changed to `[:]` (one RPC). The single `pipeline_status.copy()` under the lock is unchanged; the read-side truncation-to-latest-1000 is untouched.

## Tests

New `tests/kg/test_shared_storage_rpc_counts.py` (`@pytest.mark.offline`):
- counting proxy stand-ins asserting `set_all`/`clear_all` do **exactly one slice, zero per-index getitem, zero `__len__`**;
- namespace cache: cold-create then hot hit touches the shared dict **zero** times; workspace-key isolation; uninitialized `pipeline_status` still raises `PipelineNotInitializedError`; cache rebuilt (not reused) after finalize→reinit;
- real-Manager (`initialize_share_data(2)`): genuine nested ListProxy slice returns correct content; **cross-fork** test that a cached namespace proxy survives master-preload → worker-fork and the child's write is visible to the parent (the gunicorn preload scenario; skipped where `os.fork` is unavailable).

## Verification

- `./scripts/test.sh tests/kg tests/api tests/pipeline tests/workspace` — 1679 passed, 35 skipped.
- `pre-commit run ruff-format/ruff --files …` — passed.

## Scope notes (intentionally NOT in this PR)

- queue-stats aggregation (`aggregate_queue_stats`) is left as-is pending a real multi-queue Manager benchmark: a blanket `ns.copy()` per queue could trade one RPC for a larger serialization cost across a `/health` sweep.
- keyed-lock registry bookkeeping RPCs are a separate concern (deferred).

## Checklist

- [x] Changes tested locally
- [x] Unit tests added

### DIFF
--- a/lightrag/api/routers/document_routes.py
+++ b/lightrag/api/routers/document_routes.py
@@ -3745,10 +3745,12 @@ def create_document_routes(
             # Add processed update_status to the status dictionary
             status_dict["update_status"] = processed_update_status
 
-            # Convert history_messages to a regular list if it's a Manager.list
-            # and limit to latest 1000 entries with truncation message if needed
+            # Materialize history_messages to a regular list if it's a Manager
+            # ListProxy, then limit to latest 1000 with a truncation banner. A
+            # slice is one Manager RPC; list(proxy) walks the sequence protocol
+            # with one __getitem__ RPC per element (history can hold 10k lines).
             if "history_messages" in status_dict:
-                history_list = list(status_dict["history_messages"])
+                history_list = status_dict["history_messages"][:]
                 total_count = len(history_list)
 
                 if total_count > 1000:
--- a/lightrag/kg/json_doc_status_impl.py
+++ b/lightrag/kg/json_doc_status_impl.py
@@ -235,9 +235,12 @@ class JsonDocStatusStorage(DocStatusStorage):
         """
         async with self._storage_lock:
             if self.storage_updated.value:
-                data_dict = (
-                    dict(self._data) if hasattr(self._data, "_getvalue") else self._data
-                )
+                # DictProxy.copy() is a single Manager RPC that marshals the
+                # whole mapping server-side; dict(proxy) would walk the mapping
+                # protocol and fetch every value with its own RPC. Plain dicts
+                # (single-process mode) copy cheaply and identically — write_json
+                # only reads its argument, so the shallow copy is safe there too.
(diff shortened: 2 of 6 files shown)


danielaskdd:
@codex review

PR #3435, opened 2026-07-19
fix(utils): close compute_args_hash cache-key collision via length-prefixed encoding

### LINKED ISSUE #3392: [Bug]: compute_args_hash joins arguments without a delimiter, colliding distinct LLM-cache keys
### Describe the bug

`compute_args_hash(*args)` in `lightrag/utils.py` builds its hash input by concatenating the string form of every argument with no separator:

```python
args_str = "".join([str(arg) for arg in args])
```

With no field delimiter, two different argument tuples whose adjacent fields shift a character across the boundary produce the same `args_str`, and therefore the same MD5.

The query LLM-cache key is built from this function with free-text and integer parameters passed positionally (`lightrag/operate.py:3919` and `:5923`):

```python
args_hash = compute_args_hash(
    query_param.mode,
    query,
    query_param.response_type,
    query_param.top_k,
    query_param.chunk_top_k,
    ...
)
```

When LLM response caching is enabled, `handle_cache` looks up purely by `{mode}:{cache_type}:{args_hash}` and returns the stored content on a hit, without comparing the incoming query text against the cached one (`lightrag/utils.py:2877`). So a colliding key returns the LLM response that was cached for a semantically different query. `response_type` is a user-settable free string and `query` is arbitrary text, so the colliding inputs are constructible, not only theoretical.

### Steps to reproduce

Offline and deterministic against current `main` (HEAD `44db36f`, v1.5.5), no services needed:

```python
from lightrag.utils import compute_args_hash

# adjacent free-text boundary: (query, response_type)
a = compute_args_hash("hybrid", "abc", "x",  10, 20, 6000, 8000, 30000)
b = compute_args_hash("hybrid", "ab",  "cx", 10, 20, 6000, 8000, 30000)
assert a == b   # 7a8d68419f0e24d40fa2744d048d3a9c  (distinct queries, same key)

# adjacent integer boundary: (top_k, chunk_top_k)
c = compute_args_hash("hybrid", "q", "text", 1,  20, 6000, 8000, 30000)
d = compute_args_hash("hybrid", "q", "text", 12, 0,  6000, 8000, 30000)
assert c == d   # da767f47d42cd9eed6ac361767302816  (distinct configs, same key)
```

Both asserts pass on HEAD (the collisions reproduce). I ran this in a clean `python:3.12-slim` container importing the repo source directly, and confirmed `lightrag.utils.__file__` pointed at the checkout.

### Expected Behavior

Two semantically distinct argument tuples must map to different cache keys; an adjacent-field boundary shift must not collide. This is the same invariant `tests/llm/test_vlm_cache_key.py` already asserts for the VLM path (different content produces a different `args_hash`).

### Proposed fix, and a tradeoff I would rather you decide

The root cause is the delimiter-less join. There are two ways to close it, and the choice has a real cost, so I would prefer to agree with you before sending a PR:

1. Fix the primitive: join with a control character that will not appear in normal field text, e.g. `"\x1e".join(str(a) for a in args)` inside `compute_args_hash`. One line, covers every call site including future ones.
2. Per-site separators: insert an explicit sentinel between fields at the query cache-key sites, matching the `"\n<llm_identity>\n"` sentinels already placed at `operate.py:4186`. Leaves the shared primitive untouched.

Either option has a one-time cost: any change to the key material invalidates persisted LLM caches on upgrade, which is a single cold miss that repopulates itself. If you consider that churn not worth it, or prefer one option over the other, that is your call and I will follow it.

I have a regression test ready, modeled on `tests/llm/test_vlm_cache_key.py`: the two boundary tuples above assert distinct keys, plus a control that identical tuples still hit and an unrelated field change still misses. Glad to open a PR for whichever option you pick, or to drop it if you would rather not touch the hash.

### Additional Information

- LightRAG Version: 1.5.5 (main, HEAD 44db36f)
- Python Version: 3.12
- Storage/LLM provider: not relevant; this is pure hashing in `lightrag/utils.py`, reproduced with no backend configured

hata33:
Closes #3392

## Summary

Fixes a cache-key collision bug in `compute_args_hash` where adjacent field boundaries could shift a character and produce identical MD5 keys for semantically distinct argument tuples, causing the LLM response cache to return the wrong answer for a different query.

## Root Cause

`compute_args_hash(*args)` in `lightrag/utils.py` joined the string form of every argument with no delimiter (`''.join([str(arg) for arg in args])`). Adjacent fields lose their boundary: `('abc','x')` and `('ab','cx')` both produce `'abcx'`, hence the same MD5. Since `handle_cache` looks up purely by `{mode}:{cache_type}:{args_hash}` without comparing the incoming query text, a colliding key returns the cached LLM response for a different query.

The issue (#3392) confirmed two concrete collision families: (query, response_type) free-text boundary, and (top_k, chunk_top_k) integer boundary.

## Fix

- **Multi-argument calls use length-prefixed encoding** (`'{len}:{value}'` joined without separators), so each field's extent is unambiguously recoverable. `('abc','x')` -> `'3:abc1:x'`; `('ab','cx')` -> `'2:ab2:cx'` (distinct).
- **Single-argument (and zero-arg) calls preserve the legacy plain-str algorithm**, so `compute_mdhash_id(content)` — used to mint document/entity/relation IDs persisted across KV/vector/graph storage — keeps producing identical IDs across the upgrade.

### Why length-prefixing over a sentinel character?

The issue proposed two options (a `'\x1e'` join, or per-site sentinels). A control-character delimiter can still be constructed to collide because query text and free-form fields (`response_type`, `user_prompt`) can contain arbitrary characters. Length-prefixing is provably collision-free regardless of input content — it does not depend on any 'this character never appears' assumption, and it covers every current and future call site in one place.

## Backward Compatibility & Upgrade Note

This is **not** an API break and requires **no data migration** — persisted storage stays fully readable. The only upgrade-time effect is a one-time invalidation of the LLM response cache.

- **Persisted IDs — unchanged.** `compute_args_hash(content)` (single-arg) is verified byte-identical to raw `md5(content)`, so the `compute_mdhash_id` document/entity/relation IDs persisted across KV/vector/graph storage stay valid. No stored record breaks.
- **LLM response cache — one-time cold start.** Multi-arg hash values change, so existing cache entries no longer match their new keys. Two effects on upgrade, both benign:
  1. The first round of queries after upgrade re-hits the LLM and repopulates the cache — a one-time token/latency cost, largest for deployments carrying a big warm cache.
  2. Old-key entries become orphaned rows in the KV cache: never read again, no correctness impact, and reclaimable by clearing the LLM response cache if the extra storage matters.

No operator action is required; clearing the LLM cache is optional. The correctness risk of a wrong cache hit (returning another query's answer) far outweighs this one-time cold start.

## Validation

New regression tests in `tests/utils/test_compute_args_hash.py` (8 cases, modeled on `tests/llm/test_vlm_cache_key.py`):

- Issue case 1 (query/response_type boundary) now distinct
- Issue case 2 (top_k/chunk_top_k boundary) now distinct
- Inputs containing `'\x1e'` still distinct (immune to sentinel-based fix)
- Identical tuples still hash identically (cache HIT preserved)
- Single-arg hash equals raw MD5 — document IDs stable
- Empty-string edge case
- Surrogate Unicode handling does not regress
- Surrogate replacement does not collapse distinct fields

Existing VLM cache-key tests (`test_vlm_cache_key.py`, `test_llm_cache_identity.py`) continue to pass.

- `ruff check` + `ruff format --check`: pass
- New tests: 8 passed
- Existing cache tests: 9 passed (no regression)

### DIFF
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -643,9 +643,29 @@ def compute_args_hash(*args: Any) -> str:
         *args: Arguments to hash
     Returns:
         str: Hash string
+
+    Note:
+        - **Single-argument calls preserve the original algorithm** (plain
+          ``str(arg)``) so that ``compute_mdhash_id(content)`` keeps producing
+          the same document IDs across upgrades. Existing persisted IDs stay valid.
+        - **Two or more arguments** use length-prefixed encoding
+          (``"{len}:{str(arg)}"`` joined without separators) so that adjacent
+          field boundaries are unambiguously recoverable. This prevents the
+          cache-key collisions that the delimiter-less join could produce
+          (e.g. ``("abc","x")`` vs ``("ab","cx")`` both hashed to the same MD5).
+          Length-prefixing is used instead of a sentinel character because
+          query text and free-form fields can contain arbitrary characters,
+          so any fixed delimiter could still be constructed to collide.
     """
-    # Convert all arguments to strings and join them
-    args_str = "".join([str(arg) for arg in args])
+    if len(args) <= 1:
+        # Single-arg (or no-arg) path: keep the legacy behavior unchanged so
+        # compute_mdhash_id(content) document IDs remain stable.
+        args_str = "".join([str(arg) for arg in args])
+    else:
+        # Multi-arg path: length-prefix each field to make boundaries unique.
+        # Format: "3:abc1:x2:10" — the leading length makes the field extent
+        # unambiguous regardless of the field content.
+        args_str = "".join(f"{len(s)}:{s}" for s in (str(arg) for arg in args))
 
     # Use 'replace' error handling to safely encode problematic Unicode characters
     # This replaces invalid characters with Unicode replacement character (U+FFFD)
--- a/tests/utils/test_compute_args_hash.py
+++ b/tests/utils/test_compute_args_hash.py
@@ -1,0 +1,126 @@
+"""Regression tests for ``compute_args_hash`` cache-key collision.
+
+Verifies the length-prefixed encoding closes the delimiter-less join bug
+reported in issue #3392, where adjacent field boundaries could shift a
+character and produce identical MD5 cache keys for semantically distinct
+argument tuples.
+
+Coverage:
+- The two collision cases from the issue now produce distinct hashes.
+- Inputs containing the ``"\\x1e"`` record-separator (or any delimiter a
+  sentinel-based fix might have chosen) still do not collide.
+- Identical argument tuples still hash identically (cache HIT preserved).
+- Single-argument calls preserve the legacy hash value verbatim, so
+  ``compute_mdhash_id`` document IDs remain stable across upgrades.
+- Surrogate / invalid Unicode handling does not regress.
+"""
+
+from __future__ import annotations
+
+import hashlib
+
+import pytest
+
+from lightrag.utils import compute_args_hash, compute_mdhash_id
+
+
+pytestmark = pytest.mark.offline
+
+
+# ---------------------------------------------------------------------------
+# Collision cases from issue #3392 — must now produce DISTINCT hashes.
+# ---------------------------------------------------------------------------
+
+
+def test_adjacent_text_field_boundary_does_not_collide():
+    """Reproduces issue case 1: (query, response_type) boundary shift."""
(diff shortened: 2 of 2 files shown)


danielaskdd:
@codex review

PR #3436, opened 2026-07-20
perf(shared_storage): server-side atomic holder table for multiprocess keyed locks (~23 → 2 RPCs/cycle, kill-safe)

danielaskdd:
## Description

Follow-up to the shared-storage RPC reduction line (#3433/#3434). RPC profiling of a real multi-worker ingestion (~600k Manager round trips) showed the **keyed-lock subsystem accounted for more than half of all Manager RPCs**: `acquire`+`release` alone were 32%, plus the registry bookkeeping they drive. A single keyed-lock acquire+release cycle cost **≈23 RPCs**, of which only 8 carried the actual mutual exclusion — the other ~15 maintained three pure-bookkeeping tables.

The same profiling exposed a latent hazard: the keyed lock guarded its holder-record check-and-set with a client-held `_registry_guard = manager.RLock()`, and a `manager.RLock()`/`Lock()` whose holder is `kill -9`'d is **never released** (the server-side `threading.RLock` is left stranded by the disconnected client's connection thread) — every other process blocks forever.

This PR moves the whole check-and-set into a custom `SyncManager` type (`KeyedHolderTable`) whose methods execute atomically **inside the Manager server process**, achieving both goals at once:

1. **≈23 → 2 RPCs per uncontended acquire/release cycle** (one `try_acquire`, one `release`).
2. **The keyed path no longer holds any manager lock a dying client could strand** — the server's `threading.Lock` is scoped to each method by `with` and never spans an RPC, and the server does not die with a client, so the keyed-path kill deadlock is eliminated. (Scope note: `_internal_lock`/`_data_init_lock` remain plain Manager locks; removing them is a separate refactor, so this claim is for the keyed path only, not global.)

## Related Issues

Continuation of the Manager-RPC reduction work merged in #3433 / #3434 (which brought a real ingestion from 60s to 54s; the keyed-lock category was the dominant remainder).

## Changes Made

**Server-side holder table (`lightrag/kg/shared_storage.py`)**

- New `KeyedHolderTable` (lives in the Manager server; the holder record IS the lock: `{owner_pid, lease_id}` from the client, `start_delta` stamped by the server at grant time). `try_acquire` is a one-shot claim with dead-owner reclaim: liveness probes run outside the server lock, and a dead record is reclaimed only via a `lease_id` CAS on the exact record that was probed — N contenders produce exactly one winner. `release` is owner-checked by `lease_id`. `holder_count()` returns one int for /health without copying the table; `holders_snapshot()` serves diagnostics/tests.
- New `_HolderTableProxy` (explicit `_callmethod` wrappers — `BaseProxy` has no dynamic `__getattr__`, `_exposed_` alone is not callable) and `_LightRAGManager(SyncManager)` with a module-level `register()` so the server can import the type under both fork and spawn.
- `_KeyedLeaseLock` is now a per-acquire 0-RPC object driving the server method: uncontended acquire = 1 RPC, release = 1 RPC; the per-process async gate, backoff and cancellation points are unchanged.
- Deleted: `_lock_registry` / `_lock_registry_count` / `_lock_cleanup_data` (pure bookkeeping — nothing in the repo depended on them for correctness), `_registry_guard`, `_keyed_lock_holders`, `_get_or_create_shared_raw_mp_lock` / `_release_shared_raw_mp_lock`, the mp cleanup scheduling (`_earliest/_last_mp_cleanup_time`), the dead `self._mp_locks` declaration, and the now-unused `multiprocessing` imports. The **async-side** lock reuse + idle cleanup (`_get_or_create_async_lock` etc.) is correctness-required (same-process coroutines must share one `asyncio.Lock`) and is untouched, as is `_perform_lock_cleanup` (still used by the async side).
- `get_keyed_lock_status` (public API) is kept and still delegates to `get_lock_status`.

**Lock owner identity: server-stamped `start_delta` (clock-adjustment-safe)**

- Bare `psutil.create_time()` as identity was evaluated and rejected: it is wall-clock based, so an NTP/manual step changes a *live* process's reading → false PID-reuse verdict → double hold (psutil's own docstring notes create_time "may be affected by NTP"; `monotonic=True` exists only on private internal platform classes, not the public API, and not on Windows).
- Instead, identity is the **difference between the owner's and the Manager server's start time**, computed only inside the server (the server is its own anchor): a clock step is common-mode across one complete sample and cancels out. Linux **always** uses raw `/proc` start ticks (monotonic, immune to clock steps, integer equality criterion — detection keeps existing /proc tick precision). Off Linux, `create_time()` reads are **sandwich-sampled** (anchor read before and after the owner read; anchor mismatch = an adjustment crossed the window → bounded retries → `None` = conservatively alive; anchor readings are never cached). The non-Linux criterion is one-sided (`d1 > d0 + 1.0s` defensive tolerance): **declared limitation** — after a large *backwards* clock step a PID reuser can go undetected, which is a liveness gap (lock waits for the PID to die), never a double hold; an `abs()` criterion would trade that for a double-hold window and stays out until validated empirically.
- Because the server stamps identity at grant time, the lock path is structurally decoupled from client-side identity caches and their fork pitfalls.

**Liveness tightening**

- `_pid_alive`: zombies are now confirmed dead via `psutil.Process(pid).status() == STATUS_ZOMBIE` (a zombie executes no code and cannot hold a lock; `os.kill(pid, 0)` reports it alive, stalling reclaim until a wedged parent finally reaps). `psutil` is an **optional import** — it ships only with the `api` extra, so the core module falls back to the historical `os.kill` probe on minimal installs. **No dependency/installation contract changes** (pyproject/uv.lock untouched).
- `_my_start_id` cache is now **PID-aware**: the previous boolean "already computed" flag survived `fork`, so a worker inheriting the master's warmed cache published reservation records with `owner_pid=worker` but the *master's* start id — readers would misjudge a live worker as PID reuse. Fixed for the reservation layer (the lock path is immune by design now).
- Reservation identity is otherwise unchanged (Linux-only `/proc` start ids; cross-platform unification is a separate task). `_read_proc_starttime` / `_my_start_id` / `_process_alive` keep identical semantics and now serve only the reservation recovery layer.

**Health/diagnostics semantic change (explicit)**

- `total_mp_locks` now reports **currently held keys** (server-side `holder_count()`, one int RPC) instead of "lock objects in the registry (including idle ones awaiting cleanup)". Same key/shape, new value semantics — this PR does **not** claim a fully unchanged public API. `pending_mp_cleanup` and `mp_cleaned` are fixed at 0 and kept for response-schema compatibility (the WebUI renders the fields as strings; no rebuild needed).

**Safety argument (highlights)**

- Mutual exclusion: the holder table is the single slot; "holding ⟺ your record is in the slot". No double hold rests on (a) a live holder is never reclaimed — the server-side deadness check has no false positives by construction: uncertainty (polluted sample, missing identity, AccessDenied, `None` anywhere) always judges alive; and (b) one winner per dead record — the reclaim CAS compares the probed record's `lease_id` under the server lock.
- Kill safety: the server-side `threading.Lock` is held only inside a method that always returns; POC-verified that a SIGKILLed holder is reclaimed immediately by another process's `try_acquire` (versus a permanent hang with `manager.RLock()`).
- Connection-bound cleanup (`on_client_disconnect`) was evaluated and **rejected**: multiprocessing uses per-thread connections (connection ≠ process), so binding the lock to a connection releases early while the owner still lives — a real double-hold risk — and it requires rewriting undocumented `BaseManager.Server` internals for a purely cosmetic "prompt vs lazy" gain.
- No unbounded growth: the table only ever holds records for currently held keys (tighter than the old three tables).

**Tests**

- `tests/kg/test_keyed_lock_dead_worker_recovery.py` migrated: reads via `holders_snapshot()`, state seeded through production methods (`try_acquire` on an empty slot / `release` with a known lease); the owner-check case re-choreographed accordingly; the self-PID-reuse reclaim test rewritten on a direct `KeyedHolderTable` instance (forged records can't pass the proxy — the server always stamps the truth) and now runs on all platforms.
- New `tests/kg/test_keyed_holder_table.py` (22 tests, all `offline`): cross-process SIGKILL reclaim with a terminable contender child + join timeout (the suite can't hang on regression); zombie deadness; exact per-cycle RPC counts (1 acquire / 1 release) and proxy single-winner; reclaim-CAS single winner under 16-thread contention; real-liveness mutual-exclusion stress; 2000-distinct-key no-growth; `start_delta` semantics with injected time sources (common-mode offset invariance, polluted-window → `None` → never-dead, retry recovery, per-track criteria incl. the deliberate backwards-clock liveness gap, Linux never touching the wall clock, grant-path-only candidate stamping); PID-aware cache fix incl. a real-fork assertion; psutil-less fallback contract; spawn-start-method compatibility; new lock-status semantics.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (if necessary) — docstrings updated in place; health field semantics documented

## Additional Notes

- Verification: `./scripts/test.sh tests/kg tests/api tests/workspace tests/pipeline` — 1639 passed; the only failures (67) are pre-existing sandbox-environment issues (blocked tiktoken downloads etc.) and are **byte-identical** to the failure set on a clean `origin/main` checkout in the same environment. `ruff check .` and the pinned `pre-commit` `ruff-format`/`ruff` hooks pass.
- The end-to-end RPC re-profiling on a real ingestion workload could not be re-run here (the profiler script referenced in the plan is not in the repository); the per-cycle 23 → 2 reduction is pinned by the RPC-count regression test instead.
- Earlier "all-`False` liveness-mock stress" evidence was withdrawn by design review — that mock exceeds the algorithm's contract (with deadness forced true for live holders, reclaiming a live owner is permitted behavior), so this PR relies on the contract-shaped tests above instead.

https://claude.ai/code/session_01J3vuxN96pvghAMLmin45RC

---
_Generated by [Claude Code](https://claude.ai/code/session_01J3vuxN96pvghAMLmin45RC)_


danielaskdd:
@codex review

PR #3438, opened 2026-07-20
fix(utils): keep spaced <|> out of tuple delimiter repair

santhreal:
Entity descriptions that mention shell/prose forms like `a <|> b` were rewritten into `<|#|>` during delimiter repair, which then split the row into too many fields and dropped the entity.

Repair now only rewrites missing-core delimiters when they are glued to neighboring field text.

## Scope

The same starvation applies to relationship records (`<|>` in the 5-field description); both record shapes are now covered by regression tests.

## Known limitation

This is a heuristic that keys on whitespace: it preserves the common **spaced** free-text form (`a <|> b`, `x <|| y`) but a **glued** mention with no surrounding whitespace (`a<|>b`) is still indistinguishable from a real corrupted separator and will be repaired. For corpora with delimiter-heavy content (code, shell), the structural fix is JSON-structured extraction — set `ENTITY_EXTRACTION_USE_JSON=true`, which avoids the delimiter channel entirely. This PR hardens the default text path and the JSON→text fallback.

### DIFF
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -4820,9 +4820,9 @@ def fix_tuple_delimiter_corruption(
         record,
     )
 
-    # Fix: <|> -> <|#|>, <||> -> <|#|>
+    # Fix: <|> -> <|#|>, <||> -> <|#|> (glued only; keep free-text "a <|> b")
     record = re.sub(
-        r"<\|+>",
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3394, opened 2026-07-13
Fix ainsert_custom_chunks silently skipping knowledge-graph build

santhreal:
## Description

`ainsert_custom_chunks` ran `_process_extract_entities` inside `asyncio.gather` but **discarded its return value** and never called `merge_nodes_and_edges`. Entity and relationship extraction (and its LLM cost) ran, but nothing was written to the knowledge graph, so KG-dependent query modes (`local` / `global` / `hybrid` / `mix`) returned nothing for documents ingested through this path. Naive/chunk retrieval still worked, which is why the feature looked partially functional.

**Root cause** — in `lightrag/lightrag.py`, the extraction task's result was thrown away and there was no merge step, unlike the file pipeline (`pipeline.py`, which awaits `_process_extract_entities` and then calls `merge_nodes_and_edges`).

## Fix

Capture the extraction result and merge it into the graph, mirroring the file pipeline's merge stage: obtain `pipeline_status` / lock for the workspace and call `merge_nodes_and_edges` with the entity/relation storages and `doc_id`. An empty extraction result skips the merge.

Closes #3352

## Test

Added `test_custom_chunks_merge_extracted_entities_into_kg` (in `tests/pipeline/test_document_file_path_normalization.py`) asserting the extracted `chunk_results` reach `merge_nodes_and_edges` with the correct `doc_id`. This fails on the current code (merge is never called) and passes with the fix.

- `tests/pipeline/` → 255 passed
- Existing `test_custom_chunks_use_canonical_unknown_source_before_upsert` still passes
- `ruff format --check` and `ruff check` clean on the changed files

Mock-based test, no live services, per the contributor guidelines.

PR #3441, opened 2026-07-20
test(pipeline): mark pipeline tests offline for CI

ysys143:
### Problem

CI runs `pytest tests/ -m offline` (`.github/workflows/tests.yml`), and `tests/conftest.py` does not apply the `offline` marker automatically — its `pytest_collection_modifyitems` only adds `skip_integration`. Three files under `tests/pipeline/` declare no marker, so CI collects zero tests from them:

| file | tests |
|---|---|
| `tests/pipeline/test_content_hash_normalization.py` | 4 |
| `tests/pipeline/test_document_file_path_normalization.py` | 16 |
| `tests/pipeline/test_graph_keyed_locks.py` | 5 |

```
$ pytest tests/pipeline/test_document_file_path_normalization.py -m offline --collect-only
no tests collected (16 deselected)

$ pytest tests/pipeline/test_document_file_path_normalization.py
16 passed
```

The 16 in `test_document_file_path_normalization.py` include the custom-chunk regression tests added by #3394, #3401 and #3402 — `test_custom_chunks_merge_extracted_entities_into_kg`, `test_custom_chunks_persist_before_extraction`, `test_custom_chunks_rejects_when_pipeline_busy`, `test_custom_chunks_handoff_runs_even_when_flush_errors`, and others. #3416 substantially restructured that path, so those guardrails have not been running in CI.

### Change

Two commits:

1. **`test(pipeline): make graph keyed-lock tests self-sufficient`** — two tests in `test_graph_keyed_locks.py` call `ainsert_custom_kg`, which reads the `pipeline_status` namespace via `get_namespace_data` and raises `ValueError: Shared dictionaries not initialized` when `initialize_share_data()` was never called in the process. They passed only because other files collected earlier happened to initialize the singleton; run standalone the file failed 2/5. They now bracket their own `initialize_share_data(1)` / `finalize_share_data()` via a fixture mirroring `tests/kg/test_shared_storage_rpc_counts.py`. Doing this before adding the marker avoids handing CI an order-dependent test.

   (`initialize_pipeline_status()` is deliberately not called — `_raise_if_recovery_required` already treats a missing `pipeline_status` as a documented no-op via `PipelineNotInitializedError`; only the shared-data singleton is required.)

2. **`test(pipeline): mark pipeline tests offline for CI`** — adds `pytestmark = pytest.mark.offline` to the three files, following the precedent in `7aa5324c` ("test(utils): mark parse_optional_float tests offline for CI").

All three files need no network, database, or API credentials — every dependency is a hand-written fake or a `monkeypatch`/`unittest.mock` stub.

### Verification

CI on this branch is green on both Python versions, and reports the same counts on each:

```
3007 passed, 42 skipped, 910 deselected      # py3.12 and py3.14
```

The number this PR moves is `deselected`: **935 → 910**, i.e. the 25 tests listed above now run.

Measured locally before/after on one checkout, so the delta is attributable:

| | before | after |
|---|---|---|
| passed | 2972 | 2997 (**+25**) |
| skipped | 44 | 44 (**unchanged**) |
| deselected | 935 | 910 (**-25**) |

`skipped` staying flat is the check that the new `finalize_share_data()` teardown does not leak state into later-collected tests — no test moved from passed to skipped. 2972 + 25 = 2997 exactly, so nothing else shifted either.

(The local totals are lower than CI's because they were taken on macOS at an earlier base — macOS skips two tests Linux runs, and upstream has since added tests. The delta and the `deselected` figure are the load-bearing numbers.)

Order-independence confirmed after commit 1:

```
$ pytest tests/pipeline/test_graph_keyed_locks.py -m offline -q
5 passed                       # was: 2 failed, 3 passed

$ pytest tests/pipeline/test_graph_keyed_locks.py::test_ainsert_custom_kg_locks_every_entity_and_endpoint -q
1 passed                       # single test, alone
```

`pre-commit run --all-files` passes at the pinned `ruff` rev (`v0.15.17`).

### Note on scope

This is not limited to `tests/pipeline/` — repo-wide, 935 of 3929 tests are deselected by `-m offline`, and other directories have files in the same state (e.g. `tests/parser/test_hint_params.py`, `tests/kg/milvus_impl/test_milvus_index_config.py`). I kept this PR to `tests/pipeline/` so it stays reviewable; happy to open a separate issue for a repo-wide marker audit if that would be useful, since some of those files may genuinely not be offline-safe and would need checking individually rather than a blanket marker.

### DIFF
--- a/tests/pipeline/test_content_hash_normalization.py
+++ b/tests/pipeline/test_content_hash_normalization.py
@@ -10,10 +10,14 @@ and post-parse dedup fires.
 
 from __future__ import annotations
 
+import pytest
+
 from lightrag.utils_pipeline import (
     compute_text_content_hash,
     normalize_merged_text_for_hash,
 )
+
+pytestmark = pytest.mark.offline
 
 
 def _render(doc_hash: str, base_name: str) -> str:
--- a/tests/pipeline/test_document_file_path_normalization.py
+++ b/tests/pipeline/test_document_file_path_normalization.py
@@ -14,6 +14,8 @@ from lightrag.base import DocStatus  # noqa: E402
 from lightrag.constants import PROCESS_OPTION_CHUNK_FIXED  # noqa: E402
 from lightrag.pipeline import _PipelineMixin  # noqa: E402
 
+pytestmark = pytest.mark.offline
+
 
 class DummyRAG:
     def __init__(self):
--- a/tests/pipeline/test_graph_keyed_locks.py
+++ b/tests/pipeline/test_graph_keyed_locks.py
@@ -20,10 +20,31 @@ from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 
+from lightrag.kg.shared_storage import finalize_share_data, initialize_share_data
 
 # ---------------------------------------------------------------------------
 # Helpers
 # ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+def single_process_shared_data():
+    """Initialize the single-process shared_storage singleton.
+
+    ``LightRAG.ainsert_custom_kg`` calls ``_raise_if_recovery_required``,
+    which reads the ``pipeline_status`` namespace via ``get_namespace_data``.
+    That call raises ``ValueError`` if ``initialize_share_data()`` was never
+    called at all (as it never is running this file standalone), but is a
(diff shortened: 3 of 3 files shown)


danielaskdd:
@codex review

PR #3442, opened 2026-07-20
feat(api): add document source file download endpoint

### LINKED ISSUE #3363: [Feature Request]: Add `GET /documents/{doc_id}/file` endpoint to download MinerU files uploaded

## Summary

Currently, the REST API allows uploading documents via `POST /documents/upload`, but provides no way to retrieve the original files afterward. There is no endpoint to download an uploaded document by its ID.

## Motivation

When using LightRAG in a multi-user or production environment, users and integrations often need to:
- Verify that a file was correctly uploaded
- Re-download a document for review or archival
- Build UIs that allow browsing and downloading source files
- Integrate with document management workflows that expect bidirectional access

Right now, the only workaround is direct filesystem access to the `--input-dir` directory, which is not viable for remote deployments, containerized setups, or multi-tenant configurations.

## Proposed API

```
GET /documents/{doc_id}/file
```

**Behavior:**
- Returns the original file as a binary download (`Content-Disposition: attachment`)
- Uses the document ID already tracked in `DOC_STATUS_STORAGE`
- Returns `404` if the document does not exist or the file has been removed from disk
- Respects existing authentication (JWT / API key) like all other `/documents` routes

**Example:**
```bash
curl -X GET "http://localhost:9621/documents/{doc_id}/file" \
  -H "Authorization: Bearer <token>" \
  --output retrieved_document.pdf
```

## Implementation notes

The `DocumentManager` class in `lightrag/api/routers/document_routes.py` already tracks file metadata (name, path, status). The implementation would consist of:
1. Resolving the file path from `DOC_STATUS_STORAGE` using the provided `doc_id`
2. Checking the file exists on disk
3. Returning it via FastAPI's `FileResponse`

This should be a low-risk, additive change with no impact on existing behavior.

### Context

Setup: LightRAG with MinerU parser (`LIGHTRAG_PARSER=*:mineru-iteP`). Files are uploaded and indexed correctly, but there is currently no API-level way to retrieve them afterward.

### Alternatives considered

- Serving the `--input-dir` as a static directory: works locally but bypasses authentication and is not suitable for production
- Using the Core SDK `export_data`: only exports processed data, not the original files

### Additional Context

_No response_

amplifierplus:
## Description

Add an authenticated API endpoint for downloading the original source file associated with a document.

Uploaded source files may remain in the workspace input directory or be moved to the `__parsed__` archive after processing. The endpoint resolves both locations and returns the file using its canonical original filename.

## Related Issues

Closes #3363.

## Changes Made

- Add `GET /documents/{doc_id}/file`.
- Resolve the document source filename from `DOC_STATUS_STORAGE`.
- Support files in both the workspace input directory and the `__parsed__` archive.
- Support archived filename collision suffixes such as `_001`.
- Preserve the canonical original filename in the download response.
- Return HTTP 404 when the document record does not exist or its source file is missing.
- Reuse the existing authentication configured for document routes.
- Add regression tests covering active files, archived files, missing documents, and missing source files.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [ ] Documentation updated (not necessary)
- [x] Unit tests added

## Additional Notes

Validation completed locally:

- `69 passed` in `tests/api/routes/test_document_routes_docx_archive.py`
- Ruff lint passed
- Ruff format check passed
- `git diff --check` passed

### DIFF
--- a/lightrag/api/routers/document_routes.py
+++ b/lightrag/api/routers/document_routes.py
@@ -26,6 +26,7 @@ from fastapi import (
     Request,
     UploadFile,
 )
+from fastapi.responses import FileResponse
 from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 
 from lightrag import LightRAG
@@ -1203,6 +1204,16 @@ def get_doc_track_id(doc_status: Any) -> str:
     return str(track_id or "")
 
 
+def get_doc_file_path(doc_status: Any) -> str:
+    """Read file_path from dict or DocProcessingStatus-like objects."""
+    file_path = (
+        doc_status.get("file_path")
+        if isinstance(doc_status, dict)
+        else getattr(doc_status, "file_path", None)
+    )
+    return normalize_file_path(str(file_path or ""))
+
+
 async def get_existing_doc_by_file_path_candidates(
     doc_status: Any, file_path: Path | str
 ) -> dict[str, Any] | None:
@@ -1582,6 +1593,29 @@ def find_existing_file_by_file_path(input_dir: Path, file_path: str) -> Path | N
             if not candidate.is_file():
                 continue
             if normalize_file_path(candidate.name) == file_path:
+                return candidate
+    except FileNotFoundError:
+        return None
+    return None
+
+
+def find_document_source_file(input_dir: Path, file_path: str) -> Path | None:
+    """Find a document source in the input directory or its archive."""
+    source = find_existing_file_by_file_path(input_dir, file_path)
+    if source is not None:
+        return source
+
+    parsed_dir = input_dir / PARSED_DIR_NAME
+    try:
+        for candidate in parsed_dir.iterdir():
+            if not candidate.is_file():
+                continue
+            if (
+                canonicalize_archived_file_variant_basename(
+                    candidate.name, strip_archive_suffix=True
+                )
+                == file_path
+            ):
                 return candidate
     except FileNotFoundError:
         return None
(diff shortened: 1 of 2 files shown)


danielaskdd:
Downloading the original file poses a security risk and should not be made available until proper access control measures have been implemented.

danielaskdd:
Thanks for working on the source-file download request. The feature is useful, but after reviewing the resource semantics and security boundary, the current implementation needs a broader redesign before it can be merged.

I opened RFC #3516 to define the proposed artifact model, API namespace, exact locator lifecycle, and authorization design. I have also converted this PR to Draft while that design is discussed and implemented.

The main gaps in the current implementation are:

1. **A basename match does not prove `doc_id` ownership.** `doc_status.file_path` is a canonical, hint-stripped basename, not an exact physical locator. A file such as `report_001.pdf` only indicates an archive collision. It may be a processed source, a scan-time alias/duplicate that was never processed by this document, a post-parse duplicate, or a historical file. Scanning `__parsed__`, stripping `_001`, and returning the first match can therefore return bytes that do not belong to the requested `doc_id`.
2. **The actual archive target is not persisted.** `move_file_to_parsed_dir()` returns the real target, including collision suffixes, but normal parser call sites currently discard that return value. The source must have an exact persisted `source_location` from enqueue time, and that locator must be updated with the actual move result. Download must fail closed when the exact locator is unavailable; it must not infer numbered variants.
3. **The current endpoint does not leave a coherent artifact API for parser output.** We also need to download the `*.parsed/` directory in the future. The RFC proposes `GET /documents/{doc_id}/artifacts/{artifact_kind}`, initially supporting `source` and `parsed`, with the parsed directory safely packaged as ZIP from the persisted `full_docs.sidecar_location`.
4. **Authentication alone is insufficient for artifact download.** `combined_auth` only decides whether a JWT or the single global API key is accepted. It has no artifact permissions, and deployments without configured authentication are currently open. Source and parsed downloads should require explicit permissions before any document existence lookup.
5. **The authorization configuration and API-key model need expansion.** The RFC proposes a validated YAML/JSON policy file for roles, permissions, and bindings, plus multiple independently identifiable API keys. Clients can keep sending only `X-API-Key: <secret>`; server-side key IDs are used for role binding and audit attribution. The existing `LIGHTRAG_API_KEY` remains compatible but must not receive artifact-download permission implicitly.
6. **Location and archive safety need explicit contracts.** Source paths must remain inside the current workspace input root, parsed URIs must resolve inside the workspace artifact root, and ZIP creation must reject traversal and symlink escapes.

Please use #3516 as the design and discussion thread and rework this PR against the agreed RFC. In particular, the revised implementation should cover the complete `source_location` enqueue/archive/retry/scan/delete lifecycle, the artifact endpoint design, explicit permission checks, policy/API-key compatibility, and focused security/regression tests—not only replace the current file lookup helper.

Questions, alternatives, and implementation suggestions are welcome on #3516. Once the RFC questions are resolved and the implementation is updated comprehensively, the PR can be marked ready for review again.

amplifierplus:
Acknowledging here as well: I've picked up the design discussion on #3516 and agree this PR's basename-based lookup needs the broader rework you outlined there. I'll keep this PR closed and follow up with new, focused PRs against the RFC's incremental plan — starting with the `source_location`/`source_archive_location` persistence groundwork. Thanks for the detailed review.

PR #3443, opened 2026-07-20
test(pipeline): cover audit_kg_integrity's orphan detection path

ysys143:
### Gap

`audit_kg_integrity` reports graph contributions whose source chunks no longer resolve as `orphan_entities` / `orphan_relations`, and `README_KG_INTEGRITY_REPAIR.md` documents that class as the reason the tool exists:

> reports **irrecoverable orphans** (contributions whose source chunks no longer exist) — these are never modified automatically

But every orphan assertion in the suite is `== []`:

- `tests/pipeline/test_fault_injection_matrix.py:117-118, 262, 302`
- the orphan assertions in `tests/pipeline/test_purge_primitive.py`

No test ever puts a real orphan in front of the tool. If the orphan-detection logic silently regressed — stopped flagging nodes with an unresolvable `source_id` — nothing would fail.

The three nearest tests each miss it in a different direction:

| test | why it doesn't cover this |
|---|---|
| `test_absent_candidates_are_idempotent_noops` | reverse case — candidates present, nodes absent (superset safety) |
| `test_reprocess_with_fewer_candidates_overwrites_stale_anchor` | staleness at *write* time; never purges |
| `test_audit_tool_detects_and_repairs_missing_anchor` | entity is unanchored but still *resolvable* — lands in `missing_entity_anchors`, not `orphan_entities` |

That last one is the closest and the distinction matters: a "true" orphan is strictly stronger — a node whose `source_id` points only at chunks that no longer exist anywhere, so no owning document can be determined and the tool deliberately refuses to repair it.

### Change

One test, `test_audit_tool_detects_true_orphan_after_anchor_loss_and_purge`, added to the existing `tests/pipeline/test_fault_injection_matrix.py` (which already carries the `offline` marker).

It constructs true orphans exactly the way `kg_integrity_repair.py`'s module docstring describes a pre-write-ahead-anchor installation:

1. ingest a document normally, so the graph and both anchors exist — the fixture produces two entities and an edge between them
2. drop **both** the `full_entities` and `full_relations` anchor rows for it
3. run the ordinary whole-document purge — candidate discovery resolves to an empty set for each, so node and edge removal are both skipped, while chunk deletion proceeds
4. assert `audit_kg_integrity` reports the surviving nodes *and* the surviving edge by name

Asserted on the names and the exact edge pair rather than counts, so partial degradation in detection fails loudly.

Both classification branches are covered deliberately: `orphan_entities` and `orphan_relations` are separate code paths (`kg_integrity_repair.py:122` and `:132`), so a test that only produces an entity orphan leaves the relation branch free to regress silently — which would have reproduced the very `== []` non-coverage this PR is about.

The fixture wiring for the relation is test-local and layered on top of the shared default, so the other tests in this file keep their existing single-entity, zero-relation extraction.

### Scope — what this test does and does not claim

It asserts **the audit tool detects the orphan it was built to find**. It does *not* assert that purge leaving the node behind is correct or desired; that is a documented gap, and the test merely exploits it to get a real orphan on the board.

The docstring records this explicitly: if a future change closes the gap — e.g. `_purge_kg_contributions` learns to discover entities from the graph when the anchor row is missing — the entity would stop being orphaned and this test's assertion would need to be updated deliberately. That is an intended signal, not a regression to work around.

### Verification

Full offline suite before/after on one checkout, both Python versions CI uses:

| | py3.12 | py3.14 |
|---|---|---|
| before | 3005 passed, 44 skipped, 910 deselected | 3005 passed, 44 skipped, 910 deselected |
| after | 3006 passed, 44 skipped, 910 deselected | 3006 passed, 44 skipped, 910 deselected |

**+1 passed**, nothing else moved. Collected under `-m offline` (verified with `--collect-only`, not assumed), and passes both standalone and inside the full suite — no collection-order dependency.

Mutation-checked independently on each branch, since that is the whole point:

| mutation | result |
|---|---|
| neuter `orphan_entities.append(...)` (`:122`) | entity assertion **fails** |
| neuter `orphan_relations.append(...)` (`:132`) | relation assertion **fails**, entity assertion still passes |

The second row is what makes the two assertions independently load-bearing. A coverage test that cannot fail would not be worth adding.

`pre-commit run` passes at the pinned `ruff` rev.

### DIFF
--- a/tests/pipeline/test_fault_injection_matrix.py
+++ b/tests/pipeline/test_fault_injection_matrix.py
@@ -333,60 +333,120 @@ async def test_audit_tool_detects_and_repairs_missing_anchor(tmp_path):
         await rag.finalize_storages()
 
 
+def _wire_fake_extraction_with_relation(rag: LightRAG) -> None:
+    """Test-local extraction override: ALICE, BOB, and an edge between them
+    per chunk.
+
+    Overrides ``rag._process_extract_entities`` AFTER ``_build_rag`` has
+    already wired the shared ``_wire_fake_extraction`` (single-entity, no
+    relations) default. This function is applied only by the test below and
+    never mutates ``_wire_fake_extraction`` itself, so every sibling test in
+    this file that calls ``_build_rag`` keeps its original one-entity,
+    zero-relation wiring and is unaffected.
+    """
+
+    async def fake_extract(chunks, *args, **kwargs):
+        results = []
+        for chunk_id in chunks:
+            nodes = {
+                "ALICE": [
+                    {
+                        "entity_name": "ALICE",
+                        "entity_type": "person",
+                        "description": "ALICE description",
+                        "source_id": chunk_id,
+                        "file_path": "d.txt",
+                        "timestamp": 1,
+                    }
+                ],
+                "BOB": [
+                    {
+                        "entity_name": "BOB",
+                        "entity_type": "person",
+                        "description": "BOB description",
+                        "source_id": chunk_id,
+                        "file_path": "d.txt",
+                        "timestamp": 1,
+                    }
+                ],
+            }
+            edges = {
+                ("ALICE", "BOB"): [
+                    {
+                        "src_id": "ALICE",
+                        "tgt_id": "BOB",
+                        "description": "ALICE knows BOB",
+                        "keywords": "acquaintance",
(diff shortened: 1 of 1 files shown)


danielaskdd:
@codex review

PR #3446, opened 2026-07-20
fix(llm): do not cache token-limit-truncated LLM responses

DivyaNarahari97:
## Description

Truncated LLM responses were being written to the LLM cache. When an OpenAI(-compatible) completion stops with `finish_reason == "length"`, the adapter still returns the partial content so tolerant JSON parsing can salvage it — but that same partial payload was then persisted to the cache. On every later run the cache would replay the incomplete data (e.g. cut-off entity-extraction JSON), **even when a larger token budget would have produced the complete output**, silently degrading extraction quality.

This resolves the long-standing `# TODO LengthFinishReasonError should not persist into LLM cache` in `lightrag/llm/openai.py`.

## Related Issues

Resolves the in-code TODO in `lightrag/llm/openai.py` (caching of length-truncated responses). No external issue number.

## Changes Made

- **New marker** `TruncatedResponse` (a `str` subclass) plus `is_truncated_response()` helper in `lightrag/utils.py`. Because it *is* a `str`, all downstream text processing, sanitization, and tolerant JSON parsing keep working on the partial content unchanged — the only added meaning is "do not cache me."
- **OpenAI adapter** (`lightrag/llm/openai.py`): non-streaming completions that stop with `finish_reason == "length"` are wrapped in `TruncatedResponse` (with a warning log). The content is still returned unchanged for best-effort salvage.
- **Cache-write guards** skip persisting a truncated response:
  - `use_llm_func_with_cache` (entity/relation extraction) — the primary case from the TODO. The flag is captured *before* `remove_think_tags` rebuilds a plain `str` and drops the marker.
  - keyword-extraction cache and the `kg_query` / `naive_query` answer caches in `lightrag/operate.py`.
- Updated the resolved TODO comment and the adapter docstring note.

**Safety / backward-compat:** providers that don't emit the marker, and streaming responses, return `False` from `is_truncated_response` and therefore keep the previous "cache everything" behavior. The change is purely additive — it can never cache *less-correct* data than before, only avoid caching known-incomplete data.

## What could break

Nothing under normal operation. The only behavioral change is that a completion truncated by the token limit is no longer cached, so the next identical call re-invokes the LLM instead of replaying a partial result. That is the intended fix. Out of scope (noted for a follow-up): the multimodal/VLM cache path in `pipeline.py` (it has its own JSON-conformance retry loop and calls `str(result_text)`, which strips the marker — so no behavior change there), and emitting the marker from non-OpenAI providers.

## Additional Notes

Added regression tests:

- `tests/llm/openai_impl/test_openai_length_finish_reason.py` — the adapter flags a `finish_reason == "length"` result as truncated (and does **not** flag a normal `stop` result), while still returning the raw content.
- `tests/llm/test_utils_llm_cache.py` — `use_llm_func_with_cache` returns a truncated response to the caller but never writes it to the cache, and a subsequent call re-invokes the LLM (proving the partial result wasn't cached) and then caches the complete result.

Verification: `ruff check .` clean; `ruff format --check` clean on touched files; the targeted offline tests pass (19 in the two files above), and broader offline runs pass (`442 passed` in extraction/pipeline, `135 passed` in llm). The only failures in those wider runs are pre-existing missing-optional-dependency import errors (aioboto3, ollama, qdrant_client), unrelated to this change.

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -3492,7 +3492,10 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
                     enable_cot=True,
                     stream=param.stream,
                 )
-                if type(response) is str:
+                # isinstance, not exact type: a truncated non-streaming
+                # response arrives as TruncatedResponse (a str subclass) and
+                # must not be misclassified as a streaming iterator.
+                if isinstance(response, str):
                     return {
                         "status": "success",
                         "message": "Bypass mode LLM non streaming response",
--- a/lightrag/llm/openai.py
+++ b/lightrag/llm/openai.py
@@ -29,6 +29,7 @@ from lightrag.utils import (
     wrap_embedding_func_with_attrs,
     safe_unicode_decode,
     logger,
+    TruncatedResponse,
 )
 
 from lightrag.api import __api_version__
@@ -222,7 +223,9 @@ def create_openai_async_client(
         return AsyncOpenAI(**merged_configs)
 
 
-# TODO LengthFinishReasonError should not persist into LLM cache
+# Token-limit-truncated completions (finish_reason == "length") are returned
+# wrapped in TruncatedResponse so the cache layer skips persisting incomplete
+# output. See the "Note on truncated structured output" in the docstring below.
 @retry(
     stop=stop_after_attempt(3),
     wait=wait_exponential(multiplier=1, min=4, max=10),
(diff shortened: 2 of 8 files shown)


danielaskdd:
Pushed 8003cc6 (maintainer edit) after a coverage review:

- **Azure needs no source change** — `azure_openai_complete_if_cache` delegates to the unified `openai_complete_if_cache` (only client construction differs), so it already emits `TruncatedResponse`. Added a regression test pinning that delegation.
- **Fixed a bypass-mode edge this PR introduced**: `aquery_llm`'s bypass branch used `type(response) is str`, so a non-streaming `TruncatedResponse` (str subclass) was misclassified as streaming, handing callers a bare str as `response_iterator`. Switched to `isinstance` with a fix-proof test (fails on the previous commit, passes now).
- Added a `naive_query` guard behavior test: a truncated answer is returned but never cached, and the next identical query reaches the LLM again.

Query-mode coverage verified: local/global/hybrid/mix (kg_query), naive, and the keyword cache are all guarded; bypass writes no LLM cache. Extraction-side coverage confirmed: initial extraction, gleaning, summary merge, and parser LLM calls all flow through the `use_llm_func_with_cache` guard. The remaining gap stays as declared: the multimodal analysis cache in `pipeline.py` strips the marker at `_run_json_conformance_retry`'s `str(raw)` before it can be observed — follow-up material.

danielaskdd:
@DivyaNarahari97 Thank you for this contribution — the `TruncatedResponse` marker design turned out to be a solid foundation: the str-subclass approach kept every downstream consumer working unchanged, and the fail-open semantics made extending it straightforward.

A follow-up PR is now open: #3448 (stacked on this branch). It extends the marker to the remaining server bindings with a readable truncation signal (gemini / ollama / bedrock) and implements the multimodal analysis cache guard that was declared out of scope here. Your review would be much appreciated, especially on whether the per-driver truncation signals match your reading of each provider's API contract.

Merge order: this PR first, after which #3448's diff reduces to a single commit.

DivyaNarahari97:
@danielaskdd - Thank you! I'm really glad the approach worked well and could serve as a solid foundation. I appreciate the detailed feedback. I'll take a look at #3448, Thanks again!

PR #3447, opened 2026-07-21
fix(lightrag): fall back when COSINE_THRESHOLD env is empty

santhreal:
Setting `COSINE_THRESHOLD=` in `.env` or Compose (present but empty) crashes `from lightrag.lightrag import LightRAG` because the field default used bare `float(os.getenv(...))`.

Use `get_env_value` like the other LightRAG env defaults so empty/whitespace falls back to `DEFAULT_COSINE_THRESHOLD` (0.2).


danielaskdd:
@codex review

PR #3448, opened 2026-07-21
feat(llm): extend truncation marker to gemini/ollama/bedrock and the multimodal analysis cache

danielaskdd:
## Description

**Stacked on #3446** — this branch contains #3446's commits plus one new commit ([`46bf2f6`](https://github.com/danielaskdd/LightRAG/commit/46bf2f61)); merge #3446 first, after which this PR's diff reduces to the new commit.

Extends the `TruncatedResponse` no-cache marker from the unified OpenAI/Azure adapter to every server binding whose API exposes a truncation signal, and implements the multimodal follow-up declared in #3446.

## Changes Made

**Driver markers** (non-streaming paths, each wrapped as the last transformation before return):

| Binding | Signal | Notes |
|---|---|---|
| gemini | `candidates[0].finish_reason == FinishReason.MAX_TOKENS` | wrapped after `remove_think_tags` |
| ollama | `done_reason == "length"` | `.get` works on both the raw dict (`ollama<0.4`) and `SubscriptableBaseModel` `ChatResponse` (`>=0.4`, verified on 0.6.2); missing key → previous behavior |
| bedrock | Converse API `stopReason == "max_tokens"` | top-level dict field per AWS contract |
| lollms | — | **not coverable**: `/lollms_generate` returns bare text with no finish-reason metadata |

**Multimodal analysis cache** (`pipeline.py`):
- `_run_json_conformance_retry` kept `str(raw)`, which rebuilt a plain str and dropped the marker before the cache write could observe it; str input is now passed through as-is.
- Both `save_to_cache` sites (VLM drawings, EXTRACT text/equation) skip persisting a truncated analysis (warning logged) and leave `llm_cache_list` untouched, so document-delete cleanup never references a cache entry that was never written, and the next run re-invokes the model.

All changes stay fail-open: a driver that doesn't emit the marker, a missing signal field, or streaming responses keep the previous cache-everything behavior.

## Additional Notes

Tests (all offline):
- Marker/no-marker pairs per driver; the gemini fake `types` namespace gained `FinishReason` (the real enum exists since `google-genai>=1.0.0`, the declared floor), and an ollama missing-`done_reason` degradation case.
- End-to-end multimodal case: a truncated-but-parseable VLM analysis is used for the current run, writes no `default:analysis:` cache entry, writes no cache id back to the sidecar, and the next run re-invokes the VLM. Verified to fail against the pre-guard behavior (fix-proof).

Verification: `tests/llm` + `tests/pipeline` offline suites pass (502 passed); pre-commit (ruff format + ruff) clean.

### DIFF
--- a/lightrag/llm/bedrock.py
+++ b/lightrag/llm/bedrock.py
@@ -20,7 +20,11 @@ from tenacity import (
 from collections.abc import AsyncIterator
 from typing import Any, Union
 
-from lightrag.utils import wrap_embedding_func_with_attrs
+from lightrag.utils import (
+    TruncatedResponse,
+    logger,
+    wrap_embedding_func_with_attrs,
+)
 
 # Import botocore exceptions for proper exception handling
 try:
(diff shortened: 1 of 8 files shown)


danielaskdd:
@codex review

PR #3451, opened 2026-07-22
fix(lightrag): fall back when env-backed int defaults are empty

santhreal:
Clearing a documented int knob such as `EMBEDDING_BATCH_NUM=` in `.env` or Compose crashes `from lightrag.lightrag import LightRAG` because several field defaults still used bare `int(os.getenv(...))`.

Route those defaults through `get_env_value` like the COSINE_THRESHOLD fix so empty/whitespace falls back to the documented defaults.


danielaskdd:
@codex review

PR #3452, opened 2026-07-22
fix(parser): fall back when CHUNK_R_SEPARATORS env is empty

santhreal:
Setting `CHUNK_R_SEPARATORS=` in `.env` or Compose crashes LightRAG startup because `default_chunker_config` used bare `json.loads(os.getenv(...))`.

Fall back to `DEFAULT_R_SEPARATORS` on empty or invalid JSON, matching the multimodal sibling path.


danielaskdd:
@codex review

PR #3454, opened 2026-07-22
fix(llm): honor LLM_TIMEOUT in Bedrock driver and server wiring

danielaskdd:
## Description

The Bedrock binding ignored `LLM_TIMEOUT` entirely: nothing in the driver or the API server ever consumed a timeout value, so every `converse` / `converse_stream` call ran with botocore's default **60s** connect/read timeouts (`botocore.endpoint.DEFAULT_TIMEOUT = 60`). Any generation longer than 60 seconds (long extraction prompts, large `max_tokens`, reasoning models) failed with:

```
ERROR:root:Bedrock converse timeout error: Read timeout on endpoint URL
```

regardless of how large `LLM_TIMEOUT` was set, while the OpenAI binding honored it correctly (`kwargs["timeout"]` → `AsyncOpenAI`).

Note that the timeout could not simply be passed through `**kwargs`: `bedrock_complete_if_cache` forwards kwargs into `client.converse(**args, **kwargs)`, and `timeout` is not a Converse API parameter (it would raise `ParamValidationError`). The driver has to consume it explicitly and translate it into a botocore client `Config`.

## Related Issues

N/A (reported via Bedrock deployment: `LLM_TIMEOUT=240` but requests timed out after ~60s)

## Changes Made

- **`lightrag/llm/bedrock.py`**
  - `bedrock_complete_if_cache()` gains an explicit `timeout: int | None = None` parameter (seconds).
  - `_bedrock_client_kwargs()` maps it to `botocore.config.Config(connect_timeout=timeout, read_timeout=timeout)` — equivalent to how the OpenAI driver's scalar timeout applies to every httpx request phase. `None` keeps botocore defaults (no `config` injected), so direct library callers are unaffected.
  - Applied on both the streaming and non-streaming client constructions; docstring documents the unit and that the streaming read timeout bounds the wait for the next chunk, not the whole response.
- **`lightrag/api/lightrag_server.py`**
  - `bedrock_model_complete` now passes the global `LLM_TIMEOUT` and `role_bedrock_complete` passes the role timeout (including `<ROLE>_LLM_TIMEOUT` overrides), via `kwargs["timeout"] = ...` — same pattern as the OpenAI/Azure/Gemini wrappers, so the server-configured value overrides caller-passed ones without duplicate-keyword `TypeError`s.
- **`tests/llm/bedrock_impl/test_bedrock_llm.py`** — 7 new regression tests
  - Driver level: timeout maps to client `Config` (non-stream + stream via a new `_FakeStreamingBedrockClient` that actually consumes the async stream), timeout never leaks into Converse parameters, and no `config` is injected when timeout is absent.
  - Server level (`create_app` wiring): base `llm_model_func` passes `args.llm_timeout`, roles inherit the global timeout, role-specific timeout wins, and a caller-passed `timeout` is overridden without raising.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

- Fix-proof verified: with the source changes stashed, all 6 wiring/config tests fail on the old code (the backward-compat test passes by design); restored, the full file passes (36 passed) plus `tests/llm` + `tests/api/config/test_api_config_bedrock.py` (267 passed) and pinned ruff/ruff-format via pre-commit.
- Out of scope, noted for a follow-up discussion: botocore's default retry mode nests with the driver's outer tenacity retry (5 attempts); collapsing it via `Config(retries={"max_attempts": 1})` would be a behavior change and is intentionally not part of this bug fix. Embedding timeouts are also untouched — no provider currently receives `EMBEDDING_TIMEOUT`, so that would be a separate cross-provider change.

### DIFF
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -1839,6 +1839,9 @@ def create_app(args):
                 ) -> str:
                     if history_messages is None:
                         history_messages = []
+                    # Server-configured timeout overrides any caller-passed value,
+                    # matching the OpenAI/Azure/Gemini role wrappers.
+                    kwargs["timeout"] = role_timeout
                     if role_provider_options:
                         kwargs = {**role_provider_options, **kwargs}
                     return await bedrock_complete_if_cache(
@@ -1966,6 +1969,9 @@ def create_app(args):
         # which drops them and emits deprecation warnings when booleans are set.
         if config_cache.bedrock_llm_options:
             kwargs = {**config_cache.bedrock_llm_options, **kwargs}
+        # Server-configured timeout overrides any caller-passed value, matching
+        # the OpenAI wrapper (create_optimized_openai_llm_func).
+        kwargs["timeout"] = llm_timeout
 
         return await bedrock_complete_if_cache(
             args.llm_model,
--- a/lightrag/llm/bedrock.py
+++ b/lightrag/llm/bedrock.py
@@ -10,6 +10,9 @@ if not pm.is_installed("aioboto3"):
     pm.install("aioboto3")
 import aioboto3
 import numpy as np
+
+# botocore is a hard dependency of aioboto3, so this import is always safe.
+from botocore.config import Config as BotocoreConfig
 from tenacity import (
     retry,
     stop_after_attempt,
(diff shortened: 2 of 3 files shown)


danielaskdd:
@codex review

PR #3455, opened 2026-07-22
🐛 fix(parser): keep drawing path local-only, carry external image refs in src

danielaskdd:
## Description

The sidecar spec defines a drawing's `path` as the local cache path (always inside `*.blocks.assets/`) and `src` as the original reference from the source document. Three parser chains violated that contract:

- **native markdown**: when an external image download failed, the fallback wrote the same remote URL into **both** `path` and `src`, producing the reported duplicate:
  ```
  <drawing id="im-…-0001" format="" path="https://camo.githubusercontent.com/…" src="https://camo.githubusercontent.com/…" />
  ```
- **native docx**: linked images (`a:blip@r:link`, VML external) put the link target into `path` while `src` stayed empty (misplaced, the mirror image of the same defect).
- **Docling**: `http(s)` picture URIs travelled verbatim into `path` via `IRDrawing.path_override`, `src` stayed empty.

MinerU already implemented the correct behavior (missing image ⇒ `path=""`, URLs never reach `path`) and needs no change.

After this PR, all four chains converge on one contract: `path` is local-only — it renders as `""` whenever the image bytes were not materialized — and the original reference (remote URL or relative external target) survives only in `src`.

## Related Issues

N/A (reported directly from sidecar output inspection).

## Changes Made

**🐛 fix(parser)** — carry external image refs in `src`, keep `path` local-only
- `parser/markdown/ir_builder.py`: the external (download-failed) branch no longer sets `path_override`; the URL is kept only in `src`.
- `parser/docx/drawing_image_extractor.py`: linked (`a:blip@r:link`) and VML-external images now emit the link target as the `src` attribute instead of `path`; `src` promoted to an explicit placeholder key.
- `parser/docx/ir_builder.py`: non-asset `path` values are dropped instead of passed through; missing `format` is now inferred from `path` **or** `src` via the urlparse-aware `_infer_format_from_target` (None coerced to `""`, never the string `"None"`).
- `parser/external/docling/ir_builder.py`: the `http(s)` URI branch keeps the URL as `src` with no asset ref (renders `path=""`); the drawing itself still survives.

**♻️ refactor(sidecar)** — remove dead `IRDrawing.path_override`
- After the fix no adapter assigns it, and its docstring codified the exact contract this PR removes ("external URL travels in path"). Field and the writer's two verbatim pass-through branches deleted (internal API, no shim).

**📝 docs(sidecar)** — spec §3.3 / §4 updated in both `LightRAGSidecarFormat.md` and `-zh.md`: non-empty `path` always points inside `*.blocks.assets/`; empty `path` ⇒ not cached; `src` carries the original reference.

**🐛 fix(parser)** — native docx drawing paths rendered with the assets prefix (review follow-up)
- The docx parser dropped its `sidecar_path_style = "basename_only"` override (a purely cosmetic byte-equivalence relic of the pre-migration adapter), so `blocks.jsonl` now renders `<drawing path="<base>.blocks.assets/fig1.png">` like every other chain, matching `drawings.json` and the spec contract. The writer's `basename_only` branch stays only for `write_sidecar` public-API keyword compatibility; no in-tree adapter uses it.

**Tests**
- New `tests/parser/docx/test_drawing_image_extractor.py` (linked URL / relative target / VML external → `src`, embedded → `path` regression anchor).
- New markdown e2e: download failure renders `path="" src="<URL>"` in both `blocks.jsonl` and `drawings.json`, plus blockid/content-hash determinism anchoring.
- New docling test: `http(s)` picture URI kept as `src`, no asset, drawing not dropped.
- `test_ir_builder_security.py` rewritten to the new contract (incl. format-from-src inference and extensionless-URL `fmt==""`); `external_image_link` fixture + golden regenerated.
- Regression proof: the 5 new tests all **fail** on current `main` and pass on this branch.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Two intentional behavioral changes (not cosmetic), for reviewers:

1. **Data-identity migration**: `blockid` covers the rendered drawing tag and `content_hash` normalization does not touch remote URLs, so documents containing external images — and, via the docx `with_prefix` follow-up, docx documents containing local images — get new blockids/content hashes after re-parse; post-parse cross-filename dedup will treat them as new content once. Accepted deliberately to avoid adding drawing-tag parsing + legacy-representation compatibility to the hash normalizer (a semantic normalization of the old representation remains possible as a follow-up if cross-version dedup stability becomes a priority).
2. **docx relative external links no longer reach VLM**: previously a relative `path` like `../images/legacy.gif` could resolve against the sidecar directory and be read into VLM analysis. With `path=""` such references are explicitly excluded (also closing an out-of-sidecar relative read). This deliberately replaces the earlier "preserve the external path verbatim" contract that the `external_image_link` fixture used to encode.

Full suite: 4071 passed, 36 skipped (environment-gated spaCy markers). Golden regen is idempotent on this branch (clean worktree after re-run).


danielaskdd:
Review follow-up addressed in 612154a1: native docx still overrode `sidecar_path_style = "basename_only"`, so `blocks.jsonl` rendered `path="fig1.png"` for a file living at `<base>.blocks.assets/fig1.png` — violating the contract this PR establishes. The override was purely cosmetic byte-equivalence with the pre-migration adapter (consumers read paths from `drawings.json`, which already carried the prefix), so docx now uses the default `with_prefix` and the two golden scenarios with local assets were regenerated.

The writer's `basename_only` branch is kept solely because `block_drawing_path_style` is a keyword of the public `write_sidecar` API; no in-tree adapter passes it anymore, and its docstring now says new callers should not.

Verified: full suite 4071 passed / 36 skipped; golden regen idempotent. blockid/content_hash of docx docs with local images shift once — folded into the data-identity migration already declared in the PR notes.

PR #3458, opened 2026-07-22
feat(kg): workspace-scoped pipeline ingress infrastructure (Phase 1 of request_pending replacement)

danielaskdd:
## Summary

Phase 1 of the reviewed multi-phase design that replaces the pipeline's `request_pending` flag with a workspace-scoped **ingress mailbox**, fixing the batch-barrier defect where one long-running document blocks every later-enqueued file until the whole batch finishes. This phase lands pure infrastructure — **no pipeline behavior change; nothing consumes the ingress yet**.

### Three channels, three reliability grades (deliberately not one destructive FIFO)

| channel | semantics | crash story |
|---|---|---|
| `document` | ephemeral doc-id notifications | rebuildable from `doc_status` + next run's initial scan |
| `auto rescan` | single dirty flag, consumed atomically via `consume_auto_rescan()` | consumer must re-arm on strict-query failure |
| `manual retry` | sticky FIFO requests (`/scan`, `/reprocess_failed` intent), ACKed only after the FAILED→PENDING reset persists | survives worker SIGKILL while the Manager server lives; terminal ids (ACKED / CANCELLED_BY_CLEAR) refuse delayed replays within a bounded window |

### Implementation

- **`lightrag/kg/pipeline_ingress.py`** — `PipelineIngressMessage`, server-side `PipelineIngressMailbox` (single `threading.Condition`; every proxy method is one atomic RPC; `wait_for_documents`/`wait_for_items` **observe only, never dequeue**, so a cancelled/SIGKILLed waiter can never steal a message; waits take a **required, clamped timeout** so a dead client strands its server thread at most `MAX_MAILBOX_WAIT_SECONDS`), explicit `BaseProxy` (KeyedHolderTable pattern), and `AsyncioPipelineIngress` (real `asyncio.Queue` document channel with strictly paired `task_done`, `asyncio.Event` waits, owning-loop fast-fail).
- **`lightrag/kg/shared_storage.py`** — Manager registration, per-process registry + multiprocess Manager-dict discovery table (nested-proxy pattern like `_update_flags`), `initialize/get/finalize_pipeline_ingress`, lifecycle wiring into `initialize_share_data`/`finalize_share_data`. Multiprocess `finalize_pipeline_ingress` clears only the local cache — popping the discovery entry while sibling workers hold cached proxies would split-brain the workspace (Manager-container values unpickle `manager_owned`, holding no refcount).
- Feeder-facing waits are **document-channel-only by contract**: waiting on all three channels without consuming the control ones would busy-loop on a pending manual/auto entry.

## Tests

- `tests/kg/test_pipeline_ingress.py` — 19 tests: workspace isolation, idempotent resolution, FIFO manual retries + peek/ACK + terminal replay guard, clear-tombstone, atomic auto-rescan exchange, document-wait isolation from control channels, task_done pairing, cross-loop fast fail, bounded wait clamp, and real-Manager multiprocess coverage (cross-process publish via the discovery table, sticky survival after publisher death, cancelled waiter steals nothing, finalize keeps the discovery table intact).
- Full `tests/kg` + `tests/pipeline` regression: **1434 passed, 22 skipped**.
- storage-backend-reviewer pass done; both should-fix findings (multiprocess finalize split-brain, unbounded mailbox wait) are fixed in this PR.

Next phases (separate stacked PRs): Phase 0 (FAILED-retry semantics split + sticky manual intent), Phase 2 (in-batch feeder + dual write), Phase 3 (queue-based atomic exit + cancellation closure), Phase 4 (delete `request_pending` repo-wide).

### DIFF
--- a/lightrag/kg/pipeline_ingress.py
+++ b/lightrag/kg/pipeline_ingress.py
@@ -1,0 +1,551 @@
+"""Workspace-scoped pipeline ingress: the notification channel that replaces
+``request_pending``.
+
+One ingress instance exists per workspace (see ``shared_storage.
+get_pipeline_ingress``).  It carries three channels with three distinct
+reliability grades — deliberately NOT a single destructive FIFO:
+
+* **document** — ephemeral doc-id notifications published by enqueue after the
+  ``doc_status`` row is persisted.  Messages may be lost or duplicated freely:
+  ``doc_status`` is the source of truth and the next processing run's initial
+  scan rebuilds anything dropped.  Consumers re-read storage by id and skip
+  stale/terminal/duplicate ids.
+* **auto rescan** — a single dirty flag.  Publishers that cannot name precise
+  doc ids (busy-refused processing requests, partial doc-status commits,
+  recovery paths) set it; the pipeline supervisor is the only consumer and
+  resets it atomically via :meth:`consume_auto_rescan`.
+* **manual retry** — sticky FIFO requests carrying user intent from
+  ``/documents/scan`` and ``/documents/reprocess_failed``.  A request stays in
+  the mailbox until explicitly ACKed *after* the FAILED→PENDING reset has been
+  persisted, so a consumer crash at any point either leaves the request sticky
+  (re-executed by the next run) or leaves the documents already PENDING
+  (recovered by the automatic paths).  Terminal request ids (ACKED or
+  CANCELLED_BY_CLEAR) are remembered in a bounded FIFO set so a delayed replay
+  of the same id is refused — a *bounded-window* idempotency guarantee, not an
+  indefinite exactly-once (ids are server-generated UUIDs, so eviction of very
+  old ids is harmless in practice).
+
+Waiting primitives are split by consumer role: a feeder that only routes
+documents must wait on the document channel alone (``wait_for_documents`` /
+:meth:`AsyncioPipelineIngress.get_document`); waiting on all three channels
+(``wait_for_items``) is reserved for a supervisor that actually consumes the
+control channels — otherwise a pending manual/auto entry keeps ``has_work()``
+true and turns the wait into a busy loop.
+
+This module is dependency-free on purpose: :mod:`lightrag.kg.shared_storage`
+imports it to register :class:`PipelineIngressMailbox` on the LightRAG
+``SyncManager`` subclass and to build the per-workspace registry.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import threading
+from collections import OrderedDict, deque
+from dataclasses import dataclass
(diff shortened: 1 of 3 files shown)


danielaskdd:
Review addressed in 7e00707 — all four findings, with the high-severity one solved structurally rather than patched:

**1. Client-held creation lock (high) → hub topology.** Instead of hardening `data_init_lock`, the per-workspace mailbox proxies and the Manager-dict discovery table are gone entirely. One server-side `PipelineIngressHub` is created pre-fork (same topology as `KeyedHolderTable`); workers inherit the single hub proxy and bind namespace strings to it. Mailbox creation is atomic inside one server-side dispatch — `get_pipeline_ingress` no longer acquires any lock, so a SIGKILLed client can neither strand creation nor split-brain a workspace (workspace identity is the namespace string, not a proxy lifetime). This also structurally retires the earlier `manager_owned`-refcount concern: mailboxes are never individually proxied. Regression test: a spawn child acquires `_data_init_lock` and SIGKILLs itself; the lock is verified stranded, and ingress resolution for a fresh workspace still completes within 5s.

Note on the alternative you sketched (server-side `get_or_create` returning a proxy via `method_to_typeid`): verified empirically that the `'#PROXY'` return path requires `proxy._manager._registry` (and `_manager._state` in `_incref`), which is `None` on spawn-unpickled proxies — making it fork-topology-only without poking CPython internals. The hub avoids that entire class of issues.

**2. Closed-loop strictness (medium).** `_ensure_ingress_on_current_loop` now follows the `lightrag.py` sync-wrapper convention: owning loop closed → in-place `rebind_to_current_loop()` (queue/event rebuilt on the current loop; queued documents, manual requests, auto flag and terminal tombstones all carried over, instance identity preserved for cached references). Only a still-running foreign loop is rejected. Tests: two-`asyncio.run` migration with full state assertions, plus a live-owner-loop-in-another-thread fast-fail.

**3. Unbounded document channels (medium).** Both backends now bound the channel (`DOCUMENT_CHANNEL_CAPACITY = 4096`, per workspace). Overflow never blocks or fails the producer: the notification is dropped and **coalesced into the auto-rescan dirty flag**, and the consumer recovers the missed docs from `doc_status` — exactly the "notifications are droppable, state is not" contract. Overflows are counted in `counts()`. Tests cover fill/overflow/recovery on both backends.

**4. Document message validation (low).** Shared `_validate_document_message` at both publishers (`kind='document'`, non-empty `doc_id`, no `retry_failed`/`request_id`), same exception type everywhere; `ManagerPipelineIngress` validates pre-RPC so errors surface at the call site.

Suite: 23 tests (was 19); `tests/kg` + `tests/pipeline`: **1440 passed, 22 skipped**.

PR #3459, opened 2026-07-22
feat(pipeline): FAILED retries only via sticky manual intent (Phase 0, stacked on #3458)

danielaskdd:
## Summary

Phase 0 of the reviewed `request_pending` replacement, **stacked on #3458** (Phase 1 ingress infrastructure — merge that first; this diff shows only Phase 0 once rebased). Implements the FAILED-retry semantics ruling: a FAILED document re-enters the pipeline **only through an explicit human request**, carried crash-consistently by the workspace ingress.

### 1. Strict scheduling-read contracts (storage layer)

- `BaseDocStatusStorage.get_docs_by_statuses(..., strict: bool = False)` — strict is **complete-or-raise** (transport/pagination failure, incomplete response, undeserializable record); relaxed keeps the historical skip-and-log for UI paths. All five backends implement it; OpenSearch was the real offender (`_search_all_docs` swallowed mid-pagination exceptions returning partial results and skipped bad hits). Missing index stays a complete empty result; PIT deletion stays best-effort.
- OpenSearch KV `get_by_id`/`get_by_ids` now follow the scheduling-safe read contract: `None` = **confirmed absent**; non-missing-index transport errors raise (previously returned `None`, which the consistency validator read as "content missing" and deleted live doc_status rows). Converges OpenSearch onto the other four backends. (`filter_keys` has the same swallow pattern affecting enqueue dedup on transport errors — out of scope here, flagged as follow-up.)

### 2. Status-tuple split + quiescence decision protocol (pipeline)

- `_AUTO_RESUME_DOC_STATUSES` (PENDING + dead-process orphans) for every automatic run; `_MANUAL_RETRY_DOC_STATUSES` (+FAILED) used exactly once per sticky manual request — **processing carries no manual params, the mailbox is the single source of intent** (run-start peek, earliest-first, one per quiescence cycle).
- `_atomic_release_busy_or_consume_pending` → `_decide_pipeline_next_step`: one critical section observes manual-peek > auto-rescan-consume (sole consumer, re-armed on strict failure) > transitional `request_pending`, or releases `busy`. Storage queries stay strictly outside the lock.
- ACK timing: only after `_validate_and_fix_document_consistency` persisted the FAILED→PENDING resets (or immediately on a legitimately-empty complete scan) — a crash re-executes the request or leaves docs PENDING for automatic recovery; a doc failing again stays FAILED (no retry loop).
- Fixes the pre-existing defect where **any** upload-triggered run silently retried every FAILED document.

### 3. Sticky manual intent endpoints (API)

- New `start_committed_background_task` (dedicated helper; `start_reserved_background_task` untouched): two-state commit/ownership startup — pre-commit cancellation tears down with zero side effects; post-commit the child is never cancelled and owns its reservation + the published intent. Fence recheck and publish share ONE `pipeline_status_lock` critical section (`commit_manual_retry_request`), so a destructive clear cannot slip in between; plain busy does not refuse (the request waits for the running loop's quiescence peek). HTTP retry boundary (response lost post-commit → new id earns one more attempt) documented.
- `/documents/reprocess_failed`: pure storage-driven recovery — publish-first, no filesystem scan/classification/custom-chunk rollback, no `scanning_exclusive`; busy no longer loses the intent.
- `/documents/scan`: publishes its request only **after** the reservation is granted (refused scan = zero side effects, asserted in tests); `run_scanning_process` carries the id, with a classification-failure fallback drive (fires only when every drive branch was skipped and the request is still pending) and a cancellation contract that leaves the request sticky.

### Reviews & tests

- storage-backend-reviewer pass done; its blocker (Mongo/Redis except clauses missing `TypeError` → relaxed mode crashed instead of skipping) and both consistency gaps (terminal-id phantom commit, undocumented commit() invariant) fixed with per-backend parse-branch tests.
- New suites: OpenSearch strict reads (8) + json/Mongo/Redis/PG parse branches (4), FAILED-retry semantics (7: fix-proof upload-does-not-retry, consume+ACK+replay-guard, R1/R2 FIFO exactly-one-attempt-each, strict-failure stickiness, auto re-arm, orphan self-heal, empty-queue ACK), committed-startup protocol (10), scan manual-intent integration (3). Existing tests that relied on the implicit retry now publish an explicit request via `tests/pipeline/conftest.request_failed_retry`.
- **Full suite: 4131 passed, 36 skipped.** AGENTS.md concurrency contract updated (ingress channels, decision priority, FAILED semantics; `request_pending` demoted to transitional dual-write fallback).

Next: Phase 2 (in-batch feeder + dual write), Phase 3 (queue-based atomic exit + cancellation closure), Phase 4 (delete `request_pending` repo-wide).


danielaskdd:
## Review response (commits 9d7c6bf8, 775cb266)

Thanks for the review — all four findings addressed, plus the storage-backend-reviewer follow-up. Full suite **4144 passed, 36 skipped**.

### 1 (high) — OpenSearch item-level mget errors read as "absent" → **fixed** (9d7c6bf8)
New shared `_interpret_mget_item(item, doc_id, *, require_source)`: `found is True`→item, `found is False`→`None` (confirmed absent), item-level `error` / id mismatch / malformed → **raise**. Wired through the whole full_docs + doc_status mget surface uniformly — not just the two methods flagged:
- `_mget_optional_doc` (single-id, requires exactly one doc)
- `OpenSearchKVStorage.get_by_ids` / `filter_keys`
- `OpenSearchDocStatusStorage.get_by_ids` / `filter_keys`

`get_by_ids` uses length-check + positional zip + per-item `_id` verification (reorder/dup/omission → raise); `filter_keys` uses `require_source=False` (it queries `_source=False`). This closes both dedup-path gaps the reviewer identified (`full_docs.filter_keys` via custom-chunk create, `doc_status.filter_keys` via enqueue).

### 2 (medium) — `/scan` publishes sticky but may not drive → **fixed** (775cb266)
`pipeline_index_files` now returns whether it actually drove processing; the new-files branch marks `queue_drive_attempted` only on a real drive, so the classification-failure fallback still runs the sticky manual request when every new file is rejected.

### 3 (medium) — rollback discovery relaxed → **fixed** (775cb266)
`arollback_failed_custom_chunk_patches` discovery now passes `strict=True`; a mid-pagination/per-record failure aborts the rollback (caught scan-side, journal/FAILED kept for next scan) instead of a silent partial rollback.

### 4 (low) — JSON `status` read outside try → **fixed** (775cb266)
Moved `v["status"]` inside the try: missing-status record skipped in relaxed, raised under strict — symmetric with the other required fields.

Each fix has a fix-proof regression test.

### Deferred (tracked follow-up)
The **whole-call** `except OpenSearchException` swallow-to-empty in `filter_keys`/`get_by_ids`, and the same item-level pattern in the **graph** (`get_nodes_batch`/`has_nodes_batch`) and **vector** (`get_by_ids`/`get_vectors_by_ids`) subsystems + read-only graph traversal helpers. These need the broader swallow-vs-raise unification across those subsystems and are out of Phase 0's doc-ingest scheduling scope.

danielaskdd:
## Review round 3 — filter_keys fail-closed (commit 13b4b163)

Good catch — the whole-call fail-open in `filter_keys` does not belong in the deferred bucket: `OpenSearchDocStatusStorage.filter_keys` is the first-layer enqueue dedup (`pipeline.py:757`), squarely on the Phase 0 doc-ingest path. Fixed.

**Both `filter_keys` methods now fail closed on whole-call errors:**
- missing-index → return all keys (confirmed-empty index, unchanged);
- any other whole-call `OpenSearchException` → log and **raise** so the caller aborts before any upsert.

`OpenSearchKVStorage.filter_keys` (full_docs custom-chunk create dedup, `lightrag.py:1761`) fixed in the same commit as you suggested. `KV.get_by_ids` already raises on whole-call errors (round 1); `DocStatus.get_by_ids`'s whole-call swallow stays deferred — it has no internal callers.

**Tests added:**
- `doc_status` + KV `filter_keys` raise on a whole-call transient error;
- missing-index still returns all keys for both (regression guard for the intended fail-open);
- pipeline-level: `apipeline_enqueue_documents` aborts with **no** `full_docs`/`doc_status` upsert when the dedup `filter_keys` fails.

Full suite: **4148 passed, 36 skipped**.

PR #3461, opened 2026-07-22
fix(base): fall back when QueryParam env-backed ints are empty

santhreal:
Clearing `TOP_K=` (or the other QueryParam int knobs) crashed import because field defaults used bare `int(os.getenv)`.

Route those defaults through `get_env_value` like the LightRAG field defaults.


danielaskdd:
@codex review

PR #3462, opened 2026-07-22
fix(utils): treat empty JSON store files as missing

santhreal:
A zero-byte `kv_store_*.json` raised `JSONDecodeError` on startup while a missing file already returned `None` for `load_json(...) or {}`.

Treat empty/whitespace files like missing; corrupt JSON still raises.


danielaskdd:
@codex review

PR #3463, opened 2026-07-22
fix(parser): fall back when CHUNK_V_BUFFER_SIZE env is empty

santhreal:
Clearing `CHUNK_V_BUFFER_SIZE=` crashed `default_chunker_config` via bare `int(os.getenv)`.

Route through `get_env_value` like the other chunk env knobs.


danielaskdd:
@codex review

PR #3465, opened 2026-07-22
✨ feat(pipeline): Phase 2 — in-batch feeder (route mid-batch arrivals without the barrier)

danielaskdd:
Stacked on #3459 (Phase 0). Third phase of the `request_pending` → workspace ingress migration (order 1 → 0 → **2** → 3 → 4).

## Problem
`_run_pipeline_batch` routed a **fixed snapshot** of docs into the parse→analyze→process queues then `await`ed `q.join()` on each. A large doc blocked every doc enqueued afterward until the whole batch finished — the original defect this migration targets.

## Fix: in-batch feeder (pure accelerator)
An in-batch feeder coroutine routes documents that arrive **during** a running batch straight into its parse queues, so they process without waiting for the barrier. **`request_pending` (still dual-written by enqueue on a busy pipeline) remains the correctness anchor** — anything the feeder drops/skips/loses is a PENDING `doc_status` row recovered by the existing `request_pending → _decide_pipeline_next_step → next-batch` loop. The feeder never writes `doc_status`.

- **enqueue dual-write** — after `doc_status.upsert`, publish one document notification per new doc (after releasing `pipeline_status_lock`, so a large batch never holds it across N ingress RPCs); keep `request_pending`; arm `request_auto_rescan` on a partial commit (infallible anchor set *before* the fragile RPC).
- **`_admit_fed_document`** two-stage admission — register routing + bump counters under the lock BEFORE the bounded-queue `put`, move routing→inflight AFTER it with no `await` between (no dedup gap), roll both back on a cancelled/failed put (no ghost inflight, no double count).
- **`_pipeline_feeder`** waits ONLY on the document channel (single-process `get_document` / multiprocess `wait_for_documents`+`drain` via `to_thread`), so a pending manual/auto signal can't busy-loop it; three-way admission ROUTED / STALE-or-terminal / skip-and-recover.
- **`_run_pipeline_batch`** handshake — route initial batch → spawn feeder → join cascade → cancel feeder → second (stable) join cascade → finally cancels workers + feeder.

Deferred to **Phase 3** (needs the has_work-only exit): explicit in-batch reconciliation of skipped docs and a `CONTINUE_DOCUMENT` quiescence branch — in Phase 2 `request_pending` covers both.

## Self-review
Ran code-reviewer + silent-failure-hunter adversarially. **Both found the core correctness sound — no state-corrupting or stranding defect on any traced interleaving** (handshake, two-stage admission, dedup, dual-write ordering). Applied their observability hardening:
- feeder loop guarded — an unexpected per-iteration fault (wait RPC, engine routing, an admission read) logs + backs off + continues instead of silently killing the accelerator for the rest of the batch;
- the two safe read swallows log at debug so a persistent fault zeroing feeder throughput is diagnosable;
- `put_document` publish wrapped — a Manager-RPC failure no longer escalates a fully-committed enqueue into a caller-visible error;
- fixed a docstring that misstated the recovery contract (rollback does not re-publish; `request_pending` recovers).

## Tests
`tests/pipeline/test_pipeline_ingress_feed.py`:
- **fix-proof** — doc B enqueued while doc A blocks reaches PROCESSED without waiting for A (verified to time out with the feeder disabled, pass with it enabled);
- inflight dedup — a stale notification for an inflight doc is dropped, doc extracted once;
- document-only wait — a pending sticky manual request does not busy-loop the feeder.

Full suite: **4151 passed, 36 skipped**. Ruff format/check clean.

### DIFF
--- a/lightrag/pipeline.py
+++ b/lightrag/pipeline.py
@@ -1576,90 +1576,131 @@ class _PipelineMixin:
             ctx.routing_doc_ids.discard(doc_id)
             ctx.inflight_doc_ids.add(doc_id)
 
+    def _republish_documents(self, ingress, messages) -> None:
+        """Best-effort re-publish of drained-but-unresolved document messages.
+
+        The document channel drain is destructive (``drain_documents`` removes
+        the messages).  On a teardown — user cancel, internal error, or the
+        exception backoff — the drained messages this feeder had not yet routed
+        or confirmed stale/terminal must go back to the mailbox: a user cancel
+        clears ``request_pending``, so without the restore those PENDING docs
+        would wait for an unrelated future trigger instead of the next run.  A
+        multiprocess ``put_document`` is a Manager RPC; swallow failures — the
+        docs are PENDING and still recovered by the next initial scan.
+        """
+        for msg in messages:
+            try:
+                ingress.put_document(msg)
+            except Exception as republish_error:
+                logger.debug(
+                    "[feeder] document re-publish failed for "
+                    f"{getattr(msg, 'doc_id', None)}: {republish_error}"
+                )
+
+    async def _route_one(
+        self,
+        ctx: "_BatchRunContext",
+        msg: "PipelineIngressMessage",
+        pending: dict,
+    ) -> bool:
+        """Resolve one fed document message.
+
+        Returns True when the message is fully RESOLVED — admitted into a parse
+        queue, or a confirmed stale/terminal / duplicate / journaled doc that is
+        correctly dropped — and False when it was only SKIPPED for now (full_docs
+        not yet visible, or a transient read) and so should be re-published if
+        the batch tears down before request_pending can recover it.
+        """
+        doc_id = msg.doc_id
+        if not doc_id:
+            return True  # malformed notification: nothing to route
+        async with ctx.pipeline_status_lock:
+            duplicate = doc_id in ctx.routing_doc_ids or doc_id in ctx.inflight_doc_ids
+        if duplicate:
+            return True  # already routing/inflight — dedup
+        status_doc = pending.get(doc_id)
+        if status_doc is None:
+            return True  # not PENDING (processed/failed/deleted) — stale/terminal
+        if doc_status_custom_chunk_patch(status_doc) is not None:
+            return True  # journaled → scan/custom-chunk recovery owns it
+        try:
+            full_doc = await self.full_docs.get_by_id(doc_id)
+        except Exception as read_error:
+            logger.debug(
+                f"[feeder] full_docs read failed for {doc_id}, deferring "
+                f"(request_pending recovers): {read_error}"
+            )
+            return False  # transient → skip, re-publish on teardown
+        if not full_doc:
+            return False  # not yet visible (ryw lag) → skip, re-publish on teardown
+        await self._admit_fed_document(ctx, doc_id, status_doc)
+        return True  # admitted
+
     async def _pipeline_feeder(self, ctx: "_BatchRunContext", ingress) -> None:
         """Route documents that arrive DURING a batch into its parse queues so
         they process without waiting for the batch barrier.
 
         Pure accelerator (Phase 2): every message it discards or skips is a
         PENDING ``doc_status`` row that ``request_pending`` — dual-written by
-        enqueue on the same busy pipeline — recovers on the next batch.  So the
-        feeder never writes ``doc_status`` and never has to re-publish a skipped
-        message; correctness is anchored by the existing request_pending loop,
-        and the feeder only removes latency.
+        enqueue on the same busy pipeline — recovers on the next batch.  The
+        feeder never writes ``doc_status``; correctness is anchored by the
+        existing request_pending loop and the feeder only removes latency.
 
-        Per drained message, three outcomes:
-          * ROUTED — PENDING, not already routing/inflight, full_docs visible →
-            two-stage admission into the parse queue;
-          * STALE/TERMINAL — not PENDING (processed/failed/deleted), a
-            duplicate of a routing/inflight id, or a journaled custom-chunk doc
-            (owned by scan/custom-chunk recovery) → dropped;
-          * NEEDS_RECONCILIATION — full_docs not yet visible (read-your-writes
-            lag) or a transient read error → skipped this batch, recovered by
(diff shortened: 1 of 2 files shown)


danielaskdd:
## Review response (commit a17eefa0)

Both findings addressed. Full suite **4154 passed, 36 skipped**.

### High — feeder could keep a batch off its quiescence point (cancel + manual starvation)
Correct, and a regression the feeder introduced (batches went from bounded to unbounded). The feeder now **yields to control signals BEFORE draining more work** — it returns (stops admitting, letting the queues drain to the boundary) when:
- cancellation is requested (`ctx.pipeline_cancel_event`), so `/cancel_pipeline` completes even under a sustained stream; and
- a sticky manual retry request is waiting (`peek_next_manual_retry`), so a concurrent `/reprocess_failed` / `/scan` is consumed at the boundary instead of being starved.

It still does **not** wait on `has_work()` (that would busy-loop on a pending manual/auto signal — the document-only wait is the point).

### Medium — destructive drain + no re-publish on teardown
Also correct, and I agree it belongs in this phase per the reviewed plan. `_route_one` now reports RESOLVED (admitted / stale-terminal / dup / journaled) vs SKIPPED (full_docs not yet visible / transient read); on any teardown — user cancel, internal error, or the exception backoff — `_republish_documents` restores the unresolved tail + skipped messages to the mailbox so a cancel clearing `request_pending` can't leave those PENDING docs waiting for an unrelated trigger. On a normal quiescence stop the feeder is parked between drains, so nothing is re-published (`request_pending` covers the skips). Also routed the `doc_status` strict-read failure through the same re-publish path (previously it dropped the drain).

### Tests (all deterministic, no timing)
- `test_feeder_yields_on_cancellation` — with a full mailbox and the cancel event set, the feeder returns without draining (fix-proof: without the check it loops forever);
- `test_feeder_yields_when_manual_request_pending` — returns when a manual request is pending;
- `test_feeder_teardown_republishes_undrained_messages` — cancelled mid-route (full_docs.get_by_id gated), all drained-but-unresolved messages are back in the mailbox.

Note the earlier code-reviewer + silent-failure-hunter self-review pass had already confirmed no state-corrupting/stranding defect in the two-stage admission, inflight dedup, and double-join handshake; this round is purely the feeder exit-fairness + destructive-drain recovery you flagged.

danielaskdd:
## Review response, round 2 (commit 5121359b)

Both loops closed. Full suite **4156 passed, 36 skipped**.

### High — control signals only checked before each drain
Fixed. `_feeder_should_yield` (cancellation event OR a pending manual request) is now re-checked **before every single admission**, not just before the drain — so a cancel / `/reprocess_failed` landing while the feeder is admitting a full drain is honored **within one admission**, and the unprocessed tail is carried into the deferred set on the way out. The per-drain size is also capped (`_FEEDER_DRAIN_LIMIT = 256`) to bound the teardown re-publish burst. (`/cancel_pipeline` no longer has to wait out up to 4096 admissions.)

### Medium — SKIPPED only survived one iteration
Fixed. The deferred set is now **feeder-lifetime** (moved outside the `while`), restored to the mailbox on **any** exit via a `try/finally` — user cancel, a control-signal yield, or an unexpected escape. So a `full_docs`-not-visible SKIP from an earlier iteration is re-published on a later teardown instead of being forgotten and stranded when a cancel clears `request_pending`. The exception path re-publishes only that drain's tail for retry; the accumulated skips ride the `finally` rather than bouncing on every error.

Also removed the redundant second `full_docs.get_by_id` — `_route_one` now passes the row it already read for the visibility check into `_admit_fed_document`, so adding the per-admission manual peek is **net-neutral on RPCs**.

### Tests (deterministic)
- `test_feeder_teardown_republishes_undrained_messages` — now gates on an **entered event** (the blocking read signals it) instead of a fixed sleep, as you suggested;
- `test_feeder_republishes_earlier_iteration_skip_on_teardown` — B skipped (not visible) in one iteration, feeder parks on the next wait, then cancelled → B still re-published;
- `test_feeder_yields_within_one_admission_on_mid_drain_cancel` — cancel tripped right after the first admission's `put` → exactly **one** doc admitted, the whole tail re-published (fix-proof: without the per-admission check the feeder would admit all n).

danielaskdd:
## Review response, round 3 (commit e1b962a3)

Both addressed. Full suite **4157 passed, 36 skipped**.

### High — no bounded epoch under sustained load (overflow/skip starvation)
Fixed by tying the feeder's yield to the **auto-rescan recovery signal** (your second suggested option). `_feeder_should_yield` now also yields when `auto_rescan_pending` is dirty — **peeked** via `counts()`, never consumed (the supervisor stays its sole consumer) — checked at the top of each drain. And a **SKIP now arms `request_auto_rescan()`**. So:

- a document-channel **overflow** already arms `auto_rescan_pending` (its sole recovery signal) → feeder yields → batch boundary → `_decide_pipeline_next_step` consumes it → strict rescan recovers the dropped PENDING docs;
- a **SKIP** (full_docs not yet visible) arms the same flag → same path recovers the deferred doc.

So the batch reaches a boundary and the rescan runs even under a sustained stream — correctness no longer depends on the stream stopping. Cancel/manual stay checked **per-admission** (one-admission yield); auto-rescan is coarser (**per-drain**, ≤ `_FEEDER_DRAIN_LIMIT`), which is fine since it only needs to bound recovery, not be instantaneous. The per-admission `check_auto=False` means a skip does not abort the current drain — its visible siblings are still routed; the yield happens at the next drain top.

### Low — top-of-drain check could silently kill the feeder
Fixed. The top-of-drain `_feeder_should_yield` (which does a manual-peek / `counts()` RPC in multiprocess) is now **inside** the per-iteration `try`, so a transient RPC failure is logged, backed off, and retried like every other feeder read instead of ending the task under `gather(return_exceptions=True)`.

### Tests
- `test_feeder_yields_when_auto_rescan_pending` — with `auto_rescan_pending` set and non-PENDING notifications queued, the feeder yields without draining and leaves the flag for the supervisor (fix-proof: without the auto check it drains the stale notifications then blocks forever);
- `test_feeder_arms_auto_rescan_and_republishes_on_skip` — a skip arms auto-rescan, the feeder yields, re-publishes the deferred doc, and leaves the flag pending.

Removed the now-moot "skip → park → cancel" test: a skip now yields promptly via auto-rescan instead of parking.

PR #3466, opened 2026-07-23
feat(pipeline): has_work-only atomic exit + cancellation/recovery closure (ingress Phase 3)

danielaskdd:
Phase 3 of the pipeline-ingress migration. Stacked on #3465 (Phase 2, base of this PR); the stack is #3458 (Phase 1) → #3459 (Phase 0) → #3465 (Phase 2) → this.

## What this closes

Phase 2 left the ingress as a pure accelerator: anything the feeder missed was recovered by the transitional `request_pending` dual-write. Phase 3 makes the quiescence exit correct **on the mailbox alone**, so Phase 4 can delete `request_pending`:

1. **A doc published into a quiescing run is no longer maskable.** The atomic exit decision now covers the document channel: `_decide_pipeline_next_step` peeks `counts()["documents"]` (never drains under the lock) and returns `CONTINUE_DOCUMENT`.
2. **A stale notification can no longer hold `busy` forever.** The `CONTINUE_DOCUMENT` refetch resolves resident messages with the feeder's own drain→verify protocol at the supervisor level: **bounded destructive drain first, then the strict scan**. Proof: an enqueue persists the PENDING `doc_status` row *before* publishing, and the run holds `busy` (no other mutator), so any drained message whose doc is still live necessarily appears in the complete strict scan and comes back as the next batch — while a message absent from the scan is *provably* stale (terminal/FAILED-manual-only/deleted) and is compacted. Messages published after the drain stay in the mailbox for the next decision. On scan failure the drained messages are re-published and auto-rescan is armed; compensation failures log the drained ids and never displace the original error (`BaseException` symmetric on both sides).
3. **Cancellation = stop the whole run, ingress fully retained.** The cancellation check moved *inside* the decision's critical section (the cancel endpoint sets the flag under the same lock), checked **before** any signal consumption — a cancel can no longer race the decision into consuming an auto-rescan flag that then dies on the way out. `internal_error` aborts break to the finally (actionable halt message surfaced, per the existing `_finalize` contract); user cancels route to the loop-top handler. No new epoch either way; the backlog waits for the next explicit trigger.
4. **`ainsert_custom_chunks` exit hands off on mailbox-only work**: `request_pending OR ingress.has_work()`. Resolve/probe failures fail **toward handoff** — the driven run re-probes the mailbox itself; a transient flake self-heals, a persistent failure raises inside that run whose finally releases `busy`. Never a wedge, and never a silently deferred committed manual retry (manual requests never dual-write `request_pending`, so failing toward release would strand them until an unrelated trigger).
5. **`/documents/clear` clears the mailbox** while holding busy+destructive: un-ACKed manual requests retire as `CANCELLED_BY_CLEAR` (same-id replay refused), document/auto channels emptied. Failure degrades safely (residual messages are compacted by consumption idempotence; a surviving sticky request ACKs against the emptied doc_status).
6. **Workspace snapshot**: `run_workspace` captured at run entry; status/lock/ingress — including `_finalize` and the batch feeder — resolve through it. Coordination layer only; storages stay init-bound (documented boundary).

## Plan-item mapping (rev3 Phase 3)

- *Entry does not pre-drain*: already true since Phase 0/2 (peek protocol + strict initial scan; the feeder consumes backlog) — covered by tests here, no code change.
- *Quiescence reconciliation flag* (`new_work_since_last_reconcile`): structurally subsumed — every "published but missed" message either still sits in a channel at the decision (→ `CONTINUE_DOCUMENT`), or armed auto-rescan (feeder SKIP / overflow / partial upsert), or was fully resolved. The one remaining gap (crash *between* doc_status persist and publish) is documented on the decision as recoverable only by the next run's initial scan — the plan's declared boundary.
- *Supervisor destructive drain*: the ruling's failure mode (drain-then-release losing live messages) cannot occur here — release requires a subsequent decision with `has_work()` false and nothing in hand; live drained docs are returned by the strict scan; crash safety is identical to the feeder's drain (PENDING rows + next initial scan). Flagging explicitly for review since the plan's letter says "messages never leave the mailbox at exit": without a bounded compaction, a genuinely-stale message (enqueue-only doc deleted while idle) livelocks a `has_work`-only exit — this is the minimal protocol that resolves that contradiction with a proof, and it reuses the feeder's exact drain semantics.

## Tests (12 new; 4 fix-proofed by disabling the fix and watching them fail)

`tests/pipeline/test_pipeline_ingress_exit.py`:
- decision priority + per-step consumption semantics (manual peek consumes nothing; auto atomic; document peek doesn't drain; RELEASED clears busy/owner);
- **fix-proof** doc published at quiescence (dual-write mask killed) → processed as next batch;
- **fix-proof** stale message compacted, run releases (no livelock, no residency);
- refetch compensation (scan failure → messages republished + auto armed) and compensation-failure (original error still propagates);
- **fix-proof** user cancel preserves the armed auto-rescan flag (decision consumes nothing on the way out); internal_error same + halt message surfaced;
- journaled custom-chunk doc doesn't hold busy (notification compacted, journal intact);
- **fix-proof** custom-chunks handoff on ingress-only work; handoff when the has_work probe flakes.

`tests/api/routes/test_clear_documents_ingress.py`: all three channels cleared + CANCELLED_BY_CLEAR replay refusal; clear survives an ingress failure.

Adversarial self-review ran (code-reviewer + silent-failure-hunter agents): the drain-before-scan proof, cancel bookkeeping and clear degradation were verified sound; their findings (atomic in-decision cancel check, fail-toward-handoff, compensation observability, F401) are incorporated above.

Full suite: **4169 passed, 36 skipped**; pinned ruff (pre-commit) clean.

### DIFF
--- a/lightrag/api/routers/document_routes.py
+++ b/lightrag/api/routers/document_routes.py
@@ -3616,6 +3616,7 @@ def create_document_routes(
         from lightrag.kg.shared_storage import (
             get_namespace_data,
             get_namespace_lock,
+            get_pipeline_ingress,
         )
 
         # Get pipeline status and lock
@@ -3667,6 +3668,32 @@ def create_document_routes(
                 del pipeline_status["history_messages"][:]
                 pipeline_status["history_messages"].append(
                     "Starting document clearing process"
+                )
+
+            # We own busy+destructive: every document the mailbox refers to is
+            # about to be dropped, so clear the ingress alongside the
+            # ``request_pending`` reset above — un-ACKed manual retry requests
+            # are retired as CANCELLED_BY_CLEAR (a delayed replay of the same
+            # request id is refused), and the document/auto channels are
+            # emptied.  ``destructive_busy`` already refuses new enqueues, so
+            # nothing repopulates the mailbox during the drop.  Clearing
+            # BEFORE the drops means a later partial drop failure has already
+            # retired manual requests for still-live FAILED docs — acceptable
+            # under the destructive intent and surfaced via partial_success.
+            # A failure here only degrades: residual stale messages are
+            # compacted by the next run's consumption idempotence, and a
+            # surviving sticky request strict-scans the emptied doc_status and
+            # ACKs harmlessly.
+            try:
+                ingress = await get_pipeline_ingress(rag.workspace)
+                ingress.clear()
+            except Exception as ingress_clear_error:
+                logger.warning(
+                    "/documents/clear: failed to clear the pipeline ingress "
+                    "mailbox; safe to continue — residual document messages "
+                    "are compacted by the next run and a surviving manual "
+                    "retry request ACKs harmlessly against the emptied "
+                    f"doc_status: {ingress_clear_error}"
                 )
 
             # Use drop method to clear all data
--- a/lightrag/kg/pipeline_ingress.py
+++ b/lightrag/kg/pipeline_ingress.py
@@ -178,6 +178,8 @@ class PipelineIngress(Protocol):
 
     def put_document(self, msg: PipelineIngressMessage) -> None: ...
 
+    def put_documents(self, msgs: List[PipelineIngressMessage]) -> None: ...
+
     def request_auto_rescan(self) -> None: ...
 
     def consume_auto_rescan(self) -> bool: ...
@@ -241,12 +243,31 @@ class PipelineIngressMailbox:
         """
         _validate_document_message(msg)
         with self._cond:
-            if len(self._documents) >= self._document_capacity:
-                self._document_overflows += 1
-                self._auto_rescan_pending = True
-            else:
-                self._documents.append(msg)
+            self._put_document_locked(msg)
             self._cond.notify_all()
+
+    def put_documents(self, msgs: List[PipelineIngressMessage]) -> None:
+        """Publish a batch of document notifications in ONE server-side call.
+
+        Exists so a producer can publish inside the same
+        ``pipeline_status_lock`` critical section that the consumer's atomic
+        exit decision runs under, without holding that lock across N Manager
+        RPCs.  Per-message semantics are identical to :meth:`put_document`
+        (validation, overflow coalescing into the auto-rescan flag).
+        """
+        for msg in msgs:
+            _validate_document_message(msg)
+        with self._cond:
+            for msg in msgs:
+                self._put_document_locked(msg)
+            self._cond.notify_all()
+
+    def _put_document_locked(self, msg: PipelineIngressMessage) -> None:
+        if len(self._documents) >= self._document_capacity:
+            self._document_overflows += 1
+            self._auto_rescan_pending = True
+        else:
+            self._documents.append(msg)
 
     def request_auto_rescan(self) -> None:
         with self._cond:
@@ -438,6 +459,11 @@ class PipelineIngressHub:
     def put_document(self, namespace: str, msg: PipelineIngressMessage) -> None:
         self._mailbox(namespace).put_document(msg)
 
+    def put_documents(
+        self, namespace: str, msgs: List[PipelineIngressMessage]
+    ) -> None:
+        self._mailbox(namespace).put_documents(msgs)
+
     def request_auto_rescan(self, namespace: str) -> None:
         self._mailbox(namespace).request_auto_rescan()
 
@@ -494,6 +520,7 @@ class _PipelineIngressHubProxy(BaseProxy):
 
     _exposed_ = (
         "put_document",
+        "put_documents",
         "request_auto_rescan",
         "request_manual_retry",
         "consume_auto_rescan",
@@ -510,6 +537,11 @@ class _PipelineIngressHubProxy(BaseProxy):
 
     def put_document(self, namespace: str, msg: PipelineIngressMessage) -> None:
         self._callmethod("put_document", (namespace, msg))
+
+    def put_documents(
+        self, namespace: str, msgs: List[PipelineIngressMessage]
+    ) -> None:
+        self._callmethod("put_documents", (namespace, msgs))
 
     def request_auto_rescan(self, namespace: str) -> None:
         self._callmethod("request_auto_rescan", (namespace,))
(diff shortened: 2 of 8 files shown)


danielaskdd:
Both findings addressed in b26ee5be. Thank you — the F1 trace in particular is exactly the Phase 4 precondition this PR claimed and didn't yet deliver.

## F1 (High) — publish now inside the status-lock critical section, via a batch RPC

**Fix**: enqueue publishes its document notifications **inside the same `pipeline_status_lock` critical section that sets `request_pending`** ([pipeline.py](https://github.com/HKUDS/LightRAG/blob/feat/pipeline-ingress-phase3/lightrag/pipeline.py) `apipeline_enqueue_documents`). Serialized on the lock, a publish either lands before the consumer's exit decision (which then sees the messages via its `counts()` check) or strictly after the release — and in the latter case `busy` is already `False`, which is the designed idle-enqueue semantics (upload path drives its own `process()`; enqueue-only rests by design). The interleaving you traced (producer between flag-set and publish while the consumer releases) can no longer occur.

**Batch RPC, as suggested**: new `put_documents(msgs)` on every flavor — Protocol, `PipelineIngressMailbox` (one lock acquisition server-side), `PipelineIngressHub`, `_PipelineIngressHubProxy` (single `_callmethod`), the bound namespace view, and `AsyncioPipelineIngress` — with whole-batch validation before anything is enqueued and per-message overflow coalescing identical to `put_document`. The lock is held across exactly one Manager RPC, never N.

**On the publish-failure path**: the batch publish inside the lock stays swallow-and-log with `request_pending` (set first, infallible in-process) as the transitional anchor — the comment now explicitly marks this window as a Phase 4 item (the Phase 4 reservation-signature work moves the busy-refusal auto-arm into the same critical section, which is the flag's designed replacement).

**Test**: `test_enqueue_publishes_documents_inside_status_lock` — a class-level `NamespaceLock.__aenter__/__aexit__` trace records, per task, which namespace locks are held at the moment `put_documents` executes; the assertion requires `pipeline_status` to be among them. Fix-proofed: moving the publish back outside the lock fails it. This pins the invariant at the lock layer, which is flavor-independent — the serialization argument in multiprocess is the same lock, so I pinned the structural invariant rather than building a blocked-RPC multiprocess rig; happy to add one if you want the interleaving exercised end-to-end.

## F2 (Medium) — consumed-but-uncommitted signal restored on the cancel exit

**Fix**: the run now tracks ownership of a destructively-consumed wake-up (`uncommitted_wakeup`): set when the entry absorb or a `CONTINUE_AUTO`/`CONTINUE_DOCUMENT` decision consumed a signal AND the follow-up strict scan returned docs; cleared once `_validate_and_fix_document_consistency` takes those docs over (from there the batch, its feeder teardown and the PENDING rows own recovery), and also cleared on a complete empty scan (signal fully *served*, not lost). The loop-top cancel handler re-arms auto-rescan when it exits with an uncommitted consumption, so the cancel leaves the ingress as if the run had stopped before consuming.

For `CONTINUE_DOCUMENT` the restoration is the **auto flag rather than re-publishing the drained messages**: the refetch dropped them precisely because their live docs are fully represented by PENDING rows, and the auto flag is the canonical lost-notification recovery signal (identical to channel overflow) — the next run's entry consume + strict scan covers exactly the same set. This is within your option (b); carrying the drained list through the loop for a literal re-publish buys no additional recovery and widens the loop's state surface.

Note the entry-absorb window (initial `consume_auto_rescan` → non-empty initial scan → cancel before the first batch) had the same shape and is covered by the same ownership flag.

**Tests**: `test_cancel_after_auto_decision_restores_consumed_signal` and `test_cancel_after_document_decision_restores_consumed_signal` reproduce your sequence deterministically (a decision wrapper sets `cancellation_requested` the moment the first CONTINUE_AUTO / CONTINUE_DOCUMENT decision returns — i.e. after consumption, before the refetched docs reach a batch) and assert `auto_rescan_pending is True` after the run exits with the doc still PENDING. Both are fix-proofed: disabling the restore fails both.

Also added `test_put_documents_batch_matches_sequential_semantics` (both flavors: FIFO, whole-batch validation, overflow→auto coalescing).

Full suite: **4173 passed, 36 skipped**; pinned ruff clean.

danielaskdd:
Fixed in 1b9b0077 — the restore is now centralized exactly as suggested.

## Medium — uncommitted wake-up restored on ANY exit

**Fix**: the restore moved from the loop-top cancellation handler into the finally's cancellation-resistant bookkeeping (`_finalize`, executed via `run_to_completion`): whenever the run exits with `uncommitted_wakeup` still set — user cancel, the validation escape you traced, or any other non-cancel exception in that window — auto-rescan is re-armed **inside the same critical section that releases `busy`, before the release**, so the next acquirer observes the signal. The loop-top handler's own restore block is removed in favor of the single restore point (rather than keeping both and clearing after the handler's restore — one owner, no duplicate-arm coordination).

Details per your notes:
- `uncommitted_wakeup` and the `ingress` handle are **initialized before the `try`**, so `_finalize` can always read them — including when the run fails before the reservation or during the initial strict scan (the initial-scan failure path keeps its existing absorbed-flag compensation in the `except`; the ownership flag is still `False` there, so no double re-arm).
- The restore is a no-op on every healthy exit: `RELEASED` is only reachable with the flag already cleared (an empty refetch means *served*, the validator hand-off clears it before any batch), which the decision-priority test still pins.
- Restore failure (Manager outage at exit) logs a warning and degrades to next-explicit-trigger recovery — same class as the refetch compensation.

**Regression tests**: `test_validation_failure_after_document_decision_restores_signal` and `test_validation_failure_after_auto_decision_restores_signal` reproduce your injection (notification drained by the CONTINUE_DOCUMENT refetch / flag consumed by CONTINUE_AUTO, then `_validate_and_fix_document_consistency` raises a transient error) and assert the run escapes with the error, `busy=False`, the doc still PENDING, and `auto_rescan_pending=True`. Fix-proof: disabling the finally restore fails all **4** restore tests — these two plus both cancel-window tests from round 1, confirming the single restore point now carries every path.

Full suite: **4175 passed, 36 skipped**; pinned ruff clean.

danielaskdd:
Fixed in 7a5413d5, exactly as suggested.

## Medium — ownership starts at the instant of consumption

`uncommitted_wakeup = absorbed_auto_rescan` is now assigned immediately after `consume_auto_rescan()` (no await in between), and the entry's local `except Exception` re-arm is **removed entirely** — the finally's single cancellation-resistant restore point now covers the initial scan the same way it covers every later window: `Exception`, `CancelledError`, and any other `BaseException` all exit through `_finalize`, which re-arms the flag before releasing `busy`.

After a successful scan the ownership resolves as before: a complete empty result releases it (signal *served*), a non-empty result keeps it owned until the consistency validator takes the docs over. A failure of `consume_auto_rescan()` itself still propagates with the flag un-owned — nothing was consumed client-side, and the response-loss window remains the documented at-most-once boundary backed by PENDING rows.

**Regression test**: `test_initial_scan_cancellation_restores_absorbed_auto_signal` — the initial `get_docs_by_statuses` raises `asyncio.CancelledError`; asserts the error propagates, `busy=False`, and `auto_rescan_pending=True`. Fix-proofed: reverting to post-scan ownership fails it.

Verified on the affected suites (ingress-exit / release-closure / cancellation: 98 passed) with pinned ruff clean.

PR #3467, opened 2026-07-23
[Phase 4] Delete request_pending: the workspace ingress mailbox becomes the pipeline's only wake-up channel

danielaskdd:
Final phase of the staged `request_pending` → workspace-ingress migration (stack: #3458 P1 → #3459 P0 → #3465 P2 → #3466 P3 → this PR). The transitional dual-write flag is deleted repo-wide; wake-ups flow exclusively through the three-channel ingress mailbox, with `doc_status` PENDING rows + the next run's initial strict scan as the documented correctness backstop.

## Core changes

**Reservation (`shared_storage.py`)** — `acquire_processing_reservation` gains a required `pipeline_ingress` kwarg. A busy refusal now arms the auto-rescan flag **inside the same critical section** as the refusal decision (arming after returning BUSY would let the current holder release between the two and never see the signal — the exact lost-wakeup the plan forbade). Two deliberate semantics, both test-pinned:
- callers resolve the ingress **before** the reservation; a Manager RPC handle is never resolved lazily while `pipeline_status_lock` is held;
- an arm failure **propagates** instead of returning BUSY — a swallowed failure would tell the caller "the holder will pick your docs up" while no committed signal exists. Recovery updates committed before the arm are idempotent (recomputed by the next acquire), so no torn state.

**Enqueue (`pipeline.py`)** — documents are published unconditionally (busy or idle) via the in-lock `put_documents` batch; an idle message stays resident for the next run (the designed idle-enqueue case). Publish failure → auto-rescan fallback in the same critical section; double failure → ERROR log naming the consequence. The partial-commit path arms auto-rescan only (the idle inline recovery pass is preserved even when the arm raises).

**Quiescence decision** — the transitional consume step is gone: cancellation (consumes nothing) > manual (peek) > auto (atomic consume) > document (peek via `counts()`) > release, all under one critical section.

**Batch delete (`document_routes.py`)** — the post-delete "pending drive" check moves from the flag to an `ingress.has_work()` probe executed **inside the owner-checked release critical section** (same serialization argument as the enqueue publish: a busy-refused request either arms before the release decision and is seen, or lands after `busy=False` and drives itself). Resolve/probe failures fail **toward** driving — a spurious drive is a busy-gated no-op; skipping would silently defer accepted work.

**Custom-chunks exit (`lightrag.py`)** — the handoff decision keys purely on `has_work()`; the fail-toward-handoff arms are unchanged and now directly tested (see below).

**Contract surface** — `pipeline_status` default field, every reservation `flags` dict, `PipelineStatusResponse` (API), WebUI type + pending indicator + `requestPending` key in 11 locales, EN/CN `FileProcessingPipeline` docs, `AGENTS.md` contract table.

## Hardening from the adversarial self-review

- **`_finalize` bookkeeping gate**: an acquire that raises (now including the busy-arm RPC on a Manager outage) never owned the slot — the finally no longer logs a spurious "pipeline stopped" during a live holder's run, nor stamps a stop record into an idle `pipeline_status`. The owner-checked release semantics are byte-for-byte preserved (`holds_busy` pre-set stays: a post-stamp acquire failure must still release).
- **Log severity**: the three terminal "docs stay PENDING until the next trigger" degradations (publish+rescan double failure, exit-restore failure, partial-commit wake failure) now log at ERROR with the consequence named — they are the last trace an operator gets.
- **`get_pipeline_ingress` suspension-free contract** documented at the definition and both slot-releasing `finally` call sites (a real `await` there would let a re-cancellation skip the cancellation-resistant release).

## Tests (all new/changed regressions fix-proofed by disable→fail→restore→pass)

- in-lock busy-refusal arm + its failure propagation (`test_reservation_primitives`)
- publish-failure → auto-rescan fallback; acquire-failure bookkeeping gate (`test_pipeline_release_closure`)
- delete-exit drive: mailbox work → exactly one drive; empty → none; resolve-failure and probe-failure arms → release still runs + one drive (`test_document_routes_docx_archive`)
- custom-chunks resolve-failure → fail-toward-handoff (`test_document_file_path_normalization`)
- Found & fixed a Phase 3 test blind spot: the custom-chunks handoff tests were passing *accidentally* through the uninitialized-storage fail-toward-handoff arm (one stub's `TypeError` was even being swallowed). They now drive a controllable fake ingress; the accidental coverage is replaced by the explicit resolve-failure tests above.

## Acceptance

- `rg -n "request_pending" lightrag lightrag_webui docs tests AGENTS.md` → empty (old WebUI bundle regenerated by `bun run build`)
- Full suite: **4182 passed, 36 skipped**; pinned ruff via pre-commit clean
- Reviewer notes (accepted, documented): the flag's old role as an infallible in-process anchor is gone by design — in the double-failure window (publish + rescan both fail on a Manager outage) docs wait for the next explicit trigger, which is the plan's documented boundary; the feeder's skip path may double-publish one message on an arm failure (benign: at-least-once + downstream dedup/compaction).


danielaskdd:
## Review round 1 addressed — c4a08512

**High — direct `adelete_by_doc_id()` strands mailbox work: fixed with the atomic handoff (preferred option).**

The exclusive-`destructive_busy` alternative was rejected on contract grounds: a direct SDK delete has always admitted concurrent enqueue, and tightening that to refuse `ainsert` would be a behavior break unrelated to the wake-up migration.

What changed (mirrors the custom-chunks exit exactly):

- The delete's owner-checked release action now probes `ingress.has_work()` **inside its critical section**. Work resident → keep `busy` and return `"handoff"`; the finally then drives `apipeline_process_enqueue_documents(_holding_busy=True, token=token)`, which takes over and releases the **same** reservation at its own exit — `busy` never drops to False with mailbox-only work pending, so no clear/delete/scan reservation can slip into the gap either.
- Failure closure follows the established arms: ingress resolve failure or `has_work()` probe failure fails **toward** handoff (the driven run re-probes the mailbox itself; a transient flake self-heals, a persistent failure raises inside that run whose finally releases `busy` — never a wedge, while releasing here would silently defer the accepted work). An owner-checked `_terminal_release` backstop covers a cancelled/failed handoff drive, no-op when the handoff run already released.
- Scope guards: the ingress is resolved only when **we** acquired the slot — an `adelete` joined into a batch delete leaves the handoff to the batch exit's own `has_work` probe (no double drive) — and the resolve happens before the cancellation-resistant release under `get_pipeline_ingress`'s documented suspension-free contract.

**Regression test** (`test_direct_delete_hands_off_mailbox_work`) is the deterministic reproduction you outlined: the delete is event-gated right after taking `busy` (on its first `doc_status` read), a full public `ainsert()` runs to completion while the slot is held (asserting `has_work()` is True — the PENDING row, document message and busy-refusal auto-rescan all landed), then the gate opens. Both public calls return success, and the test asserts the new doc reaches PROCESSED via the handoff run (extraction called exactly once), `has_work()` is False, and `busy` is False. Fix-proofed: with the handoff branch disabled the doc strands in PENDING and the test fails.

Affected subsets (exit / normalization / release-closure / cancellation / internal-abort / docx-archive / reservation-primitives): **211 passed**; pinned ruff clean.

danielaskdd:
## Review round 2 addressed — 3eab9dfa

**Medium — acquire-window escapes released `busy` unconditionally: fixed with your preferred option** (single handoff-aware `try/finally`; the separate `_release_on_acquire_failure` bridge is deleted outright).

- `acquire_reservation` and the history/status-logging window now live **inside** the same `try/finally` as the deletion itself, so every escape after the slot is stamped — cancellation or RPC failure at the acquire/logging lock exit included — reaches the identical owner-checked exit decision: probe `has_work()` → handoff or release. The locals the finally reads are initialized before the `try`; non-acquired paths (refusal returns, batch-delete join) skip the exit via `we_acquired_pipeline` exactly as before.
- One deliberate behavior consequence, called out for the record: a window `Exception` now surfaces through the delete's uniform error contract (`DeletionResult(status="fail", 500)` + ERROR log) instead of raising raw — consistent with every other failure in the method. With `doc_status_data` still `None` at that point, the retry-state writer in the `except` correctly stays untouched; `CancelledError` is not an `Exception` and still propagates.

**Regression tests** (deterministic via an event-gated `acquire_reservation` wrapper that stamps `busy` for real, then parks):

- `test_direct_delete_acquire_window_failure_hands_off` — your reproduction with a non-cancel window failure: concurrent full `ainsert()` lands its work (`has_work()` True asserted), the gate releases into a `RuntimeError`, the delete returns the failure, and the handoff run processes the new doc — PROCESSED, mailbox empty, `busy=False`. Fix-proofed: forcing the old unconditional release strands the doc and the test fails.
- `test_direct_delete_acquire_window_cancel_releases_and_retains` — the cancel variant. Writing it surfaced a real asyncio subtlety worth stating: whether the handoff drive runs under cancellation depends on delivery timing — a `CancelledError` already injected (task parked on an await) lets the finally's awaits proceed, so the drive may legitimately run to completion; one still pending re-raises at the drive and the backstop releases. The test therefore asserts the invariant common to both outcomes: `CancelledError` propagates, `busy=False` (never wedged), and the work is never lost — the doc reaches PROCESSED after at most one explicit trigger, extraction exactly once. (The same timing duality has always applied to the custom-chunks handoff this mirrors.)

Affected subsets (exit / normalization / release-closure / cancellation / internal-abort / docx-archive / chunk-preservation / custom-chunk-patch): **234 passed**; pinned ruff clean.

PR #3468, opened 2026-07-23
fix(api): reject query text that strips below min_length

santhreal:
`QueryRequest` enforced `min_length=3` before strip, so values like `"  a  "` or whitespace-padded shorts passed validation and then shrank below the floor.

Validate length after strip.


danielaskdd:
@codex review

PR #3469, opened 2026-07-23
fix(api): reject whitespace-only insert text bodies

santhreal:
Insert text requests accepted whitespace-only bodies that became empty after strip, so empty documents were enqueued.

Reject whitespace-only text after strip.


danielaskdd:
@codex review

Issue #2686, opened 2026-02-08
[Question]:LightRAG embeddings very slow during document ingestion (TEI is fast standalone)


akash171198:
### Do you need to ask a question?

- [x] I have searched the existing question and discussions and this question is not already answered.
- [x] I believe this is a legitimate question, not just a bug or feature request.

### Your Question

I am facing very slow embedding performance during document ingestion in LightRAG, even though my embedding service itself is fast.

I am using TEI (Text Embeddings Inference) as the embedding backend. When I call the TEI embedding endpoint directly (via Postman/curl), embeddings are returned in milliseconds. However, when embeddings are triggered through LightRAG during document ingestion (especially when uploading multiple JSON files), the process becomes extremely slow and can take hours.

I am using PostgreSQL for vector storage (PGVector) and Neo4j for graph storage.

What are the LightRAG-side best practices or tuning recommendations to improve embedding and ingestion performance in this setup? Are there known bottlenecks or configurations that significantly impact ingestion speed?

### Additional Context

Relevant LightRAG configuration:

Concurrency and ingestion:
- MAX_ASYNC=16
- MAX_PARALLEL_INSERT=4
- EMBEDDING_FUNC_MAX_ASYNC=32
- EMBEDDING_BATCH_NUM=64

Vector and graph storage:
- LIGHTRAG_VECTOR_STORAGE=PGVectorStorage
- LIGHTRAG_GRAPH_STORAGE=Neo4JStorage

PostgreSQL vector index:
- POSTGRES_VECTOR_INDEX_TYPE=HNSW
- POSTGRES_HNSW_M=16
- POSTGRES_HNSW_EF=64

LLM / embedding context:
- OLLAMA_LLM_NUM_CTX=32768
- OLLAMA_EMBEDDING_NUM_CTX=8192

Observations:
- TEI (Text Embeddings Inference) responds quickly when called directly.
- Slowdown only occurs during LightRAG ingestion (chunking, embedding, DB insert, graph merge).
- Issue becomes more visible when uploading multiple or large JSON files.

Looking for guidance on:
- Optimal concurrency and batching values
- Whether HNSW indexing should be adjusted or deferred during bulk ingestion
- Any recommended ingestion patterns (batch size, parallelism limits)
- General production best practices for LightRAG with PostgreSQL + Neo4j

danielaskdd:
MAX_ASYNC=16
EMBEDDING_FUNC_MAX_ASYNC=32

[...]

Issue #1363, opened 2025-04-12
[Question]: Vector database upserts highly redundant / inefficient?


alineberry:
### Do you need to ask a question?

- [x] I have searched the existing question and discussions and this question is not already answered.
- [x] I believe this is a legitimate question, not just a bug or feature request.

### Your Question

I am working to apply LightRAG to a large scale database and it occurs to me that the entity and relationship vector database incremental upserts may be highly redundant and inefficient. It does not appear that LightRAG does any dense retrieval from these tables during ingestion, and that there will be many, many unnecessary upserts. Would it make sense to add a new task post-graph construction that does a bulk insertion of all entities and relationships into vector tables to solve this?

danielaskdd:
In version 1.3.2 optimizes the document indexing process by transitioning node/edge merging  and vector saving job from chunk-based to document-based. Additionally, it consolidates merge operation of multiple entities/edges with identical names into a single LLM call for improved efficiency.

alineberry:
Thank you for your response. I would like to see the code changes in 1.3.2 that accomplish this.
Also... is there a reason why the vector saving job needs to be even document-based rather than performed as a totally separate job post-graph construction? It seems even at the document level, there will be redundant upserts given that the vector databases are not used during graph construction

alineberry:
The only reason I can think of is if there are future plans to utilize the vector databases to do more robust merging during the insertion pipeline

danielaskdd:
The construction of a RAG knowledge base is an incremental process. In a production environment, it is necessary to ensure continued query services during the incremental building of the knowledge base. Building it step by step, document by document, is a reasonable approach.

alineberry:
I agree but we may use this approach which is more optimized for a large initial backfill which may take weeks to months to complete and then switch over for incremental updates.

danielaskdd:
During the document construction phase, I don't see any significant performance difference between gradually inserting text chunks, entities, and relations versus waiting until thousands of files are completed before building.  

For entity merging, delayed merging does offer some performance benefits. However, due to the context limitations of LLMs, it's practically impossible for an LLM to handle merging too many nodes at once. Therefore, we can't accumulate too many files before having the LLM merge nodes and relations either.

alineberry:
[...]

PR #3473, opened 2026-07-23
feat: support graph-first ingestion with deferred vector indexing

### LINKED ISSUE #3472: [Feature Request]: Support graph-only ingestion with deferred vector indexing
### Existing check

- [x] I searched existing LightRAG issues and pull requests for graph-only ingestion, skipped embeddings, deferred vector indexing, and no-op vector storage.
- [x] This is a feature request rather than a usage question or bug report.

LightRAG ingestion currently updates entity and relationship vector stores while merging graph records. Those VDB upserts invoke embedding providers and remote storage I/O, so graph extraction throughput can be constrained by embedding latency even when the immediate goal is only to build the knowledge graph.

Please add an officially supported graph-only ingestion mode that accepts all vector storage writes without calling the embedding function. Users can then run `lightrag-rebuild-vdb` after graph extraction to build entity, relationship, and chunk vector indexes in one dedicated phase.

A minimal implementation could provide a built-in `NoopVectorDBStorage` backend:

- `upsert`, delete, lifecycle, and inspection methods remain API-compatible no-ops.
- `query` raises a clear error explaining that vector indexing is disabled and must be rebuilt before vector-backed retrieval.
- The backend is registered in the normal vector storage registry, so users select it through the existing `vector_storage` configuration instead of patching the ingestion pipeline.
- Existing storage backends and default behavior remain unchanged.

### Motivation

This separates two resource profiles:

1. graph extraction and entity/relation merge, dominated by LLM calls;
2. embedding and vector index construction, which can use a separately tuned concurrency and batch size.

It also shortens the time spent inside graph merge operations because entity and relationship VDB writes no longer wait for embedding or remote vector storage I/O.

### Alternatives considered

- A boolean branch around every entity, relationship, and chunk VDB upsert. This spreads conditional logic across the ingestion pipeline.
- A project-local storage implementation registered by mutating `STORAGES` and `STORAGE_IMPLEMENTATIONS`. This works, but relies on LightRAG internals and must be duplicated by each user.
- Returning zero vectors from a dummy embedding function while using a real vector backend. This still writes invalid vectors and requires cleanup before rebuilding.

### Validation plan

- Unit-test the full `BaseVectorStorage` contract.
- Confirm `upsert` does not call the embedding function.
- Confirm graph-only insertion still persists graph and KV evidence.
- Confirm vector queries fail with an actionable message.
- Benchmark a controlled entity/relation ingestion workload before and after, using the same simulated embedding latency.
- Run relevant offline tests and Ruff/pre-commit checks.

letsgetai:
## Description

Add an opt-in graph-first ingestion workflow for initial or offline corpus
backfills. `NoopVectorDBStorage` lets the normal LightRAG ingestion pipeline
persist the authoritative graph, KV stores, text chunks, and document status
without embedding or writing intermediate vector revisions. After the backfill,
users switch to a persistent vector backend and rebuild the final entity,
relationship, and chunk indexes from the stored graph/KV state.

This separates expensive extraction from rebuildable vector materialization.
It is intended for backfills where query availability is not required until the
final rebuild completes; normal incremental ingestion should continue using a
persistent vector backend from the beginning.

## Related Issues

- Closes #3472.
- Related: #1363, #2686, #2785.

## Changes Made

- Add and register `NoopVectorDBStorage` without external service requirements.
- Accept vector mutations without calling the embedding function or persisting
  vectors, while leaving graph/KV/document-status ingestion unchanged.
- Raise an actionable storage-capability error when a vector-backed query
  reaches Noop; `bypass` remains available because it does not use VDB storage.
- Make the embedding requirement explicit through the base/backend capability
  contract, allowing Noop to initialize with `embedding_func=None`.
- Mark Noop as non-persistent and reject it as a rebuild target before reading
  source records or dropping the target, preventing false-success rebuilds.
- Document the complete handoff: stop writers, retain the same working
  directory/workspace/graph/KV configuration, switch to a persistent VDB,
  configure the production embedding model and dimension, rebuild all indexes,
  and only then restart the server.
- Add normal `ainsert` coverage for the Noop-to-Nano handoff, plus bypass,
  capability-contract, backend-registration, query, and rebuild-guard tests.

## Benchmark

The benchmark isolates graph/vector materialization through the public storage
path; LLM extraction is excluded. It therefore supports claims about embedding
and vector-write amplification, not general extraction speed.

| Workload | Persistent vectors | Graph-first + rebuild | Result |
|---|---:|---:|---:|
| 100 documents, elapsed time | 42.646 s | 34.380 s | -19.4% |
| 100 documents, embedded texts | 60,100 | 20,500 | -65.9% |
| 200 high-overlap documents, elapsed time | 169.020 s | 131.930 s | -21.9% |
| 200 high-overlap documents, embedded texts | 200,200 | 41,000 | -79.5% |
| 200 high-overlap documents, embedding calls | 3,400 | 650 | -80.9% |
| 200 high-overlap documents, vector write amplification | 4.88x | 1.00x | -79.5% |

Both modes produced identical final graph and vector record counts. The
graph-first 200-document run used more peak memory (471 MiB vs 303 MiB) because
the current rebuild utilities prepare graph records and payloads before batched
writes; the documentation calls out this scalability boundary.

## Test Plan

- `47 passed`: Noop backend, normal ingestion handoff, rebuild APIs, and bypass
  regression tests.
- `3972 passed, 42 skipped, 1063 deselected`: full Python 3.12 offline suite
  with the API extra on `upstream/main@73046a7e`.
- `uvx pre-commit run --all-files`: all hooks passed.
- `git diff --check`: passed.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated
- [x] Unit tests added

## Additional Notes

- No changes are made to `lightrag/lightrag.py`; query failure behavior remains
  backend-local.
- Noop is not intended for normal online incremental ingestion or vector-backed
  queries.
- This PR does not add external database integrations or migrate existing local
  graph/KV data to external stores.

### DIFF
--- a/docs/ProgramingWithCore.md
+++ b/docs/ProgramingWithCore.md
@@ -619,6 +619,7 @@ OpenSearchGraphStorage   OpenSearch
 
 **VECTOR_STORAGE**
 ```
+NoopVectorDBStorage         Disabled (graph-only ingestion)
 NanoVectorDBStorage         NanoVector (default)
 PGVectorStorage             Postgres
 MilvusVectorDBStorage       Milvus
@@ -627,6 +628,79 @@ QdrantVectorDBStorage       Qdrant
 MongoVectorDBStorage        MongoDB
 OpenSearchVectorDBStorage   OpenSearch
 ```
+
+#### Graph-only ingestion
+
+`NoopVectorDBStorage` is intended for an initial or offline corpus backfill
+where the graph and KV stores are authoritative and vector indexes can be
+materialized once from the final state. It avoids embedding and persisting
+intermediate entity, relationship, and chunk revisions during ingestion.
+
+Do not use this workflow when newly inserted documents must become queryable
+immediately. Normal incremental ingestion should use the intended persistent
+vector backend from the beginning.
+
+Configure the backfill process with the no-op backend. `embedding_func=None` is
+supported when no other configured component requires embeddings:
+
+```python
+rag = LightRAG(
+    working_dir=WORKING_DIR,
+    llm_model_func=llm_model_func,
+    embedding_func=None,
+    vector_storage="NoopVectorDBStorage",
+)
+```
+
+The backend accepts vector mutations without calling the embedding function or
+persisting vectors. Graph, full-document, text-chunk, LLM-cache, document-status,
+and graph-recovery writes continue normally.
+
+While `NoopVectorDBStorage` is active, only `bypass` queries are available.
+`local`, `global`, `hybrid`, `mix`, and `naive` modes require vector indexes and
+raise an error that points to `lightrag-rebuild-vdb`.
+
+If the semantic-vector (`V`) chunker is selected while `embedding_func=None`,
+it logs a warning and falls back to recursive-character chunking. Configure an
+embedding function during ingestion if semantic-vector chunk boundaries are
+required; this is separate from whether vectors are persisted.
+
+##### Switching to persistent vectors
+
+After the backfill, stop the server and all ingestion writers. Keep
+`WORKING_DIR`, `WORKSPACE`, graph storage, KV storage, and their connection
+settings unchanged. For example, switch from Noop to NanoVector with an OpenAI
+embedding model while pointing to the same graph and KV sources:
+
+```bash
+export WORKING_DIR=/data/lightrag/rag_storage
+export WORKSPACE=project_a
+export LIGHTRAG_GRAPH_STORAGE=NetworkXStorage
+export LIGHTRAG_KV_STORAGE=JsonKVStorage
+export LIGHTRAG_VECTOR_STORAGE=NanoVectorDBStorage
+export EMBEDDING_BINDING=openai
+export EMBEDDING_MODEL=text-embedding-3-small
+export EMBEDDING_DIM=1536
+
+lightrag-rebuild-vdb  # Select "Rebuild ALL vector storages"
+```
+
+Replace all example values with the backfill's actual storage settings and the
+production embedding model, dimension, host, and credentials. Backend-specific
+connection variables must remain available. If the same `.env` already contains
+these values, only the vector and embedding entries need to change. Run this in
+a new process or after a restart, then keep the persistent configuration for
+later queries and incremental ingestion.
+
+Rebuild cost and memory grow with the final graph and chunk data. If rebuilding
+fails or is interrupted, keep writers stopped and rerun it with the same
+configuration; the graph and KV sources remain unchanged. Start the server only
+after the tool reports a successful rebuild, for example with
+`lightrag-server`, because LightRAG has no persisted vector-index readiness
+marker.
+
+See `lightrag/tools/README_REBUILD_VDB.md` for rebuild options and operational
+details.
 
 **DOC_STATUS_STORAGE**
 ```
--- a/lightrag/base.py
+++ b/lightrag/base.py
@@ -225,20 +225,23 @@ class StorageNameSpace(ABC):
 
 @dataclass
 class BaseVectorStorage(StorageNameSpace, ABC):
-    embedding_func: EmbeddingFunc
+    requires_embedding_func: ClassVar[bool] = True
+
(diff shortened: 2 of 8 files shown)


danielaskdd:
Thanks for the contribution. The graph-first ingestion idea is valuable, and
using a registered no-op vector backend is a clean way to avoid spreading
conditional branches throughout the ingestion pipeline.

I am requesting changes before merge for the following reasons:

1. **CI is currently failing.**

   The Python 3.14 offline suite has 3 failures in
   `tests/llm/test_bypass_truncated_response.py`. `aquery_llm()` now calls
   `self._ensure_vector_query_supported()` before entering the bypass branch,
   but the existing minimal `_FakeRAG` intentionally exposes only the state
   used by bypass mode.

   Since bypass mode does not use vector storage, please fix the implementation
   so it returns to the bypass path without requiring the vector capability
   helper. Please also add an explicit Noop + bypass regression test rather
   than only modifying the test fixture.

2. **A complete graph-only ingestion guide is needed.**

   The current short section does not provide enough information to perform
   the transition safely. Please document:

   - the initial/offline-backfill use case and when not to use this backend;
   - that only bypass queries are available while Noop is active;
   - how to stop all writers and switch from Noop to a persistent VDB;
   - that WORKING_DIR, WORKSPACE, graph storage and KV storage must remain
     unchanged;
   - how to configure the embedding provider/model/dimension and run
     `lightrag-rebuild-vdb` with “Rebuild ALL”;
   - that the persistent VDB configuration must remain active for normal
     querying and future incremental ingestion;
   - that changing `vector_storage` on an existing LightRAG instance is not
     sufficient: a new instance or process restart is required;
   - that the server must not be started until rebuild succeeds, because the
     current implementation has no persisted index-readiness marker;
   - the semantic-vector chunker fallback when `embedding_func=None`;
   - rebuild cost, memory limitations, interruption and retry behavior.

3. **The main ingestion path needs regression coverage.**

   The current end-to-end test uses `ainsert_custom_kg`, while the primary user
   workflow is normal document ingestion through `ainsert`/the pipeline. Please
   add coverage proving that normal ingestion preserves graph, KV and document
   status data without invoking vector embeddings, followed by a successful
   handoff to a persistent VDB.

4. **Please make the embedding contract explicit.**

   `BaseVectorStorage` still types and documents `embedding_func` as mandatory
   for every vector backend, while Noop officially accepts `None`. This should
   be represented explicitly in the base/backend capability contract instead
   of relying on Noop skipping `_validate_embedding_func()`.

Non-blocking suggestions:

- Use `getattr(storage, "supports_vector_queries", True)` in the query guard for
  compatibility with existing custom backends, matching the rebuild guard.
- Consider separating “supports queries” from “persists vectors”.
- Disable or clearly label consistency checks against Noop, since every vector
  will intentionally appear missing.

The core design looks sound, especially the no-op storage abstraction and the
guard-before-read/drop behavior in the rebuild APIs. Once CI, documentation,
the normal-ingestion coverage, and the base contract are addressed, this
should be in good shape for another review.

letsgetai:
Thanks for the detailed review. I rebased the branch on the latest `main` and
force-pushed a clean three-commit history (implementation, tests, and docs).

The requested blockers are addressed as follows:

1. `bypass` remains independent of vector storage, with an explicit Noop +
   bypass regression test.
2. The graph-first guide now covers the intended offline-backfill scope, query
   limitations, unchanged source-storage configuration, process restart,
   persistent VDB/embedding setup, rebuild-all handoff, server-start boundary,
   semantic chunker fallback, cost/memory limits, and retry behavior.
3. The normal `ainsert` pipeline is covered through a Noop-to-Nano handoff that
   verifies graph, KV, document-status, and rebuilt vector data without invoking
   embeddings during ingestion.
4. `BaseVectorStorage` now exposes an explicit backend capability contract for
   whether `embedding_func` is required.

I also kept query failure handling backend-local and retained the rebuild guard
before source reads or target drops. The current validation results are listed
in the updated PR description. Please take another look when convenient.

danielaskdd:
@codex  review

PR #3478, opened 2026-07-24
feat(storage): cursor-page scheduling API + failure-generation write side (memory-bounding Phase 1)

danielaskdd:
## Summary

Phase 1 of the pipeline memory-bounding plan: the storage layer that every later phase builds on. When `doc_status` holds a very large number of entries, today's scheduler materializes ALL matching rows per sweep; this PR adds the bounded primitives without changing scheduler behavior yet (Phase 2+ wires them in).

### base layer
- **Three-state cursor keyset paging**: `get_docs_by_statuses_page` with `(created_at ASC, id ASC)` MUST-order, `CURSOR_START / CursorAfter(opaque) / CURSOR_END`, a lightweight `DocSchedulingRecord` projection (no `chunks_list`/large metadata), and the **consumed-position advance contract** (filtered/skipped rows advance the cursor; prefetched-but-unreturned heads never skip; a fully filtered page never terminates or re-reads).
- **`failure_generation` write side**: `reserve_failure_generation` (reserve-before-publish; holes allowed, reuse never) + `mark_doc_failed` as the single FAILED funnel (per-attempt idempotent CAS, immutable `created_at`, conditional create) + per-workspace `get_failure_generation_mode` marker (`LEGACY/MIGRATING/DUAL_WRITE_READY/ENFORCED` — marker anomalies read MIGRATING, **never** LEGACY).
- **Fail-closed `count_docs_by_statuses`** and targeted `update_doc_status_fields` (immutable `created_at`, `StorageRecordNotFoundError`).
- **Strict identity lookups as NEW methods + capability flags** (`get_doc_by_file_basename_strict`, KV `get_by_id_strict`) — existing abstractmethod signatures untouched; third-party subclasses keep working via concrete base defaults (page: single-page compat; count/strict point read: `StorageCapabilityError`).
- Capability flags: `supports_bounded_scheduling_pages` / `supports_failure_generation` / `supports_strict_doc_identity_lookup` / `supports_strict_point_reads` (base defaults `False`; all five built-ins declare `True`).

### five built-in backends
- **JSON**: bounded-heap pages; sidecar ctrl file; init migrate-then-publish-ENFORCED + counter recalibration (`max(counter, max persisted)`).
- **Redis**: atomic `__sched` sidecar (per-status ZSETs, **basename → primary-doc index** with an eligibility state machine covering the in-place primary→content-duplicate transition, ctrl hash) maintained by WATCH/MULTI on every write; `mark_doc_failed` commits counter+row in one transaction (zero reserve→publish window); streaming one-time migration; O(1) ZCARD counts; basename lookups drop from O(N) SCAN to O(1).
- **PG**: idempotent DDL (three columns + keyset index + `LIGHTRAG_DOC_SCHEDULING_CTRL` counter row — deliberately not a sequence, `nextval` escapes rollback); single-transaction `mark_doc_failed`; bucket-aware NULLS FIRST keyset.
- **Mongo**: `$or` keyset + `$inc` counter + `$nor` CAS publish; missing/null `created_at` bucket resumes via the `{"created_at": None}` predicate.
- **OpenSearch**: `search_after` live-view pages; cohort predicate includes `must_not exists(failure_generation)` (legacy docs lack the field — mapping alone does not default it); `if_seq_no/if_primary_term` CAS; independent `*_scheduling_ctrl` metadata index; three-state strict point reads.
- Basename lookups on all backends now return the **primary** (non-`is_duplicate`) row only.

### pipeline FAILED-write audit
All five FAILED write sites route through `mark_doc_failed` (transition throat, enqueue error docs, in-batch `dup-*` markers, post-parse content duplicates, FAILED custom-chunk journals), plus attempt-identity bootstrap at the batch entry with cross-transition carry-over.

## Review
- 16-round adversarially reviewed design doc (plan-level), then a storage-backend-reviewer pass on this diff: 2 blocking findings (PG/Mongo corrupt-`created_at` rows starving out of keyset sweeps) fixed with fix-proof tests; workspace isolation, contract preservation, Redis atomicity, mode-marker semantics and counter monotonicity verified clean.

## Tests
- `tests/kg`: 1251 passed (per-backend `test_*_scheduling_pages.py` suites + cross-backend base-default contract tests + upgraded shared FakeRedis with real WATCH conflict detection).
- `tests/pipeline`: 418 passed (incl. new `test_failed_write_funnel.py` fix-proofs).
- Full suite: 4366 passed; the only 4 failures are a pre-existing order-dependent flake in `tests/api/routes/test_query_stream_routes.py` reproducible identically on `main`.

### DIFF
--- a/lightrag/base.py
+++ b/lightrag/base.py
@@ -3,10 +3,12 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from enum import Enum
 import os
+import uuid
 from dotenv import load_dotenv
 from dataclasses import dataclass, field
 from typing import (
     Any,
+    ClassVar,
     Literal,
     TypedDict,
     TypeVar,
@@ -15,9 +17,14 @@ from typing import (
     List,
     AsyncIterator,
 )
-from .utils import EmbeddingFunc, get_env_value
+from .utils import EmbeddingFunc, get_env_value, logger
 from .types import KnowledgeGraph
+from .exceptions import (
+    StorageCapabilityError,
+    StorageRecordNotFoundError,
+)
 from .constants import (
+    CUSTOM_CHUNK_PATCH_METADATA_KEY,
     DEFAULT_TOP_K,
     DEFAULT_CHUNK_TOP_K,
     DEFAULT_MAX_ENTITY_TOKENS,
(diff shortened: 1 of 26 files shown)


danielaskdd:
Closing this out — the memory-bounding design implemented here has been rejected in review.

Two problems with the approach:
- It is more complex than necessary.
- The `failure_generation` / attempt-id / migration-marker machinery makes scheduling correctness depend on a **second source of truth** (the marker/counter lifecycle) alongside `doc_status`. Those two lifecycles can drift (backend swap, restore, crash), and when they do the system can misjudge already-migrated data as legacy or vice versa.

The replacement design keeps manual-retry correctness entirely in `doc_status`: freeze new enqueues, drain the pipeline to logical idle, then page `FAILED → PENDING` in an exclusive phase with **no failure producers running**, and only then start processing. No new storage fields, tables, or markers are introduced.

Phase 1 will be redone on top of `main` as a fresh PR. The genuinely useful pieces here — the bounded cursor-page scheduling API, the strict-read contracts, and the NULL/missing-`created_at` keyset reachability fix — will be carried over, minus the failure-generation write side (which becomes a `canonical → eligible primary doc-id` multimap on the Redis side). The new PR will link back to this one.

PR #3481, opened 2026-07-25
✨ feat(scan): streaming discovery, typed source resolution and a bounded job record (LR2 Phase 4)

danielaskdd:
`/scan` peaked O(N) on **both** sides — a full directory listing and a full status-store read — and resolved duplicate basenames by guessing.

**Streaming discovery.** `iter_new_files()` is a generator over a single `iterdir` (replacing per-suffix globs), flushing bounded batches of `SCAN_ENQUEUE_BATCH_SIZE` (default 100). Unlike the page size this has **no** disabled value: an unbounded scan batch is exactly what this removes, so a non-positive setting fails at startup. Peak scan memory is O(batch), not O(files).

**Ordering fix.** The reset now runs **before** discovery. The old order (enqueue → drive → drain) meant a scan processed the files it had just enqueued, and their fresh `FAILED` rows were absorbed by the same request's reset — one scan extracting the same document twice. Measured, then fixed.

**Seven typed classification exits**, on `resolve_doc_source_strict` instead of a `LIMIT 1` basename lookup: Absent → Conflict → PROCESSED → STALE_STUB → IDENTITY_UNKNOWN → RESUME → ALIAS. Two calls worth stating: `SOURCE_CONFLICT` is the **only** exit that does not archive the file (the operator needs it in place), and a stale stub is deleted **only** when a strict point read confirms absence — a capability gap or a read failure routes to a non-destructive exit rather than trusting a best-effort miss. The classifier is read-only; destructive actions stay in the loop.

**Explicit conflict repair**, since the system must never pick a winner: `GET /documents/source_conflicts` (bounded, paged) and `POST /documents/source_conflicts/repair` (dry-run by default; commit echoes back count + fingerprint as a CAS token), plus `python -m lightrag.tools.source_conflict_repair`, which initializes **only** `doc_status` so it works without a reachable vector store, graph or LLM. A stale token raises a dedicated `SourceConflictRepairCASError` → **409**; its `StorageControlPlaneError` parent means "state unknown" → 503, which would have been the wrong answer.

**`created_at` = the file's `st_mtime`** for scanned files (from the stat the scan already took), so a re-discovered old file sorts by its real age instead of jumping the queue. Rejected naive timestamps outright rather than assuming UTC, and normalized to UTC because file-backed and Redis backends compare the key as a string.

**Bounded job record** (`GET /documents/scan/status/{track_id}`): three-bucket sample caps, a record byte ceiling, owner+version CAS, lease → `ABANDONED`, TTL and evict-terminal capacity. The compensation chain releases the reservation and the job slot on every refusal path (503/429/409) and on cancellation.

**Accepted behaviour change:** among files claiming one canonical name, the winner is whoever claims first — the old preference for hint variants required a whole-directory group-by, which streaming cannot do and which would break across batches anyway. Losers are archived, not deleted.

---

**Stack** (merge bottom-up; each PR is green on its own and targets the one below):
`main` ← **#3482** P0 test rig ← **#3483** P1 storage pages ← **#3484** P2 bounded sweep ← **#3485** P2.5 dedup ← **#3486** P3 manual retry ← **#3481** P4 `/scan` ← **#3487** P5 admission ← **#3488** P6 observability ← **#3489** P7 verification

Design: LR2 *pipeline memory bounding*. When merging, retarget the child PR **before** deleting a merged branch — GitHub closes the child otherwise.


danielaskdd:
Corrected in place. This PR was opened against \`main\` with a description covering the whole LR2 design, which was a mistake on my part — its head branch only ever contained Phases 1–4, so the description did not match the diff.

It is now the **Phase 4** entry of a nine-PR stack (see the map at the bottom of the description), rebased so that two test-rig commits move to the front of the stack as #3482. Every branch in the stack now passes the full suite on its own:

| PR | branch | commits | full suite |
|---|---|---:|---|
| #3482 | phase0 | 2 | 4224 passed |
| #3483 | phase1 | 6 | 4386 passed |
| #3484 | phase2 | 2 | 4410 passed |
| #3485 | phase2p5 | 1 | 4418 passed |
| #3486 | phase3 | 7 | 4433 passed |
| **#3481** | **phase4** | **8** | **4516 passed** |
| #3487 | phase5 | 5 | 4578 passed |
| #3488 | phase6 | 6 | 4623 passed |
| #3489 | phase7 | 4 | 4666 passed |

The stack tip has a byte-identical tree to the single-PR version, so the reorder changed order only, not content.

danielaskdd:
CI is green after a fix to the stack ordering. The two failures had independent causes, both artifacts of where a commit sat rather than of what it did:

1. **Lint** — a formatting reflow was committed in Phase 6, while the unformatted code enters in Phase 2 and 2.5. Phase 4 sits in that gap, so `pre-commit` failed here while the stack tip was clean.
2. **Offline Tests** — `test_endpoints_require_authentication` pinned 401. The shared auth dependency answers 401 ("please login") when password auth is configured and 403 ("API Key required") when only an API key is, so the test tracked whether the runner had `AUTH_ACCOUNTS` in a local `.env`. It now asserts refusal plus no side effect and leaves the code choice to auth’s own tests.

Rather than patch the two spots, the stack was rebuilt running `pre-commit` after each commit and folding any change into that same commit, so formatting holds by construction. That also made the standalone Phase 6 style commit redundant (41 → 40 commits). The tip tree is byte-identical to the previously reviewed version.

Every branch was then re-verified against both of the commands CI actually runs — `pre-commit run --all-files` and `pytest tests -m offline`:

| PR | branch | commits | lint | offline tests |
|---|---|---:|---|---|
| #3482 | phase0 | 2 | clean | 3346 passed |
| #3483 | phase1 | 6 | clean | 3508 passed |
| #3484 | phase2 | 2 | clean | 3532 passed |
| #3485 | phase2p5 | 1 | clean | 3538 passed |
| #3486 | phase3 | 7 | clean | 3553 passed |
| **#3481** | **phase4** | **8** | clean | 3637 passed |
| #3487 | phase5 | 5 | clean | 3699 passed |
| #3488 | phase6 | 5 | clean | 3744 passed |
| #3489 | phase7 | 4 | clean | 3758 passed |

Worth noting for whoever merges this: the workflow triggers on `pull_request: branches: [main, dev]`, so the stacked PRs get no checks until their base becomes `main`. The green checks above were obtained by pointing this PR at `main` briefly; its base is back to `feat/lr2-bounded-scheduling-phase3`.

PR #3490, opened 2026-07-26
fix(operate): keep default weight when relationship description is numeric

santhreal:
Text-mode relation rows are five fields with no weight. A numeric description such as "1000" was taken as the edge weight.

Keep the default weight when the description is numeric, with a test.


danielaskdd:
@codex review

PR #3491, opened 2026-07-26
fix(api): require string content on query conversation_history messages

santhreal:
conversation_history entries with non-string or missing content passed QueryRequest validation and then crashed in LLM text sanitization (int has no strip).

Require string content at request validation, with a test.


danielaskdd:
@codex review

PR #3492, opened 2026-07-26
fix(utils): fall back when float env values are non-finite

santhreal:
get_env_value accepted nan/inf for float envs such as COSINE_THRESHOLD, so threshold comparisons broke.

Fall back to the default when the parsed float is non-finite, with a test.


danielaskdd:
@codex review

PR #3493, opened 2026-07-26
fix(utils): return no chunks for non-positive weighted polling max

santhreal:
RELATED_CHUNK_NUMBER=-1 reached pick_by_weighted_polling and used Python reverse slicing, so trailing chunks were returned instead of an empty selection.

Return no chunks when max_related_chunks is non-positive, with a test.


danielaskdd:
@codex review

PR #3494, opened 2026-07-26
🐛 fix(api): stop serving when the gunicorn master is gone

danielaskdd:
## Problem

gunicorn's own workers exit when their parent changes — `gunicorn/workers/sync.py` checks `self.ppid != os.getppid()` on every loop turn. `UvicornWorker` replaces `run()` with an asyncio server and never makes that check.

So a master killed without signalling its children — `kill -9 <master_pid>` reaches **one** process; POSIX does not cascade to children — leaves the workers serving forever:

- nothing restarts a worker that then crashes, and SIGHUP / reload have no master to drive them;
- the workers keep the **inherited listening socket** open, so the port stays bound and a replacement master cannot start. gunicorn sets `SO_REUSEADDR` and leaves `reuse_port` off by default, so the operator gets `EADDRINUSE` from a server they believe is dead.

(Container and systemd deployments are unaffected: PID 1 exiting tears down the cgroup. This is the bare-process / manual-kill case.)

## Fix

`LightRAGUvicornWorker` adds the check back on uvicorn's **existing** notify tick — `callback_notify`, driven roughly every `timeout / 2` seconds, the value the arbiter hands each worker — so it introduces no timer of its own.

Shutdown goes through `SIGTERM` rather than poking the server object:

- `callback_notify` has no handle on the `Server` instance (`UvicornWorker._serve` keeps it local);
- uvicorn installs its own SIGTERM handler for the whole of `serve()` (`capture_signals` → `handle_exit` → `should_exit`), which is the documented graceful path: stop accepting, drain in-flight requests, exit;
- `SIGKILL` would drop live requests.

## Test-guard change

`test_gunicorn_uses_upstream_uvicorn_worker` pinned `worker_class` to the exact string `"uvicorn.workers.UvicornWorker"`. Its docstring says the regression it guards is a worker class that injects `root_path` through `CONFIG_KWARGS`, so it now asserts **that** — the configured class resolves, is a subclass of upstream `UvicornWorker`, and carries no `root_path` in `CONFIG_KWARGS` (inherited entries included). An exact-name check never proved that anyway: a subclass could add `root_path` under any name.

## Verification

- 4 new tests, including a fix-proof: neutering the parent comparison (method and signature kept) fails the orphan tests.
- `tests/api/` 343 passed, 14 skipped.
- Full offline suite: 3350 passed, 2 skipped.

Independent of the LR2 bounded-scheduling stack (#3483–#3489) — separate branch off `main` on purpose.

### DIFF
--- a/lightrag/api/gunicorn_config.py
+++ b/lightrag/api/gunicorn_config.py
@@ -31,8 +31,9 @@ keyfile = None
 # Enable preload_app option
 preload_app = True
 
-# Use Uvicorn worker
-worker_class = "uvicorn.workers.UvicornWorker"
+# Use Uvicorn worker, subclassed so a worker whose master died stops serving
+# instead of holding the listening socket forever (see gunicorn_worker.py).
+worker_class = "lightrag.api.gunicorn_worker.LightRAGUvicornWorker"
 
 # Other Gunicorn configurations
 
--- a/lightrag/api/gunicorn_worker.py
+++ b/lightrag/api/gunicorn_worker.py
@@ -1,0 +1,58 @@
+"""Gunicorn worker class: uvicorn plus the orphan check uvicorn drops.
+
+gunicorn's own workers exit when their parent changes — ``sync.py`` checks
+``self.ppid != os.getppid()`` on every loop turn. ``UvicornWorker`` replaces
+``run()`` with an asyncio server and never makes that check, so a master that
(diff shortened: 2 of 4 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3496, opened 2026-07-27
fix(operate): distinguish naive-mode vector search failures from genuine no-context

### LINKED ISSUE #3195: [v1.5.0] /query still returns [no-context] due to embedding worker timeout even though embeddings API is reachable and documents exist
## Summary

After upgrading Docker deployment to LightRAG core 1.5.0 (API 0298), /query still returns [no-context] for content that clearly exists in /documents. Logs show embedding worker timeout inside the retrieval path.

## Environment

- Deployment: Docker
- LightRAG core: 1.5.0
- API version: 0298
- EMBEDDING_BINDING=openai
- EMBEDDING_BINDING_HOST=https://api.siliconflow.cn/v1
- EMBEDDING_MODEL=BAAI/bge-m3
- LLM_BINDING=openai
- LLM_BINDING_HOST=https://coding.dashscope.aliyuncs.com/v1
- LLM_MODEL=qwen3.6-plus
- MAX_ASYNC=2
- EMBEDDING_FUNC_MAX_ASYNC=2
- LLM_TIMEOUT=30
- EMBEDDING_TIMEOUT=12
- TIMEOUT=60

## What works

- Direct embedding provider calls succeed quickly outside LightRAG
- /documents contains the relevant documents
- App-side fallback using /documents summaries can answer the question

## Proof that documents exist

Current /documents listing includes 34 processed docs total, including 7 BMW 5-series docs. Example filenames:

- 宝马5系F18_10-17年_22款_切割版.md
- 宝马5系F18_10-17年_22款_无损版.md

## Reproduction

Request:

```json
POST /query
{
  "query": "宝马5系 老改新 多少钱 F18",
  "mode": "naive",
  "only_need_context": true,
  "top_k": 3
}
```

Actual response:

```json
{
  "response": "Sorry, I'm not able to provide an answer to that question.[no-context]",
  "references": []
}
```

## Relevant logs

```text
Embedding func: 2 new workers initialized (Timeouts: Func: 12s, Worker: 24s, Health Check: 39s)
WARNING: Embedding func: Worker timeout for task ... after 24s
ERROR: Error in _get_vector_context: Embedding func: Worker execution timeout after 24s
[naive_query] No relevant document chunks found; returning no-result.
```

## Expected

Since the documents already exist and the embedding provider is reachable, /query should return relevant context instead of [no-context].

## Actual

The embedding worker path times out during retrieval, and the query returns no context.

## Notes

As a temporary mitigation on the downstream app side, we switched customer-service retrieval to `/documents` summary-first and use `/query` only as supplemental fallback. That keeps production usable, but the underlying LightRAG /query issue remains.

AmirF194:
`_get_vector_context` wraps `chunks_vdb.query()` in a bare `except Exception` and returns `[]` on any failure, the same shape it returns for a genuine zero-match search. `naive_query` then treats an empty list as "no relevant document chunks", so an embedding-worker timeout during retrieval reads identically to the vector store honestly finding nothing, and the caller gets back the generic no-context response with no way to tell the two apart.

Add a `suppress_errors` parameter to `_get_vector_context`, defaulting to `True` so the mix-mode caller keeps its existing graceful degradation unchanged. `naive_query`'s own call site passes `suppress_errors=False`, so a backend failure now propagates as a real exception; `aquery_llm`'s existing `except Exception` clause already converts that into a `status: failure` response carrying the actual error message instead of the generic fail response text.

Verification:
- Added `tests/llm/test_naive_query_vector_search_failure.py`: `test_naive_query_propagates_vector_search_failure` fails on main (asserts `naive_query` raises when the vector store raises, and it silently returned a no-context result instead) and passes on this branch; a companion test confirms the mix-mode call site's default behavior (swallow and return `[]`) is unchanged.
- Full offline suite (`pytest tests/ -m offline`) passes: 3314 passed, 42 skipped, no new failures.
- `pre-commit run --files lightrag/operate.py tests/llm/test_naive_query_vector_search_failure.py` is clean.
- Not verified: the `kg`/`hybrid`/`mix` query modes are unaffected by design (unchanged default), but I did not separately exercise them against a live backend; this fix is scoped to the naive-mode path described in the issue.

Fixes #3195

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -4496,6 +4496,7 @@ async def _get_vector_context(
     chunks_vdb: BaseVectorStorage,
     query_param: QueryParam,
     query_embedding: list[float] = None,
+    suppress_errors: bool = True,
 ) -> list[dict]:
     """
     Retrieve text chunks from the vector database without reranking or truncation.
@@ -4508,6 +4509,10 @@ async def _get_vector_context(
         chunks_vdb: Vector database containing document chunks
         query_param: Query parameters including chunk_top_k and ids
         query_embedding: Optional pre-computed query embedding to avoid redundant embedding calls
+        suppress_errors: If True (default), a backend failure is logged and treated
+            like a zero-match result. Callers that need to tell a genuine empty
+            result apart from an infra failure should pass False and handle the
+            exception themselves.
 
     Returns:
         List of text chunks with metadata
@@ -4545,6 +4550,8 @@ async def _get_vector_context(
 
     except Exception as e:
         logger.error(f"Error in _get_vector_context: {e}")
+        if not suppress_errors:
+            raise
         return []
 
 
@@ -6036,7 +6043,9 @@ async def naive_query(
 
     if progress_callback:
         await progress_callback(QueryProgress.RETRIEVING_CHUNKS)
-    chunks = await _get_vector_context(query, chunks_vdb, query_param, None)
+    chunks = await _get_vector_context(
+        query, chunks_vdb, query_param, None, suppress_errors=False
+    )
 
     if chunks is None or len(chunks) == 0:
         logger.info(
--- a/tests/llm/test_naive_query_vector_search_failure.py
+++ b/tests/llm/test_naive_query_vector_search_failure.py
@@ -1,0 +1,72 @@
+import pytest
+
+from lightrag.base import QueryParam
+from lightrag.operate import _get_vector_context, naive_query
+
+
+class _FakeTokenizer:
+    def encode(self, content: str) -> list[int]:
+        return [ord(ch) for ch in content]
+
+    def decode(self, tokens: list[int]) -> str:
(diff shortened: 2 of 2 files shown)


AmirF194:
Closing this one: the fix landed already, just through a different door. danielaskdd's #3497 (merged 2026-07-28, commit 94c7ad9e "fix(operate): stop swallowing vector query failures") removes the same broad try/except around `_get_vector_context` that this PR targets, so a backend failure during naive-mode retrieval now propagates instead of reading as a confident zero-context result. That resolves the symptom in #3195.

One difference worth flagging rather than silently dropping: #3497's version removes the try/except unconditionally for every caller of `_get_vector_context`, including the mix-mode path through `_perform_kg_search`. This PR had kept mix-mode's existing graceful-degradation behavior unchanged (a `suppress_errors` flag defaulting to `True` there). Not saying which is right, just noting it in case the broader propagation to mix-mode wasn't the intended scope of that change.

PR #3498, opened 2026-07-27
🐛 fix(api): stop the ollama bracket prompt from corrupting the query

gaurav0107:
## Description

`parse_query_mode` mangles the bracket user-prompt syntax documented in `docs/LightRAG-API-Server.md` ("Add user prompt in chat"). Three shapes are wrong, all in the same `if bracket_match:` branch, and all of them corrupt the text that reaches retrieval — the non-bracket path handles every one of them correctly.

**1. No mode prefix — the first documented example.**

```
/[Use mermaid format for diagrams] Please draw a character relationship diagram for Scrooge
```

The branch rebuilds the query as `f"/{mode_prefix} {remaining_query}".strip()`. With an empty `mode_prefix` a bare `"/"` survives, so the retrieval query becomes `"/ Please draw ..."`.

**2. Bracket prompt with no query text.** Every key in `mode_map` is space-suffixed (`"/local "`, `"/global "`, ...). `strip()` removes that separator, so `"/local[be brief]"` matches nothing: the requested mode is silently downgraded to the default **and** the literal `"/local"` becomes the query.

**3. Multi-line questions are truncated.** The pattern's trailing group used `.`, which does not match a newline, so everything after the first line was dropped.

`cleaned_query` feeds `estimate_tokens(cleaned_query)` and `rag.aquery(cleaned_query, ...)`, so in each case the junk is embedded and retrieved on.

```
before: parse_query_mode("/[be brief] question")   -> ('/ question', mix,   False, 'be brief')
after:                                             -> ('question',   mix,   False, 'be brief')
before: parse_query_mode("/local[be brief]")       -> ('/local',     mix,   False, 'be brief')
after:                                             -> ('',           local, False, 'be brief')
before: parse_query_mode("/local[be brief] a\nb")  -> ('a',          local, False, 'be brief')
after:                                             -> ('a\nb',       local, False, 'be brief')
```

The last one is the clearest sign these are unintended: `"/local a\nb"` (same query, no bracket) already keeps both lines.

## Changes Made

`lightrag/api/routers/ollama_api.py`, two commits:

- Synthesize the `"/mode "` token only when a mode prefix was captured, and keep its separator:
  `query = f"/{mode_prefix} {remaining_query}" if mode_prefix else remaining_query`.
  The `"/...context"` keys carry no trailing space and never depended on the separator, so they are unaffected.
- Widen only the trailing capture to span newlines: `r"^/([a-z]*)\[(.*?)\]([\s\S]*)"`. The user-prompt group is left single-line, so a bracket containing a newline still falls through unmatched exactly as before.
- Docstring: added the no-query example, and changed the bare-bracket example's mode from `SearchMode.hybrid` to `SearchMode.mix` to match the function's actual fallthrough (see the question below).

Unknown mode prefixes (`"/nosuchmode[p] q"`) keep their existing pass-through behavior — pinned by a test.

One deliberate behavior delta: trailing whitespace on the query is no longer stripped in the bracket path. That makes it consistent with the non-bracket path, which only `lstrip()`s after removing the prefix.

## Tests

New `tests/api/test_ollama_query_mode.py` (18 cases, `@pytest.mark.offline`), modeled on `tests/api/routes/test_query_request_strip.py`. Covers all three bug shapes, plus regression pins for the forms that already worked (`"/local[p] query"`, the `"/...context"` keys, unknown prefixes, non-bracket prefixes, plain queries).

Fix-proofed by disabling each fix and watching only the matching tests fail: reverting the separator fix fails 9 cases, reverting the newline fix fails 2, and the "must keep working" pins stay green in both.

- New tests: 18 passed
- `ruff format --check` + `ruff check --ignore=E402` (ruff 0.15.17, the `.pre-commit-config.yaml` pin) on both changed files: pass
- Full `pytest tests/ -m offline`: 3056 passed, 271 failed, 44 skipped. The 271 failures are byte-for-byte the same set as a clean checkout of `main` on the same machine, and are all in `tests/setup/` (the setup-wizard shell tests do not run on macOS). Zero new failures.

## Question (not changed here)

`docs/LightRAG-API-Server.md` says "A chat message without a query prefix will trigger a **hybrid** mode query by default", and the docstring said the same, but the fallthrough returns `SearchMode.mix`. That applies to every prefix-less query, not just the bracket form, so I left the behavior alone and only made the docstring match the code. Happy to follow up either way — flip the default to `hybrid`, or correct the doc — whichever you intended.

## Related Issues

None — found while reading the Ollama-compatible endpoint.

### DIFF
--- a/docs/LightRAG-API-Server.md
+++ b/docs/LightRAG-API-Server.md
@@ -552,7 +552,7 @@ Open WebUI uses an LLM to do the session title and session keyword generation ta
 
 ### Choose Query mode in chat
 
-The default query mode is `hybrid` if you send a message (query) from the Ollama interface of LightRAG. You can select query mode by sending a message with a query prefix.
+The default query mode is `mix` if you send a message (query) from the Ollama interface of LightRAG. You can select query mode by sending a message with a query prefix.
 
 A query prefix in the query string can determine which LightRAG query mode is used to generate the response for the query. The supported prefixes include:
 
@@ -572,7 +572,7 @@ A query prefix in the query string can determine which LightRAG query mode is us
 /mixcontext
 ```
 
-For example, the chat message `/mix What's LightRAG?` will trigger a mix mode query for LightRAG. A chat message without a query prefix will trigger a hybrid mode query by default.
+For example, the chat message `/mix What's LightRAG?` will trigger a mix mode query for LightRAG. A chat message without a query prefix will trigger a mix mode query by default.
 
 `/bypass` is not a LightRAG query mode; it will tell the API Server to pass the query directly to the underlying LLM, including the chat history. So the user can use the LLM to answer questions based on the chat history. If you are using Open WebUI as a front end, you can just switch the model to a normal LLM instead of using the `/bypass` prefix.
 
--- a/docs/LightRAG-API-Server.md
+++ b/docs/LightRAG-API-Server.md
@@ -572,7 +572,7 @@ A query prefix in the query string can determine which LightRAG query mode is us
 /mixcontext
 ```
 
-For example, the chat message `/mix What's LightRAG?` will trigger a mix mode query for LightRAG. A chat message without a query prefix will trigger a mix mode query by default.
+For example, the chat message `/hybrid What's LightRAG?` will trigger a hybrid mode query for LightRAG. A chat message without a query prefix will trigger a mix mode query by default.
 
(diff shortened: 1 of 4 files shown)


danielaskdd:
**Regarding your question:** LightRAG's default query mode has been changed from "hybrid" to "mix." Accordingly, the default query mode for the Ollama simulation interface should also also be "mix". 

The documentation not being updated in a timely manner, please also update the corresponding documentation in this PR. Thank you.

danielaskdd:
@codex review

Issue #3502, opened 2026-07-28
[Bug]: MinerU ingestion fails when equation analysis omits the redundant equation field


cyysky2:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

When ingesting a PDF through MinerU with equation analysis enabled, the entire
document can be marked as failed even though MinerU has already extracted valid
LaTeX into the equation sidecar.

The Extract/VLM model may return valid `name` and `description` fields but omit
the expected `equation` field, or return the equation under a semantically
equivalent key such as `equ`.

LightRAG rejects this otherwise usable response with:

`missing or invalid field 'equation'`

This aborts the complete document ingestion after parsing and analyzing many
other equations successfully.

The equation text does not need to depend on the model response because the
corresponding equation sidecar already contains the authoritative LaTeX.

### Steps to reproduce

1. Configure LightRAG to parse PDFs through local MinerU with equation analysis:

   ```env
   LIGHTRAG_PARSER=pdf:mineru-iteP;*:legacy-R
   MINERU_API_MODE=local
   MINERU_LOCAL_ENDPOINT=http://127.0.0.1:8000
   MINERU_LOCAL_BACKEND=hybrid-auto-engine
   MINERU_LOCAL_PARSE_METHOD=auto
   MINERU_LOCAL_IMAGE_ANALYSIS=false
   MINERU_ENABLE_TABLE=true
   MINERU_ENABLE_FORMULA=true
2. Use an OpenAI-compatible Qwen model for extraction and multimodal analysis.

3. Upload the attached PDF sample and start document ingestion.

   The original reproduction was obtained from:

   [概率论与数理统计（浙江大学第四版）](https://github.com/TapXWorld/ChinaTextbook/blob/master/%E5%A4%A7%E5%AD%A6/%E6%A6%82%E7%8E%87%E8%AE%BA/%E6%A6%82%E7%8E%87%E8%AE%BA%E4%B8%8E%E6%95%B0%E7%90%86%E7%BB%9F%E8%AE%A1(%E6%B5%99%E5%A4%A7%E5%9B%9B%E7%89%88).pdf)

   The affected source is PDF page 142 (`page_idx=141` in MinerU), in the
   section about sample quantiles and box plots.

4. MinerU successfully emits the following equation in `content_list.json`:

   ```json
   {
     "type": "equation",
     "page_idx": 141,
[...]

PR #3503, opened 2026-07-28
fix: preserve MinerU content during document ingestion

cyysky2:
## Description

This pull request improves MinerU-based document ingestion by preventing valid textbook content from being silently dropped and by making long-running local parsing more resilient.

It also prevents equation analysis from failing when the model omits the redundant `equation` field while valid authoritative LaTeX is already available in the equation sidecar.

## Related Issues

#3502 

This addresses data-loss and ingestion-failure cases observed when processing large searchable and scanned PDFs containing mixed text, images, charts, tables, code, and equations.

## Changes Made

- Retry transient `httpx.RequestError` failures while polling local MinerU tasks, within the existing bounded polling budget.
- Preserve additional MinerU content types, including:
  - charts, seals, pictures, and drawing aliases
  - index and structured list items
  - code and algorithm captions, bodies, and footnotes
  - abstract, aside, page-footnote, phonetic, reference, and vertical text
- Retain MinerU source type and content metadata in drawing sidecars.
- Fall back to drawing sidecars when table or equation recognition fails but an image crop is available.
- Add configurable filtering for repeated layout noise through `MINERU_DROP_TYPES`.
- Summarize unsupported MinerU item types once per document instead of silently discarding them.
- Use authoritative equation-sidecar LaTeX when an otherwise valid model response omits the `equation` field.
- Add regression tests for the new mapping, fallback, reporting, and polling behavior.
- Document the new `MINERU_DROP_TYPES` setting in `env.example`.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Validation completed:

- `pre-commit run --all-files`
  - All hooks passed.
- `./scripts/test.sh tests/parser/external/mineru tests/pipeline/test_pipeline_release_closure.py`
  - 178 tests passed.
- `git diff --check`
  - Passed.

The complete test suite was also attempted. Collection was blocked by unavailable optional storage and LLM dependencies such as Ollama, Neo4j, Milvus, PostgreSQL, Redis, Anthropic, and Bedrock packages. No failures related to this change were observed in the relevant test suites.

### DIFF
--- a/env.example
+++ b/env.example
@@ -418,6 +418,10 @@ NATIVE_MD_IMAGE_DOWNLOAD_ENABLED=true
 MINERU_API_MODE=local
 # MINERU_POLL_INTERVAL_SECONDS=2
 # MINERU_MAX_POLLS=600
+### Local polling tolerates isolated transport errors, but fails after this
+### many consecutive errors. Any successful poll resets the counter. Set 1 to
+### restore fail-on-first-error behavior. Must be a positive integer.
+# MINERU_MAX_CONSECUTIVE_POLL_ERRORS=5
 # MINERU_LANGUAGE=ch
 # MINERU_ENABLE_TABLE=true
 # MINERU_ENABLE_FORMULA=true
@@ -644,6 +648,9 @@ DOCLING_DO_FORMULA_ENRICHMENT=false
 ### so deployments can bias context forward or backward as needed.
 # SURROUNDING_LEADING_MAX_TOKENS=2000
 # SURROUNDING_TRAILING_MAX_TOKENS=2000
+### Token cap for parser-provided visual text (for example MinerU chart
+### content) added to image-analysis prompts. Set <= 0 for no separate cap.
+# MM_IMAGE_CONTENT_MAX_TOKENS=2000
 
 ### Per-response cap on total entity+relationship rows/records emitted by the LLM
 # MAX_EXTRACTION_RECORDS=100
--- a/env.example
+++ b/env.example
@@ -471,6 +471,10 @@ MINERU_LOCAL_PARSE_METHOD=auto
 ###   round, trading image / chart semantic descriptions for faster parsing
 ###   and lower GPU cost.
 MINERU_LOCAL_IMAGE_ANALYSIS=false
+### Comma-separated legacy content_list item types filtered as repeated layout
+### noise. Set to `none` to keep all of them. Textbook ingestion normally
+### benefits from dropping page headers, footers, and page numbers.
+MINERU_DROP_TYPES=header,footer,page_number,header_image,footer_image
 # MINERU_LOCAL_START_PAGE_ID=0
 # MINERU_LOCAL_END_PAGE_ID=99999
 
--- a/lightrag/parser/external/mineru/client.py
+++ b/lightrag/parser/external/mineru/client.py
@@ -162,6 +162,16 @@ class MinerURawClient:
         self.poll_interval = float(os.getenv("MINERU_POLL_INTERVAL_SECONDS", "2"))
         # 600 * 2s client-side sleep ≈ 20 min worst case; raise for very large PDFs.
         self.max_polls = int(os.getenv("MINERU_MAX_POLLS", "600"))
+        self.max_consecutive_poll_errors = 5
+        if self.api_mode == "local":
+            self.max_consecutive_poll_errors = int(
+                os.getenv("MINERU_MAX_CONSECUTIVE_POLL_ERRORS", "5")
+            )
+            if self.max_consecutive_poll_errors < 1:
+                raise ValueError(
+                    "MINERU_MAX_CONSECUTIVE_POLL_ERRORS must be at least 1, "
+                    f"got {self.max_consecutive_poll_errors}"
+                )
         self.engine_version = os.getenv("MINERU_ENGINE_VERSION", "").strip()
 
         options = MinerUParserOptions.from_env(
@@ -423,6 +433,7 @@ class MinerURawClient:
         # a crafted value can't break out of ``/tasks/{id}``. The raw value is
         # kept for the error/timeout messages below where the real ID matters.
         poll_url = f"{self.local_endpoint}/tasks/{quote(task_id, safe='')}"
+        consecutive_poll_errors = 0
         for poll_index in range(self.max_polls):
             await asyncio.sleep(self.poll_interval)
             try:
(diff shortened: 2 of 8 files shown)


danielaskdd:
Thanks for the thorough PR and the well-documented issue (#3502). I verified the core mechanics against the codebase and the overall direction is right:

- The equation fallback value is semantically correct: for `kind == "equation"`, `content_text` is the equations-sidecar `content`, and the sidecar writer already strips `$$`/`$` wrappers (`_strip_latex_dollar_wrappers`), so the fallback fits what the downstream chunk renderer expects. The fallback also applies consistently to cached responses.
- Removing the outer `asset.ref not in {...}` dedup check in `_append_drawing` is safe — `_build_ir_drawing` only returns a non-None asset on first sight of a ref via the `seen` dict, so the old check was redundant.
- Dropping `header` / `footer` / `page_number` & friends by default is the correct call. `page_number` was already special-cased as layout noise; headers/footers leaking into body text through the unknown-item fallback was an oversight, not a contract. No migration note needed.
- The regression tests are behavior-level (the pipeline test drives the real `analyze_multimodal` path; the client test reuses the existing fake-httpx infrastructure). Good.

That said, there are two changes I'd like to see before merging, and two smaller points to address in comments/description.

### 1. `extras["mineru_content"]` has no consumer — chart content becomes unreachable in some configs

Nothing in the pipeline reads `IRDrawing.extras` today (`pipeline.py`, `multimodal_context.py`, `operate.py` — no consumer). Before this PR, a `chart` item fell into the unknown-item fallback, and `_coerce_text` picked up its `content` field (MinerU's chart-to-text data), so the chart's textual data entered the block text and was retrievable. After this PR, the chart becomes a drawing and its `content` only lands in a write-only sidecar field.

When VLM analysis is available, the image description compensates and this is a net improvement. But in deployments where VLM is disabled or not configured, this is a retrievability regression: chart data that used to be in chunks is now nowhere. The PR description says the extras allow "downstream prompt specialization", but that wiring doesn't exist.

Please either (a) wire `mineru_content` into the drawing analysis prompt (e.g. as additional context alongside captions/footnotes), or (b) keep emitting the chart `content` as block text in addition to the drawing, or (c) at minimum document this trade-off explicitly so operators without VLM know what changes. I'd prefer (a).

### 2. Poll retry needs a consecutive-failure bound

`_poll_local_task` now swallows every `httpx.RequestError` and continues. With the defaults (`MINERU_MAX_POLLS=600`, `MINERU_POLL_INTERVAL_SECONDS=2`), a MinerU server that is permanently down changes from "fail on the first connection error" to burning the entire budget — 600 × (2s sleep + connect timeout per attempt), which is well over 20 minutes per document before the failure surfaces.

The transient-starvation scenario you describe is real, but it doesn't produce long unbroken error streaks. Please add a consecutive-transport-error counter (reset on any successful poll) and fail after a small bound (e.g. 5). That keeps the resilience for isolated resets while preserving fast failure detection for real outages.

### 3. The `equation` field is not a "redundant echo"

The `equation_analysis` prompt (`prompt_multimodal.py`, section 4) asks the model for a *normalized* body: strip `\tag{...}`, convert `align`/`align*` → `aligned`, convert Markdown/Unicode math to LaTeX. Falling back to the raw sidecar LaTeX bypasses that normalization. For MinerU output (already LaTeX) the practical impact is small and I'm fine with the behavior — requiring `name`/`description` while backfilling the deterministic field is the right trade-off versus failing a whole document. But please adjust the code comment and PR description: the fallback returns *unnormalized* source LaTeX, it is not simply "the redundant normalized-equation echo".

### Minor

- The empty-equation / empty-table → drawing fallback means documents that previously succeeded (item silently dropped) can now fail at the `vlm_process_enable` gate in deployments with image processing enabled but no VLM role. Likely a small blast radius (such deployments would already fail on any real image), but worth a sentence in the description.
- `_visual_annotation_keys` gives `seal` the `image_caption`/`image_footnote` keys — please double-check whether MinerU emits `seal_caption`-style keys for seals; if it does, add them.

Happy to merge once items 1 and 2 are addressed.

cyysky2:
Hi! Thanks for your reply! I made the following changes according to your suggestions:
- Implemented your suggestion a) by passing `extras["mineru_content"]`  into the image-analysis prompt. It is treated as untrusted supporting evidence, capped by `MM_IMAGE_CONTENT_MAX_TOKENS`, and included in the VLM cache key.
- Added `MINERU_MAX_CONSECUTIVE_POLL_ERRORS` with a default of 5. The counter resets after every successful transport-level poll, while `MINERU_MAX_POLLS` remains the overall completion bound.
- Clarified that the equation fallback preserves potentially unnormalized source-sidecar LaTeX.
- Added no-VLM handling for newly promoted MinerU drawings: parser content remains retrievable, while empty equation/table crops are skipped without failing the document. Existing ordinary-image misconfiguration remains a hard failure.
Verified that MinerU represents seals as `type=image, sub_type=seal` with `image_caption` and `image_footnote`.
Validation passed locally (269 passed, Ruff, Ruff format, pre-commit) and on a real A800/Qwen VLM deployment (188 passed plus live VLM/no-VLM checks). 
I am not a professional and as a user, I stumbled into some issues while using this project. This is an amazing project. The changes I made may not have comprehended the whole project and I hope this PR is not causing you too much trouble.

danielaskdd:
Thanks for the update. I reviewed the current head (`892940de`) and verified that the consecutive polling-error bound, source-LaTeX clarification, prompt/cache-key integration, and seal handling are now implemented.

However, the chart-content retrievability issue is not fully resolved for the default ingestion path.

`chart` items are now always converted into drawings in `MinerUIRBuilder`. Previously, an unrecognized chart entered the unknown-item fallback, where `_coerce_text()` preserved its `content` in the document body.

The new parser-content fallback only runs inside `analyze_multimodal()`. Normal file uploads without parser directives use `process_options="F"`, and `analyze_multimodal()` returns early when none of `i`, `t`, or `e` is enabled. Consequently, the new `_parser_content_result()` path is never reached:

1. The chart content is removed from the normal block text.
2. The block contains only a drawing placeholder.
3. Multimodal analysis is skipped because `i` is not enabled.
4. No multimodal chunk containing `mineru_content` is created.

This means chart text remains silently unreachable under the default upload configuration, including typical deployments without VLM. The new no-VLM regression test does not cover this path because it calls `analyze_multimodal(..., process_options="i")` explicitly.

Please preserve chart content independently of the `i` option—for example, by also emitting it as block text—or otherwise ensure the parser-content fallback is available on the default ingestion path.

One documentation point also remains: the current PR description still calls the omitted `equation` field “redundant.” As discussed earlier, the model-produced field is normalized, while the fallback may preserve unnormalized source-sidecar LaTeX. The code comment now explains this correctly, but the PR description should be updated to match.

danielaskdd:
Could you please send the PDF file of the textbook you're experiencing the issue with to my email address (gzdaniel@me.com)? This will allow me to reproduce the problem on my end and offer more targeted suggestions for a fix.

cyysky2:
Hi, the PDF used and step to reproduce the issue is in issue #3502. I am now more convinced that this is a feature rather than a bug.

PR #3513, opened 2026-07-30
fix(operate): guard against empty entity list indexing in cached extraction rebuild

santhreal:
When cached extraction results contain empty entity lists, _rebuild_from_extraction_result raises IndexError indexing entity_list[0]. Skip empty entity lists before accessing the first element.

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -1003,6 +1003,8 @@ async def rebuild_knowledge_from_chunks(
                 # Merge entities and relationships from this extraction result
                 # Compare description lengths and keep the better version for the same chunk_id
                 for entity_name, entity_list in entities.items():
+                    if not entity_list:
+                        continue
                     if entity_name not in chunk_entities[chunk_id]:
                         # New entity for this chunk_id
                         chunk_entities[chunk_id][entity_name].extend(entity_list)
@@ -1026,6 +1028,8 @@ async def rebuild_knowledge_from_chunks(
 
                 # Compare description lengths and keep the better version for the same chunk_id
                 for rel_key, rel_list in relationships.items():
+                    if not rel_list:
+                        continue
                     if rel_key not in chunk_relationships[chunk_id]:
                         # New relationship for this chunk_id
                         chunk_relationships[chunk_id][rel_key].extend(rel_list)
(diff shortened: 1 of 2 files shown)


danielaskdd:
Thanks for the contribution, @santhreal. After reviewing this against `main`, we're going to close it. Details below so the reasoning is transparent, and so a follow-up PR can be scoped properly if the underlying problem is real.

## 1. The described bug is not reachable in the current code

The description says `_rebuild_from_extraction_result` raises `IndexError` indexing `entity_list[0]`. That indexing is not in `_rebuild_from_extraction_result` — it's in its caller, `rebuild_knowledge_from_chunks`. More importantly, the input can't be empty:

- `_process_json_extraction_result` (`lightrag/operate.py:819-989`) and `_process_extraction_result` (`lightrag/operate.py:1417-1526`) are the only two producers. Both use `defaultdict(list)` and only ever create a key via `.append(...)`; the defaultdict is never read-accessed in a way that would materialize an empty entry.
- The multimodal injection path (`lightrag/operate.py:3884`, `3918`) uses `setdefault(k, [])` immediately followed by `.append(...)`.

So the returned dicts cannot contain a key mapped to an empty list. The `elif len(...) == 0` branches already on `main` (`lightrag/operate.py:1118`, `1141`) cover the *existing* side; the new guards cover the *incoming* side, which is structurally non-empty. Same conclusion for the gleaning merge guards (`lightrag/operate.py:3831`, `3849`).

## 2. The `entity_type` Counter cannot be empty either

For `_merge_nodes_then_upsert` step 6.2:
- An empty `entity_type` is already dropped during parsing (`lightrag/operate.py:862-872`, `if not entity_type.strip(): ... continue`), so every node dict carries a non-empty type.
- `already_entity_types` is non-empty exactly when `already_node` exists (`lightrag/operate.py:2271`).
- When `nodes_data` is emptied by the KEEP source-id filter, step 5 already returns early or raises `ValueError` (`lightrag/operate.py:2344-2358`).

Separately, switching `dp["entity_type"]` to `dp.get("entity_type")` **with a truthiness filter is a behavior change, not a guard**: previously an empty-string type participated in the vote and could win; now it's excluded. That may well be an improvement, but it's undocumented in the PR description and untested.

## 3. `raw_data["data"]` is never `None`

`convert_to_user_format` (`lightrag/utils.py:5363-5372`) always returns a dict under `"data"`, and both return paths of `_build_context_str` (`lightrag/operate.py:5285`, `5334`) come from it. The `(... or {})` change is unreachable.

## 4. The tests don't exercise the fix

This is the blocking issue. All five tests re-implement the patched logic inline in the test body and never call any production function — the only import, `_build_query_context`, is unused. Reverting `lightrag/operate.py` to this PR's base commit and re-running the new file gives:

```
tests/test_cached_extraction_empty_list_index.py .....   [100%]
5 passed
```

They pass without the fix, so there is no regression coverage.

## 5. Mechanical issues

- CI **Linting and Formatting** is failing: two `F401` unused imports (`AsyncMock`, `_build_query_context`), plus `ruff format --check` wants to reformat both `lightrag/operate.py` and the new test file.
- Per `AGENTS.md`, the test belongs under `tests/extraction/`, not the `tests/` root.
- The branch base (`79b5427`) is behind `main`.

## If the problem is real

We'd happily look at a focused follow-up. What would help:

1. A traceback or reproduction path — e.g. a specific storage backend or a legacy cache payload format that actually yields an empty list. If such a path exists, the right fix is at the producer, not at every consumer's index site.
2. Tests that call `rebuild_knowledge_from_chunks` / `_merge_nodes_then_upsert` directly (mocked graph/vector storage), so they fail before the patch and pass after.
3. The `entity_type` empty-string filtering split out as its own change with its own test — that one is independently reasonable if you've observed such dirty data.

---

PR #3514, opened 2026-07-30
fix(rerank): ignore out-of-bounds document indices in chunk score aggregation

santhreal:
When doc_indices maps a chunk to a document index outside the valid range of doc_ids, aggregate_chunk_scores raises KeyError.

Safely skip document indices that fall outside the valid index map.


danielaskdd:
@codex review

PR #3528, opened 2026-07-31
feat(parser): record numbering counts past a converter's domain

danielaskdd:
## Description

Follow-up to #3526, from its review. That PR wired the CJK counting families to `_to_chinese`, which renders 1–99 and returns the decimal string above it. So a counting-family list that reaches item 100 emits `100` where Word shows a Chinese numeral — and does so without a trace.

This PR does **not** implement the missing renderings. It makes the event findable.

## Related Issues

None. Raised in review of #3526.

## Changes Made

`LIMITED_DOMAIN_FORMATS` maps each numFmt to the largest count its converter actually renders (the four counting families → 99). A count past the limit is reported once per format, mirroring the unsupported-numFmt ledger added in #3526:

- `out_of_range_formats` on the resolver,
- a log line,
- `numbering_out_of_range_formats` in `parse_warnings` (no `smart_` prefix — this is the baseline path too, so it belongs in `doc_status`, not the smart audit file).

The two ledgers are mutually exclusive: no converter at all vs. a converter with a domain. Tested.

`ideographDigital` is deliberately absent from the table — it renders every count digit-by-digit, so 100 is a correct `一〇〇`, not a degraded label. A test pins that so it cannot be "tidied" into the list.

### Why the labels are not corrected

**The degradation is legible.** `（100）` is obviously not a Chinese numeral. That is categorically different from the defect #3526 fixed, where `（1）` passes for `（一）` and nobody notices.

**"Render past 99" is not one implementation.** Per [\[MS-DOCX\] numFmt Extensions](https://learn.microsoft.com/en-us/openspecs/office_standards/ms-docx/a1bb5809-e361-4e49-8e16-7f1a67da4121) `chineseCounting` / `taiwaneseCounting` switch to a U+25CB digit-by-digit form at 100 (`一○○`) while `chineseCountingThousand` keeps counting (`一百`) — and `japaneseCounting` needs its own check. Three renderings, one shared converter today.

**Nothing can verify them.** No corpus document reaches 100 in any of these families, and two of the four (`taiwaneseCounting`, `ideographDigital`) do not appear in the corpus at all — #3526 already states they rest on the specification table and unit tests alone. Adding three branches on the same evidence widens that blind spot instead of closing it.

So: record now, and let a real document that gets there be the evidence for implementing the right one. Doing it properly means generating per-family 1/10/100/1000 samples from Word and WPS and checking each cell, not reading one spec table.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

### Verification

- `tests/parser/docx/test_numbering_resolver.py`: **46 passed** (8 new, including a 4-family parametrisation over the domain boundary at 99/100/101)
- `tests/parser/docx`: **822 passed**
- Fix-proof: with the `_note_out_of_range` call removed the new tests fail behaviourally on the ledger contents (`assert set() == {'japaneseCounting'}`), not on a missing symbol.
- `pre-commit run --files …`: ruff format + ruff pass

### Scope note

`_to_roman` has the same shape of limit (it degrades above 3999) and is not in the table. A list numbering scheme reaching 4000 is not a case worth carrying a table entry for; the CJK entries exist because 100 is reachable.

### DIFF
--- a/lightrag/parser/docx/numbering_resolver.py
+++ b/lightrag/parser/docx/numbering_resolver.py
@@ -49,6 +49,24 @@ class NumberingResolver:
         "none": lambda n: "",
     }
 
+    #: numFmt -> the largest count its converter actually renders. Above the
+    #: limit the label degrades to the decimal string, which is legible and
+    #: obviously not a Chinese numeral (unlike the silent decimal default for an
+    #: UNMAPPED numFmt, where `（1）` passes for `（一）`). The counting families
+    #: all share ``_to_chinese``'s 1-99 domain, but they do NOT share a single
+    #: rendering above it: per [MS-DOCX] "numFmt Extensions" chineseCounting /
+    #: taiwaneseCounting switch to a U+25CB digit-by-digit form at 100 (一○○)
+    #: while chineseCountingThousand keeps counting (一百) — three renderings, no
+    #: corpus document that reaches any of them, so none is implemented. This
+    #: table exists to make the event FINDABLE: a real document that gets there
+    #: is the evidence needed to implement the right one.
+    LIMITED_DOMAIN_FORMATS = {
+        "chineseCounting": 99,
+        "chineseCountingThousand": 99,
+        "japaneseCounting": 99,
+        "taiwaneseCounting": 99,
+    }
+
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3531, opened 2026-07-31
fix(api): bound the token-renewal cache and stop pre-authorization writes (GHSA-3wg5-5w54-3rfm)

danielaskdd:
## Summary

The token auto-renewal block ran **before** the authorization decision and wrote to an unbounded process-wide dict keyed on the raw JWT `sub` claim. A caller holding no credential could therefore make the server retain permanent state, and forge log records, on requests it correctly answered with 403.

Fixes **GHSA-3wg5-5w54-3rfm** (CWE-770 unbounded allocation, CWE-117 log injection).

## Motivation

Three facts combined in `get_combined_auth_dependency.combined_dependency`:

1. **Ordering.** Auto-renewal is step 2; the `X-API-Key` check is step 4. The renewal write happened on requests that were about to be rejected.
2. **Unbounded key.** `_token_renewal_cache[sub] = now` on a plain `dict` with no cap and no eviction. In the API-key-only profile (`LIGHTRAG_API_KEY` set, `AUTH_ACCOUNTS` unset) the JWT secret falls back to the public `DEFAULT_TOKEN_SECRET`, so `sub` is attacker-chosen and only bounded by the HTTP header limit — measured at roughly 100 KB per request under the plain-`uvicorn`/h11 stack this project depends on. Nothing evicted, so growth persisted until process restart, and `update_uvicorn_mode_config` forces `workers=1`, making that one process the whole API.
3. **Self-defeating rate limiter.** The block is labeled `# ===== Token Renewal Rate Limiting =====` and enforces a 60 s interval "for same user", keyed solely on `sub`. A fresh forged `sub` per request makes `last_renewal` always 0, so the interval check passed every time while the table still grew.

The renewal `INFO` line interpolated the same raw claim, so CR/LF in it emitted a record rendering as multiple lines — an attacker-chosen one indistinguishable from a genuine entry.

Worth stating explicitly, because it is easy to read this the other way round: the API-key bypass of GHSA-f4vv-55c2-5789 is **already fixed** (`f7819aa3`, released in v1.5.4) — a forged guest token no longer authenticates in the API-key-only profile. That is what makes this a real boundary crossing rather than a footnote on an already-broken profile: the attacker gets no data access, is refused on every single request, and still writes permanent server-side state.

## What's changed

- **`validate_token` bounds and type-checks the claims.** `sub` must be a non-empty `str` of at most `MAX_TOKEN_SUBJECT_LENGTH` (256); real subjects are the literal `"guest"` or an `AUTH_ACCOUNTS` key, both far shorter. This method is the single choke point every authenticated path passes through, so one cap here simultaneously bounds the cache key, the log line, and the cost of re-signing the claim.
- **Renewal moved after the authorization decision**, to the two exits of `combined_dependency` that actually accept the token. In API-key-only mode the guest token falls through to the mandatory `X-API-Key` check with no renewal bookkeeping at all.
- **`_record_token_renewal` keeps the table bounded.** Insertion order is maintained as time order, so entries older than `_RENEWAL_MIN_INTERVAL` — which can no longer influence any decision — are purged from the head, then `_MAX_TRACKED_RENEWALS` (10 000) is enforced as a hard ceiling. Both passes are amortized O(1).
- **Both renewal log lines go through `safe_log_value`**, the helper already applied to the login rate-limit and document audit logs.

### Note on the eviction policy

`login_rate_limit.py` is the in-codebase precedent for bounding a table on this request path, but its central rule — *never evict a live record* — deliberately does **not** carry over. There, evicting an active lockout would clear it, i.e. a security bypass. Here a live entry only suppresses an early renewal, so evicting one costs a single extra token mint and nothing else. Copying the sibling's constraint would leave the table full of live entries under exactly the attack it is meant to stop; enforcing the cap unconditionally is what makes it genuinely bounded rather than capped-then-full.

## Adjacent fix

Same code path, unauthenticated HTTP 500: `payload["sub"]` / `payload["exp"]` were bare subscripts, and `KeyError` is not covered by `except jwt.PyJWTError`, so a signature-valid token missing either claim escaped `validate_token` as a server error instead of a 401. A numeric but out-of-range `exp` passes PyJWT's own check and then overflows `datetime.fromtimestamp` the same way. Both now return 401. It shares the choke point with the `sub` validation above, hence the same patch.

## What's broken / behavior changes

- A JWT whose `sub` exceeds 256 characters is now rejected with 401. No supported configuration can produce one: `/login` signs the matched `AUTH_ACCOUNTS` key and `/auth-status` signs the literal `"guest"`.
- A JWT missing `sub` or `exp`, or carrying an out-of-range `exp`, now returns 401 instead of 500.
- Under sustained load from more than 10 000 distinct authenticated users within a 60 s window, an evicted entry lets one user renew before the soft interval elapses. Harmless by design — see the eviction note above.
- Renewal itself is unchanged for every profile that legitimately used it: the `X-New-Token` header, the 60 s interval, the `_TOKEN_RENEWAL_SKIP_PATHS` exclusions, and the threshold arithmetic all behave as before.

## Tests

New `tests/api/auth/test_token_renewal_bounds.py` (16 tests) drives the real `get_combined_auth_dependency` and asserts on the real `_token_renewal_cache`. The pre-existing `test_token_auto_renewal.py` re-implements the renewal algorithm against its own local dict and so cannot observe any of this; it is left untouched.

All 12 fix-proof tests fail **behaviorally** on the unpatched code — wrong cache contents, `DID NOT RAISE`, a raw newline in the emitted record — rather than with `AttributeError` on symbols the fix introduces; new symbols are referenced inside test bodies, never at module level, so a collection error cannot masquerade as proof. The 4 stability tests (renewal still fires, rate limit still suppresses, skip paths still skip) pass on both sides.

```
tests/api/auth/test_token_renewal_bounds.py   16 passed
  ... same file against unpatched code:       12 failed, 4 passed
tests/api                                    572 passed, 14 skipped
```

`pre-commit run --files` clean on all three files.

### DIFF
--- a/lightrag/api/auth.py
+++ b/lightrag/api/auth.py
@@ -32,6 +32,14 @@ _DUMMY_VERIFY_SPEC = (
 )
 
 
+# Upper bound on the JWT "sub" (username) claim accepted by validate_token.
+# Deliberately generous: real subjects are either the literal "guest" or a key
+# of AUTH_ACCOUNTS, both far shorter. The bound exists so an attacker-forged
+# claim cannot be used as an arbitrarily large payload downstream -- see the
+# comment in validate_token.
+MAX_TOKEN_SUBJECT_LENGTH = 256
+
+
 class TokenPayload(BaseModel):
     sub: str  # Username
     exp: datetime  # Expiration time
@@ -171,8 +179,43 @@ class AuthHandler:
                     detail="Insecure JWT algorithm configuration",
                 )
             payload = jwt.decode(token, self.secret, algorithms=allowed_algorithms)
-            expire_timestamp = payload["exp"]
-            expire_time = datetime.fromtimestamp(expire_timestamp, timezone.utc)
+
+            # Claim validation. In any profile that runs on DEFAULT_TOKEN_SECRET
+            # (no AUTH_ACCOUNTS, see __init__) the whole payload is attacker-
+            # forgeable, so every claim read below is untrusted input. This method
+            # is the single choke point all authenticated paths pass through, so
+            # bounding the claims here lets callers treat "username" as a short,
+            # well-typed string. A malformed claim is an invalid token, not a
+            # server error: the previous bare payload["sub"] / payload["exp"]
+            # raised KeyError, which jwt.PyJWTError does not cover, so a
+            # signature-valid token missing either claim surfaced as HTTP 500.
+            username = payload.get("sub")
+            if not isinstance(username, str) or not username:
+                raise HTTPException(
+                    status_code=status.HTTP_401_UNAUTHORIZED, detail="Invalid token"
+                )
+            if len(username) > MAX_TOKEN_SUBJECT_LENGTH:
+                # Bounds every downstream use of the claim at once: the renewal
+                # cache key, the renewal log line, and the cost of re-signing it.
+                # Without this an attacker mints a ~100 KB "sub" per request
+                # (CWE-770 / CWE-117, GHSA-3wg5-5w54-3rfm).
+                raise HTTPException(
+                    status_code=status.HTTP_401_UNAUTHORIZED, detail="Invalid token"
+                )
+
+            expire_timestamp = payload.get("exp")
+            if expire_timestamp is None:
+                raise HTTPException(
+                    status_code=status.HTTP_401_UNAUTHORIZED, detail="Invalid token"
+                )
+            try:
+                # jwt.decode already rejects a non-numeric "exp", but a numeric
+                # one far outside the datetime range reaches here intact.
+                expire_time = datetime.fromtimestamp(expire_timestamp, timezone.utc)
+            except (OverflowError, OSError, ValueError):
+                raise HTTPException(
+                    status_code=status.HTTP_401_UNAUTHORIZED, detail="Invalid token"
+                )
 
             if datetime.now(timezone.utc) > expire_time:
                 raise HTTPException(
@@ -181,7 +224,7 @@ class AuthHandler:
 
             # Return complete payload instead of just username
             return {
-                "username": payload["sub"],
+                "username": username,
                 "role": payload.get("role", "user"),
                 "metadata": payload.get("metadata", {}),
                 "exp": expire_time,
--- a/lightrag/api/auth.py
+++ b/lightrag/api/auth.py
@@ -72,20 +72,42 @@ class AuthHandler:
         self.guest_expire_hours = global_args.guest_token_expire_hours
         self.accounts = {}
         invalid_accounts = []
+        oversized_username_lengths = []
         if auth_accounts:
             for account in auth_accounts.split(","):
                 try:
                     username, password = account.split(":", 1)
                     if not username or not password:
                         raise ValueError
-                    self.accounts[username] = password
                 except ValueError:
                     invalid_accounts.append(account)
+                    continue
+                if len(username) > MAX_TOKEN_SUBJECT_LENGTH:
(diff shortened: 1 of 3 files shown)


danielaskdd:
Both findings confirmed and fixed in `8aaa22b2`. Both were regressions I introduced in `c8ea584f`, and both were real rather than theoretical — I verified the premises before fixing.

## 1. Long-username account could log in but not use its token

Correct, and the framing "not shared with `create_token`" is the right diagnosis: I enforced the cap only on read. `AuthHandler.__init__` accepted any non-empty username, `/login` signed it into `sub`, and `validate_token` then rejected that very token.

Fixed in account-config parsing (`AuthHandler.__init__`), which fails fast with an actionable message rather than letting the account exist in a broken state. Only the offending username **lengths** are logged — the rejected entry carries a password, and the pre-existing `Invalid account format` line already echoes whole entries, so I did not want to add a second path that does.

The regression test is written as an **implication** rather than pinned to the current cap:

```python
@pytest.mark.parametrize("length", [1, 64, 256, 257, 1000])
def test_accepted_account_always_yields_a_usable_token(self, monkeypatch, length):
    try:
        handler = _handler_with_accounts(monkeypatch, f"{username}:secret")
    except ValueError:
        return  # refused up front; no unusable account to speak of
    assert handler.verify_password(username, "secret")
    token = handler.create_token(username=username, role="user")
    assert handler.validate_token(token)["username"] == username
```

An account refused at configuration time satisfies it vacuously, so it keeps its meaning whatever `MAX_TOKEN_SUBJECT_LENGTH` becomes. On the parent commit the `[257]` and `[1000]` cases fail behaviorally — `validate_token` raises 401 on a token the handler happily minted. There is also an end-to-end case driving `/login`-style flow through the real dependency, as you asked.

## 2. API-key-only profile lost renewal for co-sent guest tokens

Confirmed, and it is worse than a missing convenience. I checked the two premises:

- The WebUI sends both headers unconditionally — [`lightrag.ts:427-430`](https://github.com/HKUDS/LightRAG/blob/8aaa22b2/lightrag_webui/src/api/lightrag.ts#L427-L430).
- `/auth-status` returns a guest token whenever `not auth_handler.accounts`, which **includes** the API-key-only profile — [`lightrag_server.py:2352`](https://github.com/HKUDS/LightRAG/blob/8aaa22b2/lightrag/api/lightrag_server.py#L2352).

So that guest token really is present in exactly the profile this advisory is about. And step 2 validates any presented token *before* the API key is examined, so a lapsed one 401s the request even when `X-API-Key` is correct. Renewal was what kept that from ever happening; removing it turned a silent background refresh into a periodic 401-then-recover. I pinned that causal step as its own test (`test_expired_token_rejects_even_with_a_correct_api_key`) rather than leaving the claim to a comment.

Renewal now also runs after the API key check succeeds, when a valid token accompanied it. Your reasoning that this does not reintroduce the advisory holds — it is strictly post-authorization — and `test_wrong_api_key_with_valid_token_still_writes_nothing` pins it: a valid token plus a *wrong* key is still 403 with an empty cache. `token_info` is kept in scope from step 2 for this, and is `None` unless validation actually succeeded, so the only path reaching the new call with a token is API-key-only + valid guest token.

## Verification

```
tests/api/auth                                     83 passed
tests/api                                         585 passed, 14 skipped
same file against parent commit c8ea584f:           5 failed, 24 passed
```

The 5 failures are the new fix-proofs, all behavioral (`DID NOT RAISE`, 401 on a self-minted token, missing `X-New-Token`) rather than `AttributeError` on new symbols. Ruff, `ruff format` and `pre-commit run --files` clean.

PR #3536, opened 2026-08-01
feature: add NebulaGraph(v5.0+) as graph database storage

xlshao2021:
## Description

Add NebulaGraph database v5.0+ as graph database storage

## Changes Made

Add the complete NebulaGraph implementation in nebula_impl.py, along with the necessary configuration updates to support the new storage backend.

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed
- [x ] Unit tests added (if applicable)

## Additional Notes

This PR introduces NebulaStorage as a new graph storage backend for LightRAG. Building on existing support for Neo4j, Memgraph, etc, this addition brings a high-performance, distributed graph storage solution purpose-built for large-scale enterprise deployments.

### DIFF
--- a/env.example
+++ b/env.example
@@ -1254,6 +1254,12 @@ MEMGRAPH_DATABASE=memgraph
 ### DB specific workspace should not be set, keep for compatible only
 # MEMGRAPH_WORKSPACE=forced_workspace_name
 
+### NebulaGraph Configuration
+NEBULA_GRAPH_HOSTS=192.168.8.125:9669
+NEBULA_USER=root
+NEBULA_PASSWORD=passwd
+NEBULA_GRAPH=lightrag
+
 ###########################################################
 ### Langfuse Observability Configuration
 ### Only works with LLM provided by OpenAI compatible API
--- a/lightrag/kg/__init__.py
+++ b/lightrag/kg/__init__.py
@@ -17,6 +17,7 @@ STORAGE_IMPLEMENTATIONS = {
             "MongoGraphStorage",
             "MemgraphStorage",
             "OpenSearchGraphStorage",
+            "NebulaStorage",
         ],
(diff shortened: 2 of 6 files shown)


danielaskdd:
Thanks for putting this together, @xlshao2021 — a full NebulaGraph implementation alongside the existing Neo4j/Memgraph graph backends is a meaningful contribution for folks who need a distributed, large-scale graph store.

I want to be upfront about why this is landing on `dev-nebula-graph` instead of going straight into `main`, and what happens from here.

LightRAG already supports 10+ storage backends across KV/Vector/Graph/DocStatus, and each one carries real recurring cost: every release needs end-to-end testing across all of them, and every storage-interface or schema change (which happen often as the project moves fast) has to be re-adapted, performance-tuned, and migration-tested for every backend shipped in `main`. Adding another backend directly to `main` multiplies that cost on every future change — that's a bigger tradeoff for the project's pace than the merits of any individual PR, and this PR's quality isn't in question here.

At the same time, NebulaGraph is clearly a backend people want for enterprise-scale deployments, so we don't want to just close this out. Instead we're setting up `dev-nebula-graph` as a longer-lived integration branch — a shared place where you and anyone else interested can keep contributing, keep the backend in sync as core interfaces evolve, and build up real-world usage. Once the branch shows sustained activity and maturity (active maintenance, adoption, staying current with upstream changes), we'll revisit merging it into `main` as a first-class backend.

Practically: this PR merges into `dev-nebula-graph` (base is already set correctly), not `main`. Keeping it in sync with future storage-interface changes on `main` is on the community going forward — we're glad to help and review contributions there, but core maintainers won't be carrying NebulaGraph-specific migration work as part of every `main` storage change.

Merging this into `dev-nebula-graph` now — thanks again for the contribution.

PR #3537, opened 2026-08-01
[BUG] handle vector count doubling in embedding validation (Fixes #2549)

### LINKED ISSUE #2549: [Bug]: Bug in utils.py

Rythamo8055:
Fixes #2549. Handles vector count doubling from multimodal/mineru pipelines by slicing result to match expected count instead of raising ValueError.

### DIFF
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -667,10 +667,25 @@ class EmbeddingFunc:
         if args and isinstance(args[0], (list, tuple)):
             expected_vectors = len(args[0])
             if actual_vectors != expected_vectors:
-                raise ValueError(
-                    f"Vector count mismatch: "
-                    f"expected {expected_vectors} vectors but got {actual_vectors} vectors (from embedding result)."
-                )
+                # Handle vector count doubling from multimodal/mineru pipelines
+                if actual_vectors == 2 * expected_vectors:
+                    logger.warning(
+                        f"Vector count doubling detected "
+                        f"({actual_vectors} vs {expected_vectors}). "
+                        f"Slicing result to match input."
+                    )
+                    # Reshape if it's a 2D array, slice to expected count
+                    if len(result.shape) > 1:
+                        result = result[:expected_vectors]
+                    else:
+                        new_size = expected_vectors * expected_dim
+                        result = result[:new_size]
+                else:
+                    raise ValueError(
+                        f"Vector count mismatch: "
+                        f"expected {expected_vectors} vectors but got "
+                        f"{actual_vectors} vectors (from embedding result)."
+                    )
 
         return result
 


danielaskdd:
@Rythamo8055 Thanks for the PR. I do not think this should be merged in its current form.

`actual_vectors == 2 * expected_vectors` is not evidence that the provider returned safely-discardable extra vectors. The current value is derived from `result.size // embedding_dim`, not from `result.shape[0]`. For example, a valid batch-shaped result of `(N, 2D)` caused by an output-dimension/configuration mismatch is reported as `2N` vectors. In that case this PR's 2D branch, `result[:N]`, leaves the `2D` width unchanged and merely moves the failure to the vector store. Conversely, if the result really is `(2N, D)`, there is no contract that says the first N rows correspond to the N inputs: the additional rows can be segmented, interleaved, reordered, or represent another modality/encoding. Truncating them can silently store an embedding under the wrong record.

The generic `EmbeddingFunc` contract should remain fail-fast for every mismatch, including exactly 2x. It should validate the actual batch shape, not only the total number of elements:

- `result.ndim == 2`
- `result.shape[0] == len(texts)`
- `result.shape[1] == embedding_dim`

On failure, log/raise the model, configured dimension, actual shape, input count, and any provider response indices. This makes the likely `(N, 2D)` configuration problem diagnosable instead of masking it.

If a specific provider legitimately emits multiple results per source, handle it in a dedicated provider adapter with an explicit, documented mapping (`source_id`, `chunk_index`, and/or `kind`). The adapter must either select the declared primary vector or materialize the mapped chunks before returning a canonical `(N, D)` batch. That is an explicit, observable projection—not a generic `[:N]` recovery. Token-level/multi-vector outputs likewise require a multi-vector retrieval path.

Please remove the special 2x slicing fallback and add regression tests for both `(N, 2D)` and `(2N, D)`; both must fail at the generic wrapper unless a provider-specific adapter has already normalized the response.

PR #3543, opened 2026-08-02
feat(utils): bound thread-pool submission and pin the tokenizer contract

danielaskdd:
## Summary

Preparatory infrastructure for moving CPU-bound token counting (query side) and chunking (ingestion side) off the asyncio event loop. Both follow-ups need the same submission primitive, and neither depends on the other, so it lands here first.

Two parts:

1. **The tokenizer thread-safety contract**, stated where implementers will read it. Responsibility for concurrent calls sits with the injected tokenizer; LightRAG does not serialize it.
2. **`bounded_submit`** plus the per-event-loop semaphore machinery both follow-ups submit through.

No behavioural change on its own: the submission helper has no callers yet, and the contract makes explicit what the built-in tokenizer already satisfies.

> **Revised after review.** The first version of this PR put a LightRAG-owned lock on `Tokenizer.encode`/`decode`, with a tier system, a strict-mode gate and an event-loop guard. That was the wrong layer — see [What changed after review](#what-changed-after-review) at the bottom.

## Motivation

Token counting is CPU-bound and roughly linear in caller-chosen input (~0.5 s/MiB), so doing it inline in an `async def` stops the process answering anything at all for the duration, `/health` included. The fix is to run it in worker threads — which makes concurrent calls into one tokenizer routine, and raises the question of who is responsible for them.

Separately, freeing the event loop lets more requests be in flight at once while a `ThreadPoolExecutor`'s wait queue stays unbounded, so submissions need their own ceiling. Getting that wrong is easy and the failure mode is silent, so it is introduced here with its own tests rather than buried inside a larger change.

### The tokenizer thread-safety contract

`TokenizerInterface.encode()` / `decode()` may be called concurrently from several OS threads. **An injected tokenizer is responsible for its own thread safety** — immutable state, an internal lock, thread-local state, whatever its implementation needs.

LightRAG deliberately does not serialize it. A lock owned here would be acquired by a worker thread and then *waited on by the event loop*, which is precisely the freeze the offload exists to remove; it would also penalise an implementation that is already thread-safe. The right layer is the extension point.

The built-in `TiktokenTokenizer` is the only tokenizer LightRAG ships, and concurrent use is a supported capability of tiktoken rather than an assumption made here: `encode_batch` / `decode_batch` fan `Encoding.encode` / `.decode` of a **single instance** across a `ThreadPoolExecutor` ([`core.py`](https://github.com/openai/tiktoken/blob/0.12.0/tiktoken/core.py#L174-L205)).

**Copying was never an alternative to thread safety**, and the contract says so because the opposite is easy to assume. `tiktoken` caches encodings in a process-wide registry, and `Encoding.__setstate__` rebinds a copy's `__dict__` to the registered instance's — so every `TiktokenTokenizer` in the process, deep copies included, already funnels into one `CoreBPE`:

```python
>>> a, b = TiktokenTokenizer(), TiktokenTokenizer()
>>> a.tokenizer is b.tokenizer
True
>>> c = copy.deepcopy(a)
>>> c.tokenizer is a.tokenizer, vars(c.tokenizer)["_core_bpe"] is vars(a.tokenizer)["_core_bpe"]
(False, True)          # new wrapper, same engine
```

### Dependency floor

`tiktoken` was declared without a lower bound. Since the contract now rests on a property of that library, the floor is pinned to **`>=0.7.0`** — the oldest release whose `encode_batch`/`decode_batch` fan a single `Encoding` across threads, and also the oldest carrying `o200k_base`, which the default `gpt-4o-mini` needs. Verified by reading the source of 0.5.2 / 0.6.0 / 0.7.0 / 0.8.0 / 0.12.0 / 0.13.0, not from memory.

### Deep-copy requirement, and one line that makes it cheap

`LightRAG` is a dataclass and `_build_global_config` runs `dataclasses.asdict` over it, which deep-copies non-dataclass fields. An implementation that guards itself with a `threading.Lock` would fail at construction with `TypeError: cannot pickle '_thread.lock' object`. The contract therefore also asks that an injected tokenizer survive `copy.deepcopy`, and points lock-based implementations at `__deepcopy__` returning `self` — sound precisely because being thread-safe is what makes it shareable.

`_build_global_config` now restores the tokenizer object over the copy `asdict` made, the same one-liner already used for `embedding_func`. Consumers reading `global_config["tokenizer"]` get the instance itself, no operation pays for a copy, and there is one tokenizer object to reason about instead of a per-operation copy that was never independent anyway.

### Bounded submission

- `bounded_submit(executor, semaphore, fn, ...)` — **the permit belongs to the executor future**, not the awaiting coroutine. `async with sem: await run_in_executor(...)` returns the permit on cancellation while the thread keeps running (an already-running `concurrent.futures.Future.cancel()` returns `False`), so submit-and-cancel would rebuild an unbounded queue.
- Semaphores are per event loop, because `asyncio.Semaphore` binds to the first loop that waits on it. They are stored **on the loop** rather than in a container keyed by it: once contended the semaphore strongly references its loop, so a `WeakKeyDictionary` entry whose value references its key never expires. Loops that reject attribute assignment fall back to an id-keyed table that holds the loop itself — deriving it from the semaphore fails for an uncontended one, which never learns which loop it belongs to.
- `TOKENIZER_SUBMIT_LIMIT` / `CHUNKING_SUBMIT_LIMIT` are fixed process-wide constants, not derived from `max_parallel_insert`: the helpers are module level and are called from places holding no `LightRAG` instance, several instances sharing a loop would each build their own semaphore and lose the global ceiling, and the pools have one worker, so queue depth beyond single digits buys no throughput. Over the limit the submitting coroutine waits — backpressure, not refusal.

Nothing is broken for the built-in path.

The one user-visible addition is the **injection contract** for custom tokenizers, documented in [`docs/ProgramingWithCore.md`](docs/ProgramingWithCore.md#injecting-a-custom-tokenizer) with a worked example: be thread-safe, and survive `copy.deepcopy`. A lock-based implementation needs `__deepcopy__` returning `self` — which is already true today, since `asdict` already deep-copies the field.

The `tiktoken>=0.7.0` floor is not a new practical constraint: the default `gpt-4o-mini` has always needed `o200k_base`, which no earlier release ships.

## Follow-up ordering

The two follow-ups are no longer freely orderable. **Query-side migration should merge before ingestion-side chunking.** For a custom tokenizer that is thread-safe *by internal lock*, either single-sided state leaves a window where the event loop can wait on that lock — but the windows are very different in size: an ingestion worker holds it for as long as chunking a whole document takes, while a query worker holds it for one truncation. Doing the query side first keeps the exposure to milliseconds. (Irrelevant to `tiktoken`, which has no such lock.)

## Tests

19 offline tests. Three are load-bearing and were each verified to fail against the wrong implementation:

- `test_concurrent_callers_are_not_serialized_by_the_wrapper` — two threads must genuinely overlap inside one tokenizer. Re-adding a LightRAG-owned lock drops observed concurrency from 2 to 1.
- `test_cancelling_a_submission_does_not_return_the_permit_early` — fails when the permit is released by the coroutine (observed: 2 submissions instead of 1).
- `test_a_contended_semaphore_does_not_retain_its_closed_loop` — fails under `WeakKeyDictionary[loop] -> Semaphore` (observed: closed loop still reachable after `gc.collect()`). Forcing contention is mandatory; `Semaphore.acquire()` only records `_loop` when it has to wait, so without it the leak hides.

Also pinned: concurrent `encode`/`decode` correctness on one `TiktokenTokenizer` (the property the contract now rests on, so a tiktoken upgrade that broke it fails here rather than as corrupted counts); the registry/`__setstate__` sharing that makes copying no defence; `_build_global_config` handing out the tokenizer itself; and both sides of the deep-copy boundary.

Full suite: **5304 passed, 155 skipped**. `pre-commit run ruff` / `ruff-format` clean.

## What changed after review

Review pointed out that `TiktokenTokenizer` is the only built-in implementation and that tiktoken is thread-safe, so the responsibility belongs on the extension point. Agreed — and investigating it turned up something stronger than a simplification argument.

The old design keyed its lock on the *underlying* tokenizer object. Because `tiktoken` caches encodings process-wide, every `TiktokenTokenizer` — LightRAG's, the Ollama route's, the reranker's, and every `asdict` copy — resolves to the same `Encoding`. So the old tier-1 policy would have put **one process-wide lock on the default path**, creating a global serialization point that does not exist today, and arming the event-loop gate against the built-in tokenizer.

The second branch of the old contract, *"deep-copy into fully independent state"*, was also not true of the only implementation that was said to satisfy it: the copy shares `_core_bpe`. A test asserting the wrapper differed passed while the property it was named for did not hold.

Removed: the identity-keyed lock registry and its weakref/id reclamation, the `lightrag_thread_safe` marker and three-tier policy, `TokenizerLoopBlockError`, `LIGHTRAG_STRICT_TOKENIZER_OFFLOAD`, and the serialization test file (~350 lines). Kept: `bounded_submit`, per-loop semaphores, permit ownership, and the submission ceilings.

Two of the boundaries raised alongside the proposal are handled directly rather than passed to users: the deep-copy interaction is resolved for the hot path by restoring the tokenizer in `_build_global_config` (and documented, with the `__deepcopy__` escape hatch, for the construction path that remains), and the version floor is pinned from source rather than left to a regression test alone. The third — merge ordering — is recorded above.

### DIFF
--- a/docs/ProgramingWithCore.md
+++ b/docs/ProgramingWithCore.md
@@ -75,7 +75,7 @@ Notes:
 | **doc_status_storage** | `str` | Storage type for documents process status. Supported types: `JsonDocStatusStorage`,`PGDocStatusStorage`,`MongoDocStatusStorage`,`OpenSearchDocStatusStorage` | `JsonDocStatusStorage` |
 | **chunk_token_size** | `int` | Maximum token size per chunk when splitting documents | `1200` |
 | **chunk_overlap_token_size** | `int` | Overlap token size between two chunks when splitting documents | `100` |
-| **tokenizer** | `Tokenizer` | The function used to convert text into tokens (numbers) and back using .encode() and .decode() functions following `TokenizerInterface` protocol. If you don't specify one, it will use the default Tiktoken tokenizer. | `TiktokenTokenizer` |
+| **tokenizer** | `Tokenizer` | The function used to convert text into tokens (numbers) and back using .encode() and .decode() functions following `TokenizerInterface` protocol. If you don't specify one, it will use the default Tiktoken tokenizer. An injected tokenizer must be safe to call concurrently from multiple threads and must survive `copy.deepcopy` — see [Injecting a custom tokenizer](#injecting-a-custom-tokenizer). | `TiktokenTokenizer` |
 | **tiktoken_model_name** | `str` | If you're using the default Tiktoken tokenizer, this is the name of the specific Tiktoken model to use. This setting is ignored if you provide your own tokenizer. | `gpt-4o-mini` |
 | **entity_extract_max_gleaning** | `int` | Number of loops in the entity extraction process, appending history messages | `1` |
 | **node_embedding_algorithm** | `str` | Algorithm for node embedding (currently not used) | `node2vec` |
@@ -567,6 +567,45 @@ To enhance retrieval quality, documents can be re-ranked based on a more effecti
 - **Aliyun**: `ali_rerank`
 
 Inject one of these functions into the `rerank_model_func` attribute of the LightRAG object. For detailed usage, refer to `examples/rerank_example.py`.
+
+### Injecting a Custom Tokenizer
+
+Any object with `encode(str) -> list[int]` and `decode(list[int]) -> str` can be
+wrapped in `Tokenizer` and passed as `tokenizer=`. Two requirements apply:
+
+1. **It must be safe to call concurrently from multiple threads.** Token counting
+   is CPU-bound, so LightRAG runs it in worker threads to keep the asyncio event
+   loop responsive; several of them may enter your `encode`/`decode` at once.
+   LightRAG deliberately does not serialize calls on your behalf — a lock owned
+   by LightRAG would end up being waited on by the event loop behind a worker
+   thread, which is exactly the stall the threading is there to avoid.
+2. **It must survive `copy.deepcopy`.** `LightRAG` is a dataclass and builds its
+   internal config with `dataclasses.asdict`, which deep-copies non-dataclass
+   fields.
+
+The two interact: if you achieve thread safety with an internal `threading.Lock`,
+deep-copying it raises `TypeError: cannot pickle '_thread.lock' object`. Declare
+`__deepcopy__` returning `self`, which is sound precisely because a thread-safe
+tokenizer is safe to share:
+
+```python
+class MyTokenizer:
+    def __init__(self):
+        self._lock = threading.Lock()
+
+    def __deepcopy__(self, memo):
+        return self  # thread-safe, therefore shareable
+
+    def encode(self, content: str) -> list[int]: ...
+    def decode(self, tokens: list[int]) -> str: ...
+
+rag = LightRAG(..., tokenizer=Tokenizer("my-model", MyTokenizer()))
+```
+
+The built-in `TiktokenTokenizer` satisfies both. Note that copying it is not a
+way to get isolation: `tiktoken` caches encodings in a process-wide registry, so
+every `TiktokenTokenizer` for a given model — copies included — resolves to the
+same underlying BPE engine.
 
 ### User Prompt vs. Query
 
--- a/lightrag/constants.py
+++ b/lightrag/constants.py
@@ -447,6 +447,20 @@ DEFAULT_MAX_TEXTS_PER_REQUEST = 0
 # body that lies about (or omits) Content-Length is still cut off.
 DEFAULT_MAX_REQUEST_BODY_BYTES = 0
 
+# Submission ceilings for the two single-worker CPU pools (tokenizer, chunking).
+# A ``ThreadPoolExecutor`` wait queue is unbounded, and freeing the event loop
+# means more requests can be in flight at once, so submissions need their own
+# limit. Deliberately fixed process-wide constants rather than anything derived
+# from ``max_parallel_insert``: the submitting helpers are module level and are
+# called from places holding no ``LightRAG`` instance, several instances sharing
+# one event loop would each build their own semaphore and lose the global
+# ceiling, and — the substantive reason — the pools have ONE worker, so queue
+# depth beyond single digits buys no throughput and only pins more pending data
+# in memory. Over the limit the submitting coroutine waits; it is backpressure,
+# not refusal.
+TOKENIZER_SUBMIT_LIMIT = 8
+CHUNKING_SUBMIT_LIMIT = 8
+
 # Per-workspace ceiling on manual retry requests that have been published but
 # not yet ACKed (LR2 §10.1). The channel is sticky — a request survives until an
 # exclusive reset acknowledges it — so an operator hammering /reprocess_failed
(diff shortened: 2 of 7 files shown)


danielaskdd:
Agreed, and adopted — force-pushed. Thanks, this was the right call.

Verifying it turned up something stronger than the simplification argument, so I want to record it rather than just say "done".

**The old lock would have serialized the built-in path process-wide.** It keyed on the *underlying* tokenizer object, and `tiktoken` caches encodings in a process-wide registry:

```python
>>> TiktokenTokenizer().tokenizer is TiktokenTokenizer().tokenizer
True
```

So LightRAG's tokenizer, the Ollama route's, the reranker's, and every `asdict` copy all resolve to one `Encoding` — one lock for all of them. The design meant to prevent contention would have manufactured it, on the default path, and then armed the strict gate against the built-in tokenizer.

**The contract's other branch was factually false for the only implementation said to satisfy it.** `Encoding.__getstate__` returns just the encoding name for a registered encoding and `__setstate__` does `self.__dict__ = registry.get_encoding(name).__dict__`, so a deep copy is a new wrapper over the *same* engine:

```python
>>> c = copy.deepcopy(TiktokenTokenizer())
>>> c.tokenizer is t.tokenizer, vars(c.tokenizer)["_core_bpe"] is vars(t.tokenizer)["_core_bpe"]
(False, True)
```

"Deep-copy into fully independent state" was never something LightRAG had — and a test named for it passed, because it asserted the wrapper differed. That is now pinned the other way round, so the sharing is documented rather than assumed away.

---

On your three boundaries:

**1. Thread-safe ≠ event-loop-safe.** Agreed, no change to the offload work. `bounded_submit`, the per-loop semaphores, permit ownership and the ceilings are all kept as-is.

**2. Deep-copy.** Handled rather than passed on. `_build_global_config` now restores the tokenizer over the copy `asdict` made — the one-liner already used for `embedding_func` — so the per-operation copy is gone (it was never isolation anyway, per the above) and one object is shared. The construction-time `asdict` still copies the field once, so the contract does state the deep-copy requirement and points lock-based implementations at `__deepcopy__` returning `self`, with a worked example in `docs/ProgramingWithCore.md`. Both sides are pinned: a lock-based tokenizer with `__deepcopy__` constructs fine, one without still raises.

**3. Ordering.** Agreed on query-side first, though I'd put the reasoning slightly differently: *either* single-sided state leaves a window, since after query-only migration an ingestion encode on the loop can wait on a lock a query worker holds. What makes query-first correct is the asymmetry in hold time — an ingestion worker holds it for a whole document's chunking, a query worker for one truncation. Recorded in the PR description.

**Version floor.** Set to `tiktoken>=0.7.0`, verified from source rather than memory: I read `core.py`/`model.py` from the 0.5.2, 0.6.0, 0.7.0, 0.8.0, 0.12.0 and 0.13.0 sdists. `encode_batch`/`decode_batch` have fanned a single `Encoding` across a `ThreadPoolExecutor` since well before 0.7.0, but `o200k_base` — which the default `gpt-4o-mini` needs — first appears in 0.7.0, so that is the real floor on both counts and not a new practical constraint. The concurrency regression test you suggested is in as well, so a future release that broke the property fails there rather than as corrupted token counts.

**Removed:** the identity-keyed lock registry and its weakref/id reclamation, `lightrag_thread_safe` and the three-tier policy, `TokenizerLoopBlockError`, `LIGHTRAG_STRICT_TOKENIZER_OFFLOAD`, and the serialization test file — about 350 lines net.

Full suite: 5304 passed, 155 skipped.

PR #3544, opened 2026-08-02
fix(api): bound request bodies and take token counting off the event loop

danielaskdd:
## Summary

Closes the query-side denial of service: an unauthenticated `POST /api/chat` accepts a body of any size and tokenizes it synchronously on the event loop, so the process answers nothing — `/health` included — for as long as that takes. Three independent boundaries were missing on the same path, and this PR adds all three: raw request body limits, input field limits, and moving the CPU-bound tokenizing into a worker thread.

Builds on #3543, which provides `bounded_submit` and the tokenizer contract.

## Motivation

Reported privately as a high-severity advisory. Measured against the pre-fix server:

| Attack | Observed |
|---|---|
| 64 MiB body to `/api/chat` | 65.7 s with no response to anything; 30 of 32 `/health` probes timed out |
| 192 MiB body | RSS climbs to ~2 GB and does not come back |
| 8 MiB authenticated `/query` | 3.6–4.4 s of full unavailability per request, every mode |

The client can disconnect immediately after sending; the server has no way to notice. Three gaps combine:

1. **No field limits.** `OllamaMessage.content` is a bare `str` and `messages` is unbounded; `QueryRequest` has `min_length=3` but no maximum, and `top_k` / `max_total_tokens` have `ge=1` with no upper bound.
2. **No route body limit.** `MAX_REQUEST_BODY_BYTES` reads like a global cap but only covered the three `ADMISSION_PATHS` ingestion routes, and defaulted to `0` (off).
3. **Tokenizing on the loop.** `estimate_tokens` plus every `tokenizer.encode` on the query path in `operate.py` / `utils.py` / `rerank.py`.

### Body limits, in their own middleware

New `BodyLimitMiddleware`, deliberately outside `AdmissionMiddleware` so an oversized body is rejected before a capacity slot is even reserved. Pure ASGI: on a declared-oversize `Content-Length` it answers 413 **without ever calling `receive()`**, so the body is never invited (no `100-continue`) and never buffered. For a body that lies about or omits its length, the receive wrapper counts as it streams and raises before handing the caller a truncated-but-plausible body.

Limits are tiered by route, resolved once at startup:

| `MAX_REQUEST_BODY_BYTES` | ordinary routes | `/documents/text(s)` | `/documents/upload` |
|---|---|---|---|
| unset | 1 MiB | 50 MiB | `MAX_UPLOAD_SIZE` + 1 MiB |
| `N > 0`, configured | N | N (operator intent wins) | `MAX_UPLOAD_SIZE` + 1 MiB |
| `0` | off | off | off |
| — | — | — | off when `MAX_UPLOAD_SIZE` is `0`/unset |

Whether the variable was *configured* is recorded at parse time, never inferred by comparing the value to the default — those are indistinguishable when an operator sets exactly 1 MiB on purpose, and the baseline applied such a value to `/documents/text(s)` too. Inferring it would have widened a configured ceiling fifty-fold on upgrade.

The ingestion tier is a built-in constant rather than a new env knob: batch insert has no upload equivalent to fall back on, so it needs headroom, but that is not a decision each deployment should have to make. `MAX_UPLOAD_SIZE=0` keeps its documented "no limit" meaning — the upload route is authenticated, so that is an operator's explicit choice — and now says so in a startup warning instead of silently.

`AdmissionMiddleware` loses its body handling entirely. Its `finally` ticket release is kept and now has a regression test, because `BodyLimitExceeded` from the outer middleware travels up through it.

### Input field limits

Caps on `/api/chat`, `/api/generate` and `/query*`: per-message and per-query character limits, message counts, keyword list sizes, model and response-type names, image counts, plus upper bounds on `top_k` / `chunk_top_k` and the `max_*_tokens` budgets — character limits alone do not stop `top_k=10**6`.

Two of these are load-bearing rather than cosmetic:

- **An aggregate limit**, because per-field caps compose: 32 messages × 32 KiB reconstructs the payload.
- **A serialized-size check on `conversation_history`.** That field is `List[Dict[str, Any]]` and the existing validator only checks that `role` is a non-empty `str` — it bounds neither field and ignores extra keys entirely, so the payload just moves into `role`, or into a key nobody enumerated. `len(json.dumps(...))` is the only check that is complete. Extra keys stay *allowed* (clients legitimately send OpenAI's `name` / `tool_call_id`); `extra="forbid"` would have been a gratuitous break.

Which of these is a security boundary is written into the constants block, because it is easy to assume they substitute for each other. Only the **volume** limits are: cost is proportional to bytes, and nothing on the path is superlinear per message. `MAX_MESSAGES_PER_REQUEST = 128` is a sanity guard that buys no security the volume cap does not already provide — it is set high enough not to cut off Open WebUI-style clients that send a full history.

### Token counting off the event loop

A single-worker executor plus `acount_tokens` / `aencode` / `atruncate_list_by_token_size`, with all query-path call sites migrated. One worker preserves today's concurrency exactly (the loop already serialized this work) rather than introducing parallelism a security fix has no business introducing. The synchronous helpers stay, so the ingestion path is untouched.

**Each request submits O(1) tasks.** `atruncate_list_by_token_size` and `achunk_documents_for_rerank` submit their whole loop as one task, never per element — per-element submission would make queue depth `N × in-flight requests` and turn a bounded problem into an amplifying one. This is pinned by a submission-counting test.

Two call sites worth naming:

- `process_chunks_unified` — the largest single blocker on the query path, an unyielding loop that JSON-encodes every chunk.
- `_attach_content_headings` — reached through `text_chunks_db.global_config["tokenizer"]`, which storages capture once at init, so it is one object shared by every concurrent query. Leaving it on the loop meant two concurrent queries could each freeze the other's requests; single-query load never shows this.

Two debug-only `encode` blocks became `logger.isEnabledFor`-guarded — they were doing the work regardless of log level.

## What's broken

- **`MAX_REQUEST_BODY_BYTES` now defaults to on.** Ordinary API requests over 1 MiB get 413.
- **New hard input limits** on `/api/chat`, `/api/generate` and `/query*`, including upper bounds on `top_k`, `chunk_top_k` and the `max_*_tokens` budgets.
- Status codes differ by router and that is deliberate: `/query*` uses FastAPI's native validation (422), `/api/*` handles its own parsing and returns 413. `validation_exception_handler` is already specialised for `/query/data`'s 400 contract; changing it for cosmetic uniformity is not worth the risk.

Startup now warns when body limits are disabled, when `MAX_UPLOAD_SIZE` leaves the upload route uncapped, and when a value raised for uploads is now also widening ordinary API routes. `MAX_UPLOAD_SIZE`, `MAX_REQUEST_BODY_BYTES` and `MAX_PENDING_DOCUMENTS` are shown in the startup banner, where none of them appeared before.

## Tests

Load-bearing cases, each verified to fail against the wrong implementation:

- **Rejected before tokenizing, not merely rejected.** `estimate_tokens`, `aestimate_tokens` and `rag.aquery` are patched to raise; an oversized request must return 413 with none of them touched.
- **The `role` bypass.** 32 history entries with entirely legal `content` and the payload in `role` — passes validation under a `content`-only implementation. Plus a variant with the payload in an unenumerated extra key, and a positive case proving legitimate extra keys still work.
- **`receive()` is never called** on a declared-oversize body, and a lying `Content-Length` is still caught mid-stream.
- **O(1) submissions per request**: a 1000-element truncation submits exactly 1 task.
- **Concurrent queries keep the loop alive** — two overlapping heading backfills with a heartbeat counter; 0 beats when `_attach_content_headings` runs inline.
- **Admission tickets are released** when the outer body limit fires.

Full suite: **5367 passed, 155 skipped**. `pre-commit run ruff` / `ruff-format` clean.

## Merge order

Should merge **before** the ingestion-side chunking PR. For a custom tokenizer that is thread-safe by internal lock, whichever side migrates first leaves a window where the loop can wait on that lock — but an ingestion worker holds it for a whole document's chunking, while a query worker holds it for one truncation. Doing the query side first keeps the exposure to milliseconds. (Irrelevant to the built-in `tiktoken`, which has no such lock.)

PoC code and exploitation details are deliberately omitted; the advisory will be linked once published.


danielaskdd:
Fixed and force-pushed (`c401d639` → `0b88d0ff`). You're right, and the direction of the failure is the bad one.

**Root cause.** `resolve_body_limits()` reconstructed "was this configured?" from the number, but `config.py` had already collapsed the provenance: `get_env_value("MAX_REQUEST_BODY_BYTES", DEFAULT_MAX_REQUEST_BODY_BYTES, int)` returns an `int` either way. The value can never answer who wrote it.

So a deployment with an explicit `MAX_REQUEST_BODY_BYTES=1048576` — which the baseline applies to `/documents/text` and `/documents/texts` via `AdmissionMiddleware` — would get 50 MiB on those two routes after upgrading. A configured security boundary, silently relaxed fifty-fold, with nothing in the logs.

**Fix.** Provenance is recorded at parse time instead of inferred:

```python
_configured_body_limit = get_env_value("MAX_REQUEST_BODY_BYTES", None, int)
args.max_request_body_bytes_explicit = _configured_body_limit is not None
args.max_request_body_bytes = (
    DEFAULT_MAX_REQUEST_BODY_BYTES
    if _configured_body_limit is None
    else _configured_body_limit
)
```

Using `None` as the default also settles a boundary case for free: an unparseable value falls back here, and falling back to the default *value* should mean falling back to the default *tiering* too — not pinning the ingestion routes to a number nobody chose.

`resolve_body_limits()` now reads the flag. When `args` is hand-built and the flag is absent it resolves to **configured**, i.e. the tighter reading: guessing wrong that way costs a visible 413 on a large ingest, while guessing wrong the other way silently relaxes a ceiling, which is the failure this parameter exists to prevent. That is stated in the docstring and pinned by a test.

**One more instance of the same trap, from the same review.** `env.example` ships the sample line commented out as `# MAX_REQUEST_BODY_BYTES=1048576`. Under the old logic uncommenting it verbatim was a no-op; under the correct logic it drops `/documents/text(s)` from 50 MiB to 1 MiB. That is now called out explicitly in `env.example`, `env.docker-compose-full`, and both API server docs, along with replacing "setting it to any **other** value" — which is precisely the wording that encoded the bug — with "any positive value".

**Tests.** Four added; the two in `test_body_limit_middleware.py` were verified to fail against the previous implementation (`2 failed, 20 passed`):

- `test_an_explicit_value_equal_to_the_default_still_governs_ingestion` — the regression you asked for.
- `test_missing_provenance_is_treated_as_configured` — pins the tighter default for absent provenance.
- `test_provenance_survives_a_value_equal_to_the_default` — same case at the config layer, through the real env var.
- `test_an_unparseable_value_falls_back_to_the_default_tiering`.

Full suite: **5371 passed, 155 skipped**. `pre-commit run ruff` / `ruff-format` clean.

danielaskdd:
Both fixed and force-pushed (`0b88d0ff` → `f1b02477`).

The Chinese heading is the worse of the two, for the reason you give: a heading that says the limits are off is not merely stale, it is the *specific* wrong belief that makes the commented sample line dangerous. A reader who trusts it uncomments `MAX_REQUEST_BODY_BYTES=1048576` expecting to turn something on, and instead drops `/documents/text(s)` from 50 MiB to 1 MiB — turning working bulk ingestion into 413s. Now matches the English: **服务端请求限制**（见 `env.example`）. I also dropped "入库" while I was there — the section covers `/query` and `/api/*` too, so "request" was the right scope word in the English and "入库" narrowed it.

For `document_routes.py` I did not just swap "the default" for "when explicitly disabled", because that would have left the comment arguing something weaker than the truth. The cap on the list is warranted whether or not a body ceiling exists: the ceiling bounds each factor but not their product, and the ingestion routes are the ones that allow 50 MiB. So it now reads:

```
# One request supplies both factors, and the body ceiling does not bound the
# product — the text-ingestion routes allow 50 MiB by default, and
# ``MAX_REQUEST_BODY_BYTES=0`` removes the ceiling entirely — so cap the list here.
```

Sweeping for the same class of residue turned up two more, in the limits table of `docs/FileProcessingPipeline.md` and `-zh.md`: both said `/documents/text(s)` get a built-in 50 MiB without the "while this variable is unset" qualifier, which is the same trap in a different table. Both now carry the qualifier and state that configuring any positive value — the 1 MiB default included — governs every non-upload route.

`grep` for "默认全部关闭" / "off by default" against the body-limit docs is now empty, and the two section-6 headings match across languages.

Full suite: **5371 passed, 155 skipped**. `pre-commit run ruff` / `ruff-format` clean.

Note for whoever merges: `lightrag/api/routers/document_routes.py` is also touched by #3547, which rewrites that same comment block wholesale (it no longer mentions the body ceiling at all). Expect a textual conflict on rebase, with #3547's version being the one to keep.

PR #3547, opened 2026-08-02
fix(pipeline): take chunking off the event loop and bound the separator cascade

danielaskdd:
## Summary

Closes the ingestion-side denial of service: `process_single_document` does its chunking synchronously on the event loop, so a document that is expensive to chunk holds the only thread serving HTTP for its entire duration. Reported with a 414 KiB payload plus 900 `separators`, where the `POST` returned **200 in 1.6 ms** and a following `GET /health` took **63 s** (baseline 1.5 ms).

Builds on #3543, which provides `bounded_submit` and the tokenizer contract.

## Motivation

The response is not the signal. The minute of full unavailability happens *after* the request completes, inside a background task, where neither the rate limiter nor operational monitoring can see it. The cost is `O(len(separators) × len(text))` and entirely wasted — every run of that payload produces `chunks=1`.

Three things, only the second of which the report names as the fix:

1. **`separators` is unbounded outside HTTP.** `210c5ee9` already added `max_length=64` to the request model, but the value also arrives from `CHUNK_R_SEPARATORS`, from `addon_params['chunker']['recursive_character']`, from direct SDK calls — and, most importantly, from `full_docs`, where `chunk_options` is snapshotted **at enqueue time**. A document enqueued by a pre-fix version and stuck in `PROCESSING` (an auto-resume status) still carries its 900 separators; the upgraded server reloads and re-freezes on them. That is a boot loop a restart cannot clear, and the report does not mention it. It is the same shape as `GHSA-32jh-39m7-8x84`, which the V branch already guards against via `apply_trusted_sentence_split_regex`; R had no equivalent.
2. **Chunking runs on the event loop.** For *any* sufficiently large document, not just the crafted one.
3. **No-op cascade levels re-tokenize the whole document.** Pure waste, fixed as a behaviour-preserving optimisation.

### Where the report's four branches were not the whole surface

The report lists the four synchronous branches in the `if/elif` chain (P, R, F, `chunking_func`) and treats V as the already-fixed reference. **That reference does not hold**, and fixing only those four leaves half the attack surface:

- V falls back to a **synchronous** `chunking_by_recursive_character` when `embedding_func is None` — its `await asyncio.to_thread` line is never reached. That is the R attack, reproduced under the V strategy.
- After the `await`, V's post-processing loop encodes every semantic segment and re-splits oversized ones synchronously.
- `enforce_chunk_token_limit_before_embedding` walks and encodes **all** chunks, for **every** strategy including V, after chunking is done.
- On the multimodal path, `_build_mm_chunks_from_sidecars` runs a `while True` truncate/encode loop over VLM descriptions whose length no request-side limit constrains, and `enrich_sidecars_with_surrounding` reads `CHUNK_R_SEPARATORS` directly — bypassing the R chunker and its new bound entirely.

### A single normalizer for `separators`

`normalize_r_separators()` in `lightrag/chunker/recursive_character.py`, consumed by both `chunking_by_recursive_character` and `load_chunk_separators`, so environment and persisted-snapshot values are bounded on the same path as HTTP ones. Over-long entries (>256 chars) are dropped, over-long lists (>64) truncated, both with a warning — not raised, since a misconfigured env var should not fail every document.

Three details that are easy to get wrong and are each pinned by a test:

- **`None` passes through unchanged.** `chunking_by_recursive_character` is a public chunker whose `separators=None` means "use LangChain's own cascade" — which is `['\n\n', '\n', ' ', '']`, four entries, English-only. The repo's `DEFAULT_R_SEPARATORS` is nine entries including CJK punctuation. Substituting one for the other would move split points in Chinese text; a security fix has no business changing chunking semantics.
- **Truncation preserves the terminating sentinel.** `DEFAULT_R_SEPARATORS` ends with `""`, the character-level fallback that lets `_split_text_with_spans` break anything. Naively taking the first 64 drops it, and a document that could have been split character-wise degrades into one enormous chunk. Truncation takes 63 and re-appends the sentinel — only if the caller had one, since `load_chunk_separators` deliberately omits it.
- **An empty result falls back per consumer**, not to one shared default: the chunker returns `None` (its own documented default), `load_chunk_separators` returns `DEFAULT_R_SEPARATORS` minus the sentinel (matching how it already handles invalid env values). Without a fallback, `separators[-1]` raises `IndexError` and the document simply fails.

The request model also gains the per-element cap that `210c5ee9` did not add.

### Chunking moved into a dedicated executor

Single-worker, process-wide, submitted through `bounded_submit`. One worker because chunking is *already* serialized today — while one document chunks, no other document's coroutine can run — so this preserves concurrency exactly while freeing the loop. Anything higher would be new parallelism, which belongs in a separate change.

Covered: P / R / F, V's synchronous fallback and its post-processing loop, `enforce_chunk_token_limit_before_embedding`, `backfill_chunk_sidecars`, `_build_mm_chunks_from_sidecars`, `enrich_sidecars_with_surrounding`, and the three remaining synchronous encode loops in `analyze_multimodal`, `ainsert_custom_chunks` and `ainsert_custom_kg`.

`backfill_chunk_sidecars` is the least obvious member of that list: it runs after chunking and before the storage writes, parses `blocks.jsonl`, then scans every block for every chunk — `O(chunks × blocks)` of CPU plus synchronous file I/O, which leaves the same stall a little further down the same function.

Not moved, deliberately: V's existing `await asyncio.to_thread(_semantic_groups_with_spans, ...)`. It is already off the loop and needs to call back into it for embeddings; moving it into a single-worker pool would only add deadlock surface.

The submission ceiling is not decorative: `ainsert_custom_kg` and `ainsert_custom_chunks` are public SDK entry points gated by neither `max_parallel_insert` nor the pipeline's busy flag, so "the pipeline limits concurrency" is not a ceiling that holds here.

### `chunking_func` is not moved into the executor

Its declared type is `Union[List[Dict], Awaitable[List[Dict]]]`, its docstring opens with "Synchronous or async", and the pipeline does `inspect.isawaitable(...)`. So "it must be synchronous, therefore it does not need the loop" is false: a synchronous factory that calls `asyncio.get_running_loop()` or `create_task()` at call time is a **supported** implementation, and running it in a worker thread breaks it outright with no running loop. For an `async def` implementation, threading it buys nothing anyway.

The existing identity check `self.chunking_func is chunking_by_token_size` already distinguishes the two cases: the built-in default goes to the executor, a user-supplied one keeps its exact calling convention. The docstring now states the boundary — a custom CPU-bound implementation should `to_thread` itself.

### No-op cascade levels

A separator that matches once at offset 0 yields a single fragment byte-identical to its input. The code still called `length_function` on it — a full-document `encode` — then recursed with the same text. 900 levels meant 900 full-document encodings for an output of `chunks=1`.

Detected and skipped straight to the next separator. Explicitly **not** "abandon the cascade on a no-op": text where `"\n\n"` appears only at the start and needs `"\n"` / `" "` / `""` to split is ordinary, and giving up would produce one huge chunk. Measured on the advisory payload: **65 encodes / 7.90 s → 2 encodes / 0.23 s**, byte-identical output.

## What's broken

`CHUNK_R_SEPARATORS` or `addon_params` supplying more than 64 separators, or entries longer than 256 characters, are now converged with a warning rather than used verbatim.

**Not** a breaking change, but worth stating: a custom `chunking_func` is still invoked on the event loop — the contract is unchanged — and chunking concurrency stays at 1, equivalent to today's serialized execution and not scaling with `max_parallel_insert`.

## Notes on the report

`separators` bounding at the request model landed in `210c5ee9` (2026-07-31), so that half of Fix 1 is already in. "No request-size middleware exists" is also stale — `AdmissionMiddleware` is on `main` and `ADMISSION_PATHS` includes `/documents/text`. All line numbers have drifted. What remains is Fix 2, the per-element bound, and the persisted-snapshot replay path above.

## Tests

Heartbeat fix-proofs (a `asyncio.sleep(0)` counter must keep advancing during ingestion; 0 beats before the fix) cover the paths individually, because testing only P/R/F would have missed half of them: P, R, F, V-with-`embedding_func=None`, V-with-oversized-segments, the pre-embedding hard split, the sidecar provenance backfill, and the multimodal chunk builder.

One offloaded call has no heartbeat test: `enrich_sidecars_with_surrounding`, which runs in the analyze stage rather than the process stage and would need a different harness. It is offloaded in code but guaranteed only by review, and this paragraph previously claimed otherwise — the claim is corrected rather than quietly dropped.

Also: the `chunking_func` contract (an `async def` implementation still awaited; a synchronous factory calling `get_running_loop()` at call time still works — this fails with `RuntimeError: no running event loop` under a move-everything-to-threads implementation); the persisted 900-separator snapshot replayed from `full_docs`; the cancel-and-resubmit permit test against the ungated SDK entry points; sentinel preservation driving a real split; `separators=None` producing byte-identical output; and a `length_function` call-count assertion for the cascade fix (counted, not wall-clock — wall-clock is not reliable in CI).

Full suite: **5340 passed, 155 skipped**. `pre-commit run ruff` / `ruff-format` clean.

## Merge order

Should merge **after** the query-side PR. For a custom tokenizer that is thread-safe by internal lock, moving ingestion into a worker thread first lets the event loop wait on that lock for as long as chunking a whole document takes, while the query path is still tokenizing inline. Merging the query side first reduces that window to a single truncation. (Irrelevant to the built-in `tiktoken`, which has no such lock.)

PoC code and exploitation details are deliberately omitted; the advisory will be linked once published.

---

*Supersedes #3545, which was closed and could not be reopened. Same branch, rebased onto `main` after #3546.*


danielaskdd:
All three fixed and force-pushed (`22989a07` → `4f594907`).

### P1 — the snapshot was rendered before it was bounded

Confirmed, and the durable half is worse than the logging half. `_format_chunking_params` feeds **both** the INFO line and `doc_status.metadata["chunk_opts"]`, so a 10 MB separator does not just stall the loop building the string — it pushes ~10 MB into doc_status, where it can exceed a backend's value limit and strand the document in FAILED. And the cascade the chunker went on to use was *empty*, every entry having been dropped as over-long. Cost paid, nothing gained, plus a new failure mode.

The dispatcher now normalizes before it formats, logs or splats. That is one `len()` per entry and copies no separator content, so it is safe to leave on the loop. The chunker still normalizes — it is the chokepoint for every caller that never reaches the dispatcher — and when bounding empties the list the dispatcher drops the key entirely, so the chunker takes its documented `separators=None` path instead of warning about the same emptiness twice.

This also changed an existing assertion, which is worth calling out rather than burying: `test_a_poisoned_snapshot_is_bounded_at_process_time` used to assert that the chunker *received* all 900 entries (bounding them itself). It now asserts the dispatched cascade is already ≤ 64. That is a strictly stronger guarantee, not a relaxed one.

### P1 — the sidecar backfill

Confirmed. Submitted as one coarse task — not one per chunk: the pool has a single worker and `backfill_chunk_sidecars` mutates `chunking_result` in place.

### P2 — documentation

Fixed in all three places, and the Chinese API doc turned out to be missing the paragraph entirely rather than merely stale, so it gained one.

### An inaccuracy in my own PR description, found while checking P1

The description claimed heartbeat coverage for "multimodal chunk building, and multimodal surrounding backfill". Neither test existed. Both call sites *are* offloaded in code, but the claim was about tests, and it was wrong.

`_build_mm_chunks_from_sidecars` now has one. `enrich_sidecars_with_surrounding` still does not — it runs in the analyze stage and needs a different harness — so the description now says so explicitly instead of dropping the sentence and hoping nobody re-reads it.

### Tests

Three added; two verified to fail against the previous implementation:

- `test_an_oversized_snapshot_never_reaches_the_log_or_doc_status` — asserts the persisted `chunk_opts` string stays under 4 KiB and does not contain the separator. Fails with the dispatcher normalization removed.
- `test_the_sidecar_backfill_does_not_hold_the_loop` — fails with the backfill restored to a direct call.
- `test_the_multimodal_chunk_builder_does_not_hold_the_loop`.

Both new heartbeat tests reach their branches through a shared `_with_blocks_path` helper: a raw-text document parses to `blocks_path=""`, so the sidecar-dependent branches are unreachable from a plain ingest without either injecting the path or standing up a parser that emits real artifacts.

Full suite: **5343 passed, 155 skipped**. `pre-commit run ruff` / `ruff-format` clean.

danielaskdd:
Fixed in `edfa9bd2` (a new commit, not an amend, so the delta stays reviewable).

You're right that "anything over the limit is truncated" is wrong in a way that changes what an operator does: with one 300-character separator they would expect a match on its first 256 characters, and instead the entry disappears and the split points come from a fallback.

**Writing the correction turned up a second error in the sentence I was about to write.** My first draft said "if everything is dropped, the default cascade applies". That is also wrong, and wrong in a CJK-visible way: the two consumers disagree, and neither lands on this variable's own default.

| Path | Fallback when nothing survives |
|---|---|
| R chunker (`CHUNK_R_SEPARATORS` → routing → snapshot → dispatcher) | the splitter's own **four-entry English** cascade `["\n\n", "\n", " ", ""]` |
| `load_chunk_separators` (multimodal surrounding context) | the **nine-entry** `DEFAULT_R_SEPARATORS`, minus the sentinel it strips on purpose |

Verified at runtime rather than read off the source:

```
>>> normalize_r_separators(["Z" * 300])
[]
>>> load_chunk_separators()          # CHUNK_R_SEPARATORS='["Z"*300]'
['\n\n', '\n', '。', '！', '？', '；', '，', ' ']
```

So on the R path, "every entry dropped" does **not** equal "variable unset" — unset gives the nine-entry CJK cascade, dropped gives the four-entry English one. That difference moves sentence boundaries in Chinese text, which is exactly the class of surprise the finding is about, so it is now written down rather than smoothed over.

All six files now say the same three things: over-long entry → dropped; over-long list → truncated, sentinel kept; nothing survives → consumer-specific fallback, named. The API docs link to the pipeline spec for the table; I computed the anchor slugs (`#r--recursive-character`, `#r--递归字符`) against the real headings rather than assuming them.

**One test added.** The new wording asserts the two fallbacks differ, but only the chunker half was pinned (`test_the_all_over_long_fallback_matches_separators_none`). `test_load_chunk_separators_falls_back_to_the_repo_cascade` pins the other, so a later change cannot quietly collapse them into one fallback while the docs go on claiming they differ. Drop-vs-shorten itself was already covered by `test_over_long_entries_are_dropped`.

Full suite: **5344 passed, 155 skipped**. `pre-commit run ruff` / `ruff-format` clean.

danielaskdd:
Fixed in `f57aa2c4`.

Verified before writing rather than after — both limits raise, so the HTTP path never reaches drop/truncate/fallback at all:

```
>>> RecursiveCharacterChunkParams(separators=["s%d" % i for i in range(65)])
ValidationError: too_long
>>> RecursiveCharacterChunkParams(separators=["x" * 257])
ValidationError: string_too_long
```

The old sentence listed five sources and then attributed one behaviour to all of them. It was true of four.

Rather than just striking the request body from the list, the text now says *why* the HTTP path differs, because the asymmetry looks arbitrary otherwise: the caller of an HTTP request is present to read a 422 and fix the request, while the author of an environment variable is not — failing every document over one misconfigured value would be the worse outcome. Same numeric bound everywhere, deliberately different response to breaching it.

I swept the other four files for the same conflation. Both API server documents already drew the line correctly (`A request body over the limit is rejected with HTTP 422. Every other source … is converged`). The two env templates document only their own variable, which is a non-HTTP source, so "logged at WARNING rather than rejected" is accurate there and I left them alone rather than adding an HTTP clause that does not belong in an env-var block.

Documentation only, in the two pipeline specs — no runtime or test change. `tests/chunker` plus `test_document_routes_chunking.py`: **345 passed**. `git diff --check` clean.

PR #3552, opened 2026-08-03
feat: add POST /documents/move to relocate documents across folders

cairunze3:
## Description

The WebUI lets users organise documents into folders, but there is currently no way to move an already-indexed document into a different folder. A document's folder membership is derived from its stored `file_path` (the WebUI encodes `A/B` as the `A__B__` prefix), and no API surface allows that field to be rewritten.

The only workaround today is delete + re-upload. That is unnecessarily expensive and lossy: it re-chunks, re-embeds and re-extracts the document, burning LLM/embedding budget for what is purely a metadata change, and it assigns a **new `doc_id`**, which breaks every existing reference to the document in the knowledge graph.

This PR adds a `POST /documents/move` endpoint that relocates documents by rewriting only their `file_path`. Document ids, chunks, vectors and graph data are left untouched, so **no re-indexing is required** and the knowledge graph stays intact.

## Related Issues

None — happy to open a tracking issue if maintainers prefer.

## Changes Made

All changes are confined to `lightrag/api/routers/document_routes.py`.

**New endpoint**

- `POST /documents/move` — accepts a list of `doc_ids`, a `target_folder` and a `rename_on_conflict` flag.

**New models**

- `MoveDocRequest` — validates that `doc_ids` is non-empty, contains no blank entries and has no duplicates (mirrors `DeleteDocRequest`).
- `MoveDocByIdResponse` — returns `status` (`"move_started"` / `"busy"`), `message` and `doc_ids`.

**New background task**

- `background_move_documents()` — iterates the requested documents and, for each one, rewrites the stored path via `rag.update_doc_status_fields(doc_id, {"file_path": new_file_path})`. This is the existing helper that keeps every secondary index (including the source multimap) consistent while leaving chunks/vectors/graph alone. The physical file in `INPUT_DIR` and the `__parsed__` directory is renamed on a best-effort basis.

**New helpers**

- `_encode_folder_prefix()` — encodes a `/`-separated folder path into the `__`-separated `file_path` prefix, mirroring the WebUI's `encodeFolderPrefix`.
- `_doc_basename()` — strips the folder prefix from a stored `file_path`.
- `_conflict_free_basename()` — resolves name collisions in the target folder.
- `_rename_physical_file()` — best-effort on-disk rename.

**Tests**

- `tests/api/routes/test_document_routes_move.py` — 37 offline tests covering the three pure helpers and `MoveDocRequest` validation: root/whitespace/duplicate-slash folder normalisation, prefix stripping at any nesting depth, suffix insertion before the extension (including multi-dot names like `archive.tar.gz` and extensionless names), the `rename_on_conflict=False` skip path, and rejection of empty/blank/duplicate document ids.

**Design notes**

- *Concurrency*: the endpoint reuses the exact same destructive-lock pattern as delete (`_acquire_destructive_busy` / `start_reserved_background_task` / `_release_destructive_busy`), returning `status="busy"` when another writer holds the pipeline. No new locking primitives were introduced.
- *Folder-encoding agnostic*: the backend treats `target_folder` as a plain `/`-separated path. The `__` encoding detail stays a WebUI concern, so this does not lock the server into the current scheme.
- *Conflict handling*: when a basename already exists in the target folder, `rename_on_conflict=True` (default) appends ` (1)`, ` (2)`, … before the extension; `False` skips the conflicting document. Skipped and failed documents are reported through `pipeline_status` alongside successes, so a partial move is observable rather than silent.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

**Verification performed**

- `ruff format --check` and `ruff check --ignore=E402` (using the `v0.15.17` version pinned in `.pre-commit-config.yaml`, with the repo's `pyproject.toml` config) both pass on the modified file. The `trailing-whitespace` and `end-of-file-fixer` hook conditions were verified manually — `pre-commit` itself is not installed in my environment.
- `python -m pytest tests/api` → **492 passed, 134 skipped, 1 failed**. The single failure is `test_gunicorn_worker_does_not_inject_root_path`, caused by a missing optional `uvicorn_worker` module locally. I confirmed it fails identically on a clean `origin/main` checkout, so it is unrelated to this change.

Writing the tests paid for itself: the dotfile case caught a real bug where `.env` was renamed to `.env (1).env` (an empty `rpartition` stem made the fallback duplicate the extension). That is fixed and covered.

**Open questions for reviewers**

1. **Endpoint-level tests** — the added tests cover the pure helpers and request validation. I did not add an end-to-end test of the endpoint itself because it needs a running `LightRAG` instance plus storage fixtures, and I wasn't sure which existing harness in `tests/api/` you'd prefer I extend. Happy to add one if you point me at the preferred pattern.
2. **Docs** — let me know if the API documentation should be updated as part of this PR.
3. **Destructive lock** — I reused the delete lock for safety and consistency. A move is arguably less destructive than a delete, so if a lighter-weight guard is more appropriate here, I'm glad to adjust.

The corresponding WebUI change (move dialog, API client, i18n keys) lives in a separate frontend codebase and is not included here; this PR is backend-only and the endpoint is fully usable on its own.

### DIFF
--- a/lightrag/api/routers/document_routes.py
+++ b/lightrag/api/routers/document_routes.py
@@ -4214,7 +4214,7 @@ def _encode_folder_prefix(folder_path: str) -> str:
     p = re.sub(r"/+", "/", p)
     if not p:
         return ""
-    return p.split("/").join(_MOVE_FOLDER_SEPARATOR) + _MOVE_FOLDER_SEPARATOR
+    return _MOVE_FOLDER_SEPARATOR.join(p.split("/")) + _MOVE_FOLDER_SEPARATOR
 
 
 def _doc_basename(file_path: str) -> str:
@@ -4296,9 +4296,7 @@ async def background_move_documents(
                     "latest_message": "Starting document move process",
                 }
             )
-            pipeline_status["history_messages"][:] = [
-                "Starting document move process"
-            ]
+            pipeline_status["history_messages"][:] = ["Starting document move process"]
 
         target_prefix = _encode_folder_prefix(target_folder)
         # Basenames already present in the target folder (detect collisions).
@@ -4338,9 +4336,7 @@ async def background_move_documents(
                 basename = _doc_basename(old_file_path)
                 if not basename:
                     failed_moves.append(doc_id)
-                    error_msg = (
-                        f"Cannot move document with empty file_path: {doc_id}"
-                    )
+                    error_msg = f"Cannot move document with empty file_path: {doc_id}"
                     logger.warning(error_msg)
                     async with pipeline_status_lock:
                         pipeline_status["latest_message"] = error_msg
@@ -4353,8 +4349,7 @@ async def background_move_documents(
                 if new_basename is None:
                     skipped_moves.append(doc_id)
                     skip_msg = (
-                        f"Skipped (name exists in target folder): "
-                        f"{doc_id}[{basename}]"
+                        f"Skipped (name exists in target folder): {doc_id}[{basename}]"
                     )
                     logger.info(skip_msg)
                     async with pipeline_status_lock:
@@ -4367,9 +4362,7 @@ async def background_move_documents(
                 # Rewrite the stored file_path. update_doc_status_fields keeps
                 # every secondary index (incl. the source multimap) consistent
                 # and leaves chunks/vectors/graph untouched.
-                await rag.update_doc_status_fields(
-                    doc_id, {"file_path": new_file_path}
-                )
+                await rag.update_doc_status_fields(doc_id, {"file_path": new_file_path})
 
                 # Best-effort physical rename in INPUT_DIR and __parsed__ dir.
                 _rename_physical_file(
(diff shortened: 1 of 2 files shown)


danielaskdd:
Thanks for the detailed write-up and tests. Closing this one, though — the premise doesn't hold in this codebase.

`file_path` is always normalized to a bare basename (`canonicalize_parser_hinted_basename` does `Path(file_path).name` on every write), and it has never encoded a folder path. There is no `encodeFolderPrefix` / folder-encoding scheme anywhere in `lightrag_webui` (current tree or git history) for this PR to stay compatible with.

Similarly, `INPUT_DIR` subdirectories exist solely for workspace isolation (`DocumentManager.input_dir = base_input_dir / workspace`), and `__parsed__` holds one flat sidecar directory per document's parser artifacts — neither is organized by any folder hierarchy. So there's no "moving a document between folders" to support at the storage layer; the endpoint and its physical-file-rename logic are solving a problem that doesn't exist in the current architecture.

If there's a real WebUI folder-organization feature planned, happy to look at a PR built on top of that groundwork instead.

PR #3553, opened 2026-08-03
fix(utils): return 0.0 from cosine_similarity on non-finite vectors

santhreal:
`cosine_similarity` returned NaN for NaN/Inf embedding components and emitted a divide warning, which can poison ranking.

Repro: `cosine_similarity([nan, 1], [1, 1])` was NaN on main. Treat non-finite dot products or denominators like the zero-denominator case and return `0.0`.


danielaskdd:
@codex review

PR #3554, opened 2026-08-03
fix(utils): strip whitespace before bool env conversion

santhreal:
`get_env_value(..., bool)` compared the raw string to truthy tokens. Values like `' true '` from padded env files became `False`.

Strip whitespace before the boolean membership check.


danielaskdd:
@codex review

PR #3555, opened 2026-08-03
fix(milvus): escape quotes and backslashes in filter ids

santhreal:
## Summary

Milvus filter expressions embed entity/vector ids with no escaping. An id containing `"` or `\` breaks the expression and can make get/delete miss or error.

## Problem

Filters are built by string interpolation, e.g. `id == "{id}"` and `src_id == "{entity_name}"`. Document or entity ids that include double quotes or backslashes produce invalid filter syntax (and can change the parsed expression).

## Fix

Add a small `_escape_milvus_str` helper and apply it at every filter choke point that interpolates ids or entity names (`get_by_id`, `get_by_ids`, `get_vectors_by_ids`, `delete_entity_relation`).

## Test plan

- [x] `pytest tests/kg/milvus_impl/test_milvus_escape_filter.py`


danielaskdd:
@codex review

danielaskdd:
Updated the Milvus filter-literal boundary: filter values now escape backslashes, double quotes, newlines, carriage returns, and tabs, while structured upsert data and primary-key deletes remain unchanged. Added regression coverage for both sides of this boundary; all 179 Milvus unit tests pass locally.

PR #3557, opened 2026-08-03
fix(operate): return [] when extract_entities gets no chunks

santhreal:
## Summary

`extract_entities({})` raises `ValueError: Set of Tasks/Futures is empty` because it always `asyncio.wait`s the task set, including when there are no chunks.

## Problem

After filtering or on an empty document, the pipeline can call entity extraction with an empty chunks dict. `ordered_chunks` is empty, no tasks are created, and `asyncio.wait(tasks)` fails. That aborts the extraction stage instead of producing an empty result.

## Fix

Return `[]` immediately when `ordered_chunks` is empty, before scheduling work.

## Test plan

- [x] `pytest tests/test_extract_entities_empty_chunks.py`

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -3614,6 +3614,8 @@ async def extract_entities(
     use_json_extraction = global_config.get("entity_extraction_use_json", False)
 
     ordered_chunks = list(chunks.items())
+    if not ordered_chunks:
+        return []
     # add language and example number params to prompt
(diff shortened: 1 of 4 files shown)


danielaskdd:
@codex review

santhreal:
Relocated the regression test to `tests/extraction/test_extract_entities_empty_chunks.py` to match the repository module layout.

danielaskdd:
@codex review

PR #3558, opened 2026-08-03
fix(kg): set is_truncated when get_knowledge_graph hits max_nodes

santhreal:
## Summary

`NetworkXStorage.get_knowledge_graph` can hit `max_nodes` while draining a BFS level into a local list, leave the queue empty, and still report `is_truncated=False`.

## Problem

The truncation check is `(queue and len(bfs_nodes) >= max_nodes) or has_unexplored_neighbors`. Nodes at the current depth are fully popped into `current_level_nodes` before processing. When `max_nodes` stops processing mid-level, remaining same-level nodes are not back in `queue`, so the flag stays false even though the returned subgraph is capped.

Example: star graph centered at `A` with 5 leaves, `max_nodes=3` → 3 nodes returned, `is_truncated is False`.

## Fix

Track unprocessed nodes left in the current level when the `max_nodes` break fires, and treat that as max-nodes truncation alongside a non-empty queue.

## Test plan

- [x] `pytest tests/kg/networkx_impl/test_networkx_get_kg_truncation.py`

### DIFF
--- a/lightrag/kg/networkx_impl.py
+++ b/lightrag/kg/networkx_impl.py
@@ -593,13 +593,19 @@ class NetworkXStorage(BaseGraphStorage):
 
                     # Check if we've reached max_nodes
                     if len(bfs_nodes) >= max_nodes:
-                        if idx < len(current_level_nodes) - 1:
+                        if any(
+                            n not in visited
+                            for n, _, _ in current_level_nodes[idx + 1 :]
+                        ):
                             has_unprocessed_level_nodes = True
                         break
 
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

santhreal:
Addressed review feedback: updated truncation logic to check whether any remaining entry in `current_level_nodes` or `queue` represents an unvisited node before setting `is_truncated=True`. Added regression test for multi-parent graph traversal.

danielaskdd:
@codex review

PR #3559, opened 2026-08-03
fix(utils): count delimiter tokens in split_text_by_token_limit

santhreal:
## Summary

`split_text_by_token_limit` joins units with `"\n\n"` but does not count the joiner toward the running token budget, so a rejoined chunk can exceed `max_tokens`.

## Problem

The accumulator only adds each unit's token count. When two units fit under the budget individually but not once separated by `"\n\n"`, the function still emits a single joined chunk over the limit. Callers that rely on the hard cap (LLM input windows, hard-fallback splitting) then overshoot.

## Fix

Measure `sep_tokens = len(tokenizer.encode("\n\n"))` once and charge that cost whenever a unit is appended to a non-empty group. Non-positive `max_tokens` continues to return `[]`.

## Test plan

- [x] `pytest tests/chunker/test_chunking.py -k split_text_by_token_limit_joiner`

### DIFF
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -2652,7 +2652,7 @@ def split_text_by_token_limit(
     text: str, tokenizer: Tokenizer, max_tokens: int
 ) -> list[str]:
     """Split text by token limit with sentence-first, token-window fallback."""
-    if not text:
+    if not text or max_tokens <= 0:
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

PR #3565, opened 2026-08-03
feat(tokenizer): safe encode-verified split/truncate contract, replacing #3559

danielaskdd:
## Summary

Establishes a unified, safety-verified `Tokenizer` contract for token-budget splitting and truncation, and migrates the specified call paths onto it. This closes #3559 (which only patched `split_text_by_token_limit`'s missing delimiter-token count) with a broader fix to the same class of bug, plus a performance improvement (bounded exponential retreat instead of linear/per-character re-encoding).

## Motivation

#3559 found that `split_text_by_token_limit` didn't count the `"\n\n"` delimiter's own tokens when repacking sentence units, so a chunk could exceed `max_tokens`. That is not an isolated bug — the repo has at least four independent, hand-rolled "safe token boundary" implementations with inconsistent details:

- `split_text_by_token_limit` (`utils.py`) — #3559's bug.
- `truncate_list_by_token_size` (`utils.py`) — the same class of bug, never triggered: it summed each item's own token count and never counted the separator between items.
- `_truncate_section_context` (`operate.py`) — O(max_tokens) linear re-encoding for heading-path truncation.
- `openai_embed` (`lightrag/llm/openai.py`) — `tokens[:max_token_size]` + `decode()`, fully bypassing `lightrag.utils.Tokenizer`, no independent re-verification. This is the only one of the four on a live embedding call path.

This PR establishes one contract + test suite instead of one more one-off fix, and migrates the call paths listed below. It does **not** claim to have eliminated every legacy unsafe implementation — see "Known remaining paths" below.

## What's changed

- **`Tokenizer.split_by_token_limit` / `truncate_by_token_limit`** (new): every returned `TokenSpan` is independently re-encoded and verified to fit `max_tokens` — BPE token count is not monotonic in text length, so nothing short of re-encoding the actual candidate substring can be trusted. The generic base-class implementation relies only on `encode()` (character-offset candidates, never `decode()`), so any third-party `Tokenizer` subclass gets correct (if slower) behavior with **no new required methods** — this is a purely additive capability, not a breaking interface change. `TiktokenTokenizer` overrides both with a `decode_with_offsets`-based fast path (`tiktoken>=0.7.0`, the repo's declared floor).
  - Retreat on overshoot is bounded, strictly one-directional (never oscillates), and **exponential** (1, 2, 4, 8, … tokens/chars), replacing per-token/per-character linear scans.
  - `split`'s window contract requires real forward progress (`end > covered_end` for every non-final span) — not just "the start moved", which would accept a non-covering window like `[1, 95)` following `[0, 100)`.
  - `overlap_tokens` is a best-effort target, not a "closest possible boundary" guarantee — token count isn't monotonic in char length, so an efficient generic implementation can't prove it found the closest boundary without an exhaustive search.
  - `TokenBudgetError` (new) is raised only when even the first complete Unicode code point can't fit the budget; it carries `max_tokens`, the code point's own token count, and a bounded preview.
- **`split_text_by_token_limit`**: now a thin deprecated wrapper delegating to the new split with zero overlap. The old sentence-first packer (and its `"\n\n"` rejoin logic that could produce non-byte-verbatim pieces) is removed.
- **`truncate_list_by_token_size` / `atruncate_list_by_token_size`**: gain an explicit `separator` parameter and count it against the budget (the actual #3559 bug class) — while still only ever keeping *whole* items, never a partially-truncated one. Because BPE isn't monotonic, the result is independently re-verified after mapping back to an item boundary, and shrunk further if that re-verification still fails.
- **`enforce_chunk_token_limit_before_embedding`**: now built on the safe split, with a configurable `overlap_tokens`. Hard-split children's `_source_span` is relocated via direct arithmetic when the parent's own content is a byte-verbatim slice of the source document, or via a new two-stage whitespace-normalized projection (`_parent_to_source_projection`, built once per parent and reused for every child) when it is not (e.g. the V strategy's sentence rejoining) — replacing a naive offset-add that could point at the wrong character. When the projection itself can't reconcile diverged content, this now raises `ChunkBlockMatchError` instead of silently emitting a wrong span.
- **`embedding_chunk_overlap_token_size`** (new `LightRAG` field, default 100, `EMBEDDING_CHUNK_OVERLAP_TOKEN_SIZE` env var): independent from the chunker's own `chunk_overlap_token_size`, which some strategies (V) deliberately zero out for unrelated reasons. `0` disables it; negative values are rejected at construction.
- **`process_chunks_unified`**: final truncation is now two-stage — an approximate pass on `{content, content_headings}` (no `reference_id` yet, since it can't be assigned before the survivor set is known), then re-verification against the real `reference_id`-tagged renderer (`reference_id` is recomputed from surviving chunks' `file_path` frequency, so truncating one more/fewer chunk can change it — not safe to ignore as "a couple tokens"). Both stages run in one synchronous helper submitted as a single `run_in_tokenizer_executor` job.
- **`_truncate_vdb_content`** (`operate.py`, entity/relation VDB payloads): now uses the safe contract instead of a fixed heuristic safety margin (`threshold - min(256, max(32, threshold // 16))`) plus an unverified `decode()`.

## Known remaining paths (not migrated this round)

- **`_truncate_section_context`** (`operate.py`): heading-path breadcrumb truncation, O(max_tokens) linear re-encoding. Internal display string only, not a live embedding path — lower priority.
- **`openai_embed`**'s local truncation (`lightrag/llm/openai.py`): `tokens[:max_token_size]` + `decode()`, bypasses `lightrag.utils.Tokenizer` entirely, no re-verification. This **is** a live embedding call path and should be the next follow-up; suggested fix: apply `document_prefix`/`query_prefix` first (unchanged order), use the already-resolved embedding-model tiktoken encoding, call a safe truncate on the full prefixed input, and independently re-verify the result.

Both are flagged rather than silently left — this PR is "establish the contract + migrate the specified paths", not "every unsafe implementation in the repo is now gone".

## Test plan

- [x] New `tests/llm/test_tokenizer_split_truncate_contract.py`: ASCII/CJK/emoji(ZWJ)/regional-indicator/special-token coverage, BPE non-monotonic counter-examples for both truncate and split, full-coverage/no-gap/real-progress invariants (incl. the `[0,100)→[1,95)` non-progress rejection), exponential-retreat call-count bound, third-party encode/decode-only tokenizer compatibility, no cross-call/thread state leakage.
- [x] New `tests/chunker/test_hard_fallback_span_projection.py`: the exact double-space regression example from the design doc, missing-`source_content` soft-drop, genuine-divergence hard-fail, direct-arithmetic fast path, per-parent projection reuse, overlap + sidecar backfill integration.
- [x] New `tests/pipeline/test_embedding_chunk_overlap_config.py`: default/zero/negative-rejection, independence from `chunk_overlap_token_size`.
- [x] New `tests/llm/test_process_chunks_unified_two_stage_truncation.py`: the `reference_id`-frequency-shift case, single-executor-submission assertion.
- [x] New `tests/extraction/test_truncate_vdb_content.py`.
- [x] Updated existing `truncate_list_by_token_size` call sites/tests for the new `separator` parameter; updated `tests/chunker/test_sidecar_backfill_integration.py`'s hard-split test for the new byte-verbatim-always behavior.
- [x] Full suite: `./scripts/test.sh tests` — 5540 passed, 155 skipped (integration, not run offline), 0 failed. (Updated after two rounds of review fixes — see comments below.)
- [x] `pre-commit run` (ruff format + lint) clean.


danielaskdd:
Thanks for the thorough review — all 5 findings are fixed in 930ba332.

1. **`enforce_chunk_token_limit_before_embedding` no longer swallows `TokenBudgetError`.** It used to catch it and re-emit the original, still-oversized chunk unchanged — exactly the unsafe fallback the contract exists to remove. It now lets the exception propagate, failing the document. Also removed the redundant pre-check full encode (duplicating `split_by_token_limit`'s own internal fast-path check), so an oversized chunk is now encoded in full exactly once.

2. **`_truncate_vdb_content`'s warning log no longer re-encodes the full original content** just to report its prior token count — it now reports the original character length instead.

3. **`Tokenizer.split_by_token_limit` (both implementations) now clamps the overlap target against the previous window's own token count *before* attempting it**, instead of trying a huge requested overlap and retreating from that starting point one exponential step at a time. `enforce_chunk_token_limit_before_embedding` also logs once per chunk if the requested overlap had to retreat all the way to 0.

4. **Every remaining VDB write path that built `content` directly is now migrated to `_truncate_vdb_content`**: relationship rebuild, existing-entity source-only update, custom KG ingestion, entity rename/edit, manual entity/relation create, and entity merge — covering ingestion / rebuild / custom KG / create / edit / rename / merge / source-only update.

5. **Every one of those paths (plus the three already migrated) now constructs and verifies its VDB payload BEFORE the first graph mutation** wherever that reorder is safe, so a truncation failure leaves the graph untouched. Where the graph is already documented as authoritative and mutated first (`utils_graph.py`'s merge helpers), a truncation failure is instead surfaced as `VectorStorageConsistencyError`, consistent with how VDB-operation failures are already handled there.

Also caught and fixed a regression the `ainsert_custom_kg` reorder introduced along the way: the empty-batch fast path (chunks-only `custom_kg`) must not pay for `_build_global_config`, which a bare/partially-initialized `LightRAG` instance can't satisfy — a pre-existing test (`test_ainsert_custom_kg_empty_batch_skips_keyed_lock`) caught this immediately.

24 new tests directly drive every affected function (rebuild, merge, custom KG, create, edit/rename, merge_entities) with both a truncating and a hard-failing tokenizer, asserting graph state and VDB call counts on each side, plus dedicated tests for the overlap-clamp call-count bound, `TokenBudgetError` propagation, the single-encode invariant, and `EMBEDDING_CHUNK_OVERLAP_TOKEN_SIZE` env parsing.

Full suite: 5535 passed, 155 skipped (integration, not run offline). `pre-commit` clean.

danielaskdd:
Thanks — both P1s were real, fixed in bce7e565.

1. **`aedit_relation` (utils_graph.py) still bypassed `_truncate_vdb_content` entirely.** Fixed, and reordered so content is built and verified before the (pre-existing) VDB delete and the graph edge update, instead of after both.

2. **Multi-object operations front-loaded truncation per-branch but still let one object's mutation slip in before every object's payload in the same operation was validated:**
   - `ainsert_custom_kg`: entity nodes were batch-upserted before the relationship VDB payload was validated. Restructured so relationship-endpoint discovery (`has_nodes_batch` — a read) and *both* VDB payloads are built first; every graph mutation (entities, missing endpoints, edges) now happens only afterward.
   - `_rebuild_single_relationship`: missing endpoint entities were created in the graph/VDB before the relationship's own content was validated. Split into a discovery pass (reads only, builds a plan per missing endpoint) followed by a mutation pass that runs after the relationship payload is also validated.
   - `_merge_edges_then_upsert`: same shape of issue in its two-branch (create new endpoint / update existing endpoint) per-endpoint loop — restructured the same way.

All three are now genuinely "validate every payload the operation will write, then mutate anything" — not just front-loaded within each sub-object's own branch.

9 new tests reproduce your exact repro shape — a tokenizer that truncates entity/endpoint content fine but always fails on relation content (discriminated on the literal tab in `f"{keywords}\t{src}\n{tgt}\n{desc}"`, since that's the one thing relation content always has that entity content doesn't) — for `aedit_relation`, custom KG, relationship rebuild, and edge-merge-with-new-endpoint. Each asserts nothing landed in the graph or either VDB when the relationship payload fails.

Also synced `EMBEDDING_CHUNK_OVERLAP_TOKEN_SIZE` into README.md, README-ja.md, and both API server config docs (previously only `docs/ProgramingWithCore.md` and `env.example`), and updated the PR description's stale test count.

Full suite: 5540 passed, 155 skipped (integration, not run offline). `pre-commit` clean.

danielaskdd:
Confirmed and fixed in c4c74454 — real bug, thanks for catching it.

\`existing_nodes\` was seeded only from \`has_nodes_batch()\`, which reads the graph before \`entity_nodes\` has actually been written (that write is now deferred to the mutation phase, after every payload validates). So an entity that is both an explicit batch entity and a relationship endpoint got misclassified as "missing" and queued into \`missing_nodes\` with a bare \`{description: "UNKNOWN"}\` placeholder — which then clobbered the real entity when the mutation phase wrote \`entity_nodes\` first and \`missing_nodes\` second.

Fixed exactly as suggested: \`existing_nodes.update(name for name, _ in entity_nodes)\` right after the \`has_nodes_batch\` call, before the missing-endpoint discovery loop runs.

Added the regression test you described (two new entities + their relationship in one \`custom_kg\` batch) and verified it the right way round: reverted the fix locally, confirmed the test fails with both entities collapsing to \`"UNKNOWN"\`, then restored the fix and confirmed it passes.

Full suite: 5541 passed, 155 skipped (integration, not run offline). \`pre-commit\` clean.

danielaskdd:
Fixed a real architectural regression introduced by the previous two rounds of fixes — thanks for catching this.

**The contract**: LightRAG requires the graph store to be mutated before the vector store for every entity/relation write. The graph is authoritative; a stale/missing VDB entry is always repairable from it via the offline rebuild tool (`lightrag-rebuild-vdb`). A graph write must never be gated behind VDB feasibility, since that has no equivalent recovery path.

**What broke it**: rounds 1–2 introduced a "front-load VDB payload construction + truncation before any graph mutation" pattern across 8 call sites: `acreate_entity`, `acreate_relation`, `aedit_relation` (utils_graph.py), `_rebuild_single_entity`, `_rebuild_single_relationship`, `_merge_nodes_then_upsert`, `_merge_edges_then_upsert` (operate.py), and `ainsert_custom_kg` (lightrag.py). That inverted the contract: a `TokenBudgetError` from `_truncate_vdb_content` now aborted the whole operation *before* the graph was ever touched — so a description that happens not to fit the embedding model's token budget would silently drop the entity/relation from the graph too, instead of merely leaving its VDB entry stale (the correct, recoverable failure mode).

**Fix**: reverted all 8 call sites back to graph-write-first — each object's graph mutation (`upsert_node`/`upsert_edge`/batch equivalents) now happens unconditionally, and `_truncate_vdb_content` + the VDB upsert follow immediately after. A truncation failure now only ever risks a stale/missing VDB entry. `_edit_entity_impl` and `_merge_entities_impl` already followed graph-first order (wrapped in `VectorStorageConsistencyError`) from round 1 and were left unchanged.

**Bonus simplification**: for `ainsert_custom_kg`, restoring `entity_nodes`' graph write to happen *before* the relationship-endpoint `has_nodes_batch` discovery makes the round-3 fix's `existing_nodes.update(...)` workaround unnecessary — `has_nodes_batch` now naturally sees the batch's own just-written entities, so an endpoint that's also an explicit entity is never misclassified as missing in the first place. Removed that line.

**Tests**: updated the ~10 tests in `test_vdb_content_truncation_coverage.py` that asserted the now-reverted "nothing written on truncation failure" behavior to instead assert "graph write succeeds with the full untruncated content, VDB write is skipped" — matching the restored contract. Full suite: 5541 passed, 155 skipped, pre-commit clean.

Commit: b89bedd5

danielaskdd:
Refinement on top of the previous commit (b89bedd5), per further discussion: that commit fully reverted to graph-write-then-validate, which restores the "graph before VDB" persistence-order contract but has its own downside — `TokenBudgetError` from `_truncate_vdb_content` isn't like a normal transient VDB I/O failure. It's a deterministic, non-retryable content-shape problem: the content cannot fit the embedding model's token budget, and it never will no matter how many times it's retried. That means the offline rebuild tool would hit the exact same failure trying to rebuild that entry from the graph — so graph-write-then-validate was stranding a graph node/edge that could never get a VDB entry, ever.

**Fix (e27f97ff)**: front-load only the `_truncate_vdb_content` call itself (pure validation, no I/O) to run immediately before each object's own graph mutation, while keeping the actual VDB `upsert`/`delete` I/O call after the graph write, unchanged. Net effect:
- `TokenBudgetError` (deterministic, unrepairable) now prevents that object's graph write too — nothing is left half-written for content that can never be embedded.
- A transient VDB I/O failure (network blip, backend hiccup) still leaves the normal recoverable window: graph updated, VDB stale, fixable via the offline rebuild tool.

This is applied **per-object**, not as a two-phase "validate everything, then mutate everything" restructuring across a whole multi-object operation — in a batch with several entities/relationships, an earlier object that already validated and graph-wrote successfully is not rolled back just because a later object in the same call fails validation. Concretely this matches the shape from the first review-response commit (930ba332) for `acreate_entity`, `acreate_relation`, `_rebuild_single_entity`, `_rebuild_single_relationship`, `_merge_nodes_then_upsert`, `_merge_edges_then_upsert`, and `ainsert_custom_kg`, plus the second review-response commit's shape (bce7e565) for `aedit_relation`. `_edit_entity_impl`/`_merge_entities_impl` already wrapped post-graph-write VDB failures in `VectorStorageConsistencyError` and are unchanged.

Updated the ~10 affected tests accordingly: single-object paths assert "truncation failure leaves the graph fully untouched"; multi-object paths (custom KG, relationship rebuild, edge merge with a new endpoint) assert "earlier validated objects land in the graph, the object whose own content fails validation does not." Full suite: 5541 passed, 155 skipped, pre-commit clean.

danielaskdd:
Confirmed, thanks — this was a real gap. Verified by reading the actual code at utils_graph.py:872: the order was indeed \`truncate → relationships_vdb.delete → graph.upsert_edge → relationships_vdb.upsert\`, and my e27f97ff comment claiming the delete happened after the graph write was wrong.

**Fix (8baa6969)**: moved \`chunk_entity_relation_graph.upsert_edge\` before \`relationships_vdb.delete\`. New order: truncate/validate → graph write → VDB delete → VDB upsert — matching every other call site's "graph write, then VDB I/O" contract. A transient \`delete()\` failure (storage hiccup) now leaves the edited relation content correctly saved in the graph instead of silently discarding the edit (the old order meant the offline rebuild tool could only reconstruct the pre-edit relation from the graph, since the edit never reached it).

Added \`test_edit_relation_vdb_delete_failure_still_leaves_graph_updated\` per your suggestion, and verified it's fix-proof via \`git stash\`: reverting just the utils_graph.py change reproduces the exact failure (\`AssertionError: assert 'short' == 'updated description'\`), confirming the test actually exercises this bug rather than just the code path.

Full suite: 5542 passed, 155 skipped. pre-commit clean.

PR #3556, opened 2026-08-03
fix(kg): return copies from JsonDocStatusStorage reads

santhreal:
## Summary

`JsonDocStatusStorage.get_by_id` / `get_doc_by_file_path` return live references into the in-memory store. Callers that mutate the returned dict corrupt persisted document status.

## Problem

```python
doc = await storage.get_by_id("doc1")
doc["status"] = "modified_externally"  # mutates internal _data
```

Any pipeline or API code that annotates a returned status dict then writes those changes into the live store without going through `upsert`.

## Fix

Return a shallow `dict.copy()` from both read paths so external mutation cannot touch storage state.

## Test plan

- [x] `pytest tests/kg/json_impl/test_json_doc_status_copy_on_read.py`

PR #3567, opened 2026-08-04
fix(kg): deep-copy JsonKVStorage read results to prevent aliasing

danielaskdd:
## Summary

`JsonKVStorage.get_by_id` / `get_by_ids` only shallow-copy the returned record before handing it to the caller. In single-process mode (`self._data` is a plain dict, not a `multiprocessing.Manager().dict()` proxy), a shallow copy only protects the top-level dict — any nested mutable field (list/dict) in the record still aliases the exact same object living in internal storage. Same bug class as #3556 (`JsonDocStatusStorage`), just not yet fixed on the KV side.

## Problem

```python
chunk = await text_chunks_storage.get_by_id("chunk1")
chunk["llm_cache_list"].append("new-cache-key")  # mutates internal _data in place
```

This isn't hypothetical — `lightrag/utils.py`'s `update_chunk_cache_list()` does exactly this:

```python
chunk_data = await text_chunks_storage.get_by_id(chunk_id)
...
chunk_data["llm_cache_list"].extend(new_keys)   # in-place mutation of a nested list
await text_chunks_storage.upsert({chunk_id: chunk_data})
```

Because `get_by_id`'s shallow copy leaves `llm_cache_list` aliased to the live storage row, this mutation bypasses `_storage_lock` entirely and changes internal state before `upsert()` is ever called. Any other coroutine holding an earlier `get_by_id` snapshot of the same chunk would observe the mutation appear underneath it, breaking the assumption that a read result is a point-in-time snapshot.

## Fix

Return `copy.deepcopy(...)` from both `get_by_id` and `get_by_ids` instead of a shallow copy, mirroring the fix already applied to `JsonDocStatusStorage` in #3556.

## Test plan

- [x] `pytest tests/kg/json_impl/test_json_kv_copy_on_read.py` — covers top-level + nested mutation for both read paths, plus a regression test that drives the real `update_chunk_cache_list` caller and asserts an earlier snapshot stays frozen after it runs.
- [x] Full `tests/kg/json_impl/` and `tests/pipeline/` suites pass (693 tests).
- [x] `pre-commit run ruff` clean.

### DIFF
--- a/lightrag/kg/json_kv_impl.py
+++ b/lightrag/kg/json_kv_impl.py
@@ -1,3 +1,4 @@
+import copy
 import os
 from dataclasses import dataclass
 from typing import Any, ClassVar, final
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3572, opened 2026-08-05
fix(docker): install libcairo2 so native md SVG rasterization works

danielaskdd:
## Description

The Docker image builds and installs `cairosvg` (used by the native markdown/textpack parser to rasterize embedded SVG images to PNG) but never installs `libcairo2`, the native shared library `cairosvg` binds to via cffi at runtime. As a result, `pip`/`uv install` succeeds silently, and the failure only surfaces at runtime when a document containing an SVG is processed:

```
WARNING: [native_md] cairosvg unavailable, cannot rasterize SVG: no library called "cairo-2" was found
```

This isn't Docker-specific — any `pip install lightrag` on a host without the system `libcairo` package hits the same gap, since `pip`/`uv` cannot install native shared libraries.

## Related Issues

N/A

## Changes Made

- `Dockerfile`: install `libcairo2` in the final image stage.
- `README.md` / `README-zh.md` / `README-ja.md`: document the `libcairo` system dependency (mirrors the existing spaCy/smart_heading section) with per-platform install commands.
- `docs/FileProcessingPipeline.md` / `-zh.md`: note the cffi-binding nature of `cairosvg` and the system package needed for rasterization to actually work.
- `lightrag/parser/markdown/parser.py`: replace the previous import-only check with `check_svg_rasterizer()`, which actually calls `cairosvg.svg2png()` against a trivial probe SVG (import succeeding does not prove `libcairo` is loadable) and returns a diagnostic message instead of `None`.
- `lightrag/api/utils_api.py`: surface that diagnostic as a prominent yellow warning in `display_splash_screen()` (new `_warn_about_svg_rasterizer()`, styled like the existing `_warn_about_body_limits()` warnings) instead of a log line buried in server output, so operators notice a missing `libcairo` at startup rather than the first time a user uploads an SVG-containing document.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Verified both the success path and the failure path of the new startup probe/warning locally (by monkeypatching `check_svg_rasterizer` to simulate a missing `libcairo`), and confirmed the yellow warning block renders correctly in the splash screen. Ran `tests/parser` (markdown/native_md subset, 103 passed) and the full `tests/api` suite (699 passed, 14 skipped) with no regressions.


danielaskdd:
@codex review

PR #3573, opened 2026-08-05
fix(api): clean up startup dependency warnings (SVG rasterizer + smart_heading)

danielaskdd:
## Description

Follow-up to #3572, expanded into a general cleanup of startup-time dependency warnings/errors so they read as clean, bordered notices instead of raw exception dumps or one-line summaries missing a `sudo`.

1. The SVG-rasterizer splash-screen warning embedded the raw `cairosvg`/cairocffi exception, which repeats its library-not-found lookup across every platform's naming convention (`.so`/`.dylib`/`.dll`) and produced a multi-line dump.
2. The Linux install commands shown for `libcairo` (README x3, docs, and the splash warning) were missing `sudo`, and the splash warning only had a one-line summary instead of copy-pasteable per-platform commands.
3. Missing spaCy models for `smart_heading` raised `SmartHeadingNLPError` uncaught past `create_app()`, dumping a full Python traceback at startup instead of a clean message.

## Related Issues

Follow-up to #3572.

## Changes Made

- `lightrag/parser/markdown/parser.py`: `check_svg_rasterizer()` now logs the raw exception at debug level only and returns a short, fixed diagnostic message instead of embedding the exception text.
- `README.md` / `README-zh.md` / `README-ja.md` / `docs/FileProcessingPipeline*.md`: added `sudo` to the Linux `apt-get`/`dnf` install commands for `libcairo`.
- `lightrag/api/utils_api.py`: the SVG-rasterizer splash warning now lists each platform's install command on its own line (with `sudo` on Linux).
- `lightrag/parser/docx/smart_heading/nlp.py`: `SmartHeadingNLPError` now carries `problem`/`install_hint` as separate attributes instead of only a flattened message, so a caller can lay the install command out on its own line without parsing exception text. `str(exc)` is unchanged, so existing message-matching tests still pass.
- `lightrag/api/lightrag_server.py`: `create_app()` now catches `SmartHeadingNLPError` and prints a clean bordered error box (matching the style already used for `check_frontend_build()`'s warning) before `sys.exit(1)`, instead of letting the exception propagate into a raw traceback.
- Worked around an `ascii_colors` quirk along the way: it treats `[...]` as rich-markup tags and silently drops anything it doesn't recognize, which was eating `[api]` from the `pip install lightrag-hku[api]` command until `markup=False` was passed.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Verified locally end-to-end through the real `create_app()` (monkeypatching `missing_spacy_models()` to simulate missing models) that the new box prints cleanly — no traceback, `[api]` preserved, `sys.exit(1)` raised as expected. Also re-verified the SVG-warning failure path. Ran the smart_heading test files (140 passed), the parser/docx subset (872 passed), and the full `tests/api` suite (699 passed, 14 skipped) with no regressions.


danielaskdd:
@codex review

PR #3574, opened 2026-08-05
fix(rerank): validate results at query boundary

pengpengyi92:
## Description

Validate index-based rerank results at the final query boundary before mapping them back to retrieved documents.

Custom rerank functions can return mixed or malformed result lists even when built-in HTTP providers normalize their responses. Previously, a boolean index such as `true` selected document 1 because Python treats booleans as integers, while one item missing `relevance_score` caused the entire rerank operation to fall back and discard otherwise valid results. The format check also relied only on the first item.

## Related Issues

No linked issue. This is a focused regression fix discovered while auditing the query rerank path after the recent provider-boundary validation work.

## Changes Made

- move rerank result normalization to `lightrag.utils` so provider adapters, chunk aggregation, and query mapping share one validation rule
- skip leading malformed non-object items when detecting index-based responses without broadening legacy-format detection
- ignore malformed entries while preserving valid rerank results
- retain the existing fallback to original documents when no usable result remains
- add offline regressions for boolean indices and mixed result lists

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (not necessary; no public configuration or API change)
- [x] Unit tests added

## Testing

- `python -m pytest -q tests/llm tests/chunker/test_rerank_chunking.py --disable-warnings --maxfail=3` (`371 passed`)
- `python -m ruff check --select E9,F63,F7,F82,I lightrag/rerank.py tests/llm/test_apply_rerank_result_validation.py`
- `git diff --check`

## Additional Notes

The commit is SSH-signed. This does not change provider request behavior or rerank scoring; it only validates the returned mapping before documents are selected.

### DIFF
--- a/lightrag/rerank.py
+++ b/lightrag/rerank.py
@@ -1,19 +1,19 @@
 from __future__ import annotations
 
-from collections import Counter
-from math import isfinite
 import os
+from collections import Counter
+from typing import Any, Dict, List, Optional, Tuple
+
 import aiohttp
-from typing import Any, List, Dict, Optional, Tuple
+from dotenv import load_dotenv
 from tenacity import (
     retry,
+    retry_if_exception_type,
     stop_after_attempt,
     wait_exponential,
-    retry_if_exception_type,
 )
-from .utils import logger, run_in_tokenizer_executor
 
-from dotenv import load_dotenv
+from .utils import logger, normalize_rerank_result, run_in_tokenizer_executor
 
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

pengpengyi92:
Thanks for the review and merge! I appreciate the quick turnaround and am glad the query-boundary validation is now in.

PR #3578, opened 2026-08-05
test(kg): non-interactive CLI for test_graph_storage.py, fix swallowed failures, drop redundant pg-smoke step

danielaskdd:
## Summary

- `tests/kg/test_graph_storage.py` could only be driven by an interactive menu (`input()`), which blocked on CI/non-TTY, and the menu called test functions directly — bypassing the pytest fixture, a second execution path that could drift from what pytest actually runs. Replaced with a CLI (`python tests/kg/test_graph_storage.py [names...] [--list] [-y] [-- pytest-args]`) that delegates to `pytest.main()`, so both entry points run the same code through the same fixture.
- Seven tests ended in `except Exception: return False`. `pytest-asyncio`'s `runtest()` discards a coroutine test's return value (verified against the installed plugin source, `runtest(self) -> None`), so a failed assertion was silently reported as **PASSED** with only a printed error line — no warning, nothing in the exit code. Replaced with a helper that logs the same marker and re-raises, so pytest is the actual authority on the result.
- `check_env_file()` prompted via `input()` inside the pytest fixture itself, so plain `pytest` in a terminal without a `.env` would hang waiting for an answer. Now it only warns.
- `pg-smoke.yml`'s second step ran `test_graph_storage.py` against `LIGHTRAG_GRAPH_STORAGE=PGTableGraphStorage`, duplicating the contract coverage #3575 already ported into `tests/kg/pgtable_impl/test_pgtable_smoke.py` (verified there via mutation testing to catch bugs the prior smoke suite didn't). That module hardcodes `PGTableGraphStorage` directly — no env-var dispatch — so it can't silently fall back to a different backend the way `test_graph_storage.py`'s generic dispatch can (confirmed: running this file without setting the backend env var passes all 8 tests against NetworkXStorage with zero indication anything different was expected). Removed the redundant step and its two guard scripts; `pg-smoke.yml` now only runs `test_pgtable_smoke.py`. `test_graph_storage.py` is now the manual/CLI entry point for the five backends `test_pgtable_smoke.py` doesn't cover — its docstring says so, and no CI workflow runs it.
- CLI usability, based on actually using it after the first pass: default output was noisy (`-s` replayed every `print()` from these old, print()-heavy test bodies) and it was easy to lose track of which backend was under test. The CLI now drops `-s` by default (pytest captures and only replays on failure — `-- -s` opts back into the firehose), pauses once for a TTY-gated confirmation naming the backend and tests about to run (skippable with `-y`/non-TTY), and repeats the backend name next to the final pass/fail result. This is deliberately CLI-only: a bare `pytest tests/kg/test_graph_storage.py -m integration --run-integration` is the CI-equivalent path, where the backend is already a known configuration fact (a workflow's `env:` block, or a developer's own env var) — printing it there would just make this file's output inconsistent with every other test file in the suite.
- `CLI_TESTS` (short CLI name → test function) was a hand-maintained dict alongside a dedicated `offline` guard test asserting it stayed in sync with the module's actual tests. Replaced with `_discover_cli_tests()`, deriving the mapping by introspection (same "takes the `storage` fixture" membership check the guard test used) — the drift is now structurally impossible, so the guard test is gone rather than kept for its own sake. Two derived names are slightly longer than their hand-picked predecessors (`string-escaping` → `string-escaping-regressions`, `undirected` → `undirected-property`).

## Why

`os.getenv("LIGHTRAG_GRAPH_STORAGE", "NetworkXStorage")` means running this file without explicitly setting the backend — or a CI step accidentally missing its `env:` override — silently falls back to NetworkXStorage (which needs zero configuration) and still reports every test as **passed**, with no signal that the wrong backend ran. Combined with the swallowed-`return False` bug, the previous CI step's guards (`tests>0 && skipped==0`) could not have caught either failure mode: NetworkXStorage never skips, so guard 1 is blind to it. `test_pgtable_smoke.py` is immune to this class of bug by construction (it imports and instantiates `PGTableGraphStorage` directly), which is why its ported coverage is the more reliable CI signal for that backend going forward.

## What's changed

- `tests/kg/test_graph_storage.py`:
  - Non-interactive-by-default CLI (`_discover_cli_tests()` + `argparse` + `pytest.main()`); `-y`/`--yes` and non-TTY stdin skip the one confirmation prompt.
  - Removed all `return True`/`return False` in favor of re-raising through `_report_and_reraise`.
  - `check_env_file()` no longer prompts.
  - `storage` fixture is unchanged in behavior/output — stays consistent with every other test file when driven by plain pytest.
  - Docstring documents both entry points, the CI-ownership split with `test_pgtable_smoke.py`, and the CLI's confirmation/quieter-output behavior.
- `.github/workflows/pg-smoke.yml`: removed the "Run shared graph contract suite (PGTableGraphStorage)" step and its two guards; the "Run pg_smoke tests" step is unchanged.

## What's not changed

- `test_pgtable_smoke.py` (from #3575) is untouched — its ported PGTableGraphStorage contract coverage stays as CI's source of truth for that backend.
- No behavior change to any of the 8 contract tests themselves beyond the exception-swallowing fix — same assertions, same scenarios.
- This file contributes zero tests to the default CI gate (`-m offline`) and is not invoked by any workflow — it's purely a manual/CLI tool now.

## Test plan

- [x] `pytest tests/kg/test_graph_storage.py --run-integration -q` → `8 passed`, against the default `NetworkXStorage` backend (no external service needed)
- [x] `pytest tests/kg/test_graph_storage.py -m offline -q` → `0 tests` collected from this file (fully opted out of the default CI gate)
- [x] `pytest tests/kg/test_graph_storage.py -m integration --run-integration -v` output shape verified identical to any other test file in the suite (`PASSED`/`FAILED` lines + summary, no extra prints) — confirmed both on a real terminal and piped through `cat` to simulate a non-TTY CI log
- [x] CLI: `python tests/kg/test_graph_storage.py` (all), `... basic advanced` (subset), `... --list`, `... bogus` (exits 2 on unknown name), `... -y` / `< /dev/null` (skip confirmation), `-- -s`/`-- -x --tb=long` (arg forwarding) — all verified locally
- [x] `_confirm_backend` unit-tested standalone under mocked `isatty()`/`input()` for all five paths (Enter, "n", EOFError, non-TTY, `-y`) — each behaves as designed, and non-TTY/`-y` never call `input()` at all
- [x] `main()` confirmed to return 1 and never invoke `pytest.main()` when confirmation is declined
- [x] Fix-proof: with one assertion inverted, pre-change code prints the error marker but reports `1 passed`; post-change code reports `1 failed` and the CLI exits 1
- [x] `.github/workflows/pg-smoke.yml` YAML validated (`yaml.safe_load`); diff against the removed step confirmed to be a pure deletion — the surviving "Run pg_smoke tests" step is byte-for-byte unchanged
- [x] `pre-commit run` (ruff format + ruff check) clean on all changed files


danielaskdd:
@codex review

PR #3580, opened 2026-08-05
refactor(tools): make tiktoken and spaCy cache downloads independent

danielaskdd:
## Description

`lightrag-download-cache` always downloaded the tiktoken cache regardless of flags, with `--spacy`/`--spacy-install` only appending spaCy model handling on top. There was no way to fetch just the spaCy models without also silently pulling the tiktoken cache, which was misleading and undocumented.

## Related Issues

None — spotted while reviewing the offline deployment docs.

## Changes Made

- `lightrag/tools/download_cache.py`: tiktoken and spaCy downloads are now independent selections instead of "always tiktoken, optionally plus spaCy":
  - `--cache-dir`/`--models` (tiktoken-specific detail flags) or the new explicit `--tiktoken` select the tiktoken download.
  - `--spacy-dir` (spaCy-specific detail flag) or `--spacy`/`--spacy-install` select the spaCy download.
  - With no flags at all, the default remains tiktoken-only (unchanged).
  - Combine explicitly (e.g. `--tiktoken --spacy-install`) to get both without specifying directories.
- The `Dockerfile` builder stage invocation (`--cache-dir ... --spacy-dir ...`) already passes a detail flag for each, so it keeps downloading both tiktoken and the spaCy wheels — no change needed there.
- Dropped the redundant `--spacy` alongside `--spacy-install` throughout README (en/zh/ja), `env.example`, `docs/FileProcessingPipeline(.md/-zh.md)`, `docs/DockerDeployment.md`, `docs/OfflineDeployment.md`, the smart_heading install-hint string, the server startup error hint, and the pytest hint constant — `--spacy-install` already selects the spaCy action on its own.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Unit tests added (if applicable) — no CLI-parsing behavior change required new tests; existing smart_heading hint-string tests (`tests/parser/docx/test_smart_heading_guards.py`, `tests/parser/test_smart_heading_global_default.py`, 140 tests) still pass.

## Additional Notes

Verified the new selection logic by hand against every argument combination currently used in the repo (default, `--cache-dir`, `--models`, `--spacy`, `--spacy-install`, and the Dockerfile's combined invocation) — all match the intended before/after behavior. `ruff check` passes on all touched Python files.


danielaskdd:
@codex review

PR #3581, opened 2026-08-05
Rank isolated entities in get_popular_labels, stop dropping edges, unify the graph read contract

danielaskdd:
> Supersedes #3577 — same change, squashed to one commit and rebased on current `main`, with the description reorganised by theme rather than by the order things were found.

## Description

Two problems, both reported against `PGGraphStorage` and both found in other backends once looked for.

**1. `get_popular_labels` lost entities that have no relations.** Every backend that derives degrees from its edge store had the same hole — an entity with no edge has no row to rank, so it never appeared. A graph whose entities carry no relations reported an **empty entity picker**.

**2. Edge writes could leave the graph inconsistent.** On AGE a missing endpoint made the upsert silently write nothing; on Mongo and OpenSearch only the *source* endpoint was materialized, so a target-only id could end up with edge documents and no node document.

Fixing those surfaced three consistency defects on the same read paths (§3), and the resulting contract is now written down in `BaseGraphStorage` (§4) so the seven backends cannot drift again.

## Related Issues

Reported against `PGGraphStorage`:

1. `get_popular_labels` excluded every entity with degree 0.
2. `_build_upsert_edge_sql` silently dropped an edge when an endpoint was missing — no exception, no log line.
3. `get_node_edges` returned `[]` for a node that does not exist, where NetworkX returns `None`.

Defect 1 was then found in `mongo_impl` and `opensearch_impl`; defect 3 in `mongo_impl`, `neo4j_impl` and `memgraph_impl`.

### 1. Popular labels: rank connected entities, top up from isolated ones only when short

Two phases, and the second almost never runs:

- **Phase 1** ranks the entities that *have* edges, from the edge-derived degrees. Any real graph holds far more than `limit` (default 300) connected entities, so this fills every slot on its own and the node store is never touched.
- **Phase 2** runs only when phase 1 comes up short — a small graph, which is exactly the case that used to return nothing. Its aggregate is exact, so coming up short also means the connected set is now known in full and every remaining entity is isolated by definition. The top-up is bounded by the shortfall (at most `limit` rows), which keeps it affordable even where it means a sequential scan of the node store.

| backend | phase 1 | phase 2 |
|---|---|---|
| PG/AGE | `node_degrees` ⨝ vertex table (index probes for ids that have edges) | second query, `NOT EXISTS` per endpoint column so both halves can use an index, `LIMIT` = shortfall |
| Mongo | aggregation on the edge collection | `find({_id: {$nin: …}}).sort(_id).limit(shortfall)` — rides the `_id` index, stops as soon as it has enough |
| OpenSearch | terms aggregation on the edges index | PIT scan of the node index in `entity_id` order, stops at the shortfall |

Driving the whole ranking off the node store instead is simpler to write and gives the same answer, but pays a full node-store pass on every call to produce a result phase 1 already had.

**Ties break on the label, ascending, everywhere.** `COLLATE "C"` on PG so it matches Python's code-point sort; an explicit sort key on OpenSearch, which had been falling back to aggregation bucket order; and `(-degree, label)` on **NetworkX**, the default backend, whose stable sort left ties in node insertion order — a graph that inserted `Zeta` before `Alpha` answered `?limit=1` with Zeta and dropped Alpha. With degree-0 entities now able to fill the tail, a tie at the cutoff decides which labels the caller never sees.

### 2. Edge writes

**AGE no longer loses edges silently.** The Cypher is `MATCH (source) … MATCH (target) … CREATE`: if either endpoint is absent the whole pattern fails to match, the statement **succeeds with zero rows**, and the relation is gone. Both write paths now run it through `fetch()` and check that an edge came back, raising `PGGraphEdgeWriteLostError` (a `PGGraphQueryException` subclass) otherwise. The error probes both endpoints so the message names the missing one; the probe is best-effort and can never mask the write loss. The batch path rolls the whole chunk back — same all-or-nothing behaviour as any other mid-chunk failure. Not retried: a missing endpoint is deterministic.

**Mongo and OpenSearch materialize both endpoints.** A target-only endpoint is a state where `has_node()` says absent while the edge scan says connected, and `delete_entity` 404s on an entity whose edges are right there. `NetworkXStorage.add_edge` and `PGOpsGraphStorage` create both ends; the document backends now do too. Mongo uses one `bulk_write` of `$setOnInsert` ops for the pair (deduped, so a self-loop writes one) — same round-trip count as the single `upsert_node` it replaces. OpenSearch covers the pair with one `has_nodes_batch` mget, then writes a placeholder for whichever endpoints are missing.

**That placeholder can never overwrite a real node** — two independent guards, because either alone leaves a hole:

- `has_nodes_batch` read `if doc.get("found")` straight off each mget item, so anything that was not an explicit `found: true` counted as absent, including an item-level `error` object (OpenSearch answers mget with HTTP 200 even when individual items fail on an unavailable shard) and a short or malformed `docs` array. It now goes through `_interpret_mget_item`, like every other mget read in that file, and rejects a `docs` array whose length does not match the request. This fixes every caller — entity merge/dedup read the same set.
- `upsert_node` goes through the plain index API, which **replaces** the document, so a placeholder written over a live endpoint would strip its description / entity_type / source_ids. Endpoint placeholders now use `op_type=create` with 409 tolerated, so the write is safe regardless of what the probe concluded — a concurrent writer can create the endpoint between probe and write. Non-409 bulk failures still fail the batch.

### 3. Read-path consistency defects found along the way

- **`get_node_edges` conflated "absent" with "no edges".** It returned `[]` for a node that does not exist — the same value an existing-but-isolated node yields. AGE, Neo4j and Memgraph now key on the anchor `MATCH` binding (an isolated node still yields exactly one row via `OPTIONAL MATCH`, so zero rows is the only "no such node" signal); Mongo and OpenSearch decide existence from the node store before the edge scan, which also makes an absent node cost one indexed point lookup instead of a full scan.
- **Backend errors were reported as "no labels".** `get_popular_labels` / `search_labels` logged and returned `[]` on any exception in AGE, Mongo and Memgraph. `/graph/label/popular` already turns an exception into a 500, so the swallow was the only thing between the user and an accurate error: the WebUI received 200 with an empty picker while the database was unreachable. They raise now. Deliberately preserved as real empty results: Mongo's progressive Atlas→regex fallback (a missing Atlas Search index is a capability gap, not a failure — only a failure of the final regex scan raises), blank queries, and a confirmed-empty node collection.

### 4. The contract, in `base.py`

`get_node_edges`: the three outcomes (`[]` / `None` / raise), and that `get_nodes_edges_batch` cannot express the middle one and flattens it by design. `get_popular_labels` / `search_labels`: the two-phase shape and why, the label tie-break, that errors raise, and that callers must tolerate a ranked label whose `get_node()` comes back empty (see *Scope* below).

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

**Scope — dangling endpoints are not filtered out of the ranking.** On Mongo and OpenSearch the degree ranking is keyed on edge endpoints, so an id with edge documents but no node document can appear in the picker. That is left alone on purpose:

- No in-tree path produces one. All six `upsert_edge`/`upsert_edges_batch` callers create both endpoints first; `delete_node`/`remove_nodes` delete edges **before** the node in both backends, so an interrupted delete leaves an isolated node rather than a dangling edge; and the purge path clears incident edges before removing an entity. §2 closes the remaining API-level hole going forward.
- The consequence is bounded and already handled: selecting such a label makes `get_knowledge_graph` log a warning and return an empty `KnowledgeGraph`, and the WebUI renders its "Graph Is Empty" placeholder. No crash, no silent wrong answer.
- Filtering for it would cost a join or an extra round trip on an interactive endpoint, every call, to prevent a client-side empty result. Repairing such data is `audit_kg_integrity`'s job.

**Behaviour changes reviewers should weigh.**

1. On PG/AGE, Mongo and Memgraph, `/graph/label/popular` and `/graph/label/search` now return **500 instead of 200 with an empty list** when the backend is down. Any client treating an empty list as "fine" will now see an error.
2. On Mongo and OpenSearch, an edge write also materializes the target endpoint. Existing dangling targets are not backfilled.
3. On OpenSearch, `has_nodes_batch` raises on an ambiguous mget item instead of reporting the id absent — a strictness increase shared with every other read in that file.

**No current caller is affected by the `None` change.** Every in-tree caller of `get_node_edges` guards with `if edges:` / `len(edges) if edges else 0`, and the batch sibling flattens `None` to `[]` by design. Returning `None` restores the declared contract and keeps the information recoverable; it does not change any existing call site's behaviour today.

**Tests.** Per-backend read-contract suites for AGE, Mongo, OpenSearch, Neo4j, Memgraph and NetworkX; edge-write-loss cases; `TestGraphEndpointPlaceholderSafety` covering both guards, including the end-to-end path (ambiguous probe → `upsert_edge` raises and writes nothing) and the 409 race. The popular-label tests assert behaviour rather than SQL/pipeline text: filling the limit must issue exactly one query and leave the node store untouched; coming up short must request only the shortfall.

The cross-backend integration suite `tests/kg/test_graph_storage.py` pins both contracts for every backend — two unconnected entities (`Orphan`, and `Aardvark` inserted last) must both be ranked, last, in label order rather than insertion order, and absent vs isolated must answer `None` vs `[]`. These assertions were held back until all seven backends conformed, so the shared suite never carried known-failing cases.

`tests/kg`: **1570 passed**. The 10 failures in `tests/kg/test_batch_graph_operations.py` reproduce on `main` unchanged (blocked tiktoken/pip downloads in the sandbox); the full-suite failure set is identical file-for-file with and without this branch.

https://claude.ai/code/session_01W3Y1u6KcuYQ8GLpToBMK7B

---
_Generated by [Claude Code](https://claude.ai/code/session_01W3Y1u6KcuYQ8GLpToBMK7B)_

### DIFF
--- a/lightrag/base.py
+++ b/lightrag/base.py
@@ -574,6 +574,19 @@ class BaseGraphStorage(StorageNameSpace, ABC):
         Returns:
             A list of (source_id, target_id) tuples representing edges,
             or None if the node doesn't exist
+
+        Three outcomes, three answers — implementations must not merge them:
+
+        * ``[]``   — the node exists and has no relations.
+        * ``None`` — the node is **confirmed absent**.
+        * raise    — the backend could not answer. A transport/server error is
+          neither of the above; reporting it as ``[]`` or ``None`` turns an
+          unknown into a fact that callers (entity merge/dedup, graph edits)
+          act on. Same rule as :meth:`BaseKVStorage.get_by_id_strict`.
+
+        ``get_nodes_edges_batch`` cannot express the middle case (it returns a
+        dict of lists) and flattens ``None`` to ``[]`` by design; callers that
+        need the distinction must use this single-node form.
         """
 
     async def get_nodes_batch(self, node_ids: list[str]) -> dict[str, dict]:
@@ -820,6 +833,39 @@ class BaseGraphStorage(StorageNameSpace, ABC):
 
         Returns:
             List of labels sorted by degree (highest first)
+
+        Ranks the WHOLE node set: an isolated (degree-0) entity ranks last, but
+        it still ranks. Implementations that derive degrees from their edge
+        store must therefore not let a node absent from that store fall out of
+        the result. Ties break on the label, ascending.
+
+        The expected shape is two phases, because the second one almost never
+        runs. Rank the entities that HAVE edges first — the cheap path, and any
+        graph with more than ``limit`` connected entities fills every slot there
+        — then top the result up from the isolated entities only when that comes
+        up short. The top-up is bounded by the shortfall (never more than
+        ``limit`` rows), which is what keeps it affordable even where it means a
+        sequential scan of the node store. Coming up short also means the
+        connected set is now known in full, so every remaining entity is
+        isolated by definition.
+
+        Driving the whole ranking off the node store instead is simpler to
+        write and gives the same answer, but it pays a full pass over the nodes
+        on every call — on a large graph, to produce a result the first phase
+        already had.
+
+        The result is best-effort about the reverse direction: a backend that
+        derives degrees from its edge store MAY also surface an id that has
+        edges but no node document. Only a data-quality defect produces one —
+        the write paths materialize both endpoints of every edge — and
+        confirming every ranked id against the node store would cost a join or
+        an extra round trip on a hot, interactive endpoint. Callers must
+        therefore tolerate a returned label whose :meth:`get_node` comes back
+        empty, rather than assume every label resolves.
+
+        An empty list means the graph holds no entities. A backend error MUST
+        raise instead — ``/graph/label/popular`` renders both, and the two are
+        indistinguishable to the user once the error is swallowed.
         """
 
     @abstractmethod
@@ -832,6 +878,10 @@ class BaseGraphStorage(StorageNameSpace, ABC):
 
         Returns:
             List of matching labels sorted by relevance
+
+        As with :meth:`get_popular_labels`, an empty list means "nothing
+        matched" — a backend error MUST raise rather than return it. A blank
+        query is a real empty result, not an error.
         """
 
 
--- a/lightrag/kg/memgraph_impl.py
+++ b/lightrag/kg/memgraph_impl.py
@@ -369,7 +369,9 @@ class MemgraphStorage(BaseGraphStorage):
 
         Returns:
             list[tuple[str, str]]: List of (source_label, target_label) tuples representing edges
-            None: If no edges found
+            None: If the node does not exist. An existing node with no relations
+                returns ``[]`` — the BaseGraphStorage contract, as implemented by
+                NetworkXStorage. A query error is neither value: it propagates.
 
         Raises:
             Exception: If there is an error executing the query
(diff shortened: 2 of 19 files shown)


danielaskdd:
@codex review

PR #3588, opened 2026-08-07
fix(kg): return copies from NetworkXStorage get_node and get_edge

santhreal:
## Summary

`NetworkXStorage.get_node` / `get_edge` return live NetworkX attribute dicts. Callers that mutate those dicts corrupt in-memory graph state that `index_done_callback` can persist.

## Problem

```python
node = await storage.get_node("Alice")
node["entity_type"] = "mutated"  # changes the live graph node attrs
```

`get_all_nodes` already returns copies, and JsonKV / JsonDocStatus copy-on-read already covers the same class of bug (#3556). `get_node` / `get_edge` did not.

## Fix

Return a shallow `dict(...)` copy from both read paths so external mutation cannot touch live graph attributes.

## Test plan

- [x] `pytest tests/kg/networkx_impl/test_networkx_get_node_edge_copy_on_read.py`

### DIFF
--- a/lightrag/kg/networkx_impl.py
+++ b/lightrag/kg/networkx_impl.py
@@ -230,7 +230,11 @@ class NetworkXStorage(BaseGraphStorage):
 
     async def get_node(self, node_id: str) -> dict[str, str] | None:
         graph = await self._get_graph()
-        return graph.nodes.get(node_id)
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3589, opened 2026-08-07
fix(operate): compute most common entity type and normalize keywords on rebuild

santhreal:
## Summary

Rebuilding entities and relationships from cached extractions picked arbitrary entity types and produced duplicated keyword tokens.

## Problem

1. `_rebuild_single_entity` deduplicated `entity_types` before counting, so every remaining type had count 1 and the chosen type was insertion-order arbitrary instead of majority.
2. `_rebuild_single_relationship` joined unsplit comma-separated keyword strings with `", ".join(set(keywords))`, so overlapping tokens duplicated (`"AI, ML"` + `"ML, DL"` → `"AI, ML, ML, DL"`).

## Fix

- Count entity types with `Counter` before dedup, matching `_merge_nodes_then_upsert`.
- Split keyword strings on commas, strip, dedup, and join sorted, matching `_merge_edges_then_upsert`.

## Test plan

- [x] `pytest tests/extraction/test_rebuild_synthesis.py`

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -1848,9 +1848,9 @@ async def _rebuild_single_entity(
     # Get most common entity type from raw entity_types list before deduplicating
     if entity_types:
         type_counts = Counter(entity_types)
-        entity_type = sorted(
-            type_counts.items(), key=lambda x: x[1], reverse=True
-        )[0][0]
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3590, opened 2026-08-07
fix(kg): return boolean values from get_all_update_flags_status

santhreal:
## Summary

In single-process mode, `get_all_update_flags_status` returned `MutableBoolean` objects instead of Python `bool`s. That breaks JSON encoding and callers that expect plain booleans.

## Problem

The helper branched on `_is_multiprocess` and appended the raw flag object in the single-process path. The document route then had to special-case `.value` / `hasattr` conversion before responding.

## Fix

Always read `flag.value` in `get_all_update_flags_status`, and let the document route use the returned status directly.

## Test plan

- [x] `pytest tests/kg/test_get_all_update_flags_status.py`

### DIFF
--- a/lightrag/api/routers/document_routes.py
+++ b/lightrag/api/routers/document_routes.py
@@ -5924,19 +5924,9 @@ def create_document_routes(
             )
 
             # Get update flags status for all namespaces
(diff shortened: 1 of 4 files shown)


danielaskdd:
@codex review

PR #3594, opened 2026-08-07
fix(operate): match bidirectional relationship pairs during query truncation

santhreal:
## Summary

Query token truncation dropped undirected relationships when the stored edge direction was reversed relative to `relations_context`, which also dropped their associated text chunks from the LLM context.

## Problem

`_apply_token_truncation` built `final_relation_pairs` in one direction only and required an exact `(src, tgt)` match. For undirected query modes, retrieved edges can arrive as `(tgt, src)`. Those edges were filtered out, and `convert_to_user_format` fell back to placeholders when the reverse key was missing.

## Fix

Index and match both `(entity1, entity2)` and `(entity2, entity1)` in truncation filtering and in `convert_to_user_format` fallback lookup.

## Test plan

- [x] `pytest tests/extraction/test_query_truncation_bidirectional_relation.py`

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -5085,9 +5085,7 @@ async def _apply_token_truncation(
     if relations_context:
         final_relation_pairs = {
             (r["entity1"], r["entity2"]) for r in relations_context
-        } | {
-            (r["entity2"], r["entity1"]) for r in relations_context
-        }
+        } | {(r["entity2"], r["entity1"]) for r in relations_context}
         seen_edges = set()
         for relation in final_relations:
             src, tgt = relation.get("src_id"), relation.get("tgt_id")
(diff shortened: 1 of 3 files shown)


danielaskdd:
Thanks for the patch, but I think this fixes the wrong layer.

The graph storage contract is explicitly undirected — `BaseGraphStorage` states it outright: *"All operations related to edges in graph should be undirected."* (`lightrag/base.py:492`). Every backend implements it that way: Neo4j matches `-[r:DIRECTED]-` undirected and keys results by the **requested** pair (`kg/neo4j_impl.py:881-903`), pgtable canonicalizes to `(min, max)` and maps back to the requested direction (`kg/pgtable_impl.py:1391-1413`), and the base `get_edges_batch` fallback does the same per pair (`base.py:634-650`). So a lookup in either direction returns the same edge, oriented the way the caller asked for it.

That means the query layer never sees a reversed edge, and it shouldn't have to defend against one. If a backend ever did return an edge that doesn't match the requested direction, that's a bug in **that storage implementation** and must be fixed there — teaching `operate.py` / `utils.py` to accept both orientations would just hide a broken backend from every other caller that relies on the same contract.

For what it's worth, the specific scenario in the test can't arise either: `relations_context` and `relation_id_to_original` are built in the same loop from the same `final_relations` list (`operate.py:5013-5031`), and `_perform_kg_search` already dedupes by `tuple(sorted(...))` (`operate.py:4911-4929`), so `(A,B)` and `(B,A)` can never both reach truncation. Feeding this PR's own `search_result` to the unpatched function keeps both relations and resolves them to the original DB data; the test fails pre-patch only because it hand-builds a `relations_context` entry with flipped `entity1`/`entity2`, which the pipeline cannot produce.

If you hit this against a real backend, could you say which one? I'd rather fix the undirected contract violation at its source.

---

PR #3595, opened 2026-08-07
fix(kg): hydrate track docs via shared row helper and catch TypeError

santhreal:
## Summary

`JsonDocStatusStorage.get_docs_by_track_id` reimplemented row hydration inline and only caught `KeyError`, so malformed rows raised `TypeError` instead of being skipped like the other read paths.

## Problem

Sibling helpers already route through `_doc_processing_status_from_row`. The track-id path duplicated normalization and treated only missing keys as non-fatal, so bad field types aborted the whole listing.

## Fix

Reuse `_doc_processing_status_from_row` and catch `(KeyError, TypeError)` so unhydratable rows are logged and skipped.

## Test plan

- [x] `pytest tests/kg/json_impl/test_json_doc_status_track_id_hydration.py`


danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3598, opened 2026-08-07
fix: resolve real Ollama embedding dimension instead of a single hardcoded default

### LINKED ISSUE #3596: [Bug]: Ollama embedding binding assumes 1024-dim output for all models breaks with nomic-embed-text (768-dim)

### Describe the bug

Using `EMBEDDING_BINDING=ollama `with `EMBEDDING_MODEL=nomic-embed-text` fails on the first document insert unless `EMBEDDING_DIM=768 `is set explicitly. Without it, LightRAG assumes 1024 dimensions, and the vector store rejects the real 768-dim vectors it receives.

### Steps to reproduce

1. `.env:` `EMBEDDING_BINDING=ollama`, `EMBEDDING_MODEL=nomic-embed-text` (no `EMBEDDING_DIM` set)
2. Start lightrag-server, insert any document via POST /documents/texts
3. Insert fails:
```
NanoVectorDBStorage[chunks] index flush failed: Embedding dimension mismatch detected: total elements (N) cannot be evenly divided by expected dimension (1024).
```

4. Confirmed directly against Ollama that nomic-embed-text actually returns 768-dim vectors:
```
curl localhost:11434/api/embed -d '{"model":"nomic-embed-text","input":"test"}'
 # -> 768-element embedding
```

### Expected Behavior

Auto-detect the embedding dimension from the model's actual output (one probe call at startup) instead of hardcoding a 1024 default for every Ollama embedding model. If auto-detection isn't desired, fail fast at startup with a clear "set EMBEDDING_DIM explicitly for this model" message rather than defaulting silently to a wrong value that only surfaces as a failure mid-ingestion.

### LightRAG Config Used
```
EMBEDDING_BINDING=ollama
EMBEDDING_MODEL=nomic-embed-text
EMBEDDING_BINDING_HOST=http://localhost:11434
# EMBEDDING_DIM intentionally not set — LightRAG defaults to 1024 here,
# which is wrong for nomic-embed-text (actually 768). Setting
# EMBEDDING_DIM=768 explicitly is the workaround, not part of the bug state.

LLM_BINDING=ollama
LLM_MODEL=gemma4:12b
LLM_BINDING_HOST=http://localhost:11434
```

### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version: 1.5.6
- Operating System: macOS
- Python Version: 3.11
- Related Issues: #3597

Vishwamitra:
## Description

The Ollama embedding binding used a single static `embedding_dim` (1024, tuned for `bge-m3`, the binding's default model) for every model. That default is silently wrong for any other model a user points `EMBEDDING_MODEL` at — e.g. `nomic-embed-text` is actually 768-dim — and the mismatch only surfaces later, mid-ingestion, as a confusing vector-store "dimension mismatch" error with no link back to the real cause.

This PR resolves the real dimension for known Ollama embedding models instead of guessing, and turns the failure mode for unrecognized models from silent into a clear warning.

## Related Issues

Fixes #3596

## Changes Made

- Added `KNOWN_OLLAMA_EMBEDDING_DIMS` (`lightrag/llm/binding_options.py`) — a lookup table of real output dimensions for common Ollama embedding models, keyed by model name with the tag suffix stripped (e.g. `nomic-embed-text`, not `nomic-embed-text:latest`). Placed here rather than in `lightrag/llm/ollama.py` so resolving it doesn't require importing the optional `ollama` SDK.
- Updated `create_optimized_embedding_function()` (`lightrag/api/lightrag_server.py`) to consult this table when `EMBEDDING_DIM` isn't set explicitly and the binding is `ollama`. Priority order: explicit `EMBEDDING_DIM` > known-model lookup > previous `bge-m3`-shaped provider default.
- Unrecognized models keep the previous default (no behavior change for anyone currently relying on it) but now log a clear warning naming the actual model and pointing at `EMBEDDING_DIM`, instead of failing silently until a confusing error appears mid-ingestion.
- Added `tests/api/config/test_api_config_ollama_embedding_dim.py` (5 tests): known-model resolution (the exact `nomic-embed-text` regression), tag-suffix handling, the default model is unaffected, unknown models fall back with a warning, and explicit `EMBEDDING_DIM` still wins over the lookup table.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (if necessary) — n/a, no user-facing docs describe this default

## Additional Notes

Verified against the full offline test suite (`python -m pytest tests -m offline`): 4682 passed, 0 regressions. The only failures in that run (17, in `tests/setup/`) are pre-existing and environmental — macOS ships Bash 3.2 by default, and `scripts/setup/setup.sh` requires Bash 4+; confirmed unrelated to this change (none of the failing test files reference `embedding_dim` or `ollama`).

`pre-commit run --all-files` passes clean.

### DIFF
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -713,11 +713,35 @@ def create_optimized_embedding_function(
     # Step 2: Apply priority (user config > provider default)
     # For max_token_size: explicit env var > provider default > None
     final_max_token_size = args.embedding_token_limit or provider_max_token_size
-    # For embedding_dim: user config (always has value) takes priority
-    # Only use provider default if user config is explicitly None (which shouldn't happen)
-    final_embedding_dim = (
-        args.embedding_dim if args.embedding_dim else provider_embedding_dim
-    )
+    # For embedding_dim: user config > known-model lookup (Ollama only) > provider default.
+    # provider_embedding_dim is a single static value per provider (1024 for Ollama,
+    # tuned for its default model bge-m3) -- correct for that one model, silently wrong
+    # for most others a user might point EMBEDDING_MODEL at instead. The Ollama API
+    # exposes no dimension metadata to check ahead of a real embed call, so for known
+    # models we resolve the real dimension from a lookup table instead of guessing;
+    # for unrecognized models we keep the provider default but warn, so a mismatch
+    # surfaces as a clear message here rather than a confusing failure mid-ingestion.
+    final_embedding_dim = args.embedding_dim
+    if not final_embedding_dim and binding == "ollama" and model:
+        from lightrag.llm.binding_options import KNOWN_OLLAMA_EMBEDDING_DIMS
+
+        base_model_name = model.split(":")[0]
+        known_dim = KNOWN_OLLAMA_EMBEDDING_DIMS.get(
+            model
+        ) or KNOWN_OLLAMA_EMBEDDING_DIMS.get(base_model_name)
+        if known_dim is not None:
+            final_embedding_dim = known_dim
+        else:
+            final_embedding_dim = provider_embedding_dim
+            logger.warning(
+                f"Ollama embedding model '{model}' has an unknown output dimension; "
+                f"defaulting to {provider_embedding_dim} (matches bge-m3, the Ollama "
+                f"binding's default model). If document insertion fails with an "
+                f"'Embedding dimension mismatch' error, set EMBEDDING_DIM explicitly "
+                f"to this model's real output size."
+            )
+    if not final_embedding_dim:
+        final_embedding_dim = provider_embedding_dim
     # Asymmetric embedding is explicit opt-in only. Provider-specific
     # validation decides whether task parameters or prefixes are required.
     asymmetric_opt_in = resolve_asymmetric_embedding_opt_in(
--- a/lightrag/llm/binding_options.py
+++ b/lightrag/llm/binding_options.py
@@ -527,6 +527,30 @@ class OllamaEmbeddingOptions(_OllamaOptionsMixin, BindingOptions):
     _binding_name: ClassVar[str] = "ollama_embedding"
 
 
+# Known output dimensions for common Ollama embedding models, keyed by model
+# name without a tag suffix (e.g. "nomic-embed-text", not
+# "nomic-embed-text:latest"). The Ollama API exposes no dimension metadata to
+# check ahead of a real embed call, and ollama_embed's
+# @wrap_embedding_func_with_attrs (lightrag/llm/ollama.py) needs a single
+# static default at import time -- 1024 is correct for bge-m3 (the binding's
+# default model) but silently wrong for most others. This lives here rather
+# than in lightrag/llm/ollama.py so resolving it doesn't require importing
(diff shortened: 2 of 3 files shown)


danielaskdd:
Thanks for tackling #3596 — the known-model lookup table is a real improvement over a single hardcoded default for every Ollama model. That said, I don't think this should be merged as-is: it doesn't address a more fundamental issue with how `embedding_dim` is resolved for Ollama.

**Dimension shouldn't be resolved by guessing "the model's default output size" at all** — not via a live probe call, and not via a static lookup keyed on model name. Several embedding models (via Matryoshka Representation Learning) support returning a vector *smaller* than their default dimension when the caller explicitly requests it. That requires the configured `EMBEDDING_DIM` to actually be sent to the model on every call, not just used for local bookkeeping/validation.

LightRAG's OpenAI binding already does this correctly:

```python
# lightrag/llm/openai.py
if embedding_dim is not None:
    api_params["dimensions"] = embedding_dim
```

Ollama's own client supports the equivalent — I checked the installed `ollama` package directly:

```python
async def embed(self, model='', input='', truncate=None, options=None,
                 keep_alive=None, dimensions: Optional[int] = None) -> EmbedResponse:
    ...  # forwarded straight into the POST /api/embed body
```

But `lightrag/llm/ollama.py::ollama_embed()` never forwards it:

```python
data = await ollama_client.embed(model=embed_model, input=texts, options=options)
```

`ollama_embed()`'s signature doesn't even accept an `embedding_dim` parameter, and `create_optimized_embedding_function`'s Ollama branch in `lightrag_server.py` never includes `embedding_dim` in the kwargs it builds for the Ollama call path — unlike its OpenAI/Azure/Jina/Gemini/VoyageAI branches, which all do. So today, an explicitly-set `EMBEDDING_DIM` for an Ollama binding only affects LightRAG's internal bookkeeping (default resolution + the post-hoc size check in `EmbeddingFunc.__call__`); it is never actually requested from the model.

This PR's lookup table fixes the wrong-default problem for a handful of named models, but leaves that deeper gap in place — and risks masking it, since a "resolved" dimension for a known model now looks more authoritative even though it's still never enforced against what Ollama is actually asked to return. Recommend holding off on merging until `ollama_embed()` forwards `embedding_dim` through to the client's `dimensions` parameter when explicitly set, mirroring the existing OpenAI binding.


---

danielaskdd:
Closing this PR. As discussed above, the underlying problem isn't just "which hardcoded default is correct for a given Ollama model" — it's that LightRAG never forwards `EMBEDDING_DIM` to Ollama's actual embedding call at all, so the declared dimension is never enforced against what the model returns. A lookup table for known models papers over the symptom without fixing that gap.

I've filed that as a separate, more complete issue: #3600 (Ollama embedding binding never forwards `EMBEDDING_DIM` to the model — dimension truncation is silently ignored).

@Vishwamitra would you be interested in taking that one on instead? It covers both the wrong-default problem this PR fixes and the missing `dimensions` forwarding, using the same pattern the OpenAI binding already implements (`api_params["dimensions"] = embedding_dim`). Happy to review if you'd like to pick it up.


---

Vishwamitra:
Sure i will look into this issue. Thanks for highlighting. @danielaskdd

PR #3605, opened 2026-08-08
fix(parser): bound native markdown image downloads per document (GHSA-25c3-j78v-83qx)

danielaskdd:
## Summary

Bounds the native markdown parser's external-image downloader in the three dimensions it was missing — bytes, request count, and wall clock — and makes an in-flight download cancellable. Fixes the three defects in the private advisory **GHSA-25c3-j78v-83qx** (High, CVSS 7.5, `AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`), reported by @santhreal.

Two further escape hatches found while implementing are closed here as well (see *Beyond the report*).

## Motivation

`NATIVE_MD_IMAGE_MAX_BYTES` (25 MiB) bounds **one** image. Nothing bounded a document, so a tiny upload could pin the host:

1. **No aggregate budget.** Every image's bytes stayed resident until `extract_markdown()` returned. A 4,170-byte `.textpack` produced 800 outbound GETs and 800 MiB of resident image data.
2. **The timeout was not a deadline.** `NATIVE_MD_IMAGE_DOWNLOAD_TIMEOUT` reached `opener.open` as urllib's *per-socket-operation* timeout, and the body was read in a single `resp.read(max_bytes + 1)`. A peer emitting one byte per interval resets that timeout forever. The blocking call holds a thread of the native parse pool (default 5) that no cancellation path could reclaim — `grep -rn cancel lightrag/parser/markdown/` returned nothing, so `POST /documents/cancel_pipeline` did not help and only a process restart did.
3. **No request cap.** The resolver memo keys on the literal `src`, so `?u=<i>` defeats it.

Preconditions are only "can POST a file": no authentication in the default configuration, `.textpack` routes to the native engine with no config, downloading is on by default. This is orthogonal to `GHSA-vv3m-f8x4-7377` (an SSRF *destination* bypass) — every measurement below targets an address that guard explicitly permits.

## What's changed

Seven commits, each independently green.

| # | Commit | |
|---|---|---|
| 1 | `refactor(parser): share cancel-event normalization` | Extract `normalize_cancel_events()` / `first_cancellation()` from `SyncLLMBridge`. No behaviour change. |
| 2 | `fix(parser): close the redirect-drain and non-HTTP escape hatches` | See *Beyond the report*. |
| 3 | `fix(parser): bound the native markdown image fetch by a wall clock` | Defect 2. |
| 4 | `fix(parser): add per-document image byte and request budgets` | Defects 1 and 3. |
| 5 | `fix(parser): add a per-document image download time budget` | |
| 6 | `fix(parser): honour pipeline cancellation during markdown image downloads` | Defect 2, cancellation half. |
| 7 | `docs(parser): document the per-document native markdown image budgets` | |

### New configuration

Defaults derived from `DEFAULT_MAX_PARALLEL_PARSE_NATIVE` (5), not picked to look generous:

| Variable | Default | |
|---|---|---|
| `NATIVE_MD_IMAGE_MAX_TOTAL_BYTES` | 64 MiB | Image bytes **retained** per document, all sources. 64 × 5 workers ≈ 320 MiB attacker-controlled. |
| `NATIVE_MD_IMAGE_MAX_REQUESTS` | 100 | Remote fetch **attempts** per document. |
| `NATIVE_MD_IMAGE_DOWNLOAD_TOTAL_TIMEOUT` | 120 s | Wall clock across all downloads; clamps every per-request deadline. |

`NATIVE_MD_IMAGE_DOWNLOAD_TIMEOUT` (30 s) is reinterpreted as a real per-request wall-clock deadline. Strictly a tightening.

Over-budget remote images degrade to external links with a dedicated warning counter — the document does not fail. Base64 and `.textpack` images have no URL to fall back to and are skipped, under their own counter rather than `images_skipped`. With `DOWNLOAD_REQUIRED=true` a budget stop raises, matching that switch's documented "a download error fails the document".

`0` and negatives are **not** "unlimited" — they fall back to the default. An operator typing `0` into a ceiling means strictest, and inverting that would turn a typo into the exhaustion the budget prevents.

### What each phase of a fetch can now do

| Phase | Deadline-bounded | Cancellable |
|---|---|---|
| DNS resolution | ✗ (system resolver) | ✗ |
| TCP connect | ✓ | ✓ (≤ 0.2 s poll) |
| TLS handshake | ✓ | ✓ |
| Response headers | ✓ | ✓ |
| Response body | ✓ | ✓ |
| Redirect hop | ✓ | ✓ |

**DNS is the one exception, on both counts.** It runs before any socket exists, so neither the deadline nor the watchdog can reach it. The honest bound is "deadline plus one DNS resolution", and a cancel is observed at worst one resolution plus one poll interval late. `_resolve_shared()` memoises per host so a fetch resolves once rather than twice, halving that window, and brackets the resolve with cancellation checkpoints.

Three implementation notes worth a reviewer's attention:

- **TCP connect cannot be aborted from another thread.** Measured: `shutdown()` during a blocking connect makes `connect()` return *successfully* on macOS, and `close()` does nothing. So connect uses non-blocking `connect_ex` + a selector poll driven by the calling thread. `connect_deadline` is `min(fetch deadline, now + timeout)` so the `timeout` argument is still honoured with no active context, where `create_connection(..., timeout)` used to be the bound.
- **`wrap_socket` detaches the socket it wraps** (the original object's `fileno()` becomes -1), so a handshake performed inside it leaves the watchdog with nothing to shut down. The handshake is now driven explicitly after registering the `SSLSocket`.
- **`trip_reason()` re-evaluates synchronously** — cancellation, then the document budget, then the request deadline, and only then whatever the watchdog recorded. The watchdog polls on an interval, so trusting its record alone would misclassify every race it loses: a socket timeout arriving first reads as an ordinary download failure, and a cancellation followed by a peer EOF reads as a *successful truncated image* (a half-closed socket makes `read()` return `b''`, not raise).

Ordering is now **validate/convert → charge → cache**: `raw_cache.put()` moved below the budget check, so bytes a budget rejected are never written to `.native_raw/` and cannot be served back for free on a re-parse.

## Beyond the report

Two channels that bypassed every control above, found while implementing and fixed in commit 2:

1. **Unbounded redirect-body drain.** `urllib.request.HTTPRedirectHandler.http_error_302` calls `fp.read()` with no limit before following a jump, so `302` + a huge body is an attacker-controlled allocation landing *before* the image reaches the size-capped read. We now read the `Location`, close the body, and follow.
2. **`ftp://` redirects.** The stdlib redirect allowlist permits `ftp`, and `build_opener()` installs `FTPHandler` / `FileHandler` / `DataHandler`. An `ftp://` redirect is served by `FTPHandler`, which never goes through `_GuardedHTTPConnection` — no IP pin, no deadline, no cancellation, no request budget. The redirect allowlist is narrowed to http(s) and the opener is assembled by hand with only the handlers we need.

A trap worth recording: `http_error_301 = http_error_303 = http_error_307 = http_error_308 = http_error_302` in the base class creates **five independent attributes**, so overriding only `http_error_302` in a subclass leaves the other four resolving to the original draining implementation. Verified, and all five are rebound. `max_redirections` also drops from 10 to 5.

**Behaviour change to note:** the image opener no longer speaks `ftp://`, `file://` or `data:` at the URL level. Previously an `ftp://` redirect target would be fetched.

## Verification

Full suite: **5791 passed, 195 skipped**. `pre-commit run` clean on every changed file.

Reproduced the advisory's own probes against a loopback listener, before and after:

| Probe | Pre-fix | Post-fix |
|---|---|---|
| 4,171-byte `.textpack`, 800 images | 800 GETs, **800 MiB held**, RSS 109 → **933 MiB**, 31.7 s | 100 GETs, **64 MiB held**, RSS 107 → 190 MiB, 3.6 s |
| Trickling peer, 5 s configured timeout | **still blocked at 45.1 s** | **5.05 s** |
| 5 trickling docs + 1 benign, 5-worker pool | `started=[0,1,2,3,4] finished=[] benign_started=False` | all finished, benign document ran |
| `302` advertising a 1 GiB body | 60.1 s, **image lost** | 0.0 s, image fetched |

The pre-fix column reproduces the reporter's numbers.

The RSS figures above are empirical, not a guarantee: process RSS follows the allocator's high-water mark. What the code *bounds* is live image allocations — `MAX_TOTAL_BYTES + min(MAX_BYTES, MAX_TOTAL_BYTES)`, ~89 MiB at the defaults — because with `R` retained and `C` in flight the read cap `C ≤ min(MAX_BYTES, MAX_TOTAL_BYTES − R)` makes the join moment `(R + C) + C` where `R + C ≤ MAX_TOTAL_BYTES`. A parametrized `tracemalloc` test pins that formula and fails on any read path that costs one extra live copy.

### Tests

Roughly 90 new tests across four files, plus a package-level `conftest.py` — the suite spans four modules and `tests/conftest.py` loads a developer's real `.env`, so a locally-tuned `NATIVE_MD_IMAGE_*` would otherwise decide whether these pass.

Every regression test fails **behaviourally** on the pre-fix tree, not with an `AttributeError`:

- Loopback listeners **trickle** rather than stalling outright. A peer that sends nothing is already handled by the socket timeout and proves nothing; one byte per interval is what resets it forever. The header and body cases hang indefinitely pre-fix, as does the end-to-end cancellation test.
- The end-to-end cancellation test drives the production path — `get_parser("native").parse(ParseContext(..., pipeline_cancel_event=...))` — with a 300 s deadline so the cancel, not the deadline, is what ends it.
- Redirect no-drain is asserted at handler level with a fake `fp` whose `read()` raises, parametrized over all five status codes. Deterministic; the RSS evidence stays in the manual probe above.
- Connect deadline and cancellation use an injected `EINPROGRESS` socket and a never-writable selector rather than a real blackhole address, which would be network-environment dependent.
- A real TLS handshake against a self-signed server, because nothing else proves the split handshake did not break ordinary HTTPS.

## What is not fixed here

- **Streaming asset bytes to `asset_dir`** (the report's fix 2, O(1) resident instead of O(capped total)). The byte budget already caps memory; that change additionally renames `.textpack` asset files to content-addressed form and so shifts `blockid` / `document_hash` for those documents. That is a format migration and does not belong in a security PR. Follow-up.
- **Budgets are per document, not global.** Worst case remains `max_parallel_parse_native` × the per-document budget. A process-wide budget is the natural next step and is cheap now the counters exist.
- **DNS is neither deadline-bounded nor cancellable** (above).
- A minimum-throughput floor and `NATIVE_MD_IMAGE_ALLOWED_HOSTS` (the report's fixes 5 and 6) are hardening, not bounds, and are left out.

## For the reviewer

`LLMBridgeCancelled` is now raised from a non-LLM path. It is the right class — `pipeline.py:4441` catches exactly that family to record a document as *cancelled* rather than *failed*, and a new exception type would be misreported — but the name now overreaches. Renaming the hierarchy to `ParseCancelled` / `ParsePipelineCancelled` with back-compat aliases is a wider change than a security fix should carry; happy to do it separately if you prefer.

## Upgrade / release notes

- **Byte budget applies to cache hits.** Documents parsed before this PR may hold more than `NATIVE_MD_IMAGE_MAX_TOTAL_BYTES` (default 64 MiB) of images in `.native_raw/`. On their next re-parse the excess degrades to external links, and with `NATIVE_MD_IMAGE_DOWNLOAD_REQUIRED=true` the re-parse fails deterministically even though every byte is already on disk. Recovery is one env change — raise `NATIVE_MD_IMAGE_MAX_TOTAL_BYTES` — and needs no cache invalidation: the options signature deliberately excludes the budgets, so the bundle stays valid and the next parse succeeds from cache hits alone.
- **`http(s)` → `ftp` redirects are no longer followed.** The stdlib opener carried an `FTPHandler` and its redirect allowlist included `ftp`, so a 302 to `ftp://` was followed — bypassing the IP pin entirely (FTPHandler re-resolves DNS itself, a rebinding hole). Such images now degrade to external links (or fail the document under `DOWNLOAD_REQUIRED`).


danielaskdd:
Thanks — all four verified and addressed. Pushed as `ac119d18` (behaviour) and `6e08fc72` (claims). Head is now `6e08fc72`; full suite **5798 passed, 195 skipped**.

Follow-up issue #3608 for the `LLMBridgeCancelled` rename is noted, and I'll keep the old names as aliases there.

### 2, 3, 4 — fixed as described

**2. Memo hit bypassed cancellation.** Confirmed. The check moved to the first line of `resolve()`; it also now covers the data-URL and bundle-file paths, which do no network work of their own and so had no check at all. New test uses **one URL referenced 50 times** with the cancel landing *during* the single fetch, so the memo hits are the only path left — `DID NOT RAISE` on the previous head.

**3. Redirect resolved the target before closing and charging.** Confirmed, including the `fp.closed=False` detail. Order is now close → charge → resolve, and `redirect_request` gets `None` for `fp` since the body is already gone. Two tests: one asserts the call order is `["close", "resolve"]` and that a hop the SSRF guard refuses is still counted (`requests_used == 1`); the other asserts an exhausted budget performs **zero** resolutions. Both fail on the previous head with `['resolve','close']` and `1 == 0`.

**4. `DOWNLOAD_REQUIRED` doc mismatch.** Confirmed, and I took your first option: the switch governs remote images only, matching its existing documented carve-out that base64 / `.textpack` images are always embedded regardless of it. Corrected the parenthetical, not the behaviour.

### 1 — real, but the diagnosis points elsewhere

The doubling is real and your number reproduces. Two things it is not, both measured on CPython 3.12 with a 64 MiB body and a 64 MiB cap:

| read strategy | peak |
|---|---|
| chunked list + `b"".join` (current) | 128.2 MiB |
| single bounded `fp.read(cap+1)` | 128.1 MiB |
| prealloc + `readinto` + `bytes()` | 192.1 MiB |

**It is not caused by chunking.** `BufferedReader.read(n)` accumulates and joins internally, so a single bounded read peaks identically; pre-allocating is markedly worse because it pays the full cap up front *and* the conversion. Both suggested alternatives are therefore no better, and one regresses. The doubling is inherent to producing an immutable `bytes` from a stream — removing it needs the asset bytes to stop being heap-retained at all, i.e. the streaming-to-`asset_dir` follow-up.

I did check whether a `bytearray` could avoid the final copy. It cannot be carried through: `writer.py:_materialize_assets` dispatches on `isinstance(spec.source, bytes)`, and a `bytearray` would silently fall into the `None` branch → warn-and-skip → **the asset vanishes from the sidecar**. Not worth touching the shared writer for.

**It is not a regression.** The per-image read cap is `min(MAX_BYTES, remaining_bytes)`, never more than the pre-fix cap of `MAX_BYTES`, so the transient can only shrink as a document fills its budget. Your 131 MiB needs `MAX_BYTES=64 MiB`, 4× the shipped default; on the pre-fix tree that same configuration paid the same 128 MiB **and** retained without limit.

What *was* wrong is my stated gate. The honest bound is `MAX_TOTAL_BYTES + 2 × min(MAX_BYTES, MAX_TOTAL_BYTES)` — ~114 MiB per worker at the defaults, not `1.5 × MAX_TOTAL_BYTES`. That is now in the `_ImageBudget` docstring and in `env.example` with the measurements and the reason, and it is pinned by the peak-memory regression test you asked for: parametrized over three `MAX_BYTES`/`MAX_TOTAL_BYTES` combinations (per-image cap binding, remaining budget binding, one image able to fill the budget), plus one asserting the read cap actually tracks the remaining budget (`[3 KiB, 2 KiB, 1 KiB]` across three images, never the 8 KiB per-image ceiling). A read path that regressed to three live copies fails all of them.

If you would rather the shipped defaults satisfy a ≤ 96 MiB per-worker peak outright, the lever is `MAX_BYTES` rather than `MAX_TOTAL_BYTES` — say the word and I'll lower the per-image default too, at the cost of rejecting genuinely large single images.

danielaskdd:
Confirmed and fixed in `c14bd985`. Your algebra is right and mine double-counted: with `R` retained and `C` in flight the read cap is `C ≤ min(M, T−R)`, so the join moment is `R + 2C = (R + C) + C` where `R + C ≤ T` — the in-flight copy adds `min(M, T)`, not twice it.

Measured live image data across four configurations:

| T | M | measured | `T + min(M,T)` | old `T + 2·min(M,T)` |
|---|---|---|---|---|
| 64 | 64 | 128.2 MiB | 128.0 | 192.0 |
| 64 | 25 (defaults) | 78.1 MiB | 89.0 | 114.0 |
| 8 | 2 | 10.1 MiB | 10.0 | 12.0 |
| 4 | 16 | 8.1 MiB | 8.0 | 12.0 |

The tight formula holds, with a small constant over it — the `BufferedReader`'s 64 KiB buffer, the chunk list, and the `cap + 1` extra byte. So the shipped defaults bound live image data at ~89 MiB, which is *below* the 96 MiB figure the plan started from.

**Your point about the fixture is the more important one.** The old test could not have caught this: `_FakeResponse` held each body in full, so the harness owned one extra copy per image and the `2×` allowance was partly covering the test's own data rather than the parser's. The fake now generates its body chunk by chunk and retains nothing (the first chunk still carries the index, so sha256 distinctness is preserved), and the allowance is the tight formula plus slack sized only for the read machinery.

I then checked the test actually discriminates, since a bound that cannot fail is worse than no bound. Holding one genuine extra copy across the join (`_leak = bytearray(data)`) fails all three parametrizations — 12.0 MiB against 9.0 / 11.0 / 9.0 — and removing it passes. Worth noting my first attempt at that injection, `bytes(data)`, proved nothing: CPython returns the same object for an exact `bytes` argument, so the "extra copy" was a no-op and the test passed. Same trap as the `readinto` measurement — worth being suspicious of any peak-memory experiment that doesn't move the number.

**Live allocations vs RSS** are now separated in the `_ImageBudget` docstring, in `env.example` and in the PR description: the formula bounds live Python allocations, RSS follows the allocator's high-water mark and no algebra here can promise it. The RSS figures in the probe table stay, labelled as empirical measurements rather than a guarantee.

`MAX_BYTES` keeps its default, agreed — lowering it would not establish an RSS ceiling either, for exactly the reason you give, and that question belongs with the streaming-to-`asset_dir` follow-up where the transient goes away instead of being traded down.

Head is now `c14bd985`; `tests/parser` + `tests/pipeline` **2069 passed**, `pre-commit` clean.

PR #3606, opened 2026-08-08
fix(neo4j): omit non-existent edges in get_edges_batch and support general relationships

santhreal:
### Summary
Fix `Neo4JStorage.get_edges_batch` contract violations where querying non-existent edges returned dummy records with `None` attributes instead of omitting missing pairs, and where edge lookups were overly constrained to `:DIRECTED` relationship types.

### Problem & Motivation
In `Neo4JStorage.get_edges_batch`:
1. Querying missing edge pairs populated `edges_dict[(src, tgt)]` with dummy dictionaries `{ "weight": 1.0, "source_id": None, "description": None, "keywords": None }` instead of returning an empty dict / omitting missing pairs.
2. The Cypher query matched `-[r:DIRECTED]-`, which failed to match relationships created without explicit `:DIRECTED` label, whereas `get_edge()` matches `-[r]-`.

During RAG retrieval (`_get_edge_data`), when candidate relationships from vector search were checked against graph storage, deleted or non-existent edges returned these dummy records with `None` fields, polluting retrieved KG context and introducing `None` attributes into prompt builders.

### Changes
- Updated `Neo4JStorage.get_edges_batch` to match general relationships `-[r]-` (consistent with `get_edge()`).
- Removed dummy fallback dict insertion for missing edge pairs so `get_edges_batch` returns properties only for edges present in the graph, conforming to `BaseGraphStorage` and all other graph storage implementations (`NetworkXStorage`, `PGTableGraphStorage`, `PGGraphStorage`, etc.).
- Added unit test `test_get_edges_batch_omits_missing_edges` in `tests/kg/neo4j_impl/test_neo4j_graph_read_contract.py`.

### Verification
- `pytest tests/kg/neo4j_impl/` passed (43 passed).
- Verified `test_get_edges_batch_omits_missing_edges` fails on unpatched `main` and passes with this fix.
- Full `tests/kg/` and `tests/pipeline/` suites passed (2237 passed, 0 failures).

### DIFF
--- a/lightrag/kg/neo4j_impl.py
+++ b/lightrag/kg/neo4j_impl.py
@@ -880,7 +880,7 @@ class Neo4JStorage(BaseGraphStorage):
         ) as session:
             query = f"""
             UNWIND $pairs AS pair
-            MATCH (start:`{workspace_label}` {{entity_id: pair.src}})-[r:DIRECTED]-(end:`{workspace_label}` {{entity_id: pair.tgt}})
+            MATCH (start:`{workspace_label}` {{entity_id: pair.src}})-[r]-(end:`{workspace_label}` {{entity_id: pair.tgt}})
             RETURN pair.src AS src_id, pair.tgt AS tgt_id, collect(properties(r)) AS edges
             """
             result = await session.run(query, pairs=pairs)
@@ -901,14 +901,6 @@ class Neo4JStorage(BaseGraphStorage):
                         if key not in edge_props:
                             edge_props[key] = default
                     edges_dict[(src, tgt)] = edge_props
-                else:
-                    # No edge found – set default edge properties
-                    edges_dict[(src, tgt)] = {
-                        "weight": 1.0,
-                        "source_id": None,
-                        "description": None,
-                        "keywords": None,
-                    }
             await result.consume()
             return edges_dict
 
--- a/tests/kg/neo4j_impl/test_neo4j_graph_read_contract.py
+++ b/tests/kg/neo4j_impl/test_neo4j_graph_read_contract.py
@@ -116,6 +116,8 @@ async def test_get_node_edges_returns_connected_pairs():
         ("Alpha", "Beta"),
         ("Alpha", "Gamma"),
     ]
+
+
 @pytest.mark.asyncio
 async def test_get_edges_batch_omits_missing_edges():
     """get_edges_batch must omit non-existent edge pairs rather than returning dummy attributes."""
--- a/tests/kg/neo4j_impl/test_neo4j_graph_read_contract.py
+++ b/tests/kg/neo4j_impl/test_neo4j_graph_read_contract.py
@@ -116,3 +116,11 @@ async def test_get_node_edges_returns_connected_pairs():
         ("Alpha", "Beta"),
         ("Alpha", "Gamma"),
     ]
+@pytest.mark.asyncio
+async def test_get_edges_batch_omits_missing_edges():
+    """get_edges_batch must omit non-existent edge pairs rather than returning dummy attributes."""
+    storage, calls = _make_storage([{"src_id": "A", "tgt_id": "B", "edges": []}])
+    res = await storage.get_edges_batch([{"src": "A", "tgt": "B"}])
+    assert res == {}
+    query, _ = calls[0]
+    assert "-[r]-" in query


danielaskdd:
Thanks for the careful write-up and for adding a regression test. I looked into both halves of this change and I'd like to ask for the Cypher pattern change to be dropped. Details below.

## 1. The `else` branch is unreachable, so no dummy record ever reached the retrieval path

The PR description says querying a missing pair populates `edges_dict[(src, tgt)]` with `{"weight": 1.0, "source_id": None, ...}`, and that these `None` fields pollute the KG context built in `_get_edge_data`. That does not happen against a real Neo4j server:

```cypher
UNWIND $pairs AS pair
MATCH (start:`ws` {entity_id: pair.src})-[r]-(end:`ws` {entity_id: pair.tgt})
RETURN pair.src AS src_id, pair.tgt AS tgt_id, collect(properties(r)) AS edges
```

`MATCH` filters rows before `RETURN`, and `collect()` only aggregates over rows that survived. A pair with no matching relationship produces **zero rows**, so it never appears in the result at all and `async for record` never visits it. `edges` can never be an empty list here — that would require `OPTIONAL MATCH`, which this query does not use.

So the deleted `else` is dead code. Removing it is fine as contract alignment with `BaseGraphStorage.get_edges_batch` / `PGTableGraphStorage` / `PGGraphStorage` (and it would become a genuine defect the moment someone switches this query to `OPTIONAL MATCH` to distinguish "node missing" from "edge missing"), but it fixes nothing observable today. Please reword the description accordingly — as written it will mislead future readers about whether that branch was live.

Worth noting the new test only exercises this dead branch: the record shape `{"src_id": "A", "tgt_id": "B", "edges": []}` is not something Neo4j can produce. A case that feeds `_make_storage([])` and asserts `{}` would model the real missing-edge behavior.

## 2. Dropping `:DIRECTED` is a hot-path performance regression

`get_edges_batch` is a volume path, not an occasional call:

- `_find_most_related_edges_from_entities` (`lightrag/operate.py:5661`) passes every deduplicated neighbor pair from `get_nodes_edges_batch(node_names)`. With hub entities in the result set this is easily hundreds to thousands of pairs.
- `_get_edge_data` (`lightrag/operate.py:5870`) passes `top_k` candidate pairs (default 60) on every `global` / `hybrid` / `mix` query.

Neo4j stores relationships for dense nodes (default threshold 50, which KG hub entities cross routinely) in per-type relationship groups. With the type label, degree lookup reads a single group counter and expansion walks only the `DIRECTED` chain; without it, degree is a sum across all groups and expansion walks every chain. On a pure LightRAG graph — where `upsert_edge` only ever writes `MERGE (source)-[r:DIRECTED]-(target)` (`neo4j_impl.py:1221`, `1273`; same in `memgraph_impl.py:627`, `838`) — the two forms are near-equivalent. The degradation shows up precisely on the mixed-type graphs this change is meant to support, which are also the graphs with the largest `pairs` lists. That trade puts a cost on every user's hot path in exchange for a configuration we don't otherwise support.

## 3. Dropping `:DIRECTED` also introduces a silent correctness problem

If a pair is connected by more than one relationship type — a user's own `:MENTIONS` alongside LightRAG's `:DIRECTED` — `collect(properties(r))` gathers all of them and the code takes `edges[0]`. Which one that is depends on traversal order, so `description` / `keywords` / `source_id` may silently come from a foreign relationship. That is worse than returning nothing. Note `get_edge` at least logs `"Multiple edges found between ... Using first edge"` (`neo4j_impl.py:810`); `get_edges_batch` has no such warning.

## Requested changes

1. Revert `-[r]-` back to `-[r:DIRECTED]-`, and drop the `assert "-[r]-" in query` line from the new test.
2. Keep the `else` removal if you like, but rewrite the PR description so it is presented as contract alignment rather than a fix for polluted retrieval context.

If supporting imported non-`DIRECTED` graphs is the real goal, the right shape is to normalize the relationship type at ingest, or to make the type name a single configurable constant shared by the read and write paths — that keeps the type index and avoids the cross-type mix-up. Happy to discuss that as a separate PR.

Separately, and out of scope here: `get_edge` still returns a dummy `{"weight": 1.0, "source_id": None, "description": None, "keywords": None}` when property processing raises (`neo4j_impl.py:842`). That is a live source of `None` attributes in retrieval context, unlike the branch removed here.

For reference, I ran `tests/kg` + `tests/pipeline` on this branch: 2198 passed, 143 skipped, 10 failed — all 10 failures are `tiktoken` encoding downloads blocked by my sandbox proxy, unrelated to this change.

---

danielaskdd:
Thanks for the contribution and the regression-test effort. After reviewing the Cypher execution semantics, a regular `MATCH` already omits missing pairs, so the fallback branch cannot produce dummy records in a real Neo4j query. Broadening the pattern from `:DIRECTED` to every relationship type can instead mix unrelated relationships into LightRAG retrieval.

We are therefore closing this PR and carrying the safe contract cleanup, empty-input fast path, property-copy isolation, and tests forward in #3648.

Issue #3597, opened 2026-08-07
[Bug]:Entity extraction silently returns 0 entities against thinking-capable Ollama models (e.g. gemma4:12b), think is never disabled


Vishwamitra:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

With `LLM_BINDING=ollama` and a `reasoning/"thinking`" model (e.g. gemma4:12b), document ingestion can report status: "processed" with no error, while the knowledge graph ends up empty: 0 entities, 0 relationships. **Root cause:** the model spends its num_predict budget on its hidden reasoning trace before it gets to the actual answer, and LightRAG's Ollama binding never passes think=False to suppress that.

**Suggested fix**

- Pass `think=False` unconditionally for the extract/keyword roles in ollama.py; they need deterministic structured output, not a reasoning trace, so there's no case where thinking helps here.
- Or expose think as a configurable field on OllamaLLMOptions, defaulting to False for those two roles (leaving query free to use reasoning if a user wants it for the final answer).
- Either way: when `done_reason == "length"` and content is empty, that's a strong signal something is structurally broken, not just "ran a bit long"; worth escalating past a per-chunk warning so it doesn't fail silently.

### Steps to reproduce

- Configure .env with `LLM_BINDING=ollama`, `LLM_MODEL=gemma4:12b` (or any other thinking-capable Ollama model)
- lightrag-server, then POST /documents/texts with any short document
- Wait for status: `"processed" on /documents/track_status/{track_id}`
- `rag_storage/kv_store_full_entities.json` and `kv_store_full_relations.json` show `"count": 0` for every document. Server log shows, per chunk:
```
WARNING: Ollama response truncated by token limit (done_reason=length, content_len=0), returning partial content
WARNING: ...: Complete delimiter can not be found in extraction result
```

Standalone, no LightRAG import needed. Reproduces both the failure and the fix in one run against any thinking-capable Ollama model:

```
import time
import ollama

MODEL = "gemma4:12b"  # any Ollama "thinking"-capable model reproduces this
HOST = "http://localhost:11434"

text = open("your_test_document.txt").read()[:4000]
system_prompt = (
    "You are a Knowledge Graph Specialist. Extract entities and relationships "
    "from the given text. Output a JSON list."
)
user_prompt = (
    f"---Input Text---\n{text}\n\n"
[...]

PR #3607, opened 2026-08-08
fix(extraction): surface token-limit truncation and fail on empty truncated responses

### LINKED ISSUE #3601: [Bug]: Token-limit truncation is effectively invisible to operators — extraction silently yields 0 entities and the document still ends PROCESSED
### Describe the bug

The token-limit truncation marker introduced in #3446 / #3448 works: all four bindings detect it and wrap the payload in `TruncatedResponse`, and the cache layer correctly refuses to persist a partial extraction.

- `lightrag/llm/openai.py:809` — `finish_reason == "length"`
- `lightrag/llm/ollama.py:206` — `done_reason == "length"`
- `lightrag/llm/gemini.py:559`, `lightrag/llm/bedrock.py:457`
- `lightrag/utils.py:4428` — `TruncatedResponse`

What is missing is **observability**. When an LLM response is truncated during entity/relation extraction, the run still completes, the document still reaches `PROCESSED`, `/documents/track_status/{track_id}` still reports success, and the knowledge graph is silently short — or, in the worst case (`content_len == 0`), completely empty. The only trace is a `WARNING` line in the server log that carries no document or chunk identity.

This was surfaced concretely by #3597 (thinking-capable Ollama models burning the whole `num_predict` budget on the reasoning trace), where a full ingestion of 5 chunks produced 0 entities across two runs and reported success both times. #3597/#3599 address one specific cause of the empty output; this issue is about the failure being unobservable regardless of cause, which also covers the ordinary "`num_predict` too small for a dense chunk" case that no `think` setting can fix.

### Gaps

**1. The warning never reaches `pipeline_status`, so it is invisible in the WebUI and in `/documents/pipeline_status`.**

The only message the extraction stage publishes is `operate.py:4038-4040`:

```
Chunk {n} of {total} extracted {e} Ent + {r} Rel {chunk_key}
```

An operator watching the UI sees `0 Ent + 0 Rel` with no severity marker and no cause. Every truncation warning goes to `logger.warning()` only.

**2. The truncation marker is discarded before the extraction stage can see it.**

`lightrag/utils.py:4603-4605`:

```python
res_truncated = is_truncated_response(res)
res = remove_think_tags(res)     # rebuilds a plain str, marker gone
```

`extract_entities` therefore cannot distinguish "the model was cut off" from "the model returned nothing useful". The only downstream signal is the generic `{chunk_key}: Complete delimiter can not be found in extraction result` (`operate.py:1434`), which does not name the cause.

**3. The chunk-scoped warning only exists when the extraction cache is enabled.**

The one warning that carries `chunk_id` lives inside the cache-write guard (`utils.py:4615-4618`) and is therefore conditional on `enable_llm_cache_for_entity_extract`. With that flag off, execution takes the `utils.py:4637-4658` branch, which never calls `is_truncated_response()` at all — leaving only the binding-level warning, which has no document or chunk identity whatsoever.

**4. `content_len == 0` is not escalated, and the two main bindings disagree about it.**

- OpenAI binding: empty content is checked at `openai.py:750-799`, logged with `finish_reason` / `completion_tokens` / `reasoning_tokens` diagnostics and an actionable hint (`consider raising max_tokens or disabling thinking mode`), then raised as `InvalidResponseError`. The document ends `FAILED` and the operator cannot miss it.
- Ollama binding: `ollama.py:206-213` logs a warning and returns the empty string. Extraction yields nothing, the document ends `PROCESSED`.

Same failure, two very different outcomes. An empty response with `done_reason=length` is structurally broken, not merely "ran a bit long", and should not be reported as success.

### Steps to reproduce

1. Configure any binding with a deliberately small output budget (e.g. `OLLAMA_LLM_NUM_PREDICT=64`, or an OpenAI-compatible endpoint with `OPENAI_LLM_MAX_COMPLETION_TOKENS=64`).
2. Ingest a document dense enough that extraction output exceeds that budget.
3. Observe: `/documents/track_status/{track_id}` reports `processed`; `kv_store_full_entities.json` shows `"count": 0` (or a truncated count); the WebUI pipeline panel shows only `Chunk n of m extracted 0 Ent + 0 Rel`; the server log has `WARNING: ... response truncated by token limit (…, content_len=0), returning partial content` with no document or chunk identity attached.

### Expected Behavior

A truncated extraction is visible without reading the server log, and an empty truncated response is not reported as success.

### Proposed changes

1. **Publish truncation to `pipeline_status`.** Emit the truncation event through `status_logger.log()` alongside the existing per-chunk message, including `chunk_key` and the document's `file_path`, so it appears in the WebUI and in `/documents/pipeline_status` history. Ideally aggregate a per-document count so a run that truncated 12 of 40 chunks says so once, rather than only in 12 scattered log lines.
2. **Preserve the marker past `remove_think_tags`.** Either re-wrap the sanitized string in `TruncatedResponse` at `utils.py:4605`, or return the flag alongside the content, so `extract_entities` can attribute an empty/undelimited parse to truncation instead of emitting the generic `Complete delimiter can not be found` warning.
3. **Check truncation on the cache-disabled path too.** The `utils.py:4637-4658` branch should run the same `is_truncated_response()` check and emit the same chunk-scoped warning; today truncation is unobservable at chunk granularity whenever `enable_llm_cache_for_entity_extract=false`.
4. **Align the empty-content behavior across bindings.** When `content` is empty and the finish reason is a length limit, the Ollama binding (and Gemini/Bedrock) should follow the OpenAI binding and raise, with the same diagnostic detail. A document whose extraction produced nothing because generation was cut off should end `FAILED` and be retryable, not `PROCESSED`.

Points 1–3 are additive and low-risk. Point 4 is a deliberate behavior change (documents that used to silently pass will now fail) and probably deserves its own discussion — it may be worth gating behind a setting, though "silently indexed an empty graph" is arguably never the outcome a user wants.

### Acceptance criteria

- Ingesting with a too-small output budget surfaces a truncation message in `/documents/pipeline_status` history and in the WebUI, naming the affected document and chunk.
- The same message appears with `enable_llm_cache_for_entity_extract` both enabled and disabled.
- A chunk whose extraction returned empty content with a length finish reason does not report success.
- Regression tests cover: marker survival past `remove_think_tags`, the cache-disabled branch, and the empty-content escalation per binding.

### Additional Information

- Related: #3597 (one specific cause of the empty-content case), #3599 (the PR addressing it), #3446 and #3448 (which introduced the truncation marker this issue builds on).
- Affects all bindings; the observability gaps (1–3) are provider-independent, gap 4 is Ollama/Gemini/Bedrock vs OpenAI.

pengpengyi92:
## Description

Token-limit truncation during KG building was effectively invisible to operators: a run could index a short — or completely empty — knowledge graph, and the document still reached `PROCESSED` with `/documents/track_status` reporting success. The only trace was a provider-level `WARNING` carrying no document or chunk identity.

This PR makes the condition observable, durable, and — for the one case where nothing at all was generated — fatal.

The root cause of the blindness was small: `TruncatedResponse` is a `str` subclass, but `remove_think_tags()` rebuilt it as a plain `str`, erasing the marker before `extract_entities()` could see it. With the marker preserved, the extraction stage can finally attribute what it observes.

## Related Issues

Closes #3601 (all four gaps). Related: #3597 (one concrete cause — thinking-capable Ollama models spending the whole `num_predict` budget on the reasoning trace), #3446 / #3448 (which introduced the `TruncatedResponse` marker this builds on).

### 1. Preserve the truncation marker (#3601 gaps 2–3)

- `remove_think_tags()` re-wraps its result in `TruncatedResponse` when its input was truncated, so the marker survives sanitization on **both** the cache-enabled and the cache-disabled path. The latter never inspected the marker at all.
- The now-redundant pre-sanitization flag in `use_llm_func_with_cache` is gone: the cache guard and the caller read the same flag off the same value.

### 2. Report it to operators, without flooding the status ring (#3601 gap 1)

A document whose output budget is too small does not truncate once — it truncates on **every** chunk. One status line per event would evict the rest of the run from the bounded `history_messages` ring and still leave the operator counting lines. So:

| Channel | Behavior |
|---|---|
| Server log | Every occurrence, with chunk id + file path (unbounded by design) |
| `pipeline_status` | The **first** event immediately (a long run must not hide the condition until it ends) **+ one aggregate** per stage — independent of chunk count |
| `doc_status.metadata.llm_truncation` | Durable per-document summary that outlives the status ring and reaches `/documents` and the WebUI |

New `TokenLimitTruncationTally` (`lightrag/utils.py`) accumulates per stage and folds into a document-scoped tally:

```json
"llm_truncation": {
  "events": 15, "affected": 12,
  "stages": {"initial": 12, "gleaning": 2, "summary": 1, "multimodal": 1},
  "samples": ["chunk-001", "..."], "samples_omitted": 2
}
```

`samples` is capped (`TRUNCATION_METADATA_SAMPLE_LIMIT = 10`) — the doc_status row is serialized into every documents-listing response, so the summary must stay O(1) in document size.

Metadata write rules, which are the part worth reviewing closely:
- written at the **terminal** transition (PROCESSED, both FAILED stage boundaries, and the analyze-stage FAILED transition) — the first point at which the attempt's count is final;
- deliberately **not** in `_DOC_STATUS_METADATA_CARRY_OVER_KEYS`, so a re-run with a larger budget **clears** the key rather than resurrecting a stale summary;
- **is** in `_DOC_STATUS_METADATA_ATTEMPT_KEYS`, so a document reset to PENDING is normalized to directives-only.

### 3. Cover the summary stage too

Description summaries are LLM calls and truncate identically, silently persisting a cut-off description onto the entity/relation. `truncation_tally` is threaded through `merge_nodes_and_edges` → `_merge_{nodes,edges}_then_upsert` → `_handle_entity_relation_summary` → `_summarize_descriptions`, and checked **before** `sanitize_text_for_encoding` rebuilds a plain `str` and drops the marker.

### 4. Attribute parse failures to truncation

`Complete delimiter can not be found in extraction result` and `JSON extraction result is empty or unrecoverable` now append `(response was truncated by the model's output token limit)` when the marker says so — instead of leaving the operator to correlate two unrelated-looking warnings by chunk id.

### 5. Align empty-response semantics across bindings (#3601 gap 4)

The four bindings disagreed about a response that is **empty because generation was cut off before emitting a single content token**:

| Binding | Before | After |
|---|---|---|
| OpenAI | Raised, with full diagnostics + hint | Hint travels in the exception too; type is now `EmptyTruncatedResponseError` |
| **Ollama** | **Returned `""` → document PROCESSED with an empty graph** | **Raises**, with diagnostics + hint |
| Gemini | Raised, but message was `"Gemini response did not contain any text content."` — no diagnosis | Diagnostics + hint, **and a check after `remove_think_tags`**, typed `EmptyTruncatedResponseError` |
| Bedrock | Raised `"Received empty content from Bedrock API"` — no diagnosis | Diagnostics + hint |

The Gemini addition is a genuine bug fix: a thinking-only response arrives as `<think>…</think>`, which is **non-empty** at the existing pre-strip check, then gets stripped to `""` and was returned as a truncation-flagged empty success.

Shared wording lives in two helpers (`format_response_diagnostics`, `empty_length_truncated_hint`), so all four providers describe the failure identically and each only supplies its own budget knob:

```
Received empty content from Ollama API (done_reason=length, eval_count=64,
prompt_eval_count=900, thinking_len=28): generation hit the token limit before
emitting any content (budget consumed by reasoning); consider raising
OLLAMA_LLM_NUM_PREDICT or disabling thinking mode
```

The hint goes into the **exception** (and therefore `doc_status.error_msg`), because the operator reading the failed document is the one who has to turn the knob.

The bindings that **do** retry empty responses (OpenAI, Gemini) raise the new `EmptyTruncatedResponseError`, which is deliberately absent from every retry predicate; Ollama and Bedrock reach the same fail-fast outcome through their existing non-retried error types. Either way the length case costs exactly one request.

Both bindings that track token usage now **count it before** the empty-response raise: the request that burned its entire output budget on a reasoning trace is precisely the expensive one, and it used to be the one missing from usage accounting.

### 6. Keep the attempt metadata on an extract-stage failure

The extract-stage FAILED transition passed only the process timestamps while the merge-stage one re-passed `**extraction_meta`. None of those keys is carried over, so a document that failed during extraction described itself **less** than one that failed a stage later, losing `parse_format` / `parse_engine` / `chunk_method` / `mm_chunks`.

### 7. Cover the multimodal analyze stage

An accepted-but-truncated VLM analysis is *used* — the partial description feeds KG extraction — so it must be recorded too. The old marker check lived inside the cache-write guard, so with the analysis cache off it never ran at all (the same shape as gap 3).

`analyze_multimodal` now keeps a stage tally independent of the cache flag and hands its record to the process stage through the **analyze→process dict** rather than `doc_status`: the key is per-attempt (not carried over), so a stamp written there would be dropped at the PROCESSING transition. The hand-off runs in a `finally`, on **every** exit — a sibling item's fail-fast failure or a mid-VLM cancellation leaves the document FAILED at the analyze stage, never reaching the process stage's terminal writes, so `_analyze_worker` stamps the record onto its own FAILED transition.

### 8. Cover the custom-chunk path (`ainsert_custom_chunks`)

This entry point produces documents without going through the pipeline, so it needs the same durable evidence — and its recovery model is **additive** (a resume keeps the failed attempt's graph mutations instead of purging them), which makes the ordering requirements stricter than the pipeline's:

- an **operation-scoped tally** is fed by both KG stages and merged into the pre-operation snapshot at whichever terminal transition the operation reaches; rollback restores that snapshot exactly.
- the operation journal carries an **accumulated record across resumes** (`operation_llm_truncation`), so a clean resume cannot erase the evidence of truncated output the previous attempt already merged. It is always recomputed from the journal *as loaded* plus the live tally, never from the copy a previous write folded this attempt into — the payload merge sums event counts exactly, so folding twice would double-count.
- a **write-ahead hook** (`truncation_write_ahead`, mirroring the existing `on_anchors_durable`) journals a summary truncation *before* the object carrying that summary is upserted. Invariant: **any truncation event whose subject's graph mutation has landed is already journaled** — extraction events by the applying write, which precedes all merge mutation; summary events by the hook. The pipeline's own merges pass nothing here: a reprocess purges the previous attempt, so nothing truncated survives to warn about.

### 9. Publish the stage tallies on every exit

`extract_entities` and `merge_nodes_and_edges` used to hand their counts upward from explicit success/failure call sites, all of which sit *after* the `await` on their task set. An external cancellation landing on that await (or on an inter-phase yield) bypassed them, so already-recorded truncations never reached the caller's document tally — and for the custom-chunk path the failure bookkeeping then recomputed the journal from the incomplete tally, overwriting the write-ahead snapshot. Both now publish from a `finally` covering the whole wait/drain region; the helpers are idempotent and await-free, so they are safe in a cancellation unwind.

### 10. Keep the resumed custom-chunk recovery anchors complete

Truncated extraction responses are never cached, so a resumed operation genuinely re-runs the LLM and can extract a **different sample** — while the failed attempt's graph writes stay. Two places had to become unions rather than replacements, or first-attempt-only objects would be stranded from later purges (AGENTS.md's "the journal names the complete candidate superset"):

- the applying-phase journal write unions the previous attempt's `entity_names` / `relation_pairs`;
- the commit-time anchor union now also runs for **resumed creates**, whose merge writes `full_entities` / `full_relations` from the current attempt only and whose PROCESSED write clears the journal. A first-attempt create still skips it — Phase 0 wrote exactly those candidates.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (`env.example`: `OLLAMA_LLM_NUM_PREDICT` now notes the user-visible failure semantics)
- [x] Unit tests added

## Testing

New and extended coverage (~90 tests across 15 files):

| File | Covers |
|---|---|
| `tests/utils/test_truncation_tally.py` | Tally: event vs distinct-subject counting, `absorb`, bounded sample, empty-fragment-clears semantics, `merge_truncation_metadata` precision contract |
| `tests/extraction/test_truncation_reporting.py` | First-event + aggregate contract, aggregation bound (6 chunks → still 2 lines), gleaning round, cache enabled/disabled, mid-run failure, **cancellation on the task wait / between merge phases**, clean run, per-event server logging, summary stage, parse-failure attribution |
| `tests/pipeline/test_truncation_doc_status_metadata.py` | End-to-end through a real ingestion run, plus the carry-over / reset contracts |
| `tests/pipeline/test_pipeline_analyze_multimodal.py` | Analyze-stage recording (cache on **and** off), hand-off on the success and the fail-fast exits, the worker's FAILED/cancelled stamping |
| `tests/pipeline/test_custom_chunk_{patch,rollback}.py` | Operation tally on commit and failure, merge with the base document's record, rollback restore, **journal write-ahead ordering** (intercepted at the graph upsert), accumulation across resumes with no double-count, hard-crash-then-clean-resume, candidate + anchor unions for patch and resumed create |
| `tests/api/routes/test_document_status_truncation_metadata.py` | The key survives `DocStatusResponse`'s internal-metadata strip |
| `tests/pipeline/test_doc_status_chunk_preservation.py` | FAILED row keeps every key the PROCESSING transition wrote |
| `tests/llm/{ollama,gemini,bedrock,openai}_impl/` | Per binding: empty+length raise with diagnostics and hint, the reasoning-consumed-the-budget variant, non-length empty unchanged, partial-content salvage still working, usage counted before the raise |

Every regression test is a **fix-proof**: verified to fail *behaviorally* against the pre-fix code (a specific wrong value or missing record), not merely on a missing symbol.

Runs (against this branch, up to date with `main`):

- `tests/pipeline tests/extraction tests/utils`: `937 passed`
- Binding suites + LLM cache + the API route test: `88 passed`
- `ruff check` and project pre-commit hooks: clean

The failures visible locally (`test_entity_extraction_stability.py`, `test_ollama_role_kwargs.py`) reproduce unchanged on `main` in this sandbox — blocked tiktoken/model downloads, not related to this change.

## Additional Notes

Scope and deliberate non-changes, for the reviewer:

- **Only the token-limit case escalates to a failure.** An empty response that ended *normally* is the model's answer, not a broken generation, and `_ollama_model_if_cache` also serves query traffic. The token-limit case is deterministic and actionable, which is what makes raising the right call there.
- **Non-empty truncated content is still returned for salvage**, flagged uncacheable, exactly as before. The new check only intercepts responses with nothing in them.
- **Retry predicates are untouched; the deterministic case routes around them by TYPE.** Hitting the output token limit is a property of the prompt and the configured budget, so an empty+length response fails after one request in all four bindings — via `EmptyTruncatedResponseError` where the binding retries empty responses at all (OpenAI, Gemini), via the existing non-retried types elsewhere (Ollama, Bedrock). Every other empty mode keeps its binding's existing policy: those are sampling artifacts a fresh attempt can genuinely fix. Retrying the length case would re-buy the same full-budget generation to fail identically.
- **The `merge_truncation_metadata` payload merge is documented as lossy at the margins.** `events` and `stages` are exact sums; `affected` is exact whenever both sample lists are complete, and over-counts a subject visible in neither only past the 10-sample cap. Live tallies would be exact, but only the persisted payloads survive across a base run, a patch, and a resume.
- **Behavior change to flag:** with an Ollama/Gemini binding, a document whose extraction generated nothing because the budget was exhausted now ends `FAILED` and retryable, instead of `PROCESSED` with an empty graph. #3601 raised gating this behind a setting; it is implemented ungated on the grounds that "silently indexed an empty graph" is never the wanted outcome — happy to add a switch if maintainers prefer.

### DIFF
--- a/env.example
+++ b/env.example
@@ -1056,6 +1056,10 @@ VLM_MIN_IMAGE_PIXEL=64
 ###     LLM_MODEL=qwen3.5:9b
 ### OLLAMA_LLM_NUM_CTX must be provided, and should at least larger than MAX_TOTAL_TOKENS + 2000
 OLLAMA_LLM_NUM_CTX=32768
+### OLLAMA_LLM_NUM_PREDICT caps the OUTPUT budget. A response cut off by it is
+### kept (and reported in doc_status.metadata.llm_truncation); one cut off
+### before emitting any content fails the document instead of indexing an empty
+### graph — raise this value, or disable thinking mode, if that happens.
 # OLLAMA_LLM_NUM_PREDICT=9000
 # OLLAMA_LLM_TEMPERATURE=0.85
 # OLLAMA_LLM_STOP='["</s>", "<|EOT|>"]'
--- a/lightrag/exceptions.py
+++ b/lightrag/exceptions.py
@@ -63,6 +63,22 @@ class APITimeoutError(APIConnectionError):
         super().__init__(message="Request timed out.", request=request)
 
 
+class EmptyTruncatedResponseError(RuntimeError):
+    """A token-limit-truncated LLM response that carried nothing usable.
+
+    Raised above the provider bindings, for the case none of them can see: a
+    thinking model exhausts its output budget inside the reasoning trace and
+    returns ``<think>...</think>`` with no answer after it. That payload is
+    NON-empty, so every binding's own empty-content check passes it through,
+    and it only becomes visibly empty after ``remove_think_tags``.
+
+    Same verdict as the binding-level checks (issue #3601 gap 4): nothing was
+    generated and generation was cut off, so there is nothing to salvage.
+    Failing here is what stops an empty knowledge graph from being indexed and
+    reported as success.
+    """
+
+
 class StorageNotInitializedError(RuntimeError):
     """Raised when storage operations are attempted before initialization."""
 
--- a/lightrag/exceptions.py
+++ b/lightrag/exceptions.py
@@ -66,16 +66,22 @@ class APITimeoutError(APIConnectionError):
 class EmptyTruncatedResponseError(RuntimeError):
     """A token-limit-truncated LLM response that carried nothing usable.
 
-    Raised above the provider bindings, for the case none of them can see: a
-    thinking model exhausts its output budget inside the reasoning trace and
-    returns ``<think>...</think>`` with no answer after it. That payload is
-    NON-empty, so every binding's own empty-content check passes it through,
-    and it only becomes visibly empty after ``remove_think_tags``.
+    Two raise surfaces share it (issue #3601 gap 4):
 
-    Same verdict as the binding-level checks (issue #3601 gap 4): nothing was
-    generated and generation was cut off, so there is nothing to salvage.
-    Failing here is what stops an empty knowledge graph from being indexed and
-    reported as success.
+    - the provider bindings (OpenAI/Gemini), when the response is empty and
+      the finish reason is the output token limit;
+    - ``use_llm_func_with_cache``, for the case no binding can see: a thinking
+      model exhausts its budget inside the reasoning trace and returns
+      ``<think>...</think>`` with no answer after it — NON-empty at the
+      binding's own check, visibly empty only after ``remove_think_tags``.
+
+    Deliberately absent from every binding's retry predicate, unlike the
+    retryable response-validity errors: hitting the token limit is a property
+    of the prompt and the configured output budget, so re-running the same
+    call re-buys the same full-budget generation just to fail identically.
+    Nothing was generated and generation was cut off — there is nothing to
+    salvage. Failing (once) is what stops an empty knowledge graph from being
+    indexed and reported as success.
     """
 
 
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -2101,6 +2101,11 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
                     full_text=full_text,
                     file_path=file_path,
                 )
+                # Operation-scoped truncation record, fed by both KG stages
+                # below and stamped onto whichever terminal transition this
+                # operation reaches — the same durable evidence the normal
+                # pipeline writes, for the path that does not go through it.
+                truncation_tally = TokenLimitTruncationTally()
                 try:
                     # Stage 1 (barrier): persist chunks BEFORE extraction.
                     # Extraction records per-chunk LLM cache references
@@ -2154,7 +2159,10 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
                     # extraction, never call the LLM again and merge a
                     # different result into a partially applied operation.
                     chunk_results = await self._process_extract_entities(
-                        inserting_chunks, pipeline_status, pipeline_status_lock
+                        inserting_chunks,
+                        pipeline_status,
+                        pipeline_status_lock,
+                        truncation_tally=truncation_tally,
                     )
                     staging_flush = [
                         self.text_chunks,
@@ -2213,6 +2221,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
                             entity_chunks_storage=self.entity_chunks,
                             relation_chunks_storage=self.relation_chunks,
                             file_path=file_path,
+                            truncation_tally=truncation_tally,
                         )
 
                     # ---- Commit: flush ALL derived stores, union the patch
@@ -2241,6 +2250,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
                         base_row=status_row,
                         journal=None,
                         chunks_list=final_chunks_list,
+                        metadata_extra=truncation_tally.as_metadata_extra(),
                     )
                 except BaseException as op_exc:
                     # Journal is durable: record FAILED, keep the journal and
@@ -2260,6 +2270,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
                             base_row=status_row,
                             journal=failed_journal,
                             error_msg=str(op_exc)[:500],
+                            metadata_extra=truncation_tally.as_metadata_extra(),
                         )
                     except Exception as bookkeeping_error:
                         logger.error(
@@ -2424,6 +2435,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
         file_path: str | None = None,
         chunks_list: list[str] | None = None,
         error_msg: str | None = None,
+        metadata_extra: dict[str, Any] | None = None,
     ) -> dict[str, Any]:
         """Upsert the doc_status row for a custom-chunk operation.
 
@@ -2478,6 +2490,16 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
             payload["metadata"].pop(CUSTOM_CHUNK_PATCH_METADATA_KEY, None)
         else:
             payload["metadata"][CUSTOM_CHUNK_PATCH_METADATA_KEY] = journal
+
+        # Additive, unlike the pipeline's transition metadata: that path
+        # rebuilds from a carry-over whitelist, so an absent truncation key
+        # CLEARS the previous attempt's record (a full reprocess replaces the
+        # document's whole graph contribution). Here ``base_row`` is copied
+        # verbatim, and a patch ADDS to the base document rather than
+        # replacing it — so a clean patch must leave the base run's record
+        # standing, because the graph objects it describes are still there.
+        if metadata_extra:
+            payload["metadata"].update(metadata_extra)
 
         if error_msg is not None:
             payload["error_msg"] = error_msg
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -2367,15 +2367,30 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
                             truncation_write_ahead=_journal_truncation_write_ahead,
                         )
 
-                    # ---- Commit: flush ALL derived stores, union the patch
-                    # into the base document's anchors, then write PROCESSED
-                    # (the commit record) last with the journal cleared.
+                    # ---- Commit: flush ALL derived stores, union the
+                    # accumulated candidates into the document's anchors, then
+                    # write PROCESSED (the commit record) last with the
+                    # journal cleared.
                     await self._insert_done()
 
-                    if mode == "patch":
+                    # Patch mode AND resumed creates, not patch alone. Patch
+                    # merges pass None for the anchor storages, so this union
+                    # is their only anchor write. Create merges DO write
+                    # anchors in Phase 0 — but from the current attempt's
+                    # chunk_results only (replace semantics, correct for the
+                    # pipeline's whole-document reprocess). A resumed create
+                    # can extract a DIFFERENT sample (truncated responses are
+                    # never cached), and the PROCESSED write below clears the
+                    # journal — without this union, first-attempt-only graph
+                    # objects would be named nowhere durable and stranded
+                    # from later document purges. A first-attempt create
+                    # skips it: Phase 0 already wrote exactly these
+                    # candidates.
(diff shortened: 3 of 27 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3618, opened 2026-08-10
perf(kg): drop the seed-degree round trip from PGTable traversal

ysys143:
## What

`PGTableGraphStorage._bfs_frontier` ends every traversal with one extra query for the seed's own degree. This removes it: a traversal goes from `max_depth + 2` round trips to `max_depth + 1`.

```python
# lightrag/kg/pgtable_impl.py — removed
if seed not in degrees:
    degrees[seed] = (await self.node_degrees_batch([seed])).get(seed, 0)
```

## Why it is always paid

The branch is not conditional in practice. The seed is placed into `collected` before the loop, so it rides the visited array on every hop and the `candidates` CTE anti-joins it out. It can never be a candidate row, so `degrees` can never gain a seed entry from the loop, so `seed not in degrees` is always true.

## Why the value is never read

`degrees` has exactly one consumer — the sort in `get_knowledge_graph`:

```python
node_rows.sort(key=lambda r: (
    node_label != "*" and r["id"] != node_label,   # 1
    r.get("depth", 0),                              # 2
    -degrees.get(r["id"], 0),                       # 3
    r["id"],
))
```

The seed wins on **both** leading elements independently:

- term 1 is `False` for the seed and `True` for every other row;
- term 2 is `0` for the seed, and every other collected row is at depth `>= 1` (BFS assigns shortest-hop distance).

So the tuple comparison short-circuits before term 3 no matter which of the two is relaxed, and the seed's own degree cannot change the ordering or which rows survive truncation.

The call therefore cost every traversal one round trip, plus an `O(deg(seed))` edge enumeration inside `node_degrees_batch`, to produce a value no caller reads.

## Contract

`_bfs_frontier` is private with a single caller, so the narrowed return value stays internal — no public behaviour changes, and no reference-backend parity is affected. The docstring now states that `degrees` omits the seed and that callers must read it with `.get(id, 0)` (which the one caller already does).

## Tests

Offline (`tests/kg/pgtable_impl/test_pgtable_graph_storage.py`):

- the existing hop test asserted the call had to happen once "for the seed's ordering tie-break"; it now asserts `node_degrees_batch` is **never** awaited and that `degrees` carries no seed entry, with the reasoning above inline so the inversion is deliberate and readable;
- new: exact round-trip accounting — one `_fetchrow` for the seed row plus one `_fetch` per hop, `max_depth + 1` total;
- new: the two degenerate seeds that have no hop row able to carry a seed degree — `max_depth=0` (no hop query is issued at all) and an isolated seed (the hop returns zero rows). These are regression guards: any scheme that recovers the seed degree from hop output is wrong in both, and a self-loop-only seed lands in the second case while having a non-zero degree.

Real PostgreSQL (`tests/kg/pgtable_impl/test_pgtable_smoke.py`), because a mock that hardcodes degrees would prove nothing here:

- the adversarial ordering shape — the seed has the **lowest** real degree in the graph and the budget forces a cut; it must still lead and survive;
- a self-loop-only seed (degree 2, hop yields nothing);
- an isolated seed;
- the existing degree-parity test against the in-memory reference walk now asserts the seed's absence from `degrees` explicitly, so a reintroduced round trip fails there too.

## Verification

- `pytest tests/kg/pgtable_impl/ -m offline` — 60 passed
- `pytest tests/kg/pgtable_impl/test_pgtable_smoke.py -m pg_smoke --run-integration` on PostgreSQL 18 — 43 passed
- `LIGHTRAG_GRAPH_STORAGE=PGTableGraphStorage pytest tests/kg/test_graph_storage.py -m integration --run-integration` — 8 passed, 0 skipped
- `pre-commit run --all-files` (ruff v0.15.17 as pinned) — passed
- full `pytest tests/ -m offline` in a CI-equivalent Debian/bash-5 container — 4769 passed, 45 skipped
- full `pytest tests/ -m offline` on local Python 3.12 and 3.14 venvs — 4759 passed on each

### DIFF
--- a/lightrag/kg/pgtable_impl.py
+++ b/lightrag/kg/pgtable_impl.py
@@ -1093,8 +1093,8 @@ class PGTableGraphStorage(BaseGraphStorage):
         the server-side edge enumeration: ranking a hub's neighbours by degree
         requires visiting that hub's edges, which is a lower bound for any
         correct top-k-by-degree. Those scans are index-driven (see the UNION
-        note below) and no longer duplicated by a second node_degrees_batch
-        round trip.
+        note below) and are the only degree work the traversal does: there is no
+        node_degrees_batch round trip left, per hop or after the loop.
 
         Within each depth level the highest-degree unvisited neighbours are
         admitted first, so when the budget cuts a level the retained nodes are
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

PR #3619, opened 2026-08-10
fix(parser): enforce OOXML member and DOCX image budgets

danielaskdd:
## Description

Close the residual OOXML decompression-budget bypass left after #3614. The archive-wide compression ratio could be diluted with unrelated stored padding, and individually small high-ratio members could be split below the existing exemption floor. Native DOCX image export also still materialized each embedded member in memory before writing it to disk.

This change charges each high-ratio member for the uncompressed bytes beyond its own ratio budget. The fixed floor plus archive bytes outside those members provide a 1:1 allowance, preserving legitimate repetitive XML backed by real media without letting padding buy another ratio-cap multiple of expansion. Embedded DOCX images are streamed under explicit per-image and per-document budgets. Archive-level thresholds and parser architecture remain unchanged.

## Related Issues

- GHSA-9p96-5j78-2f9x
- GHSA-2wpj-ffvv-2pq8
- Follow-up to #3614

## Changes Made

- Check each OOXML member's declared compression ratio and accumulate expansion beyond its individual ratio budget, preventing stored-padding dilution and member splitting while retaining compatibility with mixed XML/media documents.
- Preserve the existing total-size, archive-ratio, entry-count, environment override, and pre-extraction failure semantics.
- Stream native DOCX embedded images in bounded chunks through an atomic partial file instead of calling `ZipFile.read()`.
- Add live `NATIVE_DOCX_IMAGE_MAX_BYTES` (25 MiB) and `NATIVE_DOCX_IMAGE_MAX_TOTAL_BYTES` (64 MiB) limits; over-budget parts are memoized, logged once, and reported once through parse warnings while document text continues.
- Propagate destination filesystem failures such as ENOSPC/EACCES after partial-file cleanup.
- Keep ZIP-side `OSError` failures such as EMFILE retryable on later relationships; permanent missing/corrupt-part failures remain memoized. Both paths emit one sanitized log and one parse warning per affected part.
- Allocate staging files exclusively in the destination directory, preserving atomic replacement without allowing a deterministic temp path to delete or overwrite a legitimate exported asset.
- Reserve exported filenames and charge total bytes only after a successful atomic replace.
- Add fix-proof coverage for native/legacy DOCX, PPTX, XLSX, legitimate mixed content, image limits, repeated relationships, exact boundaries, transient and permanent ZIP failures, temp-name collisions, destination failures, partial-file cleanup, logging, and warning propagation.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

- Focused DOCX budget and drawing tests: 47 passed.
- Full parser suite: 1,500 passed.
- Full pipeline suite: 662 passed (2,162 combined).
- `ruff check .`: passed.
- `ruff format --check .`: 637 files already formatted.
- Full-scale regression: a 5.39 MiB archive declaring 405 MiB uncompressed at a 75.2x aggregate ratio is refused because its 1028.8x member has 378,662,400 excess bytes against a 22,020,368-byte allowance.
- No advisory state, patched-version metadata, process isolation, or parser scheduling behavior is changed in this PR.

### DIFF
--- a/env.example
+++ b/env.example
@@ -299,9 +299,13 @@ LIGHTRAG_PARSER=*:native-teP,*:legacy-R
 ### over either limit is rejected and the document is recorded FAILED.
 ### The same budget also bounds the legacy engine's .pptx and .xlsx (both are
 ### the identical OPC/ZIP bomb class), enforced before those parse locally.
-### The ratio gate applies only above DOCX_RATIO_FLOOR_BYTES (measured in
-### uncompressed bytes), because small documents legitimately compress much
-### better than large ones. Raise the three gates
+### The ratio gate is applied archive-wide and to the cumulative expansion by
+### which individual members exceed their own ratio budgets, so member
+### splitting cannot dilute it. That excess receives a fixed
+### DOCX_RATIO_FLOOR_BYTES allowance plus a 1:1 allowance for archive bytes
+### outside those members. This permits repetitive XML backed by real stored
+### media without letting padding buy another ratio-cap multiple. Raise the
+### three gates
 ### (DOCX_MAX_UNCOMPRESSED_BYTES / DOCX_MAX_COMPRESSION_RATIO / DOCX_MAX_ENTRIES)
 ### only if a legitimate document is being refused; a non-positive value
 ### disables that gate. DOCX_RATIO_FLOOR_BYTES is the opposite — it is the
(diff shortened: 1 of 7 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

Issue #3579, opened 2026-08-05
Proposal: build a verified cold-migration utility from `PGGraphStorage` (Apache AGE) to `PGTableGraphStorage` (plain PostgreSQL tables)


ysys143:
## Background

#3103 merged `PGTableGraphStorage` (plain PostgreSQL-table graph storage), now the recommended `GRAPH_STORAGE` for PostgreSQL, superseding `PGGraphStorage` (the Apache AGE-backed one). I promised a migration guide as a follow-up, and @dan1els — who validated it on a real 60k-node / 120k-edge graph (~10x faster; queries from minutes to ~10s) — already migrated with hand-written SQL and asked for a documented path. (Separately, the #3103 benchmark on an 8k / ~40k graph measured `get_knowledge_graph` ~28x.)

Switching `GRAPH_STORAGE` moves the **graph only** — vector / KV / doc-status stores are unaffected. The docs currently rule out data migration and recommend re-indexing. But re-indexing re-runs entity/relation extraction through the LLM: it costs real LLM calls and, because extraction is non-deterministic, produces a *different* graph rather than reproducing the one already in place and tuned. The graph is a straightforward relational shape, so a **verified** data migration — preserving the exact existing graph — is worth supporting.

## Why not a naive `INSERT ... SELECT`

Copying the AGE tables (`"<graph>"."base"` / `"<graph>"."DIRECTED"`) straight into `lightrag_graph_nodes` / `lightrag_graph_edges` looks trivial but silently corrupts data — likely why the docs default to re-indexing:

- Stripping quotes to read `entity_id` (`btrim(…, '"')`) mangles ids; the correct decode is `(properties::text::jsonb)->>'entity_id'`, and some valid agtype (`NaN`, `numeric`) fails `::jsonb` entirely.
- `PGTableGraphStorage` canonicalizes undirected edges with Python `min`/`max` (code-point order); a bare SQL `LEAST`/`GREATEST` under `en_US.UTF-8` diverges on uppercase/non-ASCII names and yields duplicate or unfindable edges (must force `COLLATE "C"`).
- Two separate `INSERT`s are not atomic; `DIRECTED.start_id`/`end_id` are internal vertex ids that must be joined back through `base` to `entity_id`; and count-based verification passes while ids/properties are silently wrong.

## Proposal: a version-constrained, fail-safe cold migration

[...]

PR #3620, opened 2026-08-10
Add an offline tool to migrate a graph from Apache AGE to PostgreSQL tables

ysys143:
Follow-up to the direction agreed in #3579: an offline utility that moves an already-extracted entity-relation graph from `PGGraphStorage` (Apache AGE) to `PGTableGraphStorage` without re-processing the source documents.

Changing `LIGHTRAG_GRAPH_STORAGE` today means the existing graph is simply not visible to the new backend, so the documents have to be re-indexed — an LLM-cost-bearing operation for a graph that already exists. This tool copies the graph itself.

### Operating requirement

**Every LightRAG writer must be stopped, and the tool cannot verify that.** It reads the source, writes the target, then compares; a concurrent writer inside that window invalidates the comparison, and there is no lock or detection. Every other precondition below is enforced — this one is purely operational, and the design rests on it. Same requirement as `rebuild_vdb`.

### Design

**Storage API only, no raw SQL.** Reads with `get_all_nodes` / `get_all_edges`, writes with `upsert_nodes_batch` / `upsert_edges_batch`, undoes with `remove_edges` / `remove_nodes`. Canonicalization, endpoint handling and identity rules stay in the already-tested storage layer.

**Graph only.** Vector and KV storages are untouched and stay valid, because the migrated graph keeps the same entity and relation identities. Nothing is re-embedded, nothing is re-extracted.

**Dry run by default.** `--apply` performs the migration. The dry run runs the same enumeration and the same refusal checks, so it predicts what `--apply` will do. It migrates nothing — but it is not read-only: initializing a backend runs its DDL and then `_normalize_legacy_edges`, which deletes and re-inserts edge rows. The guarantee is "no data is copied between backends", which is narrower than "nothing is written".

**Fail-closed, and the destructive step is last.** Before anything is written the tool refuses on:

- a non-allow-listed storage pair;
- a non-empty target slice (unless `--force-empty-target`);
- a node with no usable identity — `id`/`entity_id` missing, not a string, or disagreeing. A vertex created with no properties is legal in AGE and enumerates as `{"id": None}`;
- the same node id enumerated more than once, **whether or not the payloads match** — the target keeps one row per id, so node count would change;
- `a→b` and `b→a` both present, **whether or not the payloads match** — the target stores one row per canonical pair, and AGE computes degree by counting relationship rows, so collapsing an identical reciprocal preserves every property while changing degree and traversal. LightRAG's own write path matches edges undirected, so a reciprocal means the source violates the invariant this tool assumes;
- more than one directed row reaching the tool for the same ordered pair;
- a payload containing a value PostgreSQL `jsonb` cannot round-trip: `NaN`, `Infinity` and `-Infinity` (which `agtype` represents and `jsonb` rejects), or `-0.0` (which `jsonb` normalises to `0.0`, so the write succeeds and the run then fails its own type-strict verification). The rule: if the target cannot round-trip a value, refuse before the destructive step.

Under `--force-empty-target` the pre-existing slice is dropped **only after every one of those checks has passed**. A failed `drop()` aborts the run: the backend signals failure by returning a status rather than raising, so only an explicit `{"status": "success"}` is accepted.

`--force-empty-target` still has an irreducible window: the checks prove the source is migratable, not that the write will succeed. A failure after the drop leaves the target empty — old data gone, new data never landed. Take a backup.

**`--workspace` is cross-checked before any connection is opened.** The backends resolve their workspace as `POSTGRES_WORKSPACE` > the constructor argument > `"default"`, so the environment can outrank the flag — and initializing a storage mutates the workspace it lands in. The tool refuses an env-overridden workspace up front, re-checks the resolved value afterwards, and reports it.

**Verification is source-driven**, because the target has already canonicalized and cannot reveal what was lost. Property comparison is type-strict: `1`, `1.0`, `True` and `"1"` are four different values, and `-0.0` differs from `0.0`.

**What `verified` claims.** The tool only sees what the source's enumerators return, so the claim is scoped to that view: every node and edge the enumerator returned came across with its identity, payload and count intact. It does **not** prove the physical AGE graph was reproduced — AGE enumerates with `SELECT DISTINCT source, target, properties`, so byte-identical parallel relationships arrive as one row while AGE's own degree counts two. Reading raw multiplicity would mean bypassing the storage API, which this tool deliberately does not do. Payload keys named `id`, `source` or `target` are likewise overwritten by the AGE enumerator before this tool sees the row.

**id-scoped compensation.** On failure the tool removes exactly what the run wrote: migrated ids plus every edge endpoint, since the target auto-creates missing endpoints without reporting them, via `remove_edges` then `remove_nodes`. `drop()` is never the compensation path. Cancellation counts as failure — `asyncio.CancelledError` does not subclass `Exception`, so the handler catches `BaseException`, compensates, and lets cancellation propagate. If compensation itself fails, the report carries the full written id sets for manual cleanup.

**Phase 1 scope:** exactly `PGGraphStorage` → `PGTableGraphStorage`. `PRECONDITION_HOOKS` and `COMPENSATION_HOOKS` are named placeholders — declared, not yet consulted anywhere; Phase 1 behaviour is hardcoded. Adding a pair means wiring them as well as allow-listing it. Re-indexing remains the general guidance for changing storage backends; this is an advanced path for one specific pair.

Invocation is `python -m lightrag.tools.migrate_graph_storage`, matching `migrate_llm_cache`; no console script is added.

### Tests

**107 targeted tests**, passing on Python 3.12 and 3.14. Both new modules are selected by CI's `-m offline` leg (confirmed with `--collect-only`).

The full offline suite is green on this branch — **4945 passed, 47 skipped, 0 failed on both 3.12 and 3.14** — installed the way CI installs it (`uv sync --frozen --extra api --extra offline-storage --extra offline-llm --extra pytest`). `pre-commit run --all-files` passes all five hooks.

**GitHub CI has not run this branch — it is unpushed**, so these are local results.

Coverage includes canonicalization parity, every fail-closed refusal (identity, duplicates, reciprocals, non-finite payloads, pair allowlist, non-empty target), id-scoped compensation under failure injection and under cancellation, the force-mode ordering and drop-status contracts, the workspace pre-check, and a **round trip against the real NetworkX backend** exercising real enumeration, real batch writes and real `remove_*` compensation.

Deliberately not covered offline, and stated in the test module docstring: the reciprocal/parallel directed axes (NetworkX enumerates each undirected edge once, so it cannot express them — those are synthesized in unit tests) and real SQL semantics.

### Manual end-to-end run (Apache AGE 1.7.0, PostgreSQL 18)

Seeded through LightRAG's own `PGGraphStorage` write path: 5 nodes / 4 edges including a non-ASCII id (`한국`), a quote-bearing id (`He said "hi"`) and a self-loop.

```
dry-run: {'mode': 'dry-run', 'source_backend': 'PGGraphStorage',
          'target_backend': 'PGTableGraphStorage', 'workspace': 'default',
          'nodes': 5, 'edges': 4, 'verified': False, 'compensated': False,
          'target_non_empty': False, 'would_drop_target_slice': False,
          'written_node_count': 5, 'written_edge_count': 4}

apply:   {'mode': 'apply', ..., 'verified': True, 'compensated': False,
          'target_non_empty': False, 'dropped_target_slice': False,
          'written_node_count': 5, 'written_edge_count': 4}
```

Reading both backends independently afterwards (not trusting the tool's own verdict): node id sets equal, canonical edge key sets equal, and **full payload equality on both nodes and edges** with key sets compared in both directions, so a dropped or added property would fail it. All three tricky ids byte-exact.

Also confirmed against the live instance: re-running `--apply` is refused because the target is no longer empty (exit 1); a non-allow-listed pair is refused before any backend is constructed (exit 1); `--workspace staging` with `POSTGRES_WORKSPACE=production` is refused before any connection is opened (exit 1); `--force-empty-target` reports `dropped_target_slice: True` and the data re-verifies after the drop-and-re-migrate; a vertex with no properties enumerates as `{"id": None}` and is refused; `agtype` does hold `NaN` / `Infinity`, which `jsonb` rejects, and the plan refuses such a payload before the drop.

The AGE endpoint-id `::text` cast did not carry agtype quoting at any point.

### Note for reviewers

The tool constructs graph storages with a stub `EmbeddingFunc` whose function raises if called, following `rebuild_vdb`'s pattern for the no-embedding-available case. Graph storages never read `embedding_func` (`networkx_impl`, `pgtable_impl` and `PGGraphStorage` reference it zero times), and the graph branch tolerates `None` at runtime — unlike `BaseVectorStorage`, which validates against it. We pass the stub because `BaseGraphStorage` declares the field as non-Optional. Happy to switch to `None` if you prefer.

---

Full operator documentation — preconditions, the report format, every refusal and the residual risks — lives in [`lightrag/tools/README_MIGRATE_GRAPH_STORAGE.md`](https://github.com/HKUDS/LightRAG/blob/main/lightrag/tools/README_MIGRATE_GRAPH_STORAGE.md) (added by this PR).


danielaskdd:
@codex review

danielaskdd:
### Doc request: state the memory footprint / OOM risk for very large graphs

Not a code change — the design is fine as-is, but the README currently says nothing about how much memory a run needs, and this is an operator-facing tool that people will point at production graphs.

Because the tool goes through the storage API only, and `BaseGraphStorage` exposes no paginated/cursor enumeration (unlike the doc-status side, which has `get_docs_by_statuses_page`), the whole graph has to be materialized:

- `_plan_migration` holds `source.get_all_nodes()` + `sort_directed_edges(await source.get_all_edges())` as full lists, plus the derived `node_items` / `edge_items` write plan built from them;
- `verify_source_side` then reads `await target.get_all_nodes()` and `await target.get_all_edges()` **while the source lists are still alive**, so the target graph is resident at the same time;
- `report.written_node_ids` / `written_edge_keys` stay resident for the whole run, since compensation needs the complete superset.

So peak RSS is roughly *source graph + write plan + target graph*, i.e. on the order of 2x the graph's decoded size, and it grows strictly with the graph — there is no chunk that bounds it. The write path itself is properly batched (`upsert_nodes_batch` / `upsert_edges_batch` split by `POSTGRES_UPSERT_MAX_PAYLOAD_BYTES` / `POSTGRES_UPSERT_MAX_RECORDS_PER_BATCH`), which is exactly why the memory side is easy to miss: round-trips are bounded, residency is not.

I don't think this needs fixing in this PR. Streaming would require an ordered paginated enumeration API on `BaseGraphStorage`, and the fail-closed checks that make the tool trustworthy (duplicate node ids, divergent reciprocals, parallel ordered pairs) are global properties that need a whole-graph view anyway. That's a separate change with its own design cost.

What it does need is a line in the README, so an operator can size the box before starting rather than discover it as an OOM kill mid-run — which, under `--force-empty-target`, lands after the drop. Something like this under **What this tool does not do** (or as a third bullet in **Before you run it**):

> **It does not stream.** The storage API enumerates whole graphs, so a run holds the source graph, the derived write plan and — during verification — the target graph in memory at the same time. Peak memory grows with the graph, roughly 2x its decoded size; there is no setting that bounds it. For a very large graph, run the tool on a host sized accordingly, or migrate workspace by workspace. An OOM kill mid-run is handled like any other failure (compensation removes what the run wrote), but under `--force-empty-target` it still lands after the drop.

Worth pairing with a note in the report section that `nodes` / `edges` in a dry run are also a rough sizing signal for the apply run.

danielaskdd:
### Additional review findings (beyond the memory-footprint doc comment above)

Overall impression first: this is an unusually rigorous design for a migration tool — fail-closed checks ahead of every write, id-scoped compensation on failure, source-driven verification, and mode-symmetric report fields specifically to prevent a misleading preview. The five items below are ranked by severity.

**1. Workspace precheck misses `config.ini`, so the "refuse before any connection" guarantee has a gap**

[`_precheck_requested_workspace`](https://github.com/HKUDS/LightRAG/blob/feat/graph-migration-tool/lightrag/tools/migrate_graph_storage.py#L1020) only reads `os.getenv("POSTGRES_WORKSPACE")`. The actual resolution order backends use is `POSTGRES_WORKSPACE` env > `config.ini [postgres] workspace` > constructor argument (`ClientManager`, `postgres_impl.py:2456`). When the workspace override lives in `config.ini` instead of the environment, this precheck passes, both storages get constructed and `initialize()`d — DDL, `_normalize_legacy_edges`'s delete+reinsert, AGE's `create_graph` plus ten `CREATE INDEX CONCURRENTLY` statements all land — and only then does `_check_resolved_workspace` refuse. The "refuse before opening any connection" guarantee doesn't hold on this path.

**2. Test isolation: `TestCli` fails deterministically against a real developer `.env` (reproduced)**

Most tests in `tests/tools/test_migrate_graph_storage.py::TestCli` (`test_dry_run_default_writes_nothing`, `test_apply_performs_migration_and_exits_zero`, `test_shared_data_is_initialized_before_any_storage_is_built`, etc.) never `monkeypatch.delenv("WORKSPACE")` / `delenv("POSTGRES_WORKSPACE")`. `main()` unconditionally calls `load_dotenv(dotenv_path=".env", override=False)`.

I reproduced this in complete isolation (single test, no shell env vars set) against this repo's real `.env`, which happens to set `WORKSPACE=space1`:

```
FAILED tests/tools/test_migrate_graph_storage.py::TestCli::test_shared_data_is_initialized_before_any_storage_is_built
{'mode': 'dry-run', 'error': "requested workspace 'space1' but the backends resolved to '' ..."}
```

This isn't a CI-vs-local difference to shrug off — CI has no `.env` so it never surfaces there, but any contributor who follows the documented `cp env.example .env` setup and sets `WORKSPACE`/`POSTGRES_WORKSPACE` to anything non-default will hit this deterministically. The repo already has an established fix for exactly this class of bug: `tests/conftest.py`'s `_hermetic_mineru_env` autouse fixture, whose docstring explains in detail how a module-level `load_dotenv(override=False)` leaks a developer's real `.env` into test assertions, and neutralizes the affected vars for every test. This PR introduces the same leak pattern without adding the same isolation. Suggested fix: a `tests/tools/conftest.py` with an autouse fixture that `delenv`s `WORKSPACE` and `POSTGRES_WORKSPACE`, mirroring `_hermetic_mineru_env`.

**3. README understates the blast radius of a dry run**

The dry-run caveat (`README_MIGRATE_GRAPH_STORAGE.md:76`) says it "can modify the workspace it touches," but `_DDL`'s legacy-schema migration and orphan-edge sweep DELETE across **all** workspaces and namespaces — the code's own comment says this can't be scoped per workspace. The operators most likely to hit this are exactly the ones still running the legacy table shape, i.e. this tool's target audience.

**4. `jsonb` representability gate only checks floats, not string-level values Postgres also rejects**

[`_unrepresentable_paths`](https://github.com/HKUDS/LightRAG/blob/feat/graph-migration-tool/lightrag/tools/migrate_graph_storage.py#L186) checks only `NaN`/`Infinity`/`-0.0`. `jsonb` also rejects embedded NUL bytes (U+0000) and lone UTF-16 surrogates in strings. Such a payload passes this gate and only fails at the `::jsonb` cast at write time — which, under `--force-empty-target`, happens after the drop, precisely the scenario this check exists to prevent.

**5. Dead code: the "divergent" detectors are never reached by the real pipeline**

`detect_divergent_reciprocals` (`migrate_graph_storage.py:258`), `detect_divergent_duplicate_nodes` (`migrate_graph_storage.py:373`), and the `DivergentReciprocal` dataclass each have substantial docstrings describing them as "Phase 1 policy" and each have a dedicated test class. But `_plan_migration`, `migrate_graph`, and `verify_source_side` — the only production call sites — exclusively call the coarser `detect_reciprocal_pairs` / `detect_duplicate_node_ids`, which reject every reciprocal/duplicate unconditionally, regardless of whether the payloads agree. Because the coarse check always raises first, the finer "divergent" variants can never actually execute in the live pipeline; the only place they're invoked is their own unit tests (and one hand-rolled pipeline helper inside `TestDeterministicPipeline`, not the real orchestration). That's roughly 80 lines of production code plus a full test class for logic that never runs. Either wire them in (if the intent is ever to relax the current "reject all reciprocals/duplicates" policy to "reject only divergent ones") or remove them and document the stricter coarse-grained policy as the deliberate choice.

danielaskdd:
@codex review

Issue #3617, opened 2026-08-10
[Bug]: `PGGraphStorage` is broken on Apache AGE 1.8.0 — query failure plus a backend SIGSEGV


ysys143:
### TL;DR

**Do not move `PGGraphStorage` onto Apache AGE 1.8.0.**

One AGE-side change — Cypher `id()` returning `graphid`, where on 1.7.x it returned something a column definition list could cast to `bigint` — breaks `get_knowledge_graph` along both of its branches, in different ways:

1. `get_knowledge_graph("<label>")` fails on every call — `cannot cast type graphid to bigint`.
2. `get_knowledge_graph("*")` **segfaults the PostgreSQL backend**, taking the whole instance into crash recovery.

These are separate code paths, but not perfectly disjoint: the wildcard branch only issues the crashing query when its node selection is non-empty. If it selects nothing it falls through to `_bfs_subgraph` ([`postgres_impl.py#L8711-L8713`](https://github.com/HKUDS/LightRAG/blob/main/lightrag/kg/postgres_impl.py#L8711-L8713)) and hits (1) instead.

Both originate in AGE rather than in LightRAG. Reported upstream with a standalone reproduction and a backtrace: apache/age#2500, which is open and not yet triaged. The crash is present on AGE `master` too, not only the PG18 release branch.

Worth weighing for severity: (2) is not a degraded response, it terminates the backend and takes the whole instance through crash recovery, so on a deployment where the graph endpoint is reachable it is a denial-of-service reachable by an ordinary request.

No fix PR. Symptom 1 is a small adaptation, but on its own it would not make 1.8.0 usable, since symptom 2 is reachable through the other branch. I am not claiming there is nothing LightRAG could do — failing fast on a detected AGE 1.8.x, or refusing the wildcard query on affected versions, would both bound the exposure — only that I have not built or tested such a change, and that the crash itself has to be fixed in AGE. Staying off 1.8.0 is the mitigation I can actually vouch for.

**Exposure today is limited but not zero**: Docker Hub has no released `release_PG18_1.8.0` image (checked 2026-08-07; newest published PG18 release tag is `release_PG18_1.7.0`). It does publish development snapshots — `dev_snapshot_master` (2026-07-15) and `dev_snapshot_PG18` (2026-07-07) — and `master` is a lineage where I reproduced the crash, so pulling a snapshot is an exposure path alongside source builds. Reproducing the mechanism itself needs no LightRAG install; see apache/age#2500.

### Why this came up

[...]

PR #3621, opened 2026-08-10
fix(kg): refuse PGGraphStorage on Apache AGE >= 1.8.0, which crashes the backend

ysys143:
Refs #3617, apache/age#2500

### What this fixes

Apache AGE 1.8.0 changed Cypher `id()` to return `graphid`, and that breaks `PGGraphStorage.get_knowledge_graph` along both of its branches. The labelled branch fails the `graphid` -> `bigint` column cast on every call. The wildcard branch segfaults the PostgreSQL backend, which takes the whole instance — every database on the cluster, not just LightRAG's — through crash recovery.

The wildcard is not an edge case. `defaultQueryLabel = '*'` (`lightrag_webui/src/lib/constants.ts:42`), so opening the Knowledge Graph tab is what issues the crashing query, and `/graphs` is reachable without credentials in the fully-open profile. On an affected build, the first visit to that tab is what brings the server down.

### Changes

**1. Refuse to start on AGE >= 1.8.0** (`configure_age_extension`). The function now reads back `pg_extension.extversion` after `CREATE EXTENSION` and raises when the version is at or above 1.8.0, following the shape of the existing pgvector >= 0.7.0 check in `configure_vector_extension`. This is the part that actually bounds the crash, and it needs no AGE-side fix to land: it turns "opening the graph tab crashes the instance" into a startup-time message.

**2. Make `get_knowledge_graph` version-agnostic.** Two `id()` projections were never read — the initial lookup's `node_id` (the start node's id comes from the returned vertex) and the BFS expansion's `current_internal_id` (the code uses `current_node.id` from the queue) — so they are dropped outright. The two that are read, `neighbor_internal_id` and `edge_id`, are declared `agtype` instead of `bigint`. `_record_to_dict` keeps a bare numeric agtype column as its raw string while vertex ids arrive via JSON parsing as `int`; `str()` of either yields the same digits, so the `visited_node_ids` membership test and the `visited_edges` dedup keys are unchanged. This works identically on 1.7.x and 1.8.x, so no version branching is needed.

Per the review on #3617, (2) does not ship ahead of (1): on its own it would make 1.8.0 look usable while the wildcard branch still crashes the backend.

### Two notes on the gate's policy

The bound is an upper bound, not a blocklist of known-bad points. A later AGE release is only safe once someone has verified it, so passing unverified versions through would defeat the check. Raise the constant with evidence, not on release cadence.

AGE reports its version per release, not per commit — `master` at `8014174` also reports `1.8.0` — so a build that fixes the crash would still be refused. `POSTGRES_AGE_ALLOW_UNSUPPORTED_VERSION=true` skips the check for that case, with a warning that names the consequence. Without an escape hatch the gate would be a permanent block for anyone on a source build or a dev snapshot.

When `CREATE EXTENSION` itself fails the function still returns without raising, so deployments that use the PostgreSQL storages without AGE are unaffected. Verified against a stock `postgres:18` with no AGE available: behaviour is identical to `main`.

### Verification

The gate was exercised against three AGE builds, all compiled from source on the same `postgres:18` base (PostgreSQL 18.4) with the same toolchain, so the AGE version is the only variable: `release/PG18/1.7.0` at `806fa2e` proceeds; `release/PG18/1.8.0` at `e43dc1a` is refused; `master` at `8014174` is refused; and 1.8.0 with `POSTGRES_AGE_ALLOW_UNSUPPORTED_VERSION=true` proceeds with a warning. With (2) applied, `get_knowledge_graph("Entity_00042", max_depth=2, max_nodes=50)` returns an identical subgraph on 1.7.0 and 1.8.0 — 50 nodes, 54 edges.

CI was reproduced locally the way the workflows actually run it: `pre-commit run --all-files` at the config-pinned ruff (v0.15.17) passes; `uv sync --frozen --extra api --extra offline-storage --extra offline-llm --extra pytest` then `pytest tests/ -m offline` passes on both Python 3.12 and 3.14; and `pytest tests/kg/pgtable_impl/test_pgtable_smoke.py -m pg_smoke --run-integration` passes against `postgres:18-alpine`.


danielaskdd:
@codex review

danielaskdd:
Reviewed `7116ece4b0`.

**`466cd984` (`get_knowledge_graph` version-agnostic) checks out.** `current_internal_id` and `node_id` have no other consumers anywhere in the repo, so dropping the projections is safe. The `if not record["neighbor_internal_id"]` guard runs before every `str(record[...])`, and `_record_to_dict` keeps a bare numeric agtype as its raw digit string, which `str()`-compares equal to the `int` ids parsed out of vertex JSON — so `visited_node_ids` membership and `visited_edges` dedup are preserved without version branching. `id()` appears nowhere else in the file, and the only other `bigint` column declarations are `count()`, unaffected by the AGE change.

I also considered and am *not* raising the "the gate is over-broad for deployments that never open the graph tab" angle — refuse-over-warn is settled policy from the discussion on #3617.

The gate in `7116ece4` has two holes I'd want closed. Test coverage is already covered by the Codex comment on line 865, so I've kept only the part it doesn't reach (finding 4).

---

### 1. MEDIUM — `pg_extension.extversion` does not track an in-place binary upgrade

`lightrag/kg/postgres_impl.py:863`

The gate reads only `pg_extension.extversion`, which records the version of the SQL script that was last *run* against this database. An in-place binary upgrade never updates it: rebuild `Dockerfile.postgres` at AGE 1.8.0 (or pull a newer `gzdaniel/postgres-for-rag` tag) over an existing data volume, and the 1.8.0 `age.so` is loaded while `extversion` stays `'1.7.0'` — `CREATE EXTENSION IF NOT EXISTS` is a no-op and nothing in the codebase runs `ALTER EXTENSION age UPDATE`. The gate passes, and the first Knowledge Graph tab visit still takes the instance through crash recovery.

That mixed state is arguably *more* crash-prone than a clean 1.8.0 install, not less: the 1.7 catalog still declares `id()` as `RETURNS agtype` while the 1.8 C function returns a pass-by-value `graphid`.

Suggested fix: also read `pg_available_extensions.default_version`, which comes from the on-disk `age.control` and therefore tracks the binaries (`default_version = '1.8.0'` on AGE master today), and refuse on the higher of the two.

A functional probe (`SELECT id(...)` and see what happens) is **not** a safe alternative here — under a stale catalog the `bigint` cast is exactly what could segfault, so the probe would be the crash it is meant to detect.

### 2. MEDIUM — the two pre-gate raises cannot be bypassed by the override

`lightrag/kg/postgres_impl.py:866` and `:880`

Both raises land before `get_env_value(AGE_ALLOW_UNSUPPORTED_ENV, ...)` at line 890, so `POSTGRES_AGE_ALLOW_UNSUPPORTED_VERSION=true` cannot get past either one. They are new hard startup failures for two populations that start fine on `main`:

- **Unreadable row** — `CREATE EXTENSION` succeeded but the `pg_extension` lookup returns nothing or a NULL. Rare, but the failure mode is "server won't boot" and the documented escape hatch is inert.
- **Unparseable version** — reachable on precisely the source-build population the override was written for. AGE master ships `age--1.8.0--y.y.y.sql`, so `ALTER EXTENSION age UPDATE` on a master build sets `extversion = 'y.y.y'`; `int('y')` then raises and the server refuses to start with the override ignored. (A plain `CREATE EXTENSION` on master reports `1.8.0`, since `default_version = '1.8.0'` — so this needs the explicit `ALTER`, which narrows it, but it is the same dev-snapshot user the PR description says the flag exists for.)

Fix: consult the override before both raises, not only before the version comparison.

### 3. LOW — `POSTGRES_AGE_ALLOW_UNSUPPORTED_VERSION` is documented nowhere

The fatal error message names the variable, but it appears in neither `env.example` (alongside the other commented-out `POSTGRES_*` knobs around L1261–1297) nor `docs/`. Suggestion (2) from the #3617 discussion — documenting the known-good / known-bad range so users can find it before they hit the crash — is also still open.

### 4. LOW — `466cd984`'s column-definition change has nothing pinning it

Adding to the Codex comment on the gate: no test pins that the `_bfs_subgraph` column definition lists no longer cast `id()` to `bigint`, so a future edit can silently reintroduce the cast that fails on 1.8.x — a regression the gate cannot catch, because the gate refuses those versions anyway and the override path would run straight into it.

ysys143:
Thanks — finding 1 is a real hole and I've fixed it. Two of the supporting details did not reproduce on my side, so I'm recording what I measured rather than quietly adopting them; the conclusions are unchanged either way.

### finding 1 — confirmed, fixed

You're right that `pg_extension.extversion` does not track an in-place binary upgrade. Reproduced by initialising a data volume under AGE 1.7.0 and restarting the same volume on 1.8.0 binaries:

| source | value |
|---|---|
| `pg_extension.extversion` | `1.7.0` — what the old gate read, so it passed |
| `pg_available_extensions.default_version` | `1.8.0` — tracks the on-disk `age.control` |
| `id()` → `bigint` in that state | fails: `cannot cast type graphid to bigint` |

So the gate let through a deployment whose `get_knowledge_graph` is already broken. The gate now reads both and refuses on the higher of the two, exactly as you suggested. Both catalogs are world-readable (`has_table_privilege('public', …)` is true for each), so this does not assume a superuser connection.

One correction on the mechanism: in that mixed state the crashing wildcard query returns `ERROR: cache lookup failed for type 0` rather than segfaulting — no `signal 11` in the server log across repeated attempts. The deployment is broken, but on this evidence the mixed state is *less* crash-prone than a clean 1.8.0 install, not more. Refusing it is still right, since `get_knowledge_graph` fails outright.

### finding 2 — accepted, with a note on the example

Fixed: the override is now consulted before every refusal, so `POSTGRES_AGE_ALLOW_UNSUPPORTED_VERSION=true` reaches all of them. The unreadable-row path in particular was a new hard startup failure for deployments that start fine on `main`, which is not a trade I want to make silently.

The specific example does not reproduce, though. `age--1.8.0--y.y.y.sql` exists in the master tree, but the Makefile excludes it from what gets installed:

```
# Excludes the upgrade template (y.y.y) and the synthetic stamped test file.
DATA = $(filter-out age--%--y.y.y.sql $(age_upgrade_test_sql),$(wildcard age--*--*.sql))
```

A source build of master installs only `age--1.6.0--1.7.0.sql`, `age--1.7.0--1.8.0.sql` and `age--1.8.0.sql`, so `ALTER EXTENSION age UPDATE` has no `y.y.y` target to reach. I could not produce an unparseable `extversion` on any build I have. The ordering fix stands on its own merits — I'd rather the escape hatch cover a path that turns out to be unreachable than the reverse.

### findings 3 and 4, and the Codex comment — done

- `POSTGRES_AGE_ALLOW_UNSUPPORTED_VERSION` is now documented in `env.example` and `env.docker-compose-full` (both mirror the same `POSTGRES_*` section), together with the verified-good/verified-bad range stated at the width I actually measured: good `release/PG18/1.7.0` at `806fa2e`; bad `release/PG18/1.8.0` at `e43dc1a` and `master` at `8014174`. I have not tested other 1.7.x points or other PostgreSQL majors, so nothing claims a blanket "1.7.x is supported".
- New `tests/kg/postgres_impl/test_postgres_age_version_gate.py`, 16 cases, module-level `pytestmark = pytest.mark.offline` so it actually joins the offline gate: supported version starts; 1.8.0 refused; **stale-catalog-with-newer-binaries refused**; reverse mismatch refused; missing row, both-null and unparseable all fail closed; one unreadable source still lets the other decide; the override reaches all four refusal paths; the lookup still queries both catalogs; and the bound itself is pinned so raising it needs a deliberate edit.
- Your finding 4 is covered by a source pin: the `AS (...)` column definition lists in `_bfs_subgraph` are extracted and asserted free of `bigint`. I checked it actually catches the regression — reintroducing one `bigint` fails that test and nothing else.
- `configure_age_extension` still returns without raising when `CREATE EXTENSION` itself fails, so AGE-less PostgreSQL deployments are untouched. Verified against a stock `postgres:18`: identical behaviour to `main`.

### Verification

Gate exercised against three source builds on the same `postgres:18` base, plus the in-place volume: clean 1.7.0 starts; clean 1.8.0, `master`, and stale-catalog-1.7.0-on-1.8.0-binaries are all refused; each of those starts with the override set.

CI reproduced locally the way the workflows run it — `uvx pre-commit run --all-files` at the pinned ruff, then `uv sync --frozen --extra api --extra offline-storage --extra offline-llm --extra pytest` and `pytest tests/ -m offline` on both 3.12 and 3.14, plus `pg_smoke` against `postgres:18-alpine`.

danielaskdd:
@codex review

danielaskdd:
Reviewed `9ccd1923ba`. **The PR is sound and I'd merge it** — all three findings below are in the gate's edge handling, none in the `get_knowledge_graph` change.

### What checks out

- **The `agtype` change is safe.** `_record_to_dict` falls through to `d[k] = v` for a bare numeric agtype (no `::` separator), keeping it as the raw digit string, while vertex ids arrive as `int` from JSON parsing. `str()` of either yields identical digits, so the `visited_node_ids` membership test (`postgres_impl.py:8614` vs `:8517`) and the `visited_edges` / `visited_edge_pairs` dedup keys are unchanged. `RETURN` arity still matches both `AS (...)` lists after the two projections are dropped, and neither dropped name has another consumer in the repo.
- **Scope is contained.** `configure_age_extension` is reached only from `PGGraphStorage.initialize` (`:7027`), so PG deployments not using AGE for graph storage are untouched. `RuntimeError` is absent from `_transient_exceptions` (`:404`), so the refusal is not retried and does propagate to a failed startup.
- The remaining `id()` uses (`:8778`, `:8780`) are in the wildcard branch — the crashing path the gate exists to block, correctly left unfixed per the settled refuse-over-warn policy. `total_nodes bigint` (`:8718`) is `count()`, unaffected by the AGE change.
- Test layout matches `AGENTS.md`, and `asyncio_mode = "auto"` is set, so the new module does join the offline gate as claimed.

I could not execute the new test module in my environment (`asyncpg` unavailable, installs blocked), so `_parse_age_version` was verified by running it standalone.

---

## 1. MEDIUM — an unparseable version silently defers to the other source, which can be the stale one

`lightrag/kg/postgres_impl.py:124-138`, `:900-907`

`_parse_age_version` returns `None` for two conditions that are not equivalent:

| input | returns | actual meaning |
|---|---|---|
| `None` / `""` (`:130`) | `None` | source **absent** — nothing to say |
| `"1.8.0-dev"` (`int()` raises, `:134`) | `None` | source **present but unreadable** — a version nobody verified |

The docstring justifies only the first (*"so that one unreadable source does not stop the other from deciding"*). For the second, dropping the value contradicts the policy stated on `AGE_FIRST_UNSUPPORTED_VERSION` (`:101-105`): *"a later AGE release is only safe once someone has verified it, and letting unverified versions through would defeat the check."* An unreadable version number is the paradigm case of an unverified version.

Exhaustive 4×4 decision table (OK = parses & safe, BAD = parses & unsafe, NP = present-but-unparseable, AB = absent):

```
 installed  available | categories  | decision
     1.7.0      1.7.0 | OK   OK     | START
     1.7.0      1.8.0 | OK   BAD    | REFUSE: not-supported
     1.7.0  1.8.0-dev | OK   NP     | START   <== gate passes
     1.7.0       None | OK   AB     | START
     1.8.0      *     | BAD  *      | REFUSE: not-supported  (all 4 rows)
 1.8.0-dev      1.7.0 | NP   OK     | START   <== stale source decides alone
 1.8.0-dev      1.8.0 | NP   BAD    | REFUSE: not-supported
 1.8.0-dev  1.8.0-dev | NP   NP     | REFUSE: could-not-determine
 1.8.0-dev       None | NP   AB     | REFUSE: could-not-determine
      None      1.7.0 | AB   OK     | START
      None      1.8.0 | AB   BAD    | REFUSE: not-supported
      None  1.8.0-dev | AB   NP     | REFUSE: could-not-determine
      None       None | AB   AB     | REFUSE: could-not-determine
```

Row 3 is the one that matters, because of the division of labour this PR establishes itself in the `AGE_VERSION_SQL` comment (`:109-115`): `extversion` does **not** track the binaries, `default_version` does. So when `default_version` is unparseable and `extversion` is a stale `1.7.0`, the gate's only remaining input is the source known not to reflect the loaded `age.so` — which re-opens the in-place-upgrade hole the second source was added to close. That deployment's `get_knowledge_graph` is already broken, and the first Knowledge Graph tab visit still takes the instance through crash recovery.

Note the internal inconsistency: `NP + NP` and `NP + AB` fail closed, but `NP + OK` starts. The same unreadable string is treated as a danger signal when it stands alone and as noise when a (stale) parseable sibling exists.

The trigger surface is wider than one string — any non-numeric component fails the per-part `int()`:

```
1.8.0-dev   v1.8.0   1.8.0-rc1   1.9.0beta   2.0.0-beta   1.8.0+pg18   1.8.0-1.pgdg120   y.y.y
```

all drop out of the vote. PostgreSQL places no format constraint on `extversion` — it is an arbitrary string from the control file / script filename — and AGE's own versioning vocabulary does contain a non-numeric token (`age--1.8.0--y.y.y.sql` in master; your point that the Makefile's `DATA` filter makes it unreachable as an `ALTER` target on a source build does hold).

To be clear about the strength of this: **I could not confirm that any existing AGE build emits a non-numeric `default_version`.** This is a latent asymmetry, not a reproducible bug. It is worth closing for policy consistency and because the fix is small, not because anyone is hitting it today.

**Coverage gap.** `test_one_unparseable_source_still_decides` exercises only the safe direction:

```python
conn = _FakeConnection(installed="not-a-version", available="1.8.0")
#                      ^ unparseable one is stale   ^ survivor is the higher one
```

The survivor is `1.8.0`, so it refuses. The mirror case — `installed="1.7.0"` with an unparseable `available`, i.e. the only row that starts — is not covered by any of the 16 cases. The test's name reads as if the behaviour were an intended feature, but it pins only the safer half of it.

**Suggested fix** — distinguish absent from present-but-unreadable with a sentinel:

```python
_AGE_VERSION_UNPARSEABLE = object()

def _parse_age_version(raw: Any) -> tuple[int, int, int] | object | None:
    """None if the source is absent; _AGE_VERSION_UNPARSEABLE if it is present but not
    a version we understand — which is an unverified version, not a silent one."""
    if not raw:
        return None
    try:
        parts = [int(p) for p in str(raw).split(".")[:3]]
    except (ValueError, IndexError):
        return _AGE_VERSION_UNPARSEABLE
    while len(parts) < 3:
        parts.append(0)
    return (parts[0], parts[1], parts[2])
```

and order the caller's branches known-unsafe > unreadable > start, which is what keeps the existing assertions intact:

```python
candidates = (_parse_age_version(installed_raw), _parse_age_version(available_raw))
parsed = [c for c in candidates if isinstance(c, tuple)]

# An unreadable sibling source cannot make the situation safer, so decide unsafe first.
if parsed and max(parsed) >= AGE_FIRST_UNSUPPORTED_VERSION:
    ...  # existing not-supported branch, override included
if any(c is _AGE_VERSION_UNPARSEABLE for c in candidates) or not parsed:
    ...  # existing could-not-determine branch, override included
return
```

I diffed old vs new across all 16 rows: **exactly 2 rows change, both START → REFUSE, none loosened.** They are `OK + NP` and `NP + OK`. All 7 existing assertions still pass, including `test_one_unparseable_source_still_decides`' `match="not supported"`. Both new refusals land in the existing `could-not-determine` branch, so `POSTGRES_AGE_ALLOW_UNSUPPORTED_VERSION` covers them automatically — consistent with the "override reaches every refusal" design from round 2, which is the escape hatch the source-build population needs.

The `NP + OK` row (unreadable `extversion`, `available = 1.7.0`) is arguably safe on its own, since the binaries really are 1.7.0. Tightening it too is the more conservative call, and with the override behind it I think that trade is right.

One test would pin the direction that is currently open:

```python
@pytest.mark.asyncio
async def test_unparseable_binaries_source_does_not_defer_to_stale_catalog(monkeypatch):
    """default_version tracks the binaries; if it is unreadable, extversion must not
    decide alone — it is exactly the source that does not move on a binary upgrade."""
    monkeypatch.delenv(AGE_ALLOW_UNSUPPORTED_ENV, raising=False)
    conn = _FakeConnection(installed="1.7.0", available="1.8.0-dev")

    with pytest.raises(RuntimeError, match="Could not determine"):
        await PostgreSQLDB.configure_age_extension(conn)
```

## 2. LOW–MEDIUM — a failed `CREATE EXTENSION` skips the gate entirely

`lightrag/kg/postgres_impl.py:891`

The pre-existing `except Exception: ... return` now also swallows the version check. The part I'd flag specifically: a transient connection error is caught here **before** it can reach `_run_with_retry`, so the wrapper added at `:7027` — whose stated purpose is that transient connection errors get retried — is defeated for this function. The outcome is a worker running unchecked against an AGE 1.8 extension that already exists from a prior successful start, i.e. one where graph queries do work and can crash the instance.

The version is a property of the installation, not of this call's outcome, so the lookup arguably belongs outside the `try`. If you move it, the AGE-less case needs care: an absent row from `pg_available_extensions` currently reaches the fail-closed `could-not-determine` raise, and today it is only unreachable because `CREATE EXTENSION` returns early first.

One correction to my own earlier pass on this, since I checked it afterwards: the concurrent-`CREATE EXTENSION` race across gunicorn workers is largely prevented — `initialize` holds `get_data_init_lock()`, which is a genuine multiprocessing lock when `_is_multiprocess` (`shared_storage.py:1447`). Triggering that would take two separate deployments sharing one database, or external concurrent DDL. The transient-error path stands on its own.

## 3. LOW — the catalog read is unguarded and precedes the override

`lightrag/kg/postgres_impl.py:894` runs before `override_set` is computed at `:898`, so a failing `fetchrow` is a hard startup failure that `POSTGRES_AGE_ALLOW_UNSUPPORTED_VERSION=true` cannot bypass. That is the same shape as finding 2 from the previous round, which this PR went out of its way to fix for the two `raise` statements — an exception from the lookup itself is the one remaining path the override does not cover. Both catalogs are world-readable so this is unlikely; a statement timeout during startup would do it.

---

None of these block the merge. The PR is a strict improvement over `main`, where the crash is unbounded; findings 1 and 3 are about the gate not quite honouring its own stated policy at the edges, and 2 is a coverage hole in front of it.

---

ysys143:
Thanks — all three findings are addressed, and implementing them turned up more than they asked for. Two things first: the crash this gate exists to prevent is now reproduced rather than cited, and adversarial passes over our own fix found a series of further defects in it, all fixed. One of them reordered the flow I described last round — the extension is now created after the version decision, not before.

Finding 1 is implemented essentially as you proposed: an `_AgeVersionUnparseable` sentinel (`postgres_impl.py:160`, instance at `:164`), `_parse_age_version` returning `None` only for a genuinely absent source, and — once both lookups have answered — branch order known-unsafe (`:1131`) > unreadable (`:1189`) > start, the last being the fall-through past both. Two branches run ahead of that ordering: the no-usable-answer state (`:1051`) and the nothing-to-gate state (`:1085`). The parser is a strict three-component regex (`:171`) — short forms are no longer zero-padded, since a value we had to repair before comparing is not a value we read. One consequence outside your table, which lists `None` and `""` together: `_parse_age_version("")` was ABSENT at base and is now UNPARSEABLE, so `(installed="", available="1.7.0")` moves from START to REFUSE.

## The absent-row case had a second cause

Your caution that an absent row "is only unreachable because CREATE EXTENSION returns early first" is correct for the case you scoped it to: with AGE genuinely absent, the create does fail with `extension "age" is not available`, `9ccd1923` returns early, and the raise is not reached. The sharp edge is that an absent row had two causes, and only one of them is the AGE-less case.

In the other, AGE is fully present and loaded. With the extension already in `pg_extension` and its control file removed from the sharedir, `CREATE EXTENSION IF NOT EXISTS age` still succeeds — `NOTICE: extension "age" already exists, skipping`, exit 0 — so there is no early return and the path is reachable. That is the state a package upgrade, a partial uninstall, or a `pg_upgrade` into a libdir without AGE produces. `age.control` (sharedir) and `age.so` (libdir) are different files in different directories: `pg_proc` entries still point at `$libdir/age`, the binaries still load, and the query shape demonstrated below still crashes the backend. Anchored on `pg_available_extensions`, the old `LEFT JOIN` returned zero rows there — indistinguishable from a server that never had AGE. The naive quiet-return we first wrote for finding 2 would therefore have turned a live fail-closed path into a silent start.

The gate now issues two independent lookups (`:961`, `:1015`; constants at `:148`, `:149`) rather than one join, for the reason in (c) below. Each source has three states, not two: absent, present but naming no usable version, and lookup-never-answered. The third is tracked with its own flag rather than folded into `None` — as is the distinction between the two ways a source can give no usable answer, so the operator is never told about an error that did not happen (`:1031`, `:1045`). "Neither catalog names age" — the only state with nothing to gate — is both lookups answering, both with `None` (`:1085`).

Your 16-row diff held exactly as stated for the fix as you proposed it; we re-derived the table independently and got the same two flips. Two cells have since moved, for reasons outside the diff you ran: `(installed unsafe, available safe)` now raises its own remedies rather than the generic ones — still a refusal, see (b) — and `(absent, absent)` now starts. The latter is not a behavioural loosening: under a single joined row that cell was unreachable — `CREATE EXTENSION` runs first there, and either it succeeded (so `pg_extension` names age, and `installed` is not absent) or it failed and `9ccd1923` returned early before the lookup, exactly as you noted. The row-is-None path it would otherwise have reached raises rather than starts; no deployment ever arrived there.

## Problems found and fixed

Two were already present at `9ccd1923`, and the first is the direction your finding 2 was pointing at when you noted the lookup arguably belongs outside the `try`:

**(a) `CREATE EXTENSION` ran before the version check.** A refused startup still installed AGE 1.8.0 and committed `extversion='1.8.0'`. The operator then follows our own error message, pins AGE back to 1.7.0, and lands with 1.8.0 SQL objects on a 1.7.0 library. Measured on real AGE: writes still go through, but every `MATCH` fails with `ag function does not exist`, because the 1.8.0 script declares `_agtype_build_vertex(graphid, agtype, agtype)` while the older library exports `(graphid, cstring, agtype)`. Not a crash — a deployment nothing can be read back out of. The gate was manufacturing a state it then refused. Catalogs are now read first, and the create runs only once the version is acceptable (`:1205`).

**(b) The refusal message misdirected the operator in exactly that cell.** It said "Pin AGE to a verified version" to someone who had just done so, and offered the override as a way out. Measured: in this state `ALTER EXTENSION age UPDATE TO '1.7.0'` and bare `ALTER EXTENSION age UPDATE` both fail with `no update path from version "1.8.0" to version "1.7.0"` — AGE ships no downgrade script. (On an ordinary 1.8.0 install the bare form succeeds with `NOTICE: version "1.8.0" ... is already installed`; the failure is specific to this cell.) The cell now assembles its own remedies instead of the generic ones (`:1160` versus `:1177`) and names the override only to rule it out. The preamble is still shared — it carries the version and the issue link, which every refusal needs — so the SIGSEGV sentence is read first and the appendix then explains that this particular state is non-functional rather than crash-prone; only the remedies are distinct.

Two came from the fix itself:

**(c) A single malformed control file in the sharedir, belonging to an unrelated extension, makes the whole `pg_available_extensions` view raise** — measured on stock `postgres:18-alpine`, which has no AGE at all, where it made the gate refuse startup. A startup regression this PR introduced. (Constructed, not observed in the wild, though it is a plausible package-upgrade artifact.) Hence the split into two lookups, so that one catalog failing is not read as the other's answer.

**(d)** The split also makes a defect testable that a single joined row could not express: with each value coming from its own named query, swapping the two changes which message is raised, and `test_each_value_comes_from_its_own_query` catches it.

**(e) The same error your finding 1 identified, recurring at the lookup layer: a source that could not be read, treated as one that said nothing.** Twice. A failed lookup set `available_raw = None`, which read as "AGE absent" when `installed` was absent too — installing an unchecked AGE — and as "let `pg_extension` decide alone" when it was present, though `extversion` cannot see a binary swap, which is why you added the second source. Then, independently, `default_version` proved nullable while `fetchval` returns `None` for both "no row" and "row with NULL", so a control file naming no version also read as absent. Both reproduced live: a 1.7.0 data directory booted under the 1.8.0 image with `default_version` stripped, where the gate called `installed=1.7.0, available=None` verified safe with 1.8.0 binaries loaded. The lookup now uses `fetchrow` and separates the three cases (`:1031`, `:1045`); each no-usable-answer case takes the unparseable sentinel (`:1120`) and fails closed. With `installed` absent the gate neither creates nor refuses (`:1051`) — database untouched, but the process starts, so graph operations fail at runtime rather than at startup: a new silent-failure mode against `main`, and the right trade, but a trade. Closed by construction rather than by our having stopped finding instances — `test_only_a_verified_safe_or_absent_state_starts_the_gate` enumerates the whole source-by-outcome space against the policy predicate.

**(f) The override's consent was broader than its wording.** Two states can install an AGE nothing has inspected: the state in (e) where `installed` is absent and the available lookup is unusable, and the installed lookup itself failing non-transiently, which skips the available lookup entirely. In both, the override emitted only the generic "could not determine ... starting anyway" and then created the extension — but "tolerate an unknown version" and "install an unseen one" are different consents. Both states now name what actually happens (`:1073`, `:990`). Measured on a live container for the first of them: with the override set, AGE 1.8.0 — the build we measured segfaulting — was installed while the old wording said only "could not determine … starting anyway". The second is reachable only with an injected failure, since `pg_extension` is an ordinary world-readable catalog table, so it is covered by tests rather than by a live run. Behaviour is unchanged; the messages are not.

One limit worth stating, since none of this closes it: the gate trusts a control file that is *readable*. `default_version` is taken to track the binaries — the premise your finding 1 established — so a readable-but-stale control file (1.7.0 `age.control` beside a 1.8.0 `age.so`) is accepted, and that deployment starts with `get_knowledge_graph` already broken (confirmed on a container: `cannot cast type graphid to bigint`, and the wildcard failing with `cache lookup failed for type 0` rather than segfaulting). The two files ship in the same package, so this is a stated assumption rather than a gap the catalogs could close.

## Finding 2's retry point, and a regression on the way

Sharper than it first looked. `_transient_exceptions` was an instance attribute, so the static `configure_age_extension` could not reach it; our first implementation wrapped every failure in `RuntimeError`, which meant `_run_with_retry`'s `retry_if_exception_type` could never match — a strict regression against `9ccd1923`, the state you reviewed and where the unguarded catalog read was introduced. (Not against `main`, whose `configure_age_extension` has no catalog read at all.) Fixed by hoisting the tuple to module-level `TRANSIENT_DB_EXCEPTIONS` (`:105`) and re-raising transients unwrapped (`:982`, `:1018`).

Deliberate residual: `QueryCanceledError` (statement/lock timeout) is not in the repo's transient tuple, so that narrow case is a hard startup failure covered by the override. Widening the repo-wide tuple is out of scope here.

## Finding 3

`override_set` is computed before the catalog reads (`:932`), so a raising lookup is bypassable by the override rather than an unconditional startup failure. The transient check runs before the override, so a connection blip is still retried rather than swallowed.

## Tests

61 test functions, 206 cases after parametrization. The 16 cells are pinned by `test_configure_age_extension_decision_table`, with `test_decision_table_covers_every_category_pair` asserting exhaustiveness by construction. Two complementary invariants sweep the state space from both ends: `test_no_state_short_of_a_clean_start_creates_the_extension` iterates the states believed not to start and asserts none issues a `CREATE EXTENSION` — which by construction cannot catch a state that wrongly starts, the shape of every fail-open here — so `test_only_a_verified_safe_or_absent_state_starts_the_gate` enumerates the whole space, including the no-usable-answer categories, and compares each cell against the policy predicate. Your `test_unparseable_binaries_source_does_not_defer_to_stale_catalog` is included as written, and `test_one_unparseable_source_still_decides` is renamed to `test_unsafe_version_wins_even_with_one_unparseable_source`, assertions unchanged, for the reason you gave. The defects above are pinned by, among others, `test_create_extension_runs_after_the_version_decision`, `test_refused_startup_issues_no_create`, `test_stale_catalog_refusal_has_its_own_guidance`, `test_each_value_comes_from_its_own_query`, `test_unreadable_available_lookup_does_not_install_an_unchecked_age`, `test_null_default_version_does_not_let_a_stale_installed_decide`, `test_null_default_version_is_not_reported_as_age_being_absent`, `test_override_install_of_an_unchecked_age_says_so`, and `test_override_on_unreadable_installed_lookup_names_the_install`.

## Validation against real Apache AGE

AGE built from source on PostgreSQL 18 — `release/PG18/1.7.0` at `806fa2e` and `release/PG18/1.8.0` at `e43dc1a`; `postgres:18-alpine` for the AGE-less state.

The crash is reproduced — twice, independently, on AGE 1.8.0 on PostgreSQL 18.4 Debian, aarch64, using the wildcard query shape `get_knowledge_graph` issues:

```
LOG:  client backend (PID 111) was terminated by signal 11: Segmentation fault
LOG:  terminating any other active server processes
LOG:  all server processes terminated; reinitializing
```

The same pair of containers also shows the labelled branch on 1.8.0 failing with `ERROR: cannot cast type graphid to bigint for column "v"`, and AGE 1.7.0 handling both shapes — the wildcard returns rows, and `AS (v bigint)` accepts `id()`. To be precise about what breaks on 1.8.0: `id()` itself is fine, and received into an `agtype` column it returns an integer. What breaks is a `bigint` column-definition list, and the wildcard shape.

Catalog contents per state, both versions:

| state | `pg_extension` | `pg_available_extensions` |
| --- | --- | --- |
| on disk, not created | absent | `1.7.0` / `1.8.0` |
| on disk, created | `1.7.0` / `1.8.0` | same |
| created, `age.control` removed | `1.7.0` / `1.8.0` | absent |
| AGE nowhere | absent | absent |

The old anchored query returned 0 rows for state 3 on both versions, byte-identical to state 4. So an AGE 1.8.0 installation with its control file removed would have started silently — and state 3 is not merely untidy: `pg_extension` still records `extversion = 1.8.0`, 370 `ag_catalog` functions still resolve with `create_graph` and `cypher` still bound to `probin = $libdir/age`, and Cypher still executes, `create_graph('gtest')` then a MATCH returning `{"id": 844424930131969, "label": "Person", ...}::vertex`. 1.7.0 behaves identically at 348 functions.

End-to-end `configure_age_extension` against the live containers, override unset for that sweep: AGE 1.7.0 starts in all three AGE-present states; AGE 1.8.0 refuses in all three; the AGE-less server starts, logging that no version check was performed. The two fail-opens in (e) that reached a live server — the stale-installed fallback and the NULL `default_version` case — were both found and confirmed fixed on real containers, not against constructed catalog values. The override was exercised against real AGE once, for the first state in (f): with the override set and `pg_available_extensions` unreadable, `configure_age_extension` installed AGE 1.8.0 on a live container — the measurement that produced (f). No other override path was run against real AGE.

Separately, as a regression check rather than gate evidence: upstream's pg-smoke job reproduced locally against a `postgres:18-alpine` container using the workflow's own command — 43 passed. That job uses no AGE extension and does not exercise this gate.

Offline suite installed the way `tests.yml` installs it — `uv sync --frozen`, then `uv run --no-sync pytest tests/ -m offline`: 5421 passed, 47 skipped, 1284 deselected, zero failures, on Python 3.12 and on 3.14.6 alike. The gate module's 206 cases are selected by that `-m offline` leg rather than silently deselected: the same run with the gate file ignored gives 5215 passed, and 5215 + 206 = 5421. `pre-commit run --all-files` at the config-pinned ruff v0.15.17: all hooks pass. Base is current — `upstream/main` merged (83 commits) before verifying, and the merge did not move the pinned ruff rev.

danielaskdd:
@codex review

PR #3622, opened 2026-08-10
fix(graph): preserve self-loops on entity rename

pengpengyi92:
## Summary

- preserve an already-supported self-loop relation when its entity is manually renamed
- remap the source and target endpoints independently before the old node is deleted
- add a focused regression that covers the graph edge and the existing relation-vector/chunk-tracking rename cascade

## Motivation

A self-loop `(A, A)` currently enters the same rename branch as an ordinary incident edge. Only its source endpoint is changed, producing `(B, A)`; deleting node `A` then removes that edge. The same partial endpoint pair is also passed to the existing relation-vector and chunk-tracking update path.

This is limited to manual entity rename when the renamed entity has a self-loop. Ordinary incident-edge rename behavior is unchanged.

## Change

Map each endpoint from the old entity name to the new name independently. For an ordinary edge, one endpoint changes as before. For `(A, A)`, both endpoints become `(B, B)`.

## Verification

- Regression test on `origin/main` (`b93f7c31`): **red**, because `(B, B)` was absent after renaming `A` to `B`
- Same focused test after the change: **1 passed**
- Adjacent graph API validation tests: **40 passed**
- Expanded direct pytest run: **630 passed, 5 skipped, 5 failed**
  - the five failures were outside the changed paths: three Windows symlink/POSIX-mode cases, one existing prompt-text assertion mismatch in untouched code, and one Windows NetworkX atomic-replace `PermissionError`
- Control rerun with those five cases deselected: **630 passed, 5 skipped, 5 deselected**
- Repository-pinned `ruff-format` and `ruff` hooks on both changed files: **passed**
- `git diff --check`: **passed**

### DIFF
--- a/lightrag/utils_graph.py
+++ b/lightrag/utils_graph.py
@@ -348,16 +348,22 @@ async def _edit_entity_impl(
                     relations_to_delete.append(
                         compute_mdhash_id(target + source, prefix="rel-")
                     )
-                    if source == entity_name:
-                        await chunk_entity_relation_graph.upsert_edge(
-                            new_entity_name, target, edge_data
-                        )
-                        relations_to_update.append((new_entity_name, target, edge_data))
-                    else:  # target == entity_name
-                        await chunk_entity_relation_graph.upsert_edge(
-                            source, new_entity_name, edge_data
-                        )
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3625, opened 2026-08-11
fix(graph): reject self-loop relations at the manual write paths

danielaskdd:
## Description

Stacked on #3622. That PR fixes what a self-loop does to state when its entity is renamed; this one closes the root cause — the two paths that let `src == tgt` reach the graph in the first place.

> **Stacking note:** GitHub cannot use a fork branch as a PR base, so #3622's commit (`5d134a7`, authored by @pengpengyi92) is included here rather than set as the base. **Merge #3622 first**; this PR's diff then reduces to the second commit alone. Review only `9f30b77` — the commit before it is #3622 verbatim, unmodified.

LightRAG's own extraction never emits a self-loop, and the codebase enforces that in four places:

| Location | Behavior |
|---|---|
| `operate.py:768` (delimited-record parser) | `if source == target: return None` |
| `operate.py:1006` (JSON parser) | `continue` |
| `operate.py:2698` (`_merge_edges_then_upsert`) | `if src_id == tgt_id: return None` |
| `utils_graph.py:1483` (`amerge_entities`) | skips any pair that would collapse into one |

Two manual paths had no such guard, verified by direct execution against `main`:

1. **`acreate_relation`** (`POST /graph/relation/create`) — accepted `source_entity == target_entity` and wrote both the graph edge `(A, A)` and the VDB record. Only entity existence and duplicate-edge checks ran.
2. **`ainsert_custom_kg`** — accepted a relationship with `src_id == tgt_id`; the identifier-normalization phase checked type and emptiness, but not equality.

A self-loop contributes no connectivity to graph retrieval, so it is pure surface area for the rest of the code to reason about — which is exactly what the rename bug in #3622 was.

## Related Issues

Root-cause follow-up to #3622. The two are complementary and neither subsumes the other:

- This PR stops **new** self-loops from entering.
- #3622 keeps a **pre-existing** self-loop (created before these guards) from stranding a phantom relation when its entity is renamed.

For the record, the state corruption #3622 addresses is worse than its description suggests. Renaming `A`→`B` where `A` has a self-loop, measured on `b93f7c3` (chunk-tracking keys written below as `key(x,y)`, i.e. `make_relation_chunk_key`):

| | graph | relations_vdb | relation_chunks |
|---|---|---|---|
| before | `(A,A)`, `(A,C)` | `A~A`, `A~C` | `key(A,A)`, `key(A,C)` |
| after | only `(B,C)` | **`A~B`** ← phantom, `B~C` | **`key(A,B)`** ← phantom, `key(B,C)` |

The self-loop is not merely lost: a relation record is fabricated for `A~B`, naming an entity that was just deleted and an edge that never existed in the graph. Since `global`/`hybrid` retrieval queries `relations_vdb` directly, that record is retrievable.

## Changes Made

**`lightrag/utils_graph.py`**
- New `_reject_self_loop_relation()` helper beside `_require_non_empty_description`, raising `ValueError` (already mapped to HTTP 400 by `graph_routes.py:667`).
- Called from `acreate_relation` **before** the keyed lock, so a refusal writes nothing.
- Comparison is raw, which is exact rather than lax here: `upsert_edge` is called with those same strings, so raw equality is precisely the condition that produces a self-loop edge.

**`lightrag/lightrag.py`**
- `ainsert_custom_kg` rejects `src_id == tgt_id` in the identifier-validation phase that already precedes every chunk and graph write, so a rejected batch leaves storage untouched. The error names the offending index (`relationships[1] is a self-loop on 'B'`), matching the sibling validators.
- Compared **after** normalization, not on the caller's raw strings: the normalized values are what get written as the edge, so two spellings that canonicalize to the same name are the same self-loop.

**`aedit_relation` deliberately unchanged** — it cannot alter an existing relation's endpoints, so it can't create a self-loop. Editing one that already exists stays possible.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [ ] Documentation updated (if necessary) — no user-facing behavior beyond the new 400

## Additional Notes

**Tests** — `tests/api/routes/test_graph_self_loop_rejection.py` (2) and `tests/pipeline/test_custom_kg_self_loop_rejection.py` (3), placed per the `AGENTS.md` layout convention and following the existing precedent of `test_description_api_validation.py` and `test_graph_keyed_locks.py`.

Both guards are pinned as refusing *before* any write, not merely refusing:
- The `acreate_relation` test passes `None` storages — an `AttributeError` instead of the `ValueError` would mean the check landed after the lock, which dereferences `relationships_vdb.global_config`.
- The custom-KG tests assert `upsert_edge` / `upsert_node` / `relationships_vdb.upsert` were never called.

One test (`test_acreate_relation_still_accepts_distinct_endpoints`) is a control that pins no-regression on the ordinary two-entity path; it passes both with and without the guard, as intended.

**Verification**
- Red/green: the 4 guard tests fail without the change, pass with it. The control test passes both ways.
- Targeted suites (incl. #3622's own regression test, `test_description_api_validation`, `test_graph_keyed_locks`, `test_vdb_content_truncation_coverage`, `test_rebuild_vdb`): **92 passed**.
- Broad sweep (`tests/api`, `tests/pipeline`, `tests/extraction`, `tests/tools`, `tests/utils`): **1620 passed, 74 failed** vs. a **1615 passed, 74 failed** baseline on the same tree without these changes — an identical failure set, +5 being exactly the new tests. The 74 are sandbox artifacts (tiktoken encoding download blocked by the proxy, `asyncpg`/`ollama` not installable offline), not regressions.
- `ruff check` and `ruff format --check` clean on all four files.

**Behavior change worth a maintainer decision:** `ainsert_custom_kg` now raises instead of silently accepting a self-loop, so a caller currently passing one gets a `ValueError` where they previously got an inserted edge. Fail-fast matches that method's existing contract for malformed identifiers, and the message names the index so the payload is easy to fix. If you would rather it skip-and-warn like the extraction parsers do, that's a small change — say so and I'll switch it.

**Remaining inconsistency, out of scope here:** with #3622 merged, rename *preserves* a self-loop while `amerge_entities` *drops* one (verified: the drop is clean — graph, VDB, and chunk tracking all stay consistent, no phantom). Neither corrupts state, but the two policies disagree. Once these guards are in, no new self-loops arise, so this only affects graphs that already contain one. Happy to align merge with rename in a follow-up if you want a single policy.

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -3521,6 +3521,21 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
                         field=f"relationships[{index}].tgt_id",
                     )
                 )
+                # Compared after normalization, because the normalized values
+                # are the ones written as the graph edge below: two spellings
+                # that canonicalize to the same name are the same self-loop.
+                # Extraction drops self-loops (operate.py) and amerge_entities
+                # refuses to form one, so accepting them here would make this
+                # the only path that puts src == tgt into the graph.
+                if (
+                    normalized_relationship_data["src_id"]
+                    == normalized_relationship_data["tgt_id"]
+                ):
+                    raise ValueError(
+                        f"Custom KG relationships[{index}] is a self-loop on "
+                        f"'{normalized_relationship_data['src_id']}': src_id and "
+                        "tgt_id must be different entities"
+                    )
                 normalized_relationships.append(normalized_relationship_data)
 
             # Insert chunks into vector storage
--- a/lightrag/utils_graph.py
+++ b/lightrag/utils_graph.py
@@ -348,16 +348,22 @@ async def _edit_entity_impl(
                     relations_to_delete.append(
                         compute_mdhash_id(target + source, prefix="rel-")
                     )
-                    if source == entity_name:
-                        await chunk_entity_relation_graph.upsert_edge(
-                            new_entity_name, target, edge_data
-                        )
-                        relations_to_update.append((new_entity_name, target, edge_data))
-                    else:  # target == entity_name
-                        await chunk_entity_relation_graph.upsert_edge(
-                            source, new_entity_name, edge_data
-                        )
-                        relations_to_update.append((source, new_entity_name, edge_data))
+                    # Rename every matching endpoint. A self-loop has the old
+                    # entity on both sides, so changing only the first match
+                    # would create (new, old), which is then removed together
+                    # with the old node below.
+                    renamed_source = (
+                        new_entity_name if source == entity_name else source
+                    )
+                    renamed_target = (
+                        new_entity_name if target == entity_name else target
+                    )
+                    await chunk_entity_relation_graph.upsert_edge(
+                        renamed_source, renamed_target, edge_data
+                    )
+                    relations_to_update.append(
+                        (renamed_source, renamed_target, edge_data)
+                    )
 
         await chunk_entity_relation_graph.delete_node(entity_name)
 
--- a/lightrag/utils_graph.py
+++ b/lightrag/utils_graph.py
@@ -28,6 +28,29 @@ def _require_non_empty_description(
         )
 
 
+def _reject_self_loop_relation(
+    source_entity: Any, target_entity: Any, *, operation: str
+) -> None:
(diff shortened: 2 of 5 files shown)


danielaskdd:
@codex review

PR #3626, opened 2026-08-11
fix(api): reject unstorable characters in a document source

danielaskdd:
## Summary

Closes the last ingress through which an attribute no graph backend can store still reached the knowledge graph: the document source. #3623 (entry-layer allowlist) and #3624 (NetworkX backstop) are merged; this is the ingress half.

`file_path` is the one extracted attribute that arrives at graph storage with **no character cleaning at all**. The extraction handlers pass it through untouched ([operate.py:719](https://github.com/HKUDS/LightRAG/blob/main/lightrag/operate.py#L719), [:799](https://github.com/HKUDS/LightRAG/blob/main/lightrag/operate.py#L799)), unlike `description` / `entity_type` / `keywords`, which all go through `sanitize_and_normalize_extracted_text` → `sanitize_text_for_encoding`. And it is stamped onto **every** entity and relation extracted from the document.

## Motivation

The three ingress paths disagreed about what a document source may contain:

| Path | Validator | Control chars | Surrogates / U+FFFE |
|---|---|---|---|
| `/documents/upload` | `sanitize_filename` | rejected | **accepted** |
| `/documents/text{,s}` | `normalize_file_path` | **accepted** | **accepted** |
| `/documents/scan` | — none — | **accepted** | **accepted** |

`normalize_file_path` strips whitespace and canonicalizes a `[hint]` segment; that is all. So `POST /documents/text` with `{"file_source": "ab.txt"}` put that value on every node and edge the document produced, reached `upsert_node`, and — on the default NetworkX backend — poisoned the in-memory graph: the same instance-wide write outage as GHSA-c922-pw4m-4wcv, through an ingress that advisory's field allowlist does not cover, because these are not graph-mutation routes.

Measured before writing the fix, tracing a value through the normalizers it actually passes:

```
control char     survives='a\x0bb.txt'    storable=NO -> U+000B
NUL              survives='a\x00b.txt'    storable=NO -> U+0000
lone surrogate   survives='a\ud800b.txt'  storable=NO -> U+D800
non-character    survives='a￾b.txt'  storable=NO -> U+FFFE
```

Surrogates and the non-characters were accepted on **both** paths, and GraphML refuses them exactly like a control character. A surrogate is reachable in practice, not just in theory: `surrogateescape` decoding of a non-UTF-8 multipart filename header produces one.

## What's changed

One shared rule, `find_unsafe_document_source_character`, used by both paths.

It is **stricter** than the graph-attribute rule from #3623, deliberately. Tab, newline and carriage return are legal XML and legal inside a description, but a *filename* holding one is pathological — and `sanitize_filename` has always refused them. So the shared rule is that function's existing check (`ord(c) < 32`, `\x7f`) **plus** surrogates and U+FFFE/U+FFFF: a strict superset, so no upload behaviour is lost.

**Reject, never rewrite** — the reason `sanitize_filename` already documents: the source doubles as a document identifier, a filename dedup key and the `doc_id` seed, so silently rewriting it would change document identity and could collide with a legitimate document.

### `/documents/scan`

INPUT_DIR is populated by the filesystem — a volume mount, an rsync, an operator's `cp` — so scan is an ingress with no request to validate, and nothing between `os.scandir` and `doc_status` filtered characters: discovery gates on the suffix, and `normalize_file_path` only strips a parser hint. Reproduced: `scan\x0bsource.txt` in INPUT_DIR was discovered, enqueued, and its basename stamped onto every entity and relation. Post-#3624 the NetworkX guard turns that into an asynchronous FAILED document instead of an instance-wide outage — the load-bearing defence still holds — but the value should never have been accepted.

Same policy, since the upload path never renames either: a new `_ScanFileClass.UNSAFE_SOURCE` exit, decided **first** in the discovery loop, ahead of `classify_scan_file`. The file **stays in INPUT_DIR**, exactly as `SOURCE_CONFLICT` does — the basename is the document identity, so the fix is an operator rename, not this scan rewriting the operator's file or hiding it in `__parsed__`. Every scan re-warns until it is fixed.

Being first is load-bearing, and fixes a second hazard on its own: a lone surrogate basename (`surrogateescape` decoding of a non-UTF-8 directory entry) reaching *any* later exit put the raw name into `record_scan_warning` → `pipeline_status` history, or into `doc_status.file_path`. Starlette renders responses with `json.dumps(..., ensure_ascii=False).encode("utf-8")`, which **raises** on a lone surrogate, so one such file would have 500'd every later read of `/documents/pipeline_status`. `describe_rejected_document_source` escapes the name with `ascii()` for every operator-facing message, and no code path formats the raw name any more.

A backstop also sits in `pipeline_enqueue_file`, the chokepoint every file-shaped ingest passes. It refuses **without** an error document: that construction interpolates `file_path.name` raw and would persist the offending name into `doc_status.file_path`, which is exactly the outcome being prevented. The pre-fix test output shows it happening: `'file_path': 'scan\x0bsource.txt'`.

**The rejection is at the boundary and deliberately NOT inside `normalize_file_path`.** That helper also runs on *read* paths — it normalizes `doc_status.file_path` when building document listings ([document_routes.py:6032](https://github.com/HKUDS/LightRAG/blob/main/lightrag/api/routers/document_routes.py#L6032), [:6269](https://github.com/HKUDS/LightRAG/blob/main/lightrag/api/routers/document_routes.py#L6269), [:6421](https://github.com/HKUDS/LightRAG/blob/main/lightrag/api/routers/document_routes.py#L6421)) — so making it raise would turn one already-stored bad source into a 500 for the whole listing, and leave no way to delete the row. This is the same reasoning that keeps NetworkX from validating attribute *names* in #3624, and the same trap the Codex review caught there; `TestReadPathsStayNonRaising` pins it.

## Behaviour change

Two endpoints now reject input they previously accepted:

- `/documents/text`, `/documents/texts`: a `file_source` / `file_sources` entry containing a control character, DEL, a lone surrogate, or U+FFFE/U+FFFF → **422** (Pydantic field validator — the one place both routes pass through, and the convention this API already uses for model-level validation).
- `/documents/upload`: a filename containing a lone surrogate or U+FFFE/U+FFFF → **400** (`sanitize_filename`'s existing contract; control characters were already 400).
- `/documents/scan`: a file in INPUT_DIR whose basename holds one of those characters is skipped with a warning and an `unsafe_source` counter in `/documents/scan/status`, instead of being enqueued. It is not archived, renamed or deleted.

The status codes differ because each path keeps its own existing convention rather than one of them being rewritten to match the other.

An omitted `file_source` is untouched: Pydantic skips a `mode="before"` validator for a field that was never supplied, so the default `None` still reaches the endpoint and the existing "a valid file_source is required" 400 still applies.

Not covered, by design: `ainsert(file_paths=[...])`. A library caller is governed by the `BaseGraphStorage.upsert_node` contract documented in #3623 and backstopped by the NetworkX guard in #3624.

## Testing

```
./scripts/test.sh tests/api tests/parser tests/pipeline
2958 passed, 15 skipped
```

New: 10 offline tests in `tests/api/routes/test_scan_unsafe_source.py`, driving the **real** `iter_new_files` and the real classification loop — nothing enqueued, the file left in place, no `doc_status` identity lookup for the rejected name, the warning naming the code point, and the recorded history staying renderable with `ensure_ascii=False`. 7 of the 10 fail on the pre-fix tree. The end-to-end surrogate case skips where the filesystem refuses such a name (APFS does; Linux does not), with a filesystem-independent test covering the same rendering hazard.

Plus 59 offline tests in `tests/api/routes/test_document_source_characters.py` — the character rule itself, both request models (including an offender in a non-first position of `file_sources`), `sanitize_filename`'s unchanged 400 contract, ordinary sources still accepted (CJK, spaces, dots, emoji), and the read path staying non-raising.

### DIFF
--- a/lightrag/api/routers/document_routes.py
+++ b/lightrag/api/routers/document_routes.py
@@ -168,6 +168,63 @@ ARCHIVED_FILE_SUFFIX_RE = re.compile(r"_(?:\d{3}|\d{10,})$")
 _ADMISSION_RETRY_AFTER_SECONDS = 30
 
 
+# Characters a document source must not carry. Stricter than the graph-attribute
+# rule in ``lightrag.utils`` on purpose: tab/newline/carriage return are legal
+# XML and legal in a description, but a *filename* holding one is pathological,
+# and ``sanitize_filename`` has always refused them. Surrogates and the
+# non-characters are the additions -- they reach a filename through
+# ``surrogateescape`` decoding of a non-UTF-8 multipart header, and they are
+# exactly as unserializable as a control character once the value is stamped
+# onto graph nodes.
+_UNSAFE_DOCUMENT_SOURCE_CHAR_PATTERN = re.compile(
+    "[\u0000-\u001f\u007f\ud800-\udfff\ufffe\uffff]"
+)
+
+
+def find_unsafe_document_source_character(value: str) -> str | None:
+    """Return the first character that disqualifies *value* as a document source.
+
+    A document source is not only a display string: it is stamped onto the
+    ``file_path`` attribute of every entity and relation extracted from the
+    document. Nothing downstream cleans it -- the extraction handlers pass
+    ``file_path`` through untouched, unlike ``description`` / ``entity_type`` /
+    ``keywords``, which all go through ``sanitize_text_for_encoding``. So a value
+    accepted here is a value the graph backends must be able to store, and one
+    they cannot store takes out writes for the whole instance on NetworkX (see
+    GHSA-c922-pw4m-4wcv).
+
+    Rejecting rather than rewriting, for the reason ``sanitize_filename``
+    already gives: the source doubles as a document identifier, a dedup key and
+    the ``doc_id`` seed, so silently rewriting it would change document identity
+    and could collide with a legitimate document.
+
+    Args:
+        value: Candidate document source (uploaded filename or ``file_source``).
+
+    Returns:
+        The offending character, or ``None`` when the value is usable.
+    """
+    match = _UNSAFE_DOCUMENT_SOURCE_CHAR_PATTERN.search(value)
+    return match.group() if match else None
+
+
+def reject_unsafe_document_source(value: str | None) -> str | None:
+    """Raise ``ValueError`` when *value* holds a character a source may not carry.
+
+    Shaped for a Pydantic field validator (the one place ``/documents/text`` and
+    ``/documents/texts`` both pass through); ``sanitize_filename`` performs the
+    same check but raises ``HTTPException`` to match the rest of the upload path.
+    """
+    if value is None:
+        return None
+    offender = find_unsafe_document_source_character(value)
+    if offender is not None:
+        raise ValueError(
+            f"file_source must not contain the character U+{ord(offender):04X}"
+        )
+    return value
+
+
 def normalize_file_path(file_path: str | None) -> str:
     """Normalize missing document sources to a single non-null sentinel."""
     if file_path is None:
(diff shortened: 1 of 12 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

Issue #3584, opened 2026-08-06
[Bug]: gRPC RESOURCE_EXHAUSTED ("received message larger than max") when retrieving vectors by IDs from Milvus with high-dimension embeddings


El-Tatane:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Description
When running a query, LightRAG fails to retrieve chunk vectors from Milvus with a gRPC `RESOURCE_EXHAUSTED` error. The Milvus storage backend fetches all requested vectors in a **single** `query()` call, so the gRPC response can exceed the **client-side** default max receive message length (4 MiB / `4194304` bytes).
 
It is easily triggered with high-dimension embeddings such as OpenAI `text-embedding-3-large` (3072 dims). Each serialized float vector is ~12 KB, so only a few hundred rows returned in one response are enough to cross the 4 MiB client limit.

### Where
 
`MilvusVectorDBStorage` — get_vectors_by_ids in `lightrag/kg/milvus_impl.py`:
https://github.com/HKUDS/LightRAG/blob/main/lightrag/kg/milvus_impl.py
 
The offending part is the unbounded single `query()`:
 
```python
rows = self._client.query(
    collection_name=self.final_namespace,
    filter=filter_expr,
    output_fields=["id", "vector"],
)
```

### Proposed solution

Paginate the retrieval with `query_iterator(batch_size=...)` so each gRPC message stays bounded, and expose the page size via an env var. This mirrors the pattern already used by the schema-migration path in the same file (`query_iterator` + `DEFAULT_MILVUS_MIGRATION_ITERATOR_BATCH_SIZE`)

```python
iterator = self._client.query_iterator(
    collection_name=self.final_namespace,
    filter=filter_expr,
    batch_size=DEFAULT_MILVUS_MIGRATION_ITERATOR_BATCH_SIZE,
    output_fields=["id", "vector"],
)
```

### Alternative considered
 
Raising the gRPC client `max_receive_message_length` channel option. Rejected because pymilvus does not expose it cleanly through `MilvusClient`, and it only pushes the ceiling higher rather than removing the unbounded-response failure mode. The iterator approach scales regardless of collection or vector size.

### Willing to submit a PR ?
 
Yes, happy to open a PR implementing the `query_iterator` approach with the `MILVUS_QUERY_ITERATOR_BATCH_SIZE` env var if the maintainers agree with the direction.

[...]

PR #3628, opened 2026-08-11
fix(milvus): page bulk ID queries to bound response size

pengpengyi92:
## Description

Milvus `get_by_ids` and `get_vectors_by_ids` currently place every requested ID in one query. For high-dimensional vectors or large ID sets, the response can exceed a managed gateway's gRPC message limit and fail with `RESOURCE_EXHAUSTED`.

This change pages bulk ID queries so each response remains bounded while preserving the existing read-your-writes behavior.

## Related Issues

No linked issue.

## Changes Made

- Add one shared helper for paged Milvus ID queries.
- Use a 256-record cap for metadata reads.
- Derive vector page size from the embedding dimension against a conservative 2 MiB response budget.
- Preserve caller ordering, buffered/deleted ID handling, ID escaping, and atomic failure behavior.
- Avoid logging an entire potentially large ID list on failure.
- Add offline regression tests for high-dimensional paging, page-size calculation, ordering, escaping, and partial-page failure.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (not required; no public API change)
- [x] Unit tests added

## Test Results

```
py -m pytest tests/kg/milvus_impl -q
184 passed in 0.90s
```

The regression tests use a mocked Milvus client and do not require a live Milvus service.

### DIFF
--- a/lightrag/kg/milvus_impl.py
+++ b/lightrag/kg/milvus_impl.py
@@ -61,6 +61,14 @@ DEFAULT_MILVUS_UPSERT_MAX_PAYLOAD_BYTES = (
 DEFAULT_MILVUS_UPSERT_MAX_RECORDS_PER_BATCH = 128
 DEFAULT_MILVUS_DELETE_MAX_RECORDS_PER_BATCH = 1000
 
+# Managed Milvus gateways can enforce gRPC's 4 MiB response ceiling even when
+# self-hosted Milvus accepts much larger messages. Keep ID-query responses below
+# 2 MiB for a 2x safety margin. Vector pages account for float32 payload bytes;
+# metadata pages use the record cap because content length is not predictable.
+MILVUS_QUERY_MAX_RESPONSE_BYTES = 2 * 1024 * 1024
+MILVUS_QUERY_MAX_RECORDS_PER_BATCH = 256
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
Closing as a duplicate of #3629, which fixes the same issue (#3584) with an essentially identical approach — same constant names and values, same helper signature, same test file path. The two were opened 8 minutes apart, so this is parallel work, not a quality problem with your PR.

Keeping #3629 because @El-Tatane reported #3584 and its test suite covers a few more cases. That said, your `_cooperative_yield(..., every=1)` cadence was the correct one — #3629 used the default `every=64`, which needs >16k ids before yielding once, and each page there is a blocking gRPC round-trip. I've adopted your version in https://github.com/HKUDS/LightRAG/commit/3c2306b1 with credit in the commit message.

Thanks for the fix, and sorry for the collision.

pengpengyi92:
Thank you for the detailed explanation and for crediting the cooperative-yield cadence in the adopted fix. I appreciate the careful comparison and completely understand keeping #3629 for its broader coverage. Glad the contribution was still useful.

PR #3635, opened 2026-08-13
fix(llm): close OpenAI stream/client when GeneratorExit interrupts a mid-COT close

### LINKED ISSUE #3634: [Bug]: OpenAI streaming client leaks when the consumer disconnects mid-COT

### Describe the bug

When streaming from an OpenAI-compatible reasoning model with `enable_cot=True`, if the consumer stops reading the stream while a `<think>` block is still open (an HTTP client disconnecting mid reasoning-phase is the common case), closing the async generator raises `RuntimeError: async generator ignored GeneratorExit` from `aclose()`, and neither the underlying OpenAI stream nor the `AsyncOpenAI` client gets closed.

`inner()` in `openai_complete_if_cache`'s streaming branch has a `finally` block that tries to close a still-open COT tag with `yield "</think>"` as a last-resort safety net. When the generator is being closed via `GeneratorExit` (e.g. `gen.aclose()`), yielding again at that point is not allowed by the async generator protocol: Python raises `RuntimeError: async generator ignored GeneratorExit` from `aclose()` itself, and the generator frame is left suspended before it reaches the `response.aclose()` / `openai_async_client.close()` calls that follow the safety-net block. Those never run, so the stream and the OpenAI client leak.

This should be an everyday occurrence for anyone streaming a reasoning model behind an HTTP endpoint (FastAPI `StreamingResponse`, etc.), since a client disconnect during the (often multi-second) "thinking" phase is normal traffic, not an edge case.

### Steps to reproduce

Minimal repro against the real `lightrag.llm.openai.openai_complete_if_cache`, with only the network-facing `AsyncOpenAI` client mocked:

```python
import asyncio
from unittest.mock import AsyncMock, MagicMock
import lightrag.llm.openai as oa_mod

class FakeDelta:
    def __init__(self, content=None, reasoning_content=""):
        self.content = content
        self.reasoning_content = reasoning_content

class FakeChunk:
    def __init__(self, content=None, reasoning_content=""):
        self.choices = [type("C", (), {"delta": FakeDelta(content, reasoning_content)})()]
        self.usage = None

class FakeStream:
    def __init__(self, chunks):
        self._chunks, self._i, self.closed = chunks, 0, False
    def __aiter__(self):
        return self
    async def __anext__(self):
        if self._i >= len(self._chunks):
            raise StopAsyncIteration
        c, self._i = self._chunks[self._i], self._i + 1
        return c
    async def aclose(self):
        self.closed = True

async def main():
    fake_stream = FakeStream([
        FakeChunk(reasoning_content="Let me think"),
        FakeChunk(reasoning_content=" about this."),
    ])
    fake_client = MagicMock()
    fake_client.chat.completions.create = AsyncMock(return_value=fake_stream)
    fake_client.close = AsyncMock()
    oa_mod.create_openai_async_client = lambda *a, **k: fake_client

    gen = await oa_mod.openai_complete_if_cache(
        model="deepseek-reasoner", prompt="hi", enable_cot=True, stream=True,
        api_key="sk-fake", base_url="http://fake.invalid/v1",
    )
    await gen.__anext__()  # yields "<think>"
    await gen.__anext__()  # yields "Let me think"
    await gen.aclose()     # simulates the client disconnecting mid-COT

    print("stream.aclose() ran:", fake_stream.closed)
    print("client.close() ran:", fake_client.close.called)

asyncio.run(main())
```

### Expected Behavior

`gen.aclose()` completes without raising, and both `fake_stream.closed` and `fake_client.close.called` are `True` (the stream and client are closed like any other early-close case).

### Logs and screenshots

```
first yielded: '<think>'
second yielded: 'Let me think'
Traceback (most recent call last):
  ...
RuntimeError: async generator ignored GeneratorExit
stream.aclose() ran: False
client.close() ran: False
```

The `RuntimeError` is raised by `aclose()` itself (visible via `asyncio`'s "an error occurred during closing of asynchronous generator" log line), not caught anywhere in `inner()`, and both cleanup calls are skipped.

### Additional Information

- LightRAG Version: verified against `main` at `71f7b23b938298a5f768753f155c15cba1214ade`
- Operating System: Linux (reproduced in a clean `python:3.12-slim` container)
- Python Version: 3.12
- Related Issues: none found (`gh api search/issues` for `GeneratorExit`, `cot_active` in this repo turned up nothing related)

Happy to open a PR: the fix is small (skip the finally block's yield-based COT close specifically when the generator is unwinding via `GeneratorExit`, since the two earlier COT-close sites already cover the normal-completion and exception paths).

AmirF194:
## Description

`openai_complete_if_cache`'s streaming branch has a `finally`-block safety net that closes a still-open `<think>` tag with `yield "</think>"`. When the consumer disconnects while COT is open (an HTTP client going away mid reasoning-phase), `aclose()` throws `GeneratorExit` at that suspension point; yielding again is illegal there, so `aclose()` raises `RuntimeError: async generator ignored GeneratorExit` and the `finally` block is aborted before the `response.aclose()` / `openai_async_client.close()` calls that follow it run. The stream and the client leak.

## Related Issues

Fixes #3634

## Changes Made

Adds an `except GeneratorExit` clause that records the unwind reason and re-raises. The `finally` block's COT-close attempt is skipped specifically when closing via `GeneratorExit` (a yield is exactly what's disallowed there), and falls through unconditionally to the existing `response.aclose()` / `openai_async_client.close()` calls. The two other COT-close sites (after the loop ends, and inside the `except Exception` handler) are unaffected: `GeneratorExit` is a `BaseException` and never reaches the `except Exception` branch, and the post-loop close only runs on normal completion.

New regression test `tests/llm/openai_impl/test_openai_stream_generatorexit_cot_cleanup.py` drives the real `openai_complete_if_cache` through a disconnect mid-COT (only the network-facing client mocked) and asserts both the stream and the client get closed; a second test pins that a full COT stream still closes cleanly.

## Checklist

- [x] Changes tested locally
- [x] Pre-commit checks pass (`pre-commit run --all-files`)
- [x] Unit/integration tests added or updated where applicable
- [ ] Documentation updated if behavior changes
- [x] PR description explains the *why*, not just the *what*

## Additional Notes

Verified in a clean `python:3.12-slim` container, importing the real module (`lightrag.llm.openai.__file__` printed for provenance):

- The new test fails on `main` with the exact `RuntimeError: async generator ignored GeneratorExit` above, and passes on this branch, both closing the stream and the client.
- Full offline suite (`pytest tests/ -m offline`): 5213 passed, 45 skipped (pre-existing, missing spaCy models, unrelated), 0 failed on Python 3.12; the `openai_impl` subset also passes clean on 3.14.
- `coverage run --source=lightrag.llm.openai` confirms every changed line is exercised by the new test.
- `pre-commit run --files lightrag/llm/openai.py <test file>` clean.

Not verified: `GeneratorExit` arriving while the generator is suspended inside the `except Exception` handler's own COT-close `yield` (a real API error, then a disconnect before exception-path cleanup finishes) is a narrower variant of the same class of bug and is left out of scope here.

### DIFF
--- a/lightrag/llm/openai.py
+++ b/lightrag/llm/openai.py
@@ -516,6 +516,7 @@ async def openai_complete_if_cache(
             cot_active = False
             cot_started = False
             initial_content_seen = False
+            closing_via_generator_exit = False
 
             try:
                 iteration_started = True
@@ -611,6 +612,12 @@ async def openai_complete_if_cache(
                     logger.debug(f"Streaming token usage (from API): {token_counts}")
                 elif token_tracker:
                     logger.debug("No usage information available in streaming response")
+            except GeneratorExit:
+                # Consumer disconnect mid-COT: tell the finally block below to
+                # skip its yield, which would otherwise abort cleanup with
+                # "async generator ignored GeneratorExit".
+                closing_via_generator_exit = True
+                raise
             except Exception as e:
                 # Ensure COT is properly closed before handling exception
                 if enable_cot and cot_active:
@@ -645,8 +652,9 @@ async def openai_complete_if_cache(
                     )
                 raise
             finally:
-                # Final safety check for unclosed COT tags
-                if enable_cot and cot_active:
+                # Final safety check for unclosed COT tags, skipped while
+                # closing via GeneratorExit (see the except clause above).
+                if enable_cot and cot_active and not closing_via_generator_exit:
                     try:
                         yield "</think>"
                         cot_active = False
--- a/tests/llm/openai_impl/test_openai_stream_generatorexit_cot_cleanup.py
+++ b/tests/llm/openai_impl/test_openai_stream_generatorexit_cot_cleanup.py
@@ -1,0 +1,125 @@
+"""Regression test: closing the streaming generator mid-COT must not abort
+cleanup.
+
+``inner()`` in ``openai_complete_if_cache``'s streaming branch has a
+finally-block safety net that closes an still-open ``<think>`` tag with
+``yield "</think>"``. When the consumer disconnects while COT is open
+(``gen.aclose()``, e.g. an HTTP client going away mid reasoning-phase),
+Python throws ``GeneratorExit`` at the generator's suspension point; yielding
+again in response is illegal and makes ``aclose()`` raise "async generator
+ignored GeneratorExit", aborting the finally block before the
+``response.aclose()`` / ``openai_async_client.close()`` calls that follow it
+run. Mirrors ``tests/llm/anthropic_impl/test_anthropic_client_cleanup.py``'s
+``test_stream_closed_on_early_consumer_break``.
+"""
+
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+from lightrag.llm.openai import openai_complete_if_cache
+
+
+class _FakeOpenAIStream:
+    """Async-iterable fake of an OpenAI streaming response.
+
+    ``__aiter__``/``__anext__`` live on the class, matching the real
+    ``openai.AsyncStream`` shape that ``async for`` requires.
+    """
+
+    def __init__(self, chunks):
+        self._chunks = list(chunks)
+        self.aclose = AsyncMock()
+
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self):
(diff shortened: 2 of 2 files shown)


danielaskdd:
Thx for sharing.

PR #3637, opened 2026-08-13
Refactor/parser cancellation exception rename

### LINKED ISSUE #3608: refactor(parser): rename LLMBridgeCancelled hierarchy to parser-wide cancellation names
## Context

PR #3605 extends the existing cancellation hierarchy beyond the synchronous LLM bridge. Native markdown image downloads now raise `LLMBridgePipelineCancelled` so `pipeline.py` records an interrupted document as cancelled rather than failed.

The behavior is correct, but the `LLMBridgeCancelled` / `LLMBridgePipelineCancelled` / `LLMBridgeShutdown` names now describe their first consumer rather than their parser-wide role. As more blocking parser paths reuse the cancellation protocol, the misleading names will spread.

This is technical debt only. It is not part of the security fix in #3605 and must not block that PR.

## Proposed direction

- Introduce parser-wide names such as `ParseCancelled`, `ParsePipelineCancelled`, and `ParseShutdown`.
- Update the LLM bridge, native parser runtime, pipeline catch sites, type annotations, and tests to use the new names.
- Keep the existing `LLMBridge*` names as compatibility aliases for at least one release cycle so existing imports and exception identity continue to work.
- Preserve the current exception hierarchy and cancellation-source mapping.

## Acceptance criteria

- Pipeline cancellation is still recorded as cancelled, not failed.
- Parser-executor shutdown retains its current distinct handling.
- Existing imports of all three `LLMBridge*` names continue to work through aliases.
- `isinstance` and exception-catching behavior remain compatible.
- Focused parser and pipeline cancellation tests pass.
- No runtime behavior change beyond the exported names and aliases.

Follow-up to #3605.

MasaSN:
## Description

PR #3605 (security fix for GHSA-25c3-j78v-83qx) reused the existing
`LLMBridgeCancelled` / `LLMBridgePipelineCancelled` / `LLMBridgeShutdown`
exception hierarchy to signal cancellation in the markdown parser's image
downloader. That hierarchy was originally built only for the LLM bridge
component, so parser code now raises exceptions named `LLMBridge*` for
reasons that have nothing to do with the LLM bridge — misleading, though
behaviorally correct.

This PR renames the hierarchy to reflect its real parser-wide scope, while
keeping the old names fully backward compatible via aliases. No behavior
changes, no new exception types — a rename plus a compat layer.

## Related Issues

Closes #3608

## Changes Made

- New `lightrag/parser/exceptions.py`: houses `ParseCancelled` /
  `ParsePipelineCancelled` / `ParseShutdown` (renamed from `LLMBridge*`),
  plus the parser-internal helpers `normalize_cancel_events` and
  `first_cancellation`.
- `lightrag/parser/llm_bridge.py`: re-exports the old names as plain
  assignments (`LLMBridgeCancelled = ParseCancelled`, etc.) — not
  subclasses, so `except LLMBridgeCancelled` still catches a raised
  `ParseCancelled` and vice versa. Old imports from
  `lightrag.parser.llm_bridge` keep working unchanged.
- `lightrag/parser/native_base.py`: internal call sites updated to import
  and use the new names directly; cancellation-rationale comments
  and the `LLMBridgeCancelled` docstring rewritten to drop the
  "named for its first consumer" framing.
- Tests: `tests/parser/test_llm_bridge.py` migrated to the new names; new
  `tests/parser/test_parser_exceptions.py` adds explicit alias-identity
  (`is`) checks and cross-catch tests (raise via one name, catch via the
  other, both directions).
- `tests/pipeline/test_pipeline_cancellation.py`,
  `tests/parser/markdown/test_download_deadline.py`, and
  `tests/parser/docx/test_zip_budget.py` are intentionally left unchanged
  and still pass — they use the old `LLMBridge*` names, which demonstrates
  the compat layer works.

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Additional Notes

`python -m pytest tests/parser tests/pipeline` → 2124 passed, 10 failed
locally. All 10 failures are pre-existing and unrelated to this change
(missing `cairo-2` system library for SVG rasterization, Windows symlink
privilege restrictions, and one Windows path-separator assertion) — same
failure count and set observed on `main` before this change.

`pre-commit run --all-files` passes.

### DIFF
--- a/lightrag/parser/native_base.py
+++ b/lightrag/parser/native_base.py
@@ -359,7 +359,7 @@ class NativeParserBase(BaseParser):
             # raise covers both.
             #
             # The rag-level shutdown event is checked here too, and raises the
-            # distinct LLMBridgeShutdown. _shutdown_parser_executor() sets it
+            # distinct ParseShutdown. _shutdown_parser_executor() sets it
             # and then calls executor.shutdown(wait=False), i.e. it does NOT
             # wait for a running extract — its contract is that in-flight work
             # exits via the event. Without this branch a worker still inside
--- a/lightrag/parser/native_base.py
+++ b/lightrag/parser/native_base.py
@@ -45,7 +45,7 @@ class NativeExtractRuntime:
     consumes none of them simply ignores the argument.
 
     ``cancel_events`` is the full set, in the ``(event, exception_type)`` shape
-    :func:`~lightrag.parser.llm_bridge.normalize_cancel_events` accepts, so a
+    :func:`~lightrag.parser.exceptions.normalize_cancel_events` accepts, so a
     consumer stays agnostic about which source fired: the per-parse event
     (first entry), the pipeline's (``/documents/cancel_pipeline``), and the
     rag shutdown event.
@@ -258,9 +258,9 @@ class NativeParserBase(BaseParser):
         # Per-parse cancel event, polled by the LLM bridge between waits. The
         # rag-level shutdown event (when present) covers finalize_storages
         # while an extract is still in flight.
-        from lightrag.parser.llm_bridge import (
-            LLMBridgePipelineCancelled,
-            LLMBridgeShutdown,
+        from lightrag.parser.exceptions import (
+            ParsePipelineCancelled,
+            ParseShutdown,
         )
 
         cancel_event = threading.Event()
@@ -277,8 +277,8 @@ class NativeParserBase(BaseParser):
         # network read.
         cancel_events = (
             cancel_event,
-            (ctx.pipeline_cancel_event, LLMBridgePipelineCancelled),
-            (shutdown_event, LLMBridgeShutdown),
+            (ctx.pipeline_cancel_event, ParsePipelineCancelled),
+            (shutdown_event, ParseShutdown),
         )
         llm_invoke = None
         smartheading_cache_keys: list = []
@@ -344,14 +344,14 @@ class NativeParserBase(BaseParser):
             # The source is gone from INPUT_DIR by then, so that is not
             # recoverable.
             #
-            # Raise LLMBridgePipelineCancelled, NOT asyncio.CancelledError:
+            # Raise ParsePipelineCancelled, NOT asyncio.CancelledError:
             # the pipeline-level cancel (_watch_pipeline_cancellation) only
             # SETS ctx.pipeline_cancel_event, it never cancels the worker task,
             # so a CancelledError raised here would be set on a live, un-
             # cancelled future and re-raised into _parse_worker as a
             # BaseException that matches neither of its except clauses — the
             # doc would strand in PARSING and, if every worker died this way,
-            # wedge the batch on q.join() with busy=True. LLMBridgePipelineCancelled
+            # wedge the batch on q.join() with busy=True. ParsePipelineCancelled
             # is the repo-wide "blocking parse wait was cancelled" signal that
             # _parse_worker catches to mark the doc cancelled. In the task-
             # cancel case the future is already cancelled, so awaiting it raises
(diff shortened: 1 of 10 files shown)


danielaskdd:
@codex review

lightrag/parser/native_base.py
```diff
@@ -366,18 +366,18 @@ def _extract_sync():
             # validate_source_blocking when finalize_storages() runs would walk
             # on to rmtree parsed_dir and extract into storages being torn down.
             # Shutdown stays a generic parse failure for audit (the parse worker
-            # catches only PipelineCancelledException / LLMBridgePipelineCancelled
+            # catches only PipelineCancelledException / ParsePipelineCancelled
             # as "cancelled"), which is the documented intent.
             pipeline_cancelled = ctx.pipeline_cancel_event
             with cleanup_lock:
                 if shutdown_event is not None and shutdown_event.is_set():
```
danielaskdd, line 373: The comment block just above (line 362: "...and raises the distinct `LLMBridgeShutdown`. `_shutdown_parser_executor()` sets it...") still names the old `LLMBridgeShutdown`, even though the code right here raises `ParseShutdown` and the sibling comments immediately around it in this same docblock were updated by this PR ("Raise ... NOT asyncio.CancelledError" and "catches only PipelineCancelledException / ParsePipelineCancelled" two lines below). This one line got missed — suggest "distinct `ParseShutdown`" for consistency with the rest of this docblock.

danielaskdd (commented):
Nice cleanup — the rename + backward-compat alias approach (`LLMBridgeCancelled = ParseCancelled`, not a subclass) is the right shape for a non-breaking rename, and the cross-catch tests in `test_parser_exceptions.py` are a good guardrail.

Two spots outside `lightrag/parser/native_base.py` still use the old, now-misleading `LLMBridge*` names. Neither is a functional bug (the compat aliases make `except LLMBridge*` still match a raised `Parse*`), but both are exactly the kind of confusion this PR sets out to remove, and they weren't touched by this diff:

**`lightrag/pipeline.py`** (lines 82, 4495, 4499) still does:
```python
from lightrag.parser.llm_bridge import LLMBridgePipelineCancelled
...
except (PipelineCancelledException, LLMBridgePipelineCancelled):
    ...
    # LLMBridgePipelineCancelled while waiting on the batch cancel
```
`pipeline.py`'s pipeline-cancellation handling has nothing to do with the LLM bridge — it's the same "misleading name for an unrelated consumer" complaint this PR's description makes about the parser code. In fact the new `lightrag/parser/exceptions.py` docstring already asserts *"``pipeline.py`` catches :class:`ParsePipelineCancelled`"*, which isn't literally true until this import/except is migrated to the new name too.

**`lightrag/parser/markdown/parser.py`** (lines ~74-77, ~1408) still does:
```python
from lightrag.parser.llm_bridge import (
    LLMBridgeCancelled,
    first_cancellation,
    normalize_cancel_events,
)
...
except LLMBridgeCancelled:
    ...
```
This is the file the PR description itself calls out as the *original* motivating misuse ("PR #3605 ... reused the existing `LLMBridgeCancelled` ... hierarchy to signal cancellation in the markdown parser's image downloader"). It seems worth migrating this call site too, alongside `native_base.py`.

One inline nit below on a comment in `native_base.py` that also got missed.

lightrag/parser/native_base.py
MasaSN: Thanks for the thorough review! All three are fixed:

pipeline.py: import, except clause, and comment now use ParsePipelineCancelled
markdown/parser.py: renamed the import and except clause from LLMBridgeCancelled to ParseCancelled (this was indeed the original motivating file, good catch)
native_base.py: updated the docblock comment to say ParseShutdown instead of the stale LLMBridgeShutdown

Commit: 23ba9f4

danielaskdd:
@codex review

danielaskdd:
Pushed `ef34889f0` on top of this PR: since the `LLMBridge*` exception names were only ever internal (raised and caught entirely inside the parser/pipeline, never part of a public surface), we decided to complete the rename outright instead of carrying a compat layer:

- **Removed the three alias assignments** in `lightrag/parser/llm_bridge.py` (`LLMBridgeCancelled = ParseCancelled`, etc.), along with the now-unused imports.
- **Migrated the three remaining old-name test files** (`tests/pipeline/test_pipeline_cancellation.py`, `tests/parser/markdown/test_download_deadline.py`, `tests/parser/docx/test_zip_budget.py`) to the `Parse*` names.
- **Reworked `tests/parser/test_parser_exceptions.py`**: the alias-identity and cross-catch tests are gone (their subject no longer exists); the hierarchy-shape tests remain, plus a new negative test asserting no `LLMBridge*` name is importable from `llm_bridge` anymore — so the misleading old names can't quietly come back.
- Updated the `ParseCancelled` docstring in `exceptions.py` to drop the "still available as aliases" wording.

Verified: repo-wide grep shows zero remaining `LLMBridge*` exception references (only `SyncLLMBridge` itself, which is genuinely the bridge), 78 tests across the five affected files pass, and pre-commit is green.

Thanks @MasaSN for the clean rename and the thorough test coverage — merging once CI completes.

PR #3638, opened 2026-08-13
fix(parser): bound external result downloads for MinerU and Docling

### LINKED ISSUE #3610: Bound external adapter result downloads (MinerU/Docling): unbounded body buffering, uncapped zip extraction, no wall-clock deadline
## Summary

PR #3605 bounds native markdown image downloads per document (byte / request / wall-clock budgets, capped chunked reads). The two external parser adapters — MinerU and Docling — contain sibling result-download paths that predate that PR and still lack the same class of bounds. Neither file is touched by #3605, so this is follow-up hardening, not a defect in that PR.

The adapter endpoints are operator-configured (a trusted service URL), but the **size and pacing of the result bundle derive from the uploaded document**, so a huge or misbehaving conversion output can exhaust worker memory or hold a parse slot indefinitely.

### 1. MinerU result download — `lightrag/parser/external/mineru/client.py` (`_download_zip`, ~442–460)

- The whole result bundle is buffered in memory via `resp.content` → `io.BytesIO(resp.content)` with **no byte cap**.
- The inline zip extraction has path-traversal guards but **no entry-count or uncompressed-size bounds** (it does not use the shared `safe_extract_zip` helper).
- `httpx.Timeout(120.0, connect=30.0)` (~line 208) is a **per-read** timeout: a peer that trickles one chunk per interval resets it forever. There is no wall-clock deadline for the download as a whole.

### 2. Docling result download — `lightrag/parser/external/docling/client.py` (~315)

- Same unbounded in-memory body via `resp.content`.
- It already calls the shared `safe_extract_zip`, but passes **neither `max_entries` nor `max_total_bytes`** — both knobs exist on the helper (`lightrag/parser/external/_zip.py`) and default to `None` = unlimited. The native markdown textpack caller passes both.
- Same per-read `httpx.Timeout(120.0)` (~line 168), same missing wall-clock deadline.

## Suggested fixes

1. **Docling (one-liner starter):** pass `max_entries=...`/`max_total_bytes=...` to the existing `safe_extract_zip` call, mirroring the textpack caller in `lightrag/parser/markdown/parser.py`.
2. **MinerU:** migrate the ad-hoc extraction to `safe_extract_zip` with the same caps, and cap the response body (stream to a bounded buffer or reject on `Content-Length`/read-cap overflow instead of `resp.content`).
3. **Wall-clock deadline:** both clients need a whole-download deadline in addition to httpx's per-read timeout. This plumbing belongs in the shared layer (`lightrag/parser/external/_common.py`) so both adapters inherit it, rather than being duplicated per client.

## Context

- Found during review of #3605; the budget/deadline machinery introduced there is markdown-engine-local, while the bounded-extraction helper (`safe_extract_zip` caps) already exists in the shared `_zip.py` layer and is simply unused by these two callers.
- Verified against the current `main`: the caps on `safe_extract_zip` and the capped textpack call both predate #3605 (commit 9b28d5f0), so the tooling for fix (1) is already in place.

Shizoqua:
## Description

`MinerURawClient._download_zip` and `DoclingRawClient`'s result download used a plain `client.get()`, which buffers the full response in memory regardless of size, and only a per-read timeout with no overall deadline. A misbehaving endpoint could hold an unbounded response in memory, or trickle bytes to keep the download open indefinitely.

Added two helpers to `lightrag/parser/external/_common.py`:

- `stream_capped_get()`: streams the response, checking byte total per chunk, rejecting oversized responses mid-download.
- `download_deadline_seconds()`: reads `PARSER_RESULT_BUNDLE_DOWNLOAD_TIMEOUT` (default 300s, 0 disables it), enforced via `asyncio.timeout()` so it doesn't mask external cancellation.

Both clients now share these helpers.

## Related Issues

Commit `f33fe54` addressed ZIP entry-count and declared-size limits. This completes #3610 by adding the remaining download-body and wall-clock limits.

Closes #3610

## Changes Made

- Added `stream_capped_get()` and `download_deadline_seconds()` to `_common.py`.
- Routed both clients' downloads through them, replacing `client.get()`.
- `safe_extract_zip`'s declared-size check still runs afterward, unchanged.
- Consume bytes directly from the streaming helper's return value.
- Omitted the raw URL from error messages (MinerU's `full_zip_url` can carry a signed token; may end up in `doc_status.error_msg`/logs).
- Added 10 offline, mocked-`httpx` regression tests (5 per client): deadline enforcement + disabled-gate control, oversized-response rejection, cancellation-vs-deadline distinction, declared-size-lie defence-in-depth.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

### Test results

- New regression tests fail on unpatched `main`, pass after the fix. Control tests (disabled deadline, ZIP validation) pass both before and after.
- `tests/parser/external/`: 198 passed
- All 10 new tests collected under the offline marker.

## Scope Notes

- Deadline defaults to 300s, disabled by a non-positive value (matches existing zip-bomb guard convention).
- Raw URL intentionally excluded from error messages, by design.
- `safe_extract_zip`'s checks (from `f33fe54`) are unchanged; this only bounds the raw body and wall-clock time before that check runs.


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
Pushed `39f0ef528` (maintainer edit): the new download-bounds tests mutated `os.environ` directly with a try/finally-pop pattern, which *deletes* rather than *restores* any pre-existing value of the same variable in the developer's shell after the run, and was inconsistent with the autouse fixtures in the same files already using `monkeypatch`. Switched every env manipulation to `monkeypatch.setenv`, which restores the prior state on teardown. No change to what the tests assert — `tests/parser/external/` still 220 passed locally, pre-commit clean.

danielaskdd:
Thx for sharing.

PR #3646, opened 2026-08-14
fix(api): generalize EMBEDDING_DIM startup guard to all embedding bindings

### LINKED ISSUE #3644: Generalize the custom-model EMBEDDING_DIM guard from Ollama (#3636) to all embedding bindings
## Summary

PR #3636 (fixing #3596) added a startup-time check that refuses to boot when a custom (non-default) **Ollama** embedding model is configured without an explicit `EMBEDDING_DIM`. Without it, `create_optimized_embedding_function` silently falls back to the *default* model's `embedding_dim`, no matter which model the user actually configured, and the mismatch only surfaces later as a vector-store write failure. That fix is scoped to `binding == "ollama"` only.

## Problem

The same silent-fallback bug exists for every other binding that goes through this logic. In `create_optimized_embedding_function` (`lightrag/api/lightrag_server.py`):

```python
final_embedding_dim = (
    args.embedding_dim if args.embedding_dim else provider_embedding_dim
)
```

`provider_embedding_dim` always comes from the provider's `@wrap_embedding_func_with_attrs` decorator default, regardless of which `model` the user actually configured. The function's own docstring already documents the intended contract:

> When EMBEDDING_MODEL is set to a custom model: User MUST also set EMBEDDING_DIM to match the custom model's dimension

...but nothing enforces it outside the Ollama branch added in #3636. For example:

- `EMBEDDING_BINDING=jina`, `EMBEDDING_MODEL=jina-embeddings-v3` (1024 dims) with no `EMBEDDING_DIM` silently inherits the `jina-embeddings-v4` decorator default of 2048 — same failure class as #3596, just deferred to the vector store instead of caught at startup.
- Same risk for OpenAI (`text-embedding-3-small` default 1536 vs. e.g. `text-embedding-3-large`), Gemini, Bedrock, and VoyageAI whenever a non-default model is selected without `EMBEDDING_DIM`.

## Request

Generalize the check added in #3636 (`lightrag_server.py`, the `binding == "ollama"` block around `create_optimized_embedding_function`) so every applicable binding raises the same fail-fast `ValueError` at startup when a non-default `model` is configured **and** `EMBEDDING_DIM` is not explicitly set.

## Complications (this is not a drop-in copy of the Ollama check)

Each provider's `@wrap_embedding_func_with_attrs(model_name=...)` needs to actually represent a meaningful "default model" for a generic comparison to work, and that's not uniformly true today:

- `voyageai_embed` (`lightrag/llm/voyageai.py`) declares no `model_name` in its decorator at all (`EmbeddingFunc.model_name` defaults to `None`), even though the function itself has a real runtime default (`model: str = "voyage-3"`, 1024 dims). The decorator needs `model_name="voyage-3"` added before this generalizes cleanly.
- `azure_openai_embed`'s decorator uses `model_name="my-text-embedding-3-large-deployment"` — that's a docstring/example placeholder, not an actual default (Azure deployment names are always user-defined; there is no universal default to compare against). Azure likely needs its own rule instead: always require `EMBEDDING_DIM` whenever a deployment/model is explicitly configured, since there's no safe fallback dimension.
- `lollms_embed` ignores the `model` parameter entirely (the server uses its own configured vectorizer) — out of scope, there's no per-model dimension concept for it.
- The existing Ollama check normalizes away the `:latest` tag suffix (bare `bge-m3` is treated the same as `bge-m3:latest`, since that's Ollama's own implicit-tag convention). Other bindings don't have this convention, so the comparison logic isn't literally copy-pasteable as one generic string-equality helper — it should be a small per-binding table of (default model name, comparison rule) rather than one shared function.

## Suggested scope for a fix

- **openai, jina, gemini, bedrock**: same pattern as Ollama (`model provided + non-default + no EMBEDDING_DIM => raise`), once each check uses the binding's real default model name.
- **voyageai**: add the missing `model_name="voyage-3"` to the decorator first, then apply the same rule.
- **azure_openai**: needs its own rule (no valid "default" to compare against) — consider requiring `EMBEDDING_DIM` whenever a deployment/model is explicitly configured.
- **lollms**: no change needed.

Reference: #3596, #3636

ClaireXi99:
## Summary

Closes #3644.

PR #3636 added a fail-fast check for Ollama: when a custom embedding model is configured without `EMBEDDING_DIM`, the server raises immediately instead of silently using the decorator's default dimension (which causes vector-store write failures at runtime).

This PR generalizes that guard to **all** embedding bindings that go through `create_optimized_embedding_function`:

| Binding | Default model | Default dim |
|---------|--------------|-------------|
| OpenAI | text-embedding-3-small | 1536 |
| Jina | jina-embeddings-v4 | 2048 |
| Gemini | gemini-embedding-001 | 1536 |
| Bedrock | amazon.titan-embed-text-v2:0 | 1024 |
| VoyageAI | voyage-3 | 1024 |
| Ollama | bge-m3:latest | 1024 |
| Azure OpenAI | (deployment names — always requires explicit dim) | — |

**lollms** is intentionally excluded (it ignores the model parameter).

## Changes

1. **`lightrag/api/lightrag_server.py`** — Replace the Ollama-only guard with a unified check across all bindings listed in `_BINDINGS_WITH_DIM_GUARD`. Azure OpenAI gets a separate unconditional check since deployment names never match a universal default.

2. **`lightrag/llm/voyageai.py`** — Add missing `model_name="voyage-3"` to the `@wrap_embedding_func_with_attrs` decorator so the guard can compare against it (all other providers already had this).

3. **`lightrag/api/config.py`** — Reject `EMBEDDING_DIM <= 0` at parse time. The guard tested `args.embedding_dim is None`, but the effective dimension is resolved with a truthiness test (`args.embedding_dim if args.embedding_dim else provider_default`), so `EMBEDDING_DIM=0` satisfied the guard and then silently fell back to the provider's decorator dimension — exactly the failure mode the guard exists to prevent. Both guard sites now test `not args.embedding_dim`, so the guard and the dimension resolution share one notion of "set".

4. **`tests/api/config/test_embedding_dimension_guard.py`** (new) — 25 tests covering OpenAI, Jina, Gemini, Bedrock, VoyageAI, Azure OpenAI, lollms, and the non-positive `EMBEDDING_DIM` cases.

## Breaking changes

The guard compares **model names**, while the hazard it protects against is a **dimension** mismatch. That is deliberate — the server cannot know a custom model's dimension — but it means configurations that work today will now abort at startup until `EMBEDDING_DIM` is set explicitly. Affected setups:

1. **Any non-default model name on `openai` / `jina` / `gemini` / `bedrock` / `voyageai` without `EMBEDDING_DIM`.** `openai` is the generic OpenAI-compatible binding, so custom model names are the norm there (local endpoints, `text-embedding-ada-002`, third-party gateways) — including 1536-d models whose dimension happened to match the provider default.

2. **`EMBEDDING_SEND_DIM=true` with a Matryoshka model** (e.g. `text-embedding-3-large` truncated to 1536-d): the guard does not consult `args.embedding_send_dim`, so these configurations abort too. `jina` and `gemini` force `send_dimensions=True` unconditionally and fall in the same class.

3. **Azure OpenAI configured only via `AZURE_EMBEDDING_DEPLOYMENT`** (the documented alternative to `EMBEDDING_MODEL` in `env.example`) now aborts without `EMBEDDING_DIM`, even when the deployment is a 1536-d model whose default dimension is already correct.

4. **`EMBEDDING_DIM=0` / negative values** are now rejected at startup instead of silently falling back to the provider default.

**Migration in all cases: set `EMBEDDING_DIM` to the model's actual dimension.** The error message names the offending model/deployment. This is intentional — a silently wrong dimension corrupts the vector store at write time, which is far harder to diagnose than a startup failure.

Relatedly, the Azure error message now resolves the deployment name in the same order `azure_openai_embed` does (`AZURE_EMBEDDING_DEPLOYMENT` wins over the configured model), so it names the deployment that is actually used.

## Note on `voyageai_embed`'s new `model_name`

Adding `model_name` to the decorator changes `BaseVectorStorage._generate_collection_suffix()` from `None` to `voyage_3_1024d` for library users who pass `voyageai_embed` directly (Milvus / Qdrant / PostgreSQL only — Nano / Faiss / Mongo / OpenSearch do not use the suffix). All three backends have legacy→suffixed migration paths, so this is a one-time automatic migration, not data loss.

## Test plan

- [x] `pytest tests/api/config/test_embedding_dimension_guard.py tests/api/config/test_ollama_embedding_dimension.py` — 30 passed
- [x] `pytest tests/api` — 851 passed, 15 skipped (no regressions)
- [x] New tests verified to fail against the pre-fix code (behavioral fix-proof)
- [x] Verified all guarded providers have `model_name` in their decorator matching their function default


danielaskdd:
@codex review

ClaireXi99:
Thanks for the review! Addressed the Codex P1 feedback:

**Fix:** The Azure guard now resolves the effective deployment from `AZURE_EMBEDDING_DEPLOYMENT` env var when `EMBEDDING_MODEL` is unset. If either is configured, `EMBEDDING_DIM` is required.

**Changes:**
- `lightrag_server.py`: Compute `azure_effective_model = model or os.environ.get("AZURE_EMBEDDING_DEPLOYMENT")` and use it in the guard condition
- `test_embedding_dimension_guard.py`: Added regression test `test_env_deployment_without_model_requires_explicit_dim` — patches env var with `AZURE_EMBEDDING_DEPLOYMENT=text-embedding-3-large`, `model=None`, verifies ValueError is raised

All 24 tests pass.

danielaskdd:
@codex review

danielaskdd:
Thanks for generalizing this guard — the direction is right, and the per-binding test matrix made it easy to review. I pushed one commit (d5c2b16) to this branch and expanded the PR description; details below so nothing lands as a surprise.

### What I pushed

**1. `EMBEDDING_DIM=0` bypassed the guard.** The guard tests `args.embedding_dim is None`, but the effective dimension is resolved with a truthiness test a few lines later:

```python
final_embedding_dim = args.embedding_dim if args.embedding_dim else provider_embedding_dim
```

So `EMBEDDING_DIM=0` satisfied the guard and then silently fell back to the provider's decorator dimension — exactly the failure mode the guard exists to prevent. A negative value propagated into the vector stores unchecked. Fixed at two levels:

- `config.py` now rejects `EMBEDDING_DIM <= 0` at parse time (`SystemExit`, matching the other env-validation failures in `parse_args`). This also covers the case the guard can never see: a *default* model with `EMBEDDING_DIM=0`, where no custom-model check fires at all.
- Both guard sites now test `not args.embedding_dim`, so the guard and the dimension resolution share one notion of "set".

**2. The Azure error message named the wrong deployment.** `azure_openai_embed` resolves `os.getenv("AZURE_EMBEDDING_DEPLOYMENT") or model` (lightrag/llm/openai.py:1268), i.e. the env var *wins* over the configured model. The guard computed `model or AZURE_EMBEDDING_DEPLOYMENT`, so when both were set the message named a deployment that is not the one actually used. Swapped the operands (the trigger condition is unchanged — only the reported name) and corrected the trailing note, which said the opposite.

Regression tests added for all of the above; they were verified to fail against the pre-fix code. `tests/api` is green (851 passed, 15 skipped).

### What I documented rather than changed

The guard compares **model names**, while the hazard is a **dimension** mismatch. That is unavoidable — the server cannot know a custom model's dimension — but it means some configurations that work today will abort at startup until `EMBEDDING_DIM` is set:

- Any non-default model name on `openai`/`jina`/`gemini`/`bedrock`/`voyageai`. `openai` is the generic OpenAI-compatible binding, so custom model names are the norm there (local endpoints, gateways, `text-embedding-ada-002`), including 1536-d models whose dimension already matched.
- `EMBEDDING_SEND_DIM=true` with a Matryoshka model (e.g. `text-embedding-3-large` truncated to 1536-d) — the guard does not consult `args.embedding_send_dim`. `jina` and `gemini` force `send_dimensions=True` unconditionally and fall in the same class.
- Azure setups configured only via `AZURE_EMBEDDING_DEPLOYMENT` (the documented alternative in `env.example`).

I think this trade is correct — a silently wrong dimension corrupts the vector store at write time, which is much harder to diagnose than a startup failure — so I wrote these up as a **Breaking changes** section in the PR description with the migration (set `EMBEDDING_DIM`) rather than trying to special-case them. If you disagree on any of the three, say so and we can narrow the guard instead.

Also noted in the description: adding `model_name` to `voyageai_embed` changes `BaseVectorStorage._generate_collection_suffix()` from `None` to `voyage_3_1024d` for library users passing `voyageai_embed` directly (Milvus/Qdrant/PostgreSQL). All three backends have legacy→suffixed migration paths, so it is a one-time automatic migration, not data loss — worth being explicit about for anyone reading the changelog.

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3654, opened 2026-08-15
fix(lmdeploy): preserve do_sample setting

Shizoqua:
## Description

The LMDeploy binding reads `do_sample` from the caller but then unconditionally replaces it with `True` on supported LMDeploy versions.

As a result, omitting the parameter or explicitly passing `do_sample=False` still enables sampling, despite the documented default being greedy decoding.

This change preserves the caller's value when constructing LMDeploy's `GenerationConfig`.

## Related Issues

No related issue.

## Changes Made

- Removed the unconditional assignment that forced `do_sample=True`.
- Added offline regression tests for the default, explicit `False` and explicit `True` cases.
- Added control coverage confirming the existing pre-0.6 version guard remains unchanged.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [ ] Documentation updated, not necessary
- [x] Unit tests added

## Test Results

- New regression tests fail on unmodified `main` and pass after the fix.
- Python 3.10: 6 passed.
- Python 3.11: 6 passed.
- Python 3.12: 6 passed.
- All six tests are collected under the offline marker.
- `pre-commit run --all-files` passed.
- `git diff --check` passed.

## Additional Notes

The tests use a mocked LMDeploy pipeline and do not require a model, GPU, network connection or external service.

The existing behaviour for LMDeploy versions earlier than 0.6 remains unchanged.

### DIFF
--- a/lightrag/llm/lmdeploy.py
+++ b/lightrag/llm/lmdeploy.py
@@ -142,7 +142,6 @@ async def lmdeploy_model_if_cache(
             f"v0.6.0, but currently using lmdeloy {lmdeploy.__version__}"
         )
     else:
-        do_sample = True
         gen_params.update(do_sample=do_sample)
 
     lmdeploy_pipe = initialize_lmdeploy_pipeline(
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3656, opened 2026-08-16
fix(parser): prevent quadratic Markdown image parsing

danielaskdd:
## Description

Replace the native Markdown image regex substitution with a linear scanner so malformed image prefixes cannot repeatedly rescan a long line and block document ingestion.

## Related Issues

GHSA-8wpg-h5wf-jww4

## Changes Made

- Parse supported inline image syntax with shared forward delimiter searches instead of global regex replacement.
- Preserve angle-bracket and bare-source parsing behavior, including optional titles and nested candidate handling.
- Add grammar compatibility and malformed-input performance regression coverage.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

The regression test covers both the reported unclosed-source input and an unclosed-alt input with a trailing closing parenthesis, which would remain quadratic under the advisory's proposed shortcut.


danielaskdd:
@codex review

Issue #3603, opened 2026-08-08
Per-chunk text "breadcrumb" / metadata hook — `chunking_func` is silently bypassed on `/documents/scan` & `/documents/upload`


geraldtscharre:
## Summary

There is no working way — through the REST API — to prepend per-chunk metadata (e.g. source filename, path, document as-of date) to the chunk text that the extraction LLM sees. The one documented customization point for per-chunk text, `chunking_func`, is silently bypassed on the `/documents/scan` and `/documents/upload` paths, with no error and no log line. This makes per-chunk provenance/temporal grounding impossible for API-driven ingestion.

## The problem (silent bypass)

`chunking_func` is described in `lightrag/lightrag.py` as a "Legacy chunking-function customization point," and its docstring is explicit:

> "If `process_options` explicitly contains a chunking selector char (`F`/`R`/`V`/`P`), the dispatcher routes to a chunker that follows the new file-chunker contract … This `chunking_func` is NOT called in that case — it is a legacy escape hatch and is intentionally bypassed when the user opted into a specific strategy."

Meanwhile, in `lightrag/api/routers/document_routes.py`, the scan/upload enqueue path does:

```python
api_process_options = process_options or PROCESS_OPTION_CHUNK_FIXED
```

So both endpoints **always** name a strategy (`F` by default) → `chunking_func` is **never** invoked. The only way to reach the no-selector path where `chunking_func` actually runs is the SDK (`ainsert`) with an empty/non-chunking `process_options` — not the HTTP API most deployments use.

Net effect: a user who supplies a `chunking_func` that, say, prepends a `SOURCE / PATH / DATE` breadcrumb to every chunk gets **0 of their chunks augmented**, silently. The chunks are created, entities extracted, the graph builds — and the customization simply never ran.

## Why this matters (use case)

The extraction LLM sees only chunk *text*, never chunk *metadata* (`file_path`, headings, `created_at`). For multi-document corpora this loses information that matters at **extraction time**, not query time:

- **Temporal grounding:** a document's as-of date (often only present in the filename or a header line) lands in chunk 1 and is lost on chunks 2..N. Entities extracted from later chunks get no date grounding, so the graph can't self-document *when* a fact was attested.
- **Source provenance:** which document a chunk came from — useful for any multi-doc corpus, essential when the same entity is mentioned across many documents with different roles/statuses.
[...]

PR #3662, opened 2026-08-17
fix: warn when selectors bypass custom chunking

er-s-an:
## Description

Warn when an explicit F/R/V/P selector bypasses a user-supplied `chunking_func`, and update the field documentation to describe the implemented R and V strategies.

This implements only the immediate scope agreed in #3603. It does not add a chunker registry or a document-metadata contract.

## Related Issues

Related to #3603.

## Changes Made

- detect a non-default `chunking_func` at selector dispatch time and log the selected strategy plus document ID
- avoid warning when the built-in default is configured
- replace the stale R/V fallback text with the current recursive-character and semantic-vector implementations
- add an offline regression test for both the custom and built-in cases

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated
- [x] Unit tests added

## Validation

- `./scripts/test.sh tests/chunker/test_chunking_raw_lightrag_parity.py -q` — 10 passed
- exact new parametrized test — 2 passed
- `uv run ruff check ...` and `uv run ruff format --check ...`
- `uv run pre-commit run --files ...`

The optional spaCy language models were not installed; the test runner reported only the unrelated smart-heading skip notice.

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -573,11 +573,12 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
         char (``F``/``R``/``V``/``P``), the dispatcher routes to a
         chunker that follows the new file-chunker contract — see
         :mod:`lightrag.chunker` (``chunking_by_fixed_token`` for ``F``,
-        ``chunking_by_paragraph_semantic`` for ``P``; ``R``/``V`` are
-        not yet implemented and fall back to ``F``). **This
-        ``chunking_func`` is NOT called in that case** — it is a
-        legacy escape hatch and is intentionally bypassed when the user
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

danielaskdd:
This PR intentionally addresses only the immediate observability gap: it warns when an explicit `F/R/V/P` selector bypasses a configured custom `chunking_func`.

The broader follow-up is now tracked in [RFC #3664](https://github.com/HKUDS/LightRAG/issues/3664), covering an explicit `C` strategy, trusted Server-side registry injection, `sidecar_location`-based callback context, legacy compatibility, and the warned `C -> F` fallback when no callback is injected.

Please move further design discussion and any interest in implementing that work to #3664.

Issue #3664, opened 2026-08-18
RFC: Add C strategy for Server-injected custom chunking callbacks


danielaskdd:
## Summary

Add an explicit `C` process-options chunking strategy as a compatibility bridge for the existing instance-level `LightRAG.chunking_func` extension point.

The current Server does not inject a user-defined `chunking_func`, while API ingestion normally selects `F/R/V/P` and therefore bypasses the legacy callback. This RFC defines a safe, observable route for trusted Server-side customization without changing the semantics of `F`.

Related to #3603. The immediate observability fix is #3662.

## Goals

- Preserve the existing `chunking_func` mechanism for deployments that need it before a full named-chunker contract is available.
- Make custom chunking an explicit per-document choice: `process_options=C`.
- Let LightRAG Server resolve a trusted registered callback during startup and inject it into the constructed `LightRAG` instance.
- Give a context-aware callback access to the complete parser sidecar through `full_docs.sidecar_location`.
- Keep existing six-argument callbacks working unchanged.
- Make a missing Server injection visible rather than silently behaving as a custom chunker.

### 1. `C` selector

Extend the chunking-selector set from `F/R/V/P` to `F/R/V/P/C`.

- `F/R/V/P` retain their present meanings and dispatch behavior.
- `C` means “use the configured instance `chunking_func`.”
- `C` uses the fixed-token chunk-options snapshot for compatibility with the old callback arguments: separator, separator-only flag, overlap, and token size.
- `C` is accepted consistently by SDK enqueue, filename hints / parser rules, scan/upload routes, and the text-insert API (as a `custom` chunking strategy).
- The persisted selector remains authoritative on retry and resume.

### 2. Server registry injection

At trusted Server startup, resolve an operator-selected callback from a chunker registry and pass it as `chunking_func=` when constructing `LightRAG`.

- Registry keys are allowlisted, pre-registered names; HTTP request input never supplies Python code, module paths, or import paths.
- Registry lookup and callback construction happen only at startup, in the trusted server process.
- The selected callback should have a stable identifier/version recorded with the document's processing snapshot, so retries and post-restart processing can detect an unavailable or incompatible deployment rather than silently changing behavior.
- This RFC does not prescribe the eventual multi-chunker registry grammar; it only establishes the Server-to-`LightRAG` injection bridge. A later registry design may extend `C` with a named form such as `C(name=...)`.

### 3. Required fallback when no callback was injected

[...]

PR #3665, opened 2026-08-18
feat: add explicit custom chunking selector

er-s-an:
## Description

Implements the maintainer-approved first slice of #3664: an explicit `C` selector that invokes an injected `LightRAG.chunking_func` through the existing six-argument contract, while keeping the undecided registry and callback-identity work out of scope.

## Related Issues

- First implementation slice of #3664
- Builds on the selector-bypass warning from #3662

## Changes Made

- Accept and persist `C` in process options, filename hints, parser rules, and chunk-option snapshots, with an explicit `C -> fixed_token` mapping.
- Add `chunking.strategy="custom"` to the text API. It reuses the fixed-token parameter schema and returns 422 when no non-default callback is configured.
- Reject upload hints/rules selecting `C` before writing the file when the callback is unavailable; retain warn-and-fallback behavior for scans, persisted rows, and reprocessing.
- Invoke synchronous and asynchronous custom callbacks on the event loop. Callback errors fail the document with context and never trigger fallback.
- Record the actual path as `custom_chunking_func` or `custom_chunking_fallback_fixed_token`, suppress the #3662 bypass warning for C, and derive sidecar-backfill eligibility from the path that actually ran.
- Document filename-hint and API behavior in English and Chinese.

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Additional Notes

Validation on the rebased branch:

- `ruff check` on all touched Python files: passed
- focused C/API/pipeline contract suite: 254 passed
- related chunker/parser/API/pipeline regression suite: 880 passed

A full local test collection is not available in this checkout because optional provider/storage dependencies (for example Ollama, Neo4j, Milvus, MongoDB, Redis, PostgreSQL, Anthropic, and Bedrock clients) are not installed. The implementation intentionally does not add callback identity/version persistence, registry grammar, or a context-aware callback contract; those remain follow-up RFC work.

### DIFF
--- a/docs/FileProcessingPipeline-zh.md
+++ b/docs/FileProcessingPipeline-zh.md
@@ -119,7 +119,7 @@ LightRAG 的文件处理配置由两部分合成：内容抽取引擎决定原�
 
 ### 2.1 文件处理选项
 
-处理选项以文件为粒度控制多模态分析、知识图谱构建和文本分块的行为；既可在 `LIGHTRAG_PARSER` 中作为规则默认值批量设置（见 [§2.4](#24-默认规则lightrag_parser)），也可通过文件名 hint 对单个文件覆盖（见 [§2.5](#25-单文件覆盖文件名-hint)）。所有选项都是可选的；缺省值见下表。同一文件最多指定一种分块方式（F/R/V/P），其它选项可任意组合。
+处理选项以文件为粒度控制多模态分析、知识图谱构建和文本分块的行为；既可在 `LIGHTRAG_PARSER` 中作为规则默认值批量设置（见 [§2.4](#24-默认规则lightrag_parser)），也可通过文件名 hint 对单个文件覆盖（见 [§2.5](#25-单文件覆盖文件名-hint)）。所有选项都是可选的；缺省值见下表。同一文件最多指定一种分块方式（F/R/V/P/C），其它选项可任意组合。
 
 | 选项 | 类型 | 默认 | 含义 |
 | --- | --- | --- | --- |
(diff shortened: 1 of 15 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
Reviewed the whole diff and found no correctness bugs. Specifically checked and cleared: no double-await / unbound `chunking_result` on the `C` path; the `sidecar_backfill_eligible` refactor is behaviour-identical for F/R/V/P and the legacy path, with C-custom correctly ineligible (rewritten text can't be located in `blocks.jsonl`) and C's fallback correctly eligible with `_emit_source_span=True`; the `chunking != "C"` warning suppression does not weaken the #3662 bypass warning; `chunking_by_fixed_token` really does delegate to `chunking_by_token_size`, so "exact fixed-token fallback" holds; the upload precheck cannot diverge from the later enqueue resolution (filename hints match on basename, `LIGHTRAG_PARSER` rules on suffix only); the new 422 releases the enqueue reservation; and the `C → fixed_token` mapping is consistent across `chunk_strategy_key`, `slim_chunk_options`, `resolve_chunk_options`, `_CHUNKING_PARAMS_MODEL` and both `ParamSpec.targets` sets.

I pushed c2e71e2 for three non-blocking items rather than leaving them as review comments:

**1. `env.example` did not document the `C` selector.** Both `chunk_token_size` and `chunk_overlap_token_size` accept `C` now (`_ALL_CHUNK_SELECTORS` and the overlap spec were both extended), but the canonical config template still listed the parameter-bearing selectors as `F/R/V/P` / `F/R/P`, so an operator writing `LIGHTRAG_PARSER=pdf:legacy-C(chunk_ts=800)` had no way to discover the feature. Fixed the three lines.

**2. `docs/ProgramingWithCore.md:111`** described `addon_params["chunker"]` as "Dict with F/R/V/P chunking settings". Left the list alone — the dict genuinely has only four sub-dictionaries — and added a clause noting that `C` reuses `fixed_token`.

**3. The upload endpoint resolved the parser directives twice.** `doc_manager.is_supported_file(safe_filename)` goes through `resolve_file_parser_engine` → `resolve_file_parser_directives` → `resolve_parser_directives` with the same arguments, so the new `C` gate two lines later repeated the whole hint parse plus `LIGHTRAG_PARSER` rule scan — and, because the earlier call raises first, its `except FilenameParserHintError` branch was dead code. `is_supported_file` now takes an optional pre-resolved `directives=`, and the endpoint resolves once and reuses it for both gates. Status-code precedence is unchanged (hint 400 → unsupported-type 400 → chunking 422 → 413), and the engine value is provably identical because all three entry points share one implementation. Also documented the new 422 in the endpoint docstring's `Raises:` list, which still enumerated only 400/409/413/500.

Considered and deliberately not reported: the 422-before-409/413 ordering for duplicate/oversized uploads carrying a `C` hint (a precedence choice, no misbehaviour); the now-unreachable `chunk_method = "fixed_token_fallback"` initializer (equally unreachable before, since `sanitize_process_options` guarantees the selector is one of F/R/V/P/C); the upload-422 vs. scan-warn-and-fallback asymmetry (explicit, documented design in §2.7); and `_validate_effective_chunk_overlap`'s split-only skip now applying to `custom`.

Ran `tests/api/routes` after the change: 509 passed, 5 skipped.

PR #3667, opened 2026-08-18
fix(rerank): reject negative chunk overlap

Shizoqua:
## Description

`chunk_documents_for_rerank()` calculates the next window start as `end - overlap_tokens`. When `overlap_tokens` is negative, the next window starts after the previous window ends. This silently drops content before documents are sent to the reranker. This change rejects negative overlap values before short-document handling and before the tokenizer or character-fallback paths are selected.

## Related Issues

No related issue.

## Changes Made

- Raise `ValueError` when `overlap_tokens` is negative.
- Add regression coverage confirming that negative overlap is rejected before conditional chunking.
- Preserve the existing upper-bound clamp-and-warning behaviour.
- Leave zero and valid positive overlap behaviour unchanged.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (not necessary)
- [x] Unit tests added

## Test Results

- `tests/chunker/test_rerank_chunking.py`: 27 passed
- `tests/chunker/test_overlap_validation.py`: 7 passed
- Full `tests/chunker` run: 296 passed, with 3 pre-existing network-dependent failures reproduced identically on unmodified `main`
- `git diff --check`: passed

## Additional Notes

The validation is placed at the start of `chunk_documents_for_rerank()` so it also covers short documents that do not enter either windowing loop. The asynchronous wrapper delegates directly to this function and requires no separate implementation.

### DIFF
--- a/lightrag/rerank.py
+++ b/lightrag/rerank.py
@@ -57,6 +57,13 @@ def chunk_documents_for_rerank(
         # max_tokens=0 makes the chunk window zero-width and the loop never advances
         raise ValueError(f"max_tokens must be >= 1, got {max_tokens}")
 
+    if overlap_tokens < 0:
+        # Checked up front, before the tokenizer/fallback branching below, so a
+        # negative value is always rejected -- including for a short document
+        # that never enters either windowing loop and would otherwise let the
+        # invalid value pass through unnoticed.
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3668, opened 2026-08-18
fix(lollms): honour configured binding host

Shizoqua:
## Description

`create_llm_model_kwargs()` builds the same Ollama-shaped dictionary for both the `lollms` and `ollama` bindings, including a `host` key. Ollama accepts a `host` parameter, so this works correctly. The LoLLMs binding expects the corresponding parameter to be named `base_url`.

As a result, a configured `LLM_BINDING_HOST` for LoLLMs silently enters `**kwargs` and is never used. Requests therefore continue using the default `http://localhost:9600` endpoint regardless of the configured host.

This change moves the helper to module scope as `_create_llm_model_kwargs` and separates the binding-specific host keys. LoLLMs receives `base_url`, while Ollama continues receiving `host`. The existing `timeout`, `options` and `api_key` values remain unchanged.

## Related Issues

No related issue.

## Changes Made

- Move `create_llm_model_kwargs` out of `create_app()` to module scope as `_create_llm_model_kwargs`.
- Return `base_url` for the `lollms` binding while preserving `host` for `ollama`.
- Add tests covering custom and default LoLLMs hosts, Ollama behaviour, `timeout` and `api_key` preservation, and the final outgoing request URL.
- Leave `lightrag/llm/lollms.py` unchanged.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (not necessary for this fix)
- [x] Unit tests added

## Test Results

- `tests/api/config/test_api_config_lollms_host.py`: 6 passed
- Full `tests/api/config` run: 128 passed, with 4 pre-existing failures caused by unavailable `aioboto3` and `voyageai` packages and reproduced identically on unmodified `main`
- `git diff --check`: passed

## Additional Notes

The defect was reproduced through the real `create_app()` path. `LightRAG.__init__` was temporarily patched during testing to capture the received arguments. The unmodified code supplied `host` for LoLLMs, while the corrected code supplies `base_url`.

### DIFF
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -1307,6 +1307,41 @@ def _build_scheduling_status(pipeline_snapshot: dict, ingress_counts: dict) -> d
     }
 
 
+def _create_llm_model_kwargs(binding: str, args, llm_timeout: int) -> dict:
+    """
+    Create LLM model kwargs based on binding type.
+    Uses lazy import for binding-specific options.
+    """
+    if binding in ["lollms", "ollama"]:
+        try:
+            from lightrag.llm.binding_options import OllamaLLMOptions
+
+            options = OllamaLLMOptions.options_dict(args)
+        except ImportError as e:
+            raise Exception(f"Failed to import {binding} options: {e}")
+        if binding == "ollama":
+            # Imported lazily (the module installs the ollama package on
+            # import) and only for the binding that actually forwards
+            # think= -- lollms never reaches the ollama client.
+            from lightrag.llm.ollama import ensure_think_supported
+
+            ensure_think_supported(options, context="the base LLM binding")
+        kwargs = {
+            "timeout": llm_timeout,
+            "options": options,
+            "api_key": args.llm_binding_api_key,
+        }
+        if binding == "lollms":
+            # lollms_model_if_cache()'s parameter is named base_url, not
+            # host -- unlike ollama's AsyncClient(host=...). Passing "host"
+            # here would silently land in its **kwargs and never be read.
+            kwargs["base_url"] = args.llm_binding_host
+        else:
+            kwargs["host"] = args.llm_binding_host
+        return kwargs
+    return {}
+
+
 def create_app(args):
     # Check frontend build first and get status
     webui_assets_exist, is_frontend_outdated = check_frontend_build()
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
Thanks for the fix — the root cause diagnosis is correct and I verified it end to end: `lollms_model_if_cache()` really does declare `base_url` as a named parameter (`lightrag/llm/lollms.py:48`) and builds `f"{base_url}/lollms_generate"`, so the old `"host"` key was swallowed by `**kwargs` and never read. The role path (`role_lollms_complete`) and the embedding path already passed `base_url`, so the base binding was the only inconsistent one; all three now agree. Nothing else consumes `"host"` out of the base `llm_model_kwargs` (the role metadata `host` comes from `role_llm_configs`, and only `lightrag/llm/ollama.py` pops it — that path still receives it).

I pushed two follow-up commits to this branch.

**1. `refactor(api): pin lollms llm kwargs options to an explicit empty dict`**

`_create_llm_model_kwargs()` kept one shared branch for `lollms` and `ollama`, so the `"options"` value handed to lollms came from `OllamaLLMOptions.options_dict(args)`. That dict is empty for lollms only as a side effect of where the ollama arguments are registered (`lightrag/api/config.py:555` adds them when `LLM_BINDING=ollama`) — an implicit dependency on unrelated code for a value that should just be stated.

The branches are now split: the lollms branch builds its kwargs directly with `"options": {}` and no longer imports `binding_options` at all, while the ollama branch keeps the lazy import and the `ensure_think_supported()` check. `lollms_model_if_cache()` never reads `"options"` either way, so behavior is unchanged. The existing assertion `"options" in kwargs` was tightened to `kwargs["options"] == {}` to pin it.

**2. `test(api): isolate lollms host tests from a developer-local .env`**

The new tests pinned `sys.argv` but not the environment, so a developer's `.env` decided the outcome. On my machine `VLM_PROCESS_ENABLE=true`, and because the tests set `LLM_BINDING=lollms`, `parse_args()` hits the VLM-incompatibility guard and raises `SystemExit` (`lightrag/api/config.py:805`) — **4 of the 6 new tests failed**. The same root cause fires through a second door: any `{ROLE}_LLM_BINDING` in `.env` (e.g. `EXTRACT_LLM_BINDING=openai`) differs from the base `lollms` and `SystemExit`s on the missing role model/api key (`lightrag/api/config.py:790`). CI passes only because it runs without a `.env`, so the "128 passed" claim was machine-dependent.

The fixture now clears the relevant variables, deriving the role prefixes from the `ROLES` registry so a newly added role cannot silently reintroduce the leak. One subtlety worth flagging for future tests here: copying the plain `delenv` list from `tests/api/config/test_ollama_think_startup_validation.py` is not enough, because `lightrag.api.lightrag_server` is imported lazily *after* the fixture runs and its module-level `load_dotenv(dotenv_path=".env", override=False)` calls re-populate every deleted key straight back out of `.env`. The order is therefore load-bearing: pin `sys.argv` → import the server module (so all module-level `load_dotenv` calls fire) → *then* delete. The `config._global_args` save/restore from that sibling file is not needed, since these tests call `parse_args()` directly and never go through `create_app`.

Verification with a real `.env` present:

- `tests/api/config/test_api_config_lollms_host.py`: 6 passed (was 4 failed / 2 passed)
- same file with a deliberately hostile ambient env (`VLM_PROCESS_ENABLE=true EXTRACT_LLM_BINDING=openai`): 6 passed
- `tests/api/config`: 132 passed
- `tests/api`: 857 passed, 15 skipped
- `pre-commit` (pinned ruff format + check): clean

danielaskdd:
@codex review

Shizoqua:
Thanks @danielaskdd for verifying the fix.

Issue #249, opened 2024-11-11
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument index in method wrapper_CUDA__index_select)


Z-oo883:
I am using a huggingface demo, but with a local model.How to deal with it?please help me , thank you very much!!
**code**:
import os

from lightrag import LightRAG, QueryParam
from lightrag.llm import hf_model_complete, hf_embedding
from lightrag.utils import EmbeddingFunc
from transformers import AutoModel, AutoTokenizer

WORKING_DIR = "./dickens"

if not os.path.exists(WORKING_DIR):
    os.mkdir(WORKING_DIR)

rag = LightRAG(
    working_dir=WORKING_DIR,
    llm_model_func=hf_model_complete,
    llm_model_name="/data/Qwen2.5-14B-Instruct",
    embedding_func=EmbeddingFunc(
        embedding_dim=1024,
        max_token_size=5000,
        func=lambda texts: hf_embedding(
            texts,
            tokenizer=AutoTokenizer.from_pretrained(
                r"/data/project/raag/bge-large-zh-v1.5/", model_max_length=512
            ),
            embed_model=AutoModel.from_pretrained(
                r"/data/project/raag/bge-large-zh-v1.5/"
            ),
        ),
    ),
)

with open(r"/data/project/raag/caiwu.txt", "r", encoding="utf-8") as f:
    rag.insert(f.read())

# Perform naive search
print(
    rag.query("What are the top themes in this story?", param=QueryParam(mode="naive"))
)

# Perform local search
print(
    rag.query("What are the top themes in this story?", param=QueryParam(mode="local"))
)

# Perform global search
print(
    rag.query("What are the top themes in this story?", param=QueryParam(mode="global"))
)

# Perform hybrid search
print(
    rag.query("What are the top themes in this story?", param=QueryParam(mode="hybrid"))
)
error:
root@c15e0721d1a6:~# CUDA_VISIBLE_DEVICES=0  python /data/project/raag/light_rag.py
INFO:lightrag:Logger initialized for working directory: /data/project/raag/dickens
DEBUG:lightrag:LightRAG init with param:
  working_dir = /data/project/raag/dickens,
  chunk_token_size = 1200,
  chunk_overlap_token_size = 100,
  tiktoken_model_name = gpt-4o-mini,
  entity_extract_max_gleaning = 1,
  entity_summary_to_max_tokens = 500,
[...]

PR #3678, opened 2026-08-20
fix(hf): place generation inputs on model device

Shizoqua:
## Description

`hf_model_if_cache()` moved tokenizer outputs to a hard-coded CUDA device before generation. This caused CPU-only systems to fail immediately and produced device mismatches when `device_map="auto"` placed or offloaded the model outside CUDA.

The function already created an `inputs` dictionary using `hf_model.device`, but passed the original CUDA-pinned tokenizer output to `generate()` instead.

This change removes the hard-coded CUDA placement and passes the device-adjusted inputs to the model.

## Related Issues

Related to #249.

## Changes Made

* Remove the hard-coded `.to("cuda")` from the tokenizer output.
* Move each input tensor to `hf_model.device`.
* Pass the device-adjusted `inputs` dictionary to `generate()`.
* Preserve the existing prompt-length slicing and response-decoding behaviour.
* Add dependency-light regression tests without downloading a model or requiring real PyTorch in offline CI.

## Checklist

* [x] Changes tested locally
* [x] Code reviewed
* [x] Documentation updated (not necessary for this fix)
* [x] Unit tests added

## Test Results

* New device-placement tests: 4 passed
* Full `tests/llm/hf_impl` suite: 12 passed
* Fails-before behaviour reproduced on unmodified `main`
* CPU-only execution verified
* CUDA execution verified
* CPU-offloaded model placement verified
* Full `tests/llm` comparison showed exactly the four new tests changing from fail to pass, with all remaining environment-related failures unchanged
* `git diff --check`: passed

## Additional Notes

This change does not modify model placement, `device_map`, quantisation, embeddings or the Hugging Face pooling implementation. Existing single-GPU CUDA behaviour remains unchanged because `hf_model.device` already resolves to CUDA in that configuration.

### DIFF
--- a/lightrag/llm/hf.py
+++ b/lightrag/llm/hf.py
@@ -110,10 +110,13 @@ async def hf_model_if_cache(
 
     input_ids = hf_tokenizer(
         input_prompt, return_tensors="pt", padding=True, truncation=True
-    ).to("cuda")
+    )
+    # Move to wherever the model actually is, rather than assuming CUDA.
+    # hf_model is loaded with device_map="auto" (see initialize_hf_model),
+    # so hf_model.device already reflects accelerate's placement.
     inputs = {k: v.to(hf_model.device) for k, v in input_ids.items()}
     output = hf_model.generate(
-        **input_ids, max_new_tokens=512, num_return_sequences=1, early_stopping=True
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3687, opened 2026-08-20
fix(hf): honor generation token limits

Shizoqua:
## Description

`hf_model_if_cache()` always passed `max_new_tokens=512` to the Hugging Face model, ignoring both LightRAG’s generic `max_tokens` option and the Hugging Face-native `max_new_tokens` setting.

This meant callers could not control the generation length through either supported parameter.

This change resolves the generation limit using the following precedence:

1. Explicit `max_new_tokens`
2. Generic `max_tokens`
3. Existing default of `512`

## Related Issues

No related issue.

## Changes Made

* Honour the Hugging Face-native `max_new_tokens` setting.
* Map LightRAG’s generic `max_tokens` option to `max_new_tokens`.
* Give `max_new_tokens` precedence when both options are provided.
* Preserve the existing default of 512 tokens when neither option is supplied.
* Add regression tests covering the default, both individual options, and their precedence.

## Checklist

* [x] Changes tested locally
* [x] Code reviewed
* [x] Documentation updated (not necessary for this fix)
* [x] Unit tests added

## Test Results

* Device-placement and token-limit tests: 8 passed
* Full `tests/llm/hf_impl` suite: 16 passed
* Fails-before behaviour reproduced on unmodified code
* Default 512-token behaviour verified
* Generic `max_tokens` behaviour verified
* Native `max_new_tokens` behaviour verified
* Native-option precedence verified when both options are supplied
* `git diff --check`: passed

## Additional Notes

This change only affects generation-token limit handling in `hf_model_if_cache()`. It does not modify model placement, `device_map`, quantisation, embeddings, pooling behaviour, or response decoding.

### DIFF
--- a/lightrag/llm/hf.py
+++ b/lightrag/llm/hf.py
@@ -75,6 +75,8 @@ async def hf_model_if_cache(
     messages.extend(history_messages)
     messages.append({"role": "user", "content": prompt})
     kwargs.pop("hashing_kv", None)
+    max_tokens = kwargs.pop("max_tokens", 512)
+    max_new_tokens = kwargs.pop("max_new_tokens", max_tokens)
     input_prompt = ""
     try:
         input_prompt = hf_tokenizer.apply_chat_template(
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3689, opened 2026-08-21
fix(lmdeploy): honor native max_new_tokens

Shizoqua:
## Description

`lmdeploy_model_if_cache()` raised a duplicate-keyword error when callers supplied the native `max_new_tokens` option.

This change consumes both token-limit options before constructing `GenerationConfig`, with `max_new_tokens` taking precedence over the generic `max_tokens` alias.

## Related Issues

No related issue.

## Changes Made

* Support native `max_new_tokens`.
* Preserve `max_tokens` as the generic alias.
* Keep the existing default of 512 tokens.
* Add tests covering defaults, aliases and precedence.

## Checklist

* [x] Changes tested locally
* [x] Code reviewed
* [x] Documentation updated (not necessary)
* [x] Unit tests added

## Test Results

* LMDeploy tests: 10 passed
* Fails-before behaviour reproduced on unmodified `main`
* `git diff --check`: passed

## Additional Notes

When both options are provided, the provider-native `max_new_tokens` value takes precedence.

### DIFF
--- a/lightrag/llm/lmdeploy.py
+++ b/lightrag/llm/lmdeploy.py
@@ -128,7 +128,8 @@ async def lmdeploy_model_if_cache(
             stacklevel=2,
         )
     kwargs.pop("response_format", None)
-    max_new_tokens = kwargs.pop("max_tokens", 512)
+    max_tokens = kwargs.pop("max_tokens", 512)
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3690, opened 2026-08-21
fix(ollama): map max_tokens to num_predict

Shizoqua:
## Description

The Ollama adapter discarded the generic `max_tokens` value before generation. This change maps it to Ollama’s native `options.num_predict` setting.

## Related Issues

No related issue.

## Changes Made

- Map `max_tokens` to `options.num_predict`.
- Preserve an explicitly configured native `num_predict` value.
- Avoid mutating caller-owned options.
- Add regression tests for mapping and precedence behaviour.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (not required)
- [x] Unit tests added

## Test Results

- New regression tests: 7 passed
- Full Ollama test suite: 27 passed
- Fails-before behaviour reproduced on unmodified `main`
- `git diff --check`: passed

## Additional Notes

Existing native Ollama configuration remains unchanged because `num_predict` takes precedence over the generic alias.

### DIFF
--- a/lightrag/llm/ollama.py
+++ b/lightrag/llm/ollama.py
@@ -1,4 +1,4 @@
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, Mapping
 import os
 import re
 import warnings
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3696, opened 2026-08-21
fix(gemini): map max_tokens to max_output_tokens

### LINKED ISSUE #3695: Gemini binding silently drops the generic max_tokens kwarg
## Description

On the Gemini binding, LightRAG's provider-agnostic `max_tokens` knob (set via `llm_model_kwargs`, the same mechanism documented for Ollama's `num_ctx` in `docs/ProgramingWithCore.md`) is silently discarded instead of capping the generated output.

`gemini_complete_if_cache` in `lightrag/llm/gemini.py` ends its signature with a `**_: Any` catch-all, so a caller-supplied `max_tokens` value never reaches the function body at all. No warning or error is raised; the request is simply sent without any output-length limit.

This defeats the safety mechanism described in `docs/LightRAG-API-Server.md` ("Set the max_tokens to prevent excessively long or endless output loop... particularly crucial for locally deployed, smaller-parameter models") for anyone using Gemini through the generic `llm_model_kwargs` knob, which is exactly the pattern PRs #3690 (ollama), #3689 (lmdeploy) and #3687 (hf) already fixed for those bindings.

## Reproduction

```python
import asyncio
from unittest.mock import MagicMock
import lightrag.llm.gemini as gemini_mod

captured = {}
class FakeAio:
    class models:
        @staticmethod
        async def generate_content(**kwargs):
            captured.update(kwargs)
            resp = MagicMock()
            part = MagicMock(text="hello", thought=False)
            candidate = MagicMock()
            candidate.content.parts = [part]
            resp.candidates = [candidate]
            resp.usage_metadata = MagicMock(candidates_token_count=1, thoughts_token_count=0)
            return resp
class FakeClient:
    def __init__(self, *a, **k):
        self.aio = FakeAio()
gemini_mod._get_gemini_client = lambda *a, **k: FakeClient()

asyncio.run(gemini_mod.gemini_complete_if_cache(
    model="gemini-2.0-flash", prompt="hi", api_key="fake", max_tokens=7,
))
print("config sent to generate_content:", captured.get("config"))
```

Only the network boundary (`google.genai.Client`) is faked; the real kwarg-handling and `_build_generation_config` code runs. Output on current `main` (`300d9df1`):

```
config sent to generate_content: None
```

No `config` object is built at all, so `max_tokens=7` never reaches the Gemini API in any form.

## Expected

The generic `max_tokens` kwarg should map to Gemini's native `max_output_tokens` generation-config field, the same precedence pattern (native value wins if already set) already established by the ollama/lmdeploy/hf fixes.

AmirF194:
## Description

`gemini_complete_if_cache` (`lightrag/llm/gemini.py`) ends its signature with a `**_: Any` catch-all, so LightRAG's provider-agnostic `max_tokens` kwarg (forwarded via `llm_model_kwargs`, the same mechanism already fixed for ollama/lmdeploy/hf in #3690/#3689/#3687) never reaches the function body. No config object is built and no output-length limit is ever sent to the API.

Fixes #3695

## Changes Made

- Added an explicit `max_tokens` parameter to `gemini_complete_if_cache`.
- Before building the generation config, map `max_tokens` into the native `max_output_tokens` field, only when `generation_config` does not already set it (mirrors the "native value wins" precedence rule the ollama/lmdeploy/hf fixes already established, so `GEMINI_LLM_MAX_OUTPUT_TOKENS`/explicit `generation_config` callers are unaffected).

## Verification

- New regression tests in `tests/llm/gemini_impl/test_gemini_llm.py` drive the real `gemini_complete_if_cache`, mocking only the `google.genai` network boundary: generic `max_tokens` maps to `max_output_tokens`, an explicit `generation_config["max_output_tokens"]` takes precedence, other `generation_config` keys are preserved, and no config is built when neither is set. 2 of the 4 parametrized cases fail on unmodified `main` and pass on this branch (confirmed both ways via `git stash` in the same container).
- Full offline suite (`pytest tests/ -m offline`) in a clean `python:3.12-slim` and `python:3.14-slim` container (matching the CI matrix): 5749 passed, 2 pre-existing failures unrelated to this change (reproduce identically on unmodified `main`, `tests/pipeline/test_chunk_options_regex_scrub.py` and `tests/pipeline/test_strict_capability_gate.py`).
- `pre-commit run --files lightrag/llm/gemini.py tests/llm/gemini_impl/test_gemini_llm.py` (ruff format, ruff, trailing-whitespace, end-of-file-fixer): clean.
- Not verified: behavior against a live Gemini API. The fix and its tests only exercise the local kwarg-to-config mapping; the network boundary is mocked throughout.

### DIFF
--- a/lightrag/llm/gemini.py
+++ b/lightrag/llm/gemini.py
@@ -301,6 +301,7 @@ async def gemini_complete_if_cache(
     generation_config: dict[str, Any] | None = None,
     timeout: int | None = None,
     image_inputs: list[Any] | None = None,
+    max_tokens: int | None = None,
     **_: Any,
 ) -> str | AsyncIterator[str]:
     """
@@ -346,6 +347,9 @@ async def gemini_complete_if_cache(
         hashing_kv: Storage interface (for interface parity with other bindings).
         enable_cot: Whether to include Chain of Thought content in the response.
         timeout: Request timeout in seconds (will be converted to milliseconds for Gemini API).
+        max_tokens: Generic output-length cap (as forwarded via LightRAG's
+            provider-agnostic ``llm_model_kwargs``). Mapped to Gemini's native
+            ``max_output_tokens`` unless ``generation_config`` already sets it.
         **_: Additional keyword arguments (ignored).
 
     Returns:
@@ -390,6 +394,14 @@ async def gemini_complete_if_cache(
         prompt_sections.append(history_block)
     prompt_sections.append(f"[user] {prompt}")
     combined_prompt = "\n".join(prompt_sections)
+
+    if max_tokens is not None and (
+        generation_config is None or generation_config.get("max_output_tokens") is None
+    ):
+        generation_config = {
+            **(generation_config or {}),
+            "max_output_tokens": max_tokens,
+        }
 
     config_obj = _build_generation_config(
         generation_config,
--- a/tests/llm/gemini_impl/test_gemini_llm.py
+++ b/tests/llm/gemini_impl/test_gemini_llm.py
@@ -464,3 +464,81 @@ async def test_gemini_counts_usage_before_the_empty_truncated_raise(
         "the exhausted request's tokens vanished from usage accounting "
         "because the raise preceded token_tracker.add_usage"
     )
+
+
+def _make_capturing_client(fake_response):
+    captured: dict = {}
+
+    async def _fake_generate_content(**kwargs):
+        captured.update(kwargs)
+        return fake_response
+
+    client = SimpleNamespace(
+        aio=SimpleNamespace(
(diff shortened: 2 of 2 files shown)


AmirF194:
Heads up: #3697 landed about 20 minutes after this PR and fixes the same gap (Gemini's `max_tokens` kwarg silently dropped instead of mapping to `max_output_tokens`). Same file, same approach, functionally equivalent diffs.

This PR was opened first and its test suite covers the same cases, so I will leave it open, but happy to stand down in favor of #3697 if that is easier to review.

danielaskdd:
@codex review

dianatofficial (commented):
Nice improvement. Refactoring this module enhances readability and code maintainability.

PR #3697, opened 2026-08-21
Map the generic max_tokens kwarg to Gemini's max_output_tokens

### LINKED ISSUE #3695: Gemini binding silently drops the generic max_tokens kwarg

yzxcj797:
Fixes #3695.

## Root cause

`gemini_complete_if_cache` ended its signature with a `**_: Any` catch-all, so LightRAG's provider-agnostic `max_tokens` knob — the same `llm_model_kwargs` mechanism documented for Ollama's `num_ctx`, and the output-loop safety cap the API-server docs recommend — never reached the function body. No warning, no error: the request was simply sent without any output-length limit.

This is the Gemini member of the family #3687 (hf) / #3689 (lmdeploy) / #3690 (ollama) already fixed.

## Fix

- `max_tokens` is accepted as an explicit kwarg (documented) instead of vanishing into `**_`;
- it maps to Gemini's native `max_output_tokens` generation-config field, with an explicitly configured native value taking precedence — the identical contract the sibling bindings implement;
- the caller's `generation_config` dict is copied before the merge, never mutated.

## Tests

`tests/llm/gemini_impl/test_gemini_max_tokens.py` (offline, fake `google.genai` namespace per the existing harness):

- generic limit maps to `max_output_tokens=37`;
- other generation-config fields (temperature) are preserved;
- a native value survives alone and takes precedence over the alias;
- the caller's dict is not mutated;
- nothing is injected when neither form is configured.

Differential: 3 of the 6 tests fail on `main` (alias dropped / native-only shapes pass). `tests/llm/gemini_impl/` is 19/19 green with the change.

### DIFF
--- a/lightrag/llm/gemini.py
+++ b/lightrag/llm/gemini.py
@@ -301,6 +301,7 @@ async def gemini_complete_if_cache(
     generation_config: dict[str, Any] | None = None,
     timeout: int | None = None,
     image_inputs: list[Any] | None = None,
+    max_tokens: int | None = None,
     **_: Any,
 ) -> str | AsyncIterator[str]:
     """
@@ -346,6 +347,9 @@ async def gemini_complete_if_cache(
         hashing_kv: Storage interface (for interface parity with other bindings).
         enable_cot: Whether to include Chain of Thought content in the response.
         timeout: Request timeout in seconds (will be converted to milliseconds for Gemini API).
+        max_tokens: Optional generic output-length cap, mapped to Gemini's
+            native ``max_output_tokens`` generation-config field when the
+            caller has not set a native value themselves.
         **_: Additional keyword arguments (ignored).
 
     Returns:
@@ -390,6 +394,15 @@ async def gemini_complete_if_cache(
         prompt_sections.append(history_block)
     prompt_sections.append(f"[user] {prompt}")
     combined_prompt = "\n".join(prompt_sections)
+
+    # Generic output-length knob: map LightRAG's provider-agnostic max_tokens
+    # to Gemini's native config field, with an explicitly configured native
+    # max_output_tokens taking precedence (#3695, same contract as the
+    # ollama/lmdeploy/hf bindings).
+    if max_tokens is not None:
+        generation_config = dict(generation_config or {})
+        if generation_config.get("max_output_tokens") is None:
+            generation_config["max_output_tokens"] = max_tokens
 
     config_obj = _build_generation_config(
         generation_config,
--- a/tests/llm/gemini_impl/test_gemini_max_tokens.py
+++ b/tests/llm/gemini_impl/test_gemini_max_tokens.py
@@ -1,0 +1,213 @@
+"""Offline tests for Gemini's generic output-token alias (#3695).
+
+LightRAG's provider-agnostic ``max_tokens`` knob used to vanish into the
+``**_`` catch-all of ``gemini_complete_if_cache`` — the request was sent with
+no output-length limit at all. The binding now maps it to Gemini's native
+``max_output_tokens`` generation-config field, with an explicitly configured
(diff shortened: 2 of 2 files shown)


danielaskdd:
@codex review

dianatofficial (commented):
The parameter mapping from `max_tokens` to `max_output_tokens` for Gemini is accurate and prevents unhandled kwarg errors. Approved.

dianatofficial (commented):
Approved. The fix handles null/edge cases gracefully without regressions.

danielaskdd:
Thanks for this — the diagnosis and the fix are both correct.

We went with #3696, which was opened ~20 minutes earlier and lands a functionally equivalent change to `gemini_complete_if_cache`. Verified equivalence by running this PR's test file against #3696's source and vice versa: both suites pass against either implementation, so the choice came down to the test layout rather than the fix itself. #3696 appends to the existing `tests/llm/gemini_impl/test_gemini_llm.py` and reuses its `_load_gemini_module` harness, whereas this PR copies ~120 lines of that harness into a new file. That harness is the fragile part of the Gemini tests (it has to restore both the `sys.modules` entry and the parent-package attribute, or it leaks the stubbed `genai.types` namespace into unrelated tests), so we would rather not maintain two copies of it.

Your `test_caller_generation_config_is_not_mutated` case was the one genuinely additive piece here, and it is being ported onto the existing harness with attribution to this PR — so it is not lost.

Sorry for the duplicated effort. Really appreciate the contribution.

---

PR #3699, opened 2026-08-21
fix: keep Nano/Faiss pending ops as a replay log until save is durable

### LINKED ISSUE #3688: [Bug]: index_done_callback's unconditional reload silently drops applied-but-unsaved ops (Nano/Faiss)
## Summary

In both file-based vector backends (`NanoVectorDBStorage`, `FaissVectorDBStorage`) the flush helper clears its pending buffer **before** the save, while `index_done_callback` reloads the on-disk snapshot **unconditionally** — and the reload *replaces* the in-memory store rather than merging into it. So any operation that was applied to the in-memory store but whose save then failed has **no replay record left**, and the next reload silently discards it.

This was surfaced by the automated review on #3680 for the *delete* half of the Nano backend (that PR is fixing it there, see the review discussion). This issue tracks the **general** case, which is pre-existing and independent of #3680:

- **Nano upserts** — pre-existing on `main`.
- **Faiss upserts and deletes** — pre-existing on `main`; the delete half is also covered by #3681.

## The chain

Nano (`lightrag/kg/nano_vector_db_impl.py`):

1. `_flush_pending_locked` materializes the pending upserts into `self._client`, sets `_client_dirty = True`, and **clears `_pending_upserts`** (`del self._pending_upserts[doc_id]` at the end of the helper).
2. `_save_to_disk_locked()` raises (IO error, full disk, permissions). The exception propagates; `_client_dirty` stays `True`; the pending buffer is already empty.
3. Another process commits → our `storage_updated.value` is flipped to `True`.
4. The next `index_done_callback` starts with an **unconditional** `_reload_client_from_disk_locked(for_write=True)`, which does `self._client = NanoVectorDB(...)` — a wholesale replacement with the on-disk snapshot. The materialized-but-unsaved rows are gone, nothing replays them, and the following save succeeds. **Silent data loss.**

Faiss (`lightrag/kg/faiss_impl.py`) is structurally identical: `_flush_pending_locked` clears `_pending_upserts` at its end (`faiss_impl.py:775`-ish), `index_done_callback`'s reload at `faiss_impl.py:951` is unconditional, and `_reload_index_from_disk_locked` re-inits `self._index` and replaces `self._id_to_meta`. On Faiss this covers deletes too, because `delete` there is still eager (#3681).

## Why both retry paths currently lose something

Once you are in the "applied but not saved" state and another writer has committed, **neither** retry path is lossless today:

- `index_done_callback` reloads unconditionally → **our** applied-but-unsaved ops are lost (the chain above).
- `finalize` guards the same reload with `if not self._client_dirty` / `if not self._index_dirty`, so it skips the reload — but then its save writes our stale snapshot over the top, so **the other writer's committed rows** are lost instead.

This is exactly the trade-off #3680's "Considered and rejected" paragraph describes, and it is the reason the naive fix — adding the `_client_dirty` guard to `index_done_callback`'s reload as well — is **not** acceptable: it just moves the loss from one party to the other.

## Proposed direction

Treat the pending buffer as a **replay log**, not a work queue: clear it only after the save is durable, and re-apply it after a reload.

- Move the buffer clear out of `_flush_pending_locked` to **after** `_save_to_disk_locked()` / `_save_faiss_index()` succeeds, in both `index_done_callback` and `finalize`.
- A reload followed by a replay is then lossless in both directions: the reload picks up the other writer's committed rows, and the replay re-applies ours on top.
- **Replay is cheap and safe for both op kinds.** Deletes are idempotent (a replayed id that is no longer present simply matches nothing). Upserts replay **without re-embedding**, because `_PendingNanoDoc.vector` / `_PendingFaissDoc.vector` is already populated by the flush that consumed it — this is exactly the invariant the deferred-embedding protocol (#3145) already documents ("once `vector` is set it is an already-normalized float32 ndarray"; the flush honours it and does not re-normalize). For the same id, our replayed upsert wins over the reloaded row, which is the correct outcome.

### Details that need an explicit decision

1. **Dirty-flag bookkeeping.** A flush whose ops matched nothing has nothing to persist, so the buffer can be cleared immediately in that case (`deleted_count == 0` on the delete side); only matched-but-unsaved work needs retention. Keep `_client_dirty` / `_index_dirty` set strictly from "did the materialized store actually change", as it is today.
2. **Interaction with `_discard_pending_index_ops`.** The pipeline abort path deliberately drops pending buffers (`lightrag.py:2558`), because a failed ingestion is regenerable by reprocessing. Retention does not conflict with that: a retained buffer dropped there is re-issued by the retry. And deletion paths already avoid the discard on purpose — see the `_insert_done_with_cleanup` docstring at `lightrag.py:3437` ("deletion paths must use plain `_insert_done` so failed deletes stay buffered for a later retry"), which is the same intent this change generalizes.
3. **Read-your-writes stays consistent.** A retained-but-applied buffer means the read paths would report a pending upsert from the buffer *and* find the same row in the store — both describe the same content, so the existing "pending wins" precedence is already correct. Same for a retained delete: the id is genuinely absent from the store.
4. **Bounded retention.** The buffer is retained only across a *failed* save, so it does not grow unbounded in the healthy path. A permanently failing save keeps it resident — same exposure the `_client_dirty` retry already has today.

## Tests

Both backends, in `tests/kg/nano_impl/` and `tests/kg/faiss_impl/`:

- **Fix-proof**: apply ops → make the save raise → have a second writer commit → run `index_done_callback` again → assert the ops are present in the on-disk snapshot **and** the other writer's rows survived. Must fail on the current behaviour (it loses one side or the other), not merely on a missing symbol.
- Replay does not re-embed (spy on the embedding func: zero calls on the retry).
- A flush that matched nothing clears the buffer immediately (no perpetual retention).
- `_discard_pending_index_ops` still drops a retained buffer.

## References

- #3680 — the Nano delete half, where the automated review surfaced this window
- #3681 — the Faiss deferred-delete port, which will inherit this window
- #3145 — the deferred-embedding protocol whose cached-vector invariant makes upsert replay free

*Line references are against `main` at 877873a0b.*

YashvantHange:
## Description

`index_done_callback` in the two file-based vector backends materializes the pending ops, drops the buffer, and *then* reloads the on-disk snapshot unconditionally. If the save between those steps fails and another writer commits, the reload replaces the in-memory store and there is no replay record left, so the applied-but-unsaved rows are silently lost. `finalize` has the mirror-image bug: it skips the reload while dirty, so its save overwrites the other writer's committed rows instead.

The fix treats the pending buffer as a **replay log** rather than a work queue: it is cleared only once a save is durable, and it is re-applied after a reload. A reload-then-replay is then lossless in both directions — the reload picks up the other writer's rows, the replay puts ours back on top.

## Related Issues

Fixes #3688

## Changes Made

- `nano_vector_db_impl.py` / `faiss_impl.py`: keep `_pending_upserts` after a flush; clear them in `_commit_locked` only after the save succeeds.
- Retain matched deletes in `_pending_deletes` and replay them on a foreign-commit reload.
- Do not retain deletes that matched nothing — there is nothing to persist, so the buffer clears immediately and cannot grow unbounded.
- `finalize` now reloads and replays instead of skipping the reload while dirty, which closes the "overwrite the other writer" half of the trade-off.
- Replay never re-embeds: `_PendingNanoDoc.vector` / `_PendingFaissDoc.vector` is already populated by the flush that consumed it, per the deferred-embedding invariant from #3145.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (if necessary) — no user-facing behavior change to document

## Additional Notes

Tests cover the four cases called out in the issue, on both backends:

| Case from #3688 | Test |
|---|---|
| Fix-proof: failed save + foreign commit, both sides survive | `test_index_done_replays_unsaved_upsert_after_foreign_commit`, `test_index_done_replays_unsaved_delete_after_foreign_commit` |
| Replay does not re-embed (zero extra embedding calls) | asserted inside the upsert test via a counting embedder |
| A flush that matched nothing clears the buffer immediately | `test_delete_that_matches_nothing_is_not_retained` |
| `drop_pending_index_ops` still drops a retained buffer | `test_drop_pending_does_not_rollback_materialized` |

Verified locally:

- `pytest tests/kg/nano_impl tests/kg/faiss_impl -m offline` — 54 passed
- `pre-commit run --files <the four changed files>` — all hooks pass

### DIFF
--- a/lightrag/kg/faiss_impl.py
+++ b/lightrag/kg/faiss_impl.py
@@ -576,19 +576,22 @@ class FaissVectorDBStorage(BaseVectorStorage):
                 if meta.get("src_id") == entity_name
                 or meta.get("tgt_id") == entity_name
             ]
-        if relations:
-            deleted_custom_ids = []
-            for fid in relations:
-                meta = self._id_to_meta.get(fid)
-                if meta is not None and "__id__" in meta:
-                    deleted_custom_ids.append(meta["__id__"])
-            self._remove_faiss_ids_locked(relations)
-            self._pending_deletes.update(deleted_custom_ids)
-            self._index_dirty = True
+            if relations:
+                deleted_custom_ids = []
+                for fid in relations:
+                    meta = self._id_to_meta.get(fid)
+                    if meta is not None and "__id__" in meta:
+                        deleted_custom_ids.append(meta["__id__"])
+                self._remove_faiss_ids_locked(relations)
+                self._pending_deletes.update(deleted_custom_ids)
+                self._index_dirty = True
 
-            # Materialized rebuild succeeded — safe to prune matching
+            # Materialized side is settled — safe to prune matching
             # buffered upserts (their records carry src_id / tgt_id from
-            # the relationships vdb meta_fields).
+            # the relationships vdb meta_fields). Unconditional: an entity
+            # whose relations exist only in the pending buffer has nothing
+            # materialized to rebuild but still needs those buffered
+            # upserts cancelled.
             pending_ids = [
                 doc_id
                 for doc_id, pdoc in self._pending_upserts.items()
--- a/lightrag/kg/faiss_impl.py
+++ b/lightrag/kg/faiss_impl.py
@@ -182,10 +182,12 @@ class FaissVectorDBStorage(BaseVectorStorage):
         unflushed pending data is intentionally not queryable.
 
         A flush failure (embedding error, count mismatch, or save IO
-        error) raises through ``index_done_callback``; the pending buffer
-        is preserved on flush failure, and if only the save failed
-        ``_index_dirty`` stays ``True`` so a subsequent ``finalize``
-        retries the save without re-embedding.
+        error) raises through ``index_done_callback``. The pending buffer
+        is a **replay log**: it is not cleared until the save is durable.
+        If only the save failed, ``_index_dirty`` stays ``True`` and the
+        buffered docs keep their already-normalized vectors so a later
+        commit can reload a foreign snapshot and re-apply our ops without
+        re-embedding (issue #3688).
 
     Non-pipeline write paths:
         The pipeline ``busy`` gate serializes ``upsert`` / ``delete`` /
@@ -241,15 +243,13 @@ class FaissVectorDBStorage(BaseVectorStorage):
         # Maps <int faiss_id> → metadata (including your original ID).
         self._id_to_meta = {}
 
-        # Minimal pending area for deferred embedding: custom-id -> _PendingFaissDoc.
-        # Holds only records not yet embedded+materialized into self._index;
-        # it never duplicates rows already added to the Faiss index. Flushed
-        # under _storage_lock by _flush_pending_locked().
+        # Replay log for deferred embedding: custom-id -> _PendingFaissDoc.
+        # Cleared only after a durable save (issue #3688).
         self._pending_upserts: dict[str, _PendingFaissDoc] = {}
+        # Custom ids whose materialized delete has not been saved yet.
+        self._pending_deletes: set[str] = set()
         # True when self._index / self._id_to_meta have materialized changes
-        # that have not been successfully saved to disk yet. This lets
-        # finalize retry a save even after a previous flush cleared the
-        # pending buffer (see _flush_pending_locked / index_done_callback).
+        # that have not been successfully saved to disk yet.
         self._index_dirty = False
 
         # Sweep orphan tmp siblings left behind by hard kills mid-save.
@@ -303,6 +303,7 @@ class FaissVectorDBStorage(BaseVectorStorage):
         self._id_to_meta = {}
         self._load_faiss_index()
         self.storage_updated.value = False
+        self._replay_pending_deletes_locked()
         return True
 
     async def _get_index(self):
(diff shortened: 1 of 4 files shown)


dianatofficial (commented):
Good catch on keeping Nano/Faiss pending operations as a replay log until save durability is confirmed. Great reliability improvement.

dianatofficial (commented):
Good optimization! Deferring heavy computation and cleaning up resources helps performance.

danielaskdd:
@codex review

danielaskdd:
Closing this PR — not because the analysis was wrong, but because `main` has moved underneath it in a way that makes a direct merge both impossible and undesirable:

1. **The delete half of #3688 is already fixed on `main`** by the deferred-delete series (#3680, #3693, and follow-ups through #3705), using a *fingerprint-scoped* redo log (`_unsaved_deletes`): a replay removes only the exact row version it deleted, so a new row another writer published under the same id survives. This PR's coarser design — retaining the whole `_pending_deletes` id set and replaying by id — would regress that precision: replaying a bare id can destroy a foreign writer's newer row that happens to reuse the id.

2. **The branch conflicts wholesale with current `main`** (`mergeable: CONFLICTING`; a trial merge produces 12+ conflict hunks in `nano_vector_db_impl.py` alone, spanning the whole file). The base predates the deferred-delete restructuring, so the two designs cannot be reconciled hunk-by-hunk.

3. **What remains open is the upsert half only**: materialized-but-unsaved upserts still have no replay record, exactly as the Nano docstring on `main` documents ("Upserts have no equivalent log … see issue #3688"). Per the earlier discussion (see #3703's closing note as well), the fix needs to be redone as a focused PR on current `main` that adds an upsert-side redo log mirroring `_unsaved_deletes`, rather than reworking this branch.

A follow-up PR scoped to the upsert side is being opened against current `main`; #3688 stays open (with a scope-narrowing note) until it lands.

Thank you for the thorough failure-model analysis and the test matrix — both carried directly into the rework.

Issue #3657, opened 2026-08-16
[Feature Request]: Add an `openai_responses` LLM binding (OpenAI Responses API)


danielaskdd:
### Do you need to file a feature request?

- [x] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

### Feature Request Description

Add a dedicated `openai_responses` LLM binding that talks to OpenAI's Responses API (`/v1/responses`), alongside the existing `openai` binding (Chat Completions, `/v1/chat/completions`).

This is a **forward-looking** proposal, not an outage fix. See the fact-check below before prioritizing it.

#### Fact-check: Chat Completions is NOT deprecated

This request originated from a report that "starting with gpt-5.6, OpenAI no longer supports the completions endpoint". That premise does **not** hold up:

- OpenAI's own migration guide still states *"While Chat Completions remains supported, Responses is recommended for all new projects"*, and its examples use `gpt-5.6` against **both** APIs.
- The real restriction is **parameter-combination scoped**, not endpoint scoped: since GPT-5.4 there are restrictions around combining tool calling with reasoning settings in Chat Completions. Reports in the OpenAI developer community show errors of the form `Function tools with reasoning_effort are not supported for gpt-5.6-sol`. (The exact wording/semantics differ between the community reports and the migration guide, so this issue deliberately does not pin down the precise rule.)
- On the Azure side, "the portal only offers Responses" is a misreading — Microsoft's official answer is that the same deployment is reachable through both `/openai/responses` and `/openai/deployments/{id}/chat/completions`; 404s there are usually an `api-version`/endpoint-format problem.

**LightRAG does not trigger that restriction.** The LLM call path never sends function tools — `openai_complete_if_cache` only sends `messages` plus an optional dict `response_format` (`lightrag/llm/openai.py`). The `"tools"` entry in `lightrag/llm/bedrock.py` is just a passthrough allow-list key for the Converse API, not an actual tool-calling flow.

So there is **no service-continuity risk** today, and no need to rush this.

#### Why it is still worth doing

- **Cost/performance on reasoning models**: OpenAI reports substantially better prompt-cache utilization on Responses vs Chat Completions, plus better reasoning-model quality when reasoning items are preserved across turns.
- **Encrypted reasoning**: lets a stateless deployment keep reasoning context without `store: true`.
[...]

PR #3701, opened 2026-08-21
feat(llm): add openai_responses binding for the OpenAI Responses API

YashvantHange:
## Description

Adds `openai_responses` as an additional LLM binding talking to the OpenAI Responses API (`/v1/responses`), alongside the existing `openai` binding rather than replacing its transport.

This is deliberately **additive**. The `openai` binding doubles as the generic OpenAI-*compatible* binding, and much of that ecosystem (vLLM, gateways, third-party providers) does not implement `/v1/responses` — changing its transport would regress those deployments. The new transport is opted into explicitly with `LLM_BINDING=openai_responses`, and `lightrag/llm/openai.py` is not modified at all.

## Related Issues

Refs #3657 (approach and open questions confirmed in a comment on the issue before implementing)

## Changes Made

- **`lightrag/llm/openai_responses.py`** (new): the driver, matching the `openai_complete_if_cache` contract. Client construction is imported from the Chat Completions module so the two bindings cannot drift on base-URL resolution, default headers, or the DashScope workspace header.
- **`lightrag/llm/binding_options.py`**: `OpenAIResponsesLLMOptions` under its own `openai_responses_llm` namespace, deriving `OPENAI_RESPONSES_LLM_*` and the role-scoped forms.
- **`lightrag/api/config.py`**: `get_default_host()` mapping, `--llm-binding` choice, options selection branch.
- **`lightrag/api/lightrag_server.py`**: `LLMConfigCache` init, startup allow-list, dispatch branch, role-binding branch, and the role-level completion function.
- **`env.example`**: a dedicated section documenting when to prefer this binding.

Four behaviours are reimplemented rather than reused, because the wire shapes differ:

| Concern | Chat Completions | Responses |
|---|---|---|
| System prompt | `messages[0]` role=system | `instructions` parameter |
| Output budget | `max_completion_tokens` | `max_output_tokens` |
| Reasoning | flat `reasoning_effort` | nested `reasoning: {effort}` |
| Structured output | `response_format` | `text: {format}` |
| Truncation | `finish_reason == "length"` | `status` + `incomplete_details.reason` |
| Usage | `prompt_tokens` / `completion_tokens` | `input_tokens` / `output_tokens` |
| Streaming | `choices[].delta` | typed events |

### The silent-fallthrough guard

`create_llm_model_func()` ends in a bare `else:  # openai and compatible`, so a binding missing from any one exact in-list check produces **no error** — it just sends requests to `/v1/chat/completions`. The binding is added to every one of those checks, and `test_the_responses_binding_does_not_reach_the_chat_completions_fallthrough` asserts the dispatch branch precedes the fallthrough so a future regression fails loudly.

Note there are two such fallthroughs; the earlier one is the *embedding* dispatch, which this binding never reaches because it is not an embedding choice. Embeddings are unchanged — `/v1/embeddings` is not affected by this API.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (`env.example`)
- [x] Unit tests added

## Additional Notes

**Scope** — LLM role only, per the open questions in the issue:

- **Embeddings**: out of scope, `/v1/embeddings` is unchanged.
- **VLM role**: out of scope. Image input is `input_image` on Responses, a different content-part shape; the driver handles it, but wiring the VLM role is better as a follow-up once the text path is settled.
- **Azure Responses**: out of scope, and suggested to stay a separate binding rather than a flag — mirroring how `azure_openai` is its own binding rather than a switch on `openai`.

**No behaviour change to existing paths.** The diff is 1344 insertions against 2 deletions, and both deleted lines are documentation strings that gained the new binding name — no existing logic was removed. Verified by running `tests/llm` + `tests/api` before and after the change: identical failure set, `1098 → 1129` passing, i.e. exactly the 31 new tests and nothing else moved.

**Verified locally** (per AGENTS.md, the directories mirroring the changed modules):

- `pytest tests/llm/openai_responses_impl tests/api/config/test_api_config_openai_responses.py -m offline` — 31 passed
- `pytest tests/llm tests/api -m offline` — 1129 passed, with the same 9 pre-existing failures as on an unmodified checkout (Windows-only: they shell out to `bash`)
- `pre-commit run --files <the eight changed files>` — all hooks pass

### DIFF
--- a/env.example
+++ b/env.example
@@ -939,7 +939,7 @@ MAX_PARALLEL_INSERT=3
 
 ###########################################################################
 ### Gloabal LLM Configuration
-###   LLM_BINDING type: openai, ollama, lollms, azure_openai, bedrock, gemini
+###   LLM_BINDING type: openai, openai_responses, ollama, lollms, azure_openai, bedrock, gemini
 ###   LLM_BINDING_HOST: Service endpoint (left empty if using the provider SDK default endpoint)
 ###   LLM_BINDING_API_KEY: api key
 ### If LightRAG deployed in Docker:
@@ -1070,6 +1070,23 @@ VLM_MIN_IMAGE_PIXEL=64
 
 ### Role-specific + Provider-sepecific LLM options
 # VLM_OPENAI_LLM_MAX_TOKENS=20000
+
+### OpenAI Responses API Specific Parameters (/v1/responses):
+###     LLM_BINDING=openai_responses
+###     LLM_BINDING_HOST=https://api.openai.com/v1
+###     LLM_MODEL=gpt-5.6
+### A sibling of the `openai` binding, not a replacement. Chat Completions
+### remains supported; Responses is worth choosing for better prompt-cache
+### utilisation on reasoning models, or for a gateway that exposes only
+### /v1/responses. Many OpenAI-compatible servers do not implement it, so the
+### `openai` binding stays the right default for those.
+### Options use their own OPENAI_RESPONSES_LLM_* namespace: the Responses API
+### spells these differently and does not accept the Chat Completions forms.
+# OPENAI_RESPONSES_LLM_MAX_OUTPUT_TOKENS=9000
+# OPENAI_RESPONSES_LLM_REASONING_EFFORT=minimal
+# OPENAI_RESPONSES_LLM_TEMPERATURE=0.9
+### Whether OpenAI retains the response server-side
+# OPENAI_RESPONSES_LLM_STORE=false
 
 ### Azure OpenAI Specific Parameters:
 ###     LLM_BINDING=azure_openai
--- a/lightrag/api/config.py
+++ b/lightrag/api/config.py
@@ -16,6 +16,7 @@ from lightrag.llm.binding_options import (
     OllamaEmbeddingOptions,
     OllamaLLMOptions,
     OpenAILLMOptions,
+    OpenAIResponsesLLMOptions,
 )
 from lightrag.base import OllamaServerInfos
 import sys
@@ -81,6 +82,7 @@ def get_default_host(binding_type: str) -> str:
         "lollms": os.getenv("LLM_BINDING_HOST", "http://localhost:9600"),
         "azure_openai": os.getenv("AZURE_OPENAI_ENDPOINT", "https://api.openai.com/v1"),
         "openai": os.getenv("LLM_BINDING_HOST", "https://api.openai.com/v1"),
+        "openai_responses": os.getenv("LLM_BINDING_HOST", "https://api.openai.com/v1"),
         # Let boto3 select the regional Bedrock endpoint unless the user
         # explicitly overrides LLM_BINDING_HOST / EMBEDDING_BINDING_HOST.
         "bedrock": os.getenv("LLM_BINDING_HOST", "DEFAULT_BEDROCK_ENDPOINT"),
@@ -503,6 +505,7 @@ def parse_args() -> argparse.Namespace:
             "ollama",
             "openai",
             "openai-ollama",
+            "openai_responses",
             "azure_openai",
             "bedrock",
             "gemini",
@@ -556,6 +559,8 @@ def parse_args() -> argparse.Namespace:
         OllamaLLMOptions.add_args(parser)
     elif llm_binding_value in ["openai", "azure_openai"]:
         OpenAILLMOptions.add_args(parser)
+    elif llm_binding_value == "openai_responses":
+        OpenAIResponsesLLMOptions.add_args(parser)
     elif llm_binding_value == "gemini":
         GeminiLLMOptions.add_args(parser)
     elif llm_binding_value == "bedrock":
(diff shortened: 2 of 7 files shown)


dianatofficial (commented):
LGTM! Checked the diff for the `openai_responses` binding. Clean implementation and proper integration with existing response models.

dianatofficial (commented):
LGTM! Code changes are clean, well-tested, and maintain backwards compatibility.

danielaskdd:
@codex review

YashvantHange:
@danielaskdd please do re-review

danielaskdd:
@codex review

YashvantHange:
@danielaskdd please do re-review

YashvantHange:
Addressed the `response.failed` finding in 2807af87.

The report is right. `_stream_response` recognised `response.completed`, `response.incomplete`, and the separate `error` event; `response.failed` matched none of them and fell through every branch, so a generation that failed after the request was accepted ended the iterator *successfully* on whatever partial text had already been yielded, discarding both the provider's reason and the terminal usage. It now accounts the usage the attempt spent and raises `InvalidResponseError` carrying the error code and message — the same type the `error` event already raises, so its place in the retry set is unchanged.

**The same terminal exists on the non-streaming path**, which the report did not cover. `responses.create()` does not raise on a failed generation; it returns a payload with `status: "failed"` and the reason in `error`. That was landing in the empty-content branch, which reported "model produced no output" in place of the provider's actual reason, and returned the content as a successful short answer when the failed payload carried any. `_handle_non_streaming` now checks the status before the content walk.

Streaming usage accounting moved into a small helper so the failure terminal and the normal end cannot drift on when the attempt is billed — this module states that ordering rule ("account before any validation raises") in two places, and a second hand-written copy is how it stops being true.

Six tests cover it: the failure terminal's message, its usage, and its interaction with an already-open `<think>` tag, plus the three non-streaming cases. Five fail against the pre-fix driver; the sixth pins the accounting order and is a guard rather than a demonstration.

danielaskdd:
@codex review

YashvantHange:
All three findings addressed in b118c8f1, each on the non-streaming path too.

**json schema is flattened for `text.format`.** The report is right: `{"type": "json_schema", "json_schema": {...}}` was forwarded unchanged, and Responses wants `name` / `schema` / `strict` directly on the format object, so every schema-constrained request 400d. `_normalize_text_format` unwraps it, leaves `json_object` alone, and forwards an already-flat format untouched for a caller who wrote the Responses spelling themselves. This is the same normalization `_normalize_gemini_response_schema` does next door, kept provider-local for the same reason.

**Non-budget `response.incomplete` terminals raise.** `content_filter` and its siblings are stopped generations, not truncation, and they were falling through to the same end-of-stream path as `response.completed` — empty finished successfully, partial was presented as a complete answer. They now raise with the reason. Budget exhaustion keeps its partial output, because there the model stopped inside a limit the caller set; that distinction is now the stated rule rather than a side effect of `_is_truncated`. A terminal the driver cannot read reports `unknown` and takes the same raise path: incomplete is not complete, and guessing the reason would be worse.

The same terminal exists non-streaming — `status: "incomplete"` with a non-budget reason and partial content was returned as a clean answer — and is checked before the content walk.

**Refusals surface as errors.** The event and its text are tracked apart. An empty `response.refusal.delta` is still a refusal, and letting it stand in for the text would both suppress the raise and disarm the `response.refusal.done` fallback that carries the actual reason; a refusal with no text anywhere still raises rather than reporting the turn as empty. If a server populates the terminal payload with `refusal` content parts and emits no refusal events, `_extract_refusal` picks them up — the same fallback shape `_extract_text` already uses for servers whose SDK objects lack `output_text`. Non-streaming, those content parts were being skipped by the body walk and diagnosed as "model produced no output", with the provider's reason discarded.

One deliberate choice worth flagging: refusals and stopped generations raise `InvalidResponseError`, which is in the retry set. A refusal is largely deterministic for a given prompt, so an argument exists for the non-retryable type. I kept it retryable because this driver cannot tell a policy verdict from a transient safety-stack failure, and because `error` and `response.failed` are already retryable here; budget exhaustion is the one non-retryable case precisely because its outcome is fixed by the caller's own `max_output_tokens`. Happy to flip it if you would rather refusals fail on the first attempt.

Twelve tests cover the three defects and their non-streaming siblings; nine fail against the pre-fix driver, and the rest pin the boundaries — the flat schema passing through unchanged, `.done` not duplicating the deltas it repeats, and budget truncation still returning its partial content.

danielaskdd:
@codex review

YashvantHange:
Confirmed and fixed in bb226647.

Reproduced it directly first: a stream that emits only a `response.completed` terminal — no `response.output_text.delta`, no reasoning delta, nothing — runs to completion with an empty chunk list and no exception. `_handle_non_streaming` raises `InvalidResponseError` for the exact same payload (see `test_empty_output_without_truncation_raises_retryable`), so the asymmetry is real: the non-streaming path already treats "completed but nothing usable" as an error, the streaming path didn't.

The gap was structural: the post-loop validation checks refusal, then non-budget incomplete, then truncated-and-empty — but a clean `response.completed` satisfies none of those, so it fell off the end of the function and the generator just stopped. Added the missing case: after the truncated branch, if neither body text nor reasoning text was ever seen, raise `InvalidResponseError` — same type and same "no usable output" framing `_handle_non_streaming` uses.

Two new tests: the empty-completion case (fails against the pre-fix driver, confirmed by reverting locally and re-running), and a reasoning-only completion to pin that the check keys on body **or** reasoning text — a `<think>`-only turn is still usable output and must not raise.

YashvantHange:
@danielaskdd Fixed the empty-completion finding in bb226647 (details above). Could you re-review when you have a moment?

danielaskdd:
@codex review

PR #3702, opened 2026-08-22
test(utils): cover validate_file_path_security and subtract_source_ids

YashvantHange:
## Description

Adds direct tests for two `lightrag/utils.py` helpers that guard behaviour with real consequences and had no test referencing them.

**Tests only — no production code is modified.** The diff is 262 insertions across two new files and zero deletions.

## Related Issues

None. `CONTRIBUTING.md` lists "Testing — add test coverage for untested code paths" as a way to contribute.

### `validate_file_path_security`

The containment boundary for two callers that turn externally-influenced strings into filesystem paths: the `/documents` upload and scan routes (`document_routes.py`), and sidecar artifact resolution (`pipeline.py`, whose docstring delegates containment to it outright). A hole here is a read outside the input directory.

Covered:

- Accepted: plain and nested relative paths, `./` and interior `/./` normalization, backslash-to-separator normalization, surrounding whitespace.
- Rejected: parent traversal in both separator styles, a bare `..`, absolute paths, blank and `None` input.
- Two properties that are easy to break while *looking* like hardening:
  - `a/../b.pdf` resolves back inside the base and must stay **accepted** — containment is decided on the resolved path, not on the presence of a `..` segment. Replacing this with a blanket `".." in path` rejection would break ordinary inputs.
  - A sibling directory sharing a name prefix (`/base_evil` against base `/base`) is **not** contained. A prefix comparison on the path string would wrongly accept it; `Path.is_relative_to` compares whole components.
- The documented "does not check existence" contract: a contained but absent path is returned rather than conflated with unsafe.

### `subtract_source_ids`

Decides what attribution a graph object retains after a purge removes chunks — `_purge_kg_contributions` computes `remaining_sources` from it in both the entity and relation branch, and an empty result **deletes** the object while a non-empty one keeps it.

Covered:

- Falsy ids are dropped on **both** branches. The implementation returns early when the removal set is empty, so the two paths filter independently; an empty id leaking through either would keep an object alive with no real attribution — the unattributable orphan the purge contract exists to prevent.
- Surviving order is preserved and is not influenced by the removal collection's iteration order, because the result is written back as the object's chunk attribution and `apply_source_ids_limit` truncates it positionally.
- Duplicates are deliberately *not* collapsed (`merge_source_ids` is the normalization point; collapsing here would change the count the limit sees).
- Input types: generator source, set/tuple removal collections, and that the source iterable is not mutated.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [ ] Documentation updated (if necessary) — not applicable, tests only
- [x] Unit tests added

## Additional Notes

**The tests were verified to have teeth.** Rather than assume, each file was run against a deliberately mutated `lightrag/utils.py` before committing:

| Mutation | Result |
|---|---|
| `is_relative_to` containment check disabled | 7 tests fail |
| falsy-id filter removed from `subtract_source_ids` | 1 test fails |

`lightrag/utils.py` was restored afterwards and is untouched in this PR.

**Portability.** Assertions compare `Path` objects and use `relative_to` rather than string matching, since separators and resolution differ between POSIX and Windows. Inputs whose meaning is platform-specific — trailing-dot names such as `"..."`, which Windows strips and POSIX keeps as a literal directory name — are deliberately not asserted on.

**Verified locally** (per AGENTS.md, the directory mirroring the module under test):

- `pytest tests/utils -m offline` — 196 passed (156 before, +40 from these two files)
- `pre-commit run --files <the two new files>` — all hooks pass

**One observation, no change made.** `validate_file_path_security` reads `file_path_str.strip()` before entering its `try`, so a non-string argument raises `AttributeError` rather than returning `None` like every other invalid input. No current caller passes a non-string, so this is not reachable today and I have not changed it — flagging it in case you would prefer that guard moved inside the `try`.

### DIFF
--- a/tests/utils/test_subtract_source_ids.py
+++ b/tests/utils/test_subtract_source_ids.py
@@ -1,0 +1,107 @@
+"""Tests for ``subtract_source_ids`` removal semantics.
+
+The helper decides what a graph object still has attribution from after a purge
+removes chunks: ``lightrag/lightrag.py`` computes ``remaining_sources`` from it
+in both the entity and the relation branch of ``_purge_kg_contributions``, and
+an object whose remaining sources are empty is *deleted* while a non-empty
+result keeps it. Two properties are therefore load-bearing rather than cosmetic:
+
+* **Falsy ids are dropped.** A stray ``""`` surviving into the result would make
+  ``remaining_sources`` truthy and strand an object that owns no real chunk —
+  precisely the unattributable orphan the purge contract exists to prevent.
+* **Order is preserved.** The surviving list is written back as the object's
+  chunk attribution, and ``apply_source_ids_limit`` truncates it positionally,
+  so reordering would silently change which ids survive a later cap.
+
+It had no direct tests; this file pins both, alongside the sibling
+``test_merge_source_ids.py`` for the collection half of the same contract.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from lightrag.utils import subtract_source_ids
+
+pytestmark = pytest.mark.offline
+
+
+class TestRemoval:
+    def test_listed_ids_are_removed(self):
+        assert subtract_source_ids(["a", "b", "c"], ["b"]) == ["a", "c"]
+
+    def test_removing_every_id_yields_an_empty_list(self):
+        # The caller reads this as "no attribution left" and deletes the object.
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3703, opened 2026-08-22
fix(faiss): keep pending upserts as the replay record until the save succeeds

yzxcj797:
## What?

The Faiss half of #3688: `_flush_pending_locked` cleared the pending upsert buffer **before** the save ran, so a failed save combined with another process's commit silently destroyed the materialized-but-unsaved rows — the empty pending buffer left nothing to replay them from. (The Nano half of the issue is being fixed in #3680; this PR covers the Faiss backend, whose structure is identical.)

Fixes the Faiss half of #3688

## The chain (pre-existing on main)

1. `index_done_callback` flushes: rows materialize into `self._index`/`_id_to_meta`, `_index_dirty = True`, **pending cleared**;
2. `_save_faiss_index` raises (IO error / full disk) — the exception propagates, pending stays empty, dirty stays True;
3. another process commits → `storage_updated` flips;
4. the next `index_done_callback`'s reload-from-disk **replaces** the in-memory index with the foreign snapshot — the materialized rows are gone, nothing replays them, the following save succeeds. Silent data loss.

The callback's own docstring already promised "a later `finalize` retries the save without re-embedding", but that self-healing only covered the no-other-writer case; any foreign commit between the failed save and the retry destroyed the rows instead.

## How?

The pending buffer now survives until the save **succeeds** — it doubles as the replay record:

- `_flush_pending_locked` no longer clears the entries it flushed (a NOTE explains why);
- both callers (`index_done_callback`, `finalize`) snapshot the flushed ids and drain them only **after** `_save_faiss_index` returns;
- a reload replaces `self._index`/`_id_to_meta` but never touches `_pending_upserts`, so re-running the flush against the reloaded snapshot rebuilds exactly those rows: `_remove_faiss_ids_locked` by custom-id + re-add is idempotent, and the cached vectors mean no re-embedding.

Read paths are unaffected: they run under the same non-reentrant `_storage_lock`, so the brief window where a row is both materialized and pending is invisible.

## Testing

New `tests/kg/faiss_impl/test_faiss_save_failure_replay.py`:

- a failed save keeps the buffer; a **foreign reload** (`storage_updated` flipped) followed by a retry persists the row — the exact #3688 chain;
- a successful save drains the buffer;
- `finalize` replays after a save failure and persists the row.

The three deferred-embedding tests that pinned the old contract ("pending is empty after a failed save") are updated to the replay contract.

**Differential**: on `main` the new replay test fails with `pending must survive a failed save — it is the replay record (#3688)`. With this change the full `tests/kg/faiss_impl` suite passes: 32/32.

Related: my open #3693 (deferred deletes) touches the delete paths of the same file; this PR is based on `main` and independent of it.

### DIFF
--- a/lightrag/kg/faiss_impl.py
+++ b/lightrag/kg/faiss_impl.py
@@ -773,16 +773,15 @@ class FaissVectorDBStorage(BaseVectorStorage):
 
         self._index_dirty = True
 
-        # Clear only the entries we just flushed. Today the non-reentrant
-        # _storage_lock locks out concurrent upserts for the entire flush
-        # (including the asyncio.gather await), so the `is pdoc` identity
-        # check is always True — it's kept as defensive scaffolding so that
-        # if the lock scope is ever relaxed (e.g. embedding moved outside the
-        # lock), a concurrent upsert that re-set vector=None would not be
-        # silently dropped here.
-        for doc_id, pdoc in pending_items:
-            if self._pending_upserts.get(doc_id) is pdoc:
-                del self._pending_upserts[doc_id]
+        # NOTE: the flushed entries are intentionally NOT removed from
+        # ``_pending_upserts`` here — the CALLER clears them once the
+        # subsequent save succeeds. Until then the pending buffer is the
+        # replay record for the materialized-but-unsaved rows: if the save
+        # raises and another process commits, the next callback's
+        # reload-from-disk replaces ``self._index`` / ``self._id_to_meta``,
+        # and re-running this flush against the reloaded snapshot rebuilds
+        # exactly these rows (remove-by-custom-id + re-add is idempotent).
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

danielaskdd:
Closing in favor of #3699. Both PRs address the same #3688 save-failure replay problem and overlap in the Faiss pending-upsert path, so they are alternative implementations rather than complementary changes. #3699 covers the complete failure model across both Faiss and Nano, including upserts, applied-but-not-yet-durable deletes, foreign-writer reloads, and dirty finalize, and its author has now pushed fixes for the review feedback. Keeping both PRs open would duplicate and conflict with the same Faiss changes. Thank you for the focused fix and regression coverage.

PR #3715, opened 2026-08-24
fix(openai): use safe token truncation for embedding input

### LINKED ISSUE #3714: [Bug]: openai_embed's local truncation bypasses the safe Tokenizer contract
## Description

`openai_embed`'s built-in truncation slices the raw tiktoken token list and decodes it back, bypassing `lightrag.utils.Tokenizer` entirely:

```python
tokens = encoding.encode(text)
if len(tokens) > max_token_size:
    truncated_tokens = tokens[:max_token_size]
    truncated_texts.append(encoding.decode(truncated_tokens))
```

This is the site #3565 flagged under "Known remaining paths (not migrated this round)":

> **`openai_embed`**'s local truncation (`lightrag/llm/openai.py`): `tokens[:max_token_size]` + `decode()`, bypasses `lightrag.utils.Tokenizer` entirely, no re-verification. This **is** a live embedding call path and should be the next follow-up […]

Two distinct wrong behaviours:

**(a) A literal special-token string fails the whole batch.** `Encoding.encode` defaults to `disallowed_special=ALL`, so any text merely *containing* the literal `<|endoftext|>` raises `ValueError`. Once the truncation block is entered the `encode` runs for **every** non-empty text, so length is irrelevant. This is exactly the failure `Tokenizer.encode` exists to prevent — its own comment reads *"This crashes document indexing on any user content that happens to contain those strings"* — and `openai_embed` never goes through it.

**(b) Truncation can cut inside a code point.** `decode(tokens[:n])` decodes with `errors="replace"`, so a budget landing mid-character yields a string ending in U+FFFD. What gets embedded is then **not a prefix of the input**.

Neither result is re-encoded to confirm it actually fits, which the contract added by #3565 requires because BPE token count is not monotonic in text length.

### Which deployments this reaches

The truncation block only runs when `max_token_size` actually arrives, so the scope is narrower than it looks. Verified on `main` @ 5183dec5 by driving each path with a 20 001-token input:

| Path | `max_token_size` injected? | Truncation block runs? |
|---|---|---|
| Direct library use — `LightRAG(embedding_func=openai_embed)`, as in `examples/lightrag_openai_demo.py` | yes, 8192 from the decorator | **yes** |
| API server | no | no |
| `azure_openai_embed` | no | no |

`EmbeddingFunc.__call__` injects only when `inspect.signature(self.func)` declares the parameter (`lightrag/utils.py:697-701`). For the server that `self.func` is `optimized_embedding_function(texts, embedding_dim=None, context="document")` (`lightrag/api/lightrag_server.py:789-791`), which does not, so `EMBEDDING_TOKEN_LIMIT` never reaches this block; `azure_openai_embed`'s own signature omits the parameter too. Those two look like separate gaps and are not what this issue is about.

## Steps to reproduce

```python
import tiktoken
from lightrag.utils import TiktokenTokenizer

# (a)
enc.encode("an example of the <|endoftext|> marker in prose")
# ValueError: Encountered text corresponding to disallowed special token '<|endoftext|>'

# (b)
s = "hello world " * 10 + "\U0001F44D" * 5
cut = enc.decode(enc.encode(s)[:24])
print(repr(cut[-6:]), "�" in cut, s.startswith(cut))
# 'rld \U0001F44D�' True False

# what the safe contract gives instead
tk = TiktokenTokenizer(model_name="gpt-4o-mini")
span = tk.truncate_by_token_limit(s, 24)
print(repr(s[:span.end][-6:]), "�" in s[:span.end], s.startswith(s[:span.end]))
# 'd \U0001F44D\U0001F44D\U0001F44D\U0001F44D' False True
```

End to end, through the binding itself:

```python
import asyncio
import lightrag.llm.openai as O

captured = {}
def fake_client(**_):
    class C:
        async def __aenter__(self): return self
        async def __aexit__(self, *a): return False
    c = C()
    async def create(**p):
        captured["input"] = p["input"]
        return SimpleNamespace(data=[SimpleNamespace(embedding=[0.0] * 1536) for _ in p["input"]])
    c.embeddings = SimpleNamespace(create=create)
    return c

O.create_openai_async_client = fake_client
asyncio.run(O.openai_embed(["an example of the <|endoftext|> marker"], api_key="k"))
# ValueError: Encountered text corresponding to disallowed special token '<|endoftext|>'
```

## Expected Behavior

Same as every other truncation path #3565 migrated: special-token strings encode as ordinary text, the truncated text is a verbatim character-boundary prefix of the input, and its token count is independently re-verified against the budget.

## LightRAG Config Used

Direct library use with the OpenAI binding, e.g. `examples/lightrag_openai_demo.py`. `max_token_size` comes from `@wrap_embedding_func_with_attrs(..., max_token_size=8192)` on `openai_embed`.

## Logs and screenshots

```
```

## Additional Information

- Operating System: reproduced on macOS 15 (arm64); platform-independent
- Python Version: 3.13.5, tiktoken 0.12.0 (the version `uv.lock` pins, so CI resolves the same)
- Related Issues: #3565 established the contract and named this path as the next follow-up

JessYanCoding:
## Description

`openai_embed`'s own truncation slices the raw tiktoken token list and decodes it back, bypassing `lightrag.utils.Tokenizer`. #3565 established the verified split/truncate contract, migrated the paths it covered, and listed this one under "Known remaining paths (not migrated this round)" with the approach to take:

> **`openai_embed`**'s local truncation (`lightrag/llm/openai.py`): `tokens[:max_token_size]` + `decode()`, bypasses `lightrag.utils.Tokenizer` entirely, no re-verification. This **is** a live embedding call path and should be the next follow-up; suggested fix: apply `document_prefix`/`query_prefix` first (unchanged order), use the already-resolved embedding-model tiktoken encoding, call a safe truncate on the full prefixed input, and independently re-verify the result.

Two defects:

- **A literal special-token string fails the whole batch.** `Encoding.encode` defaults to `disallowed_special=ALL`, so text merely containing the literal `<|endoftext|>` raises `ValueError`. Once the block is entered the `encode` runs for every non-empty text, so length is irrelevant. `Tokenizer.encode` exists precisely to encode those as ordinary text — its comment reads *"This crashes document indexing on any user content that happens to contain those strings"* — and this path never reached it.
- **Truncation can cut inside a code point.** `decode(tokens[:n])` decodes with `errors="replace"`, so a budget landing mid-character yields a trailing U+FFFD; what gets embedded is then not a prefix of the input. Nothing re-encoded the result to confirm it fits, which the contract requires because BPE token count is not monotonic in text length.

**Scope.** The block only runs where `max_token_size` actually arrives: direct library use (`LightRAG(embedding_func=openai_embed)`, as in `examples/lightrag_openai_demo.py`), where `EmbeddingFunc` injects the decorator's 8192 because `openai_embed`'s signature declares the parameter. The API server's `optimized_embedding_function(texts, embedding_dim=None, context="document")` does not declare it, so `EmbeddingFunc.__call__`'s `inspect.signature` check (`utils.py:697-701`) skips the injection and the block never runs there; `azure_openai_embed` does not forward the parameter either. Both look like separate gaps — happy to open an issue for them, but they are not touched here.

Fixes #3714

## Changes Made

- The truncation loop now builds a `Tokenizer` over the encoding `_get_tiktoken_encoding_for_model` already resolves, and takes `truncate_by_token_limit(text, max_token_size)`, slicing the original string. The result is a verbatim character-boundary prefix whose token count is independently re-verified. Prefix application order is untouched, as prescribed.
- The **base** `Tokenizer` is used rather than `TiktokenTokenizer`, for two reasons stated in the code comment: `TiktokenTokenizer` resolves the encoding from the model name itself and raises `ValueError` for the deployment / gateway names this binding accepts (which is why `_get_tiktoken_encoding_for_model` has a `cl100k_base` fallback at all), and its `decode_with_offsets` fast path is aimed at the many windows of a `split`, not a single `truncate` — see the measurements below.
- `TokenBudgetError` is logged with the offending input's batch position and re-raised rather than swallowed, matching `_truncate_vdb_content` (`lightrag/operate.py:318-330`), the sibling call site #3565 migrated. It is only reachable when a single code point cannot fit the budget.
- The debug line keeps a magnitude signal (`chars dropped`) so operators can still see how much was cut.

## Verification

New `tests/llm/openai_impl/test_openai_embed_safe_truncation.py` (9 tests) drives the real `openai_embed`, mocking only `create_openai_async_client`, and asserts on the exact `input` list that reaches the embeddings API.

**Three-state check**, same working tree, reverting only `lightrag/llm/openai.py`:

| State | Result |
|---|---|
| impl + tests | `9 passed` |
| revert impl only | `4 failed, 5 passed` — genuine assertion failures, not an `ImportError` (the module imports only symbols present in both revisions) |
| restored | `27 passed` across `tests/llm/openai_impl/` |

**Mutation check** — each mutation applied to the implementation, suite re-run, then reverted:

| Mutation | Caught |
|---|---|
| prefixes applied *after* truncation instead of before | `1 failed` |
| `except TokenBudgetError` handler deleted | `1 failed` |
| tokenizer built from a hardcoded model name | `9 failed` |

- Full offline suite (`pytest tests/ -m offline`), Python 3.13.5: **5888 passed, 48 skipped, 1304 deselected, 0 failed** — the `main` baseline (5879 passed, 0 failed) plus this PR's 9 new tests.
- `pre-commit run --files lightrag/llm/openai.py tests/llm/openai_impl/test_openai_embed_safe_truncation.py`: clean.

**Performance**, measured on this machine (`text-embedding-3-small` / `cl100k_base`, best of 3):

| Case | old slice+decode | base `Tokenizer` | `TiktokenTokenizer` fast path |
|---|---|---|---|
| 40 texts under budget (the common case) | — | 9.8 ms | 9.6 ms |
| 20 texts at ~12 000 tokens, budget 8192 | — | **51.7 ms** | 124.9 ms |
| tokens kept at budget 8192 | 8192 | 8192 | 8192 |

A 3000-case fuzz over ASCII / CJK / ZWJ-emoji / combining-mark / literal-special-token inputs at budgets 1–128 found zero contract violations for either tokenizer and identical `TokenBudgetError` behaviour; the fast path keeps 0–14 more tokens on 35% of *small* budgets and exactly the same amount at 8192. So the base class is the right choice here: same output at realistic budgets, 2.4× faster on the branch that does work.

Not verified: behavior against a live embeddings endpoint. The change and its tests exercise only the local text→text truncation; the network boundary is mocked throughout. The skipped tests are the pre-existing spaCy-model skips, identical on `main`.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (not necessary — no documented interface changes)
- [x] Unit tests added

## Additional Notes

Deliberately **not** in this PR, flagged rather than silently left:

- **The loop is still synchronous CPU work inside an `async def`.** #3565 submits its analogous work through `run_in_tokenizer_executor`. That is a real improvement, but the blocking is pre-existing (the old slice+decode ran on the event loop too) and moving this loop into the tokenizer executor changes the concurrency of every embedding batch — a bigger, separately-reviewable change. With the base `Tokenizer` the added block on the truncation branch is ~1.3× the old cost, not the ~3× the fast path would have cost; texts already within budget are unchanged.
- **`_truncate_section_context`** (`operate.py`), the other path #3565 listed as remaining — an internal display string rather than a live embedding path, which that PR itself ranked lower priority.

### DIFF
--- a/lightrag/llm/openai.py
+++ b/lightrag/llm/openai.py
@@ -31,6 +31,8 @@ from lightrag.utils import (
     empty_length_truncated_hint,
     logger,
     TruncatedResponse,
+    Tokenizer,
+    TokenBudgetError,
 )
 
 from lightrag.api import __api_version__
@@ -1046,22 +1048,45 @@ async def openai_embed(
         texts = [document_prefix + text for text in texts]
     # Apply text truncation if max_token_size is provided
     if max_token_size is not None and max_token_size > 0:
-        encoding = _get_tiktoken_encoding_for_model(model)
+        # Wrap the resolved encoding in LightRAG's verified tokenizer instead of
+        # slicing the raw token list: Tokenizer.encode treats special-token
+        # strings such as "<|endoftext|>" as ordinary text, and
+        # truncate_by_token_limit returns a character-boundary prefix whose token
+        # count is independently re-verified. The base class is used rather than
+        # TiktokenTokenizer because the latter resolves the encoding from the
+        # model name itself and raises ValueError for the deployment / gateway
+        # names this binding accepts; its decode_with_offsets fast path only pays
+        # off across the many windows of a split, not a single truncate. The BPE
+        # table is already memoized by _get_tiktoken_encoding_for_model.
+        tokenizer = Tokenizer(
+            model_name=model, tokenizer=_get_tiktoken_encoding_for_model(model)
+        )
         truncated_texts = []
         truncation_count = 0
 
-        for text in texts:
+        for index, text in enumerate(texts):
             if not text:
                 truncated_texts.append(text)
                 continue
 
-            tokens = encoding.encode(text)
-            if len(tokens) > max_token_size:
-                truncated_tokens = tokens[:max_token_size]
-                truncated_texts.append(encoding.decode(truncated_tokens))
+            try:
+                span = tokenizer.truncate_by_token_limit(text, max_token_size)
+            except TokenBudgetError as e:
+                # Only reachable when a single Unicode code point does not fit
+                # the budget -- surfaced rather than swallowed, matching
+                # _truncate_vdb_content.
+                logger.error(
+                    f"Embedding input {index + 1}/{len(texts)} cannot fit "
+                    f"token limit {max_token_size}: {e}"
+                )
+                raise
+
+            if span.end < len(text):
+                truncated_texts.append(text[span.start : span.end])
                 truncation_count += 1
                 logger.debug(
-                    f"Text truncated from {len(tokens)} to {max_token_size} tokens"
+                    f"Text truncated to {span.token_count} tokens "
+                    f"(limit {max_token_size}, {len(text) - span.end} chars dropped)"
                 )
             else:
                 truncated_texts.append(text)
--- a/tests/llm/openai_impl/test_openai_embed_safe_truncation.py
+++ b/tests/llm/openai_impl/test_openai_embed_safe_truncation.py
@@ -1,0 +1,211 @@
+"""``openai_embed``'s own truncation must honour the safe-tokenizer contract.
+
+#3565 introduced ``lightrag.utils.Tokenizer``'s verified split/truncate
+contract and migrated the paths it covered, listing this one under "Known
+remaining paths": ``openai_embed`` sliced the raw tiktoken token list and
+``decode()``d it back, bypassing ``Tokenizer`` entirely with no
+re-verification. Two consequences, both live wherever the truncation block runs at all. That is
+direct library use -- ``LightRAG(embedding_func=openai_embed)``, as in
+``examples/lightrag_openai_demo.py`` -- where ``EmbeddingFunc`` injects the
+decorator's ``max_token_size=8192`` because ``openai_embed``'s own signature
+declares the parameter. (The API server wraps the binding in an
+``optimized_embedding_function`` whose signature does not, so nothing is
+injected there and the block is skipped entirely; ``azure_openai_embed``
+likewise never forwards the parameter. Both are separate gaps, out of scope
+here.)
+
+1. ``Encoding.encode`` defaults to ``disallowed_special=ALL``, so a chunk whose
+   text merely *contains* the literal ``"<|endoftext|>"`` raised ``ValueError``
+   and failed the whole embedding batch -- regardless of its length, since the
+   encode ran for every non-empty text once the block was entered.
+   ``Tokenizer.encode`` exists precisely to treat those strings as ordinary
+   text.
+2. ``decode(tokens[:n])`` decodes with ``errors="replace"``, so a budget landing
+   inside a multi-token code point yielded a string ending in U+FFFD. What was
+   embedded was then not a prefix of the input at all.
+
+Nothing re-encoded the result either, which the contract requires because BPE
+token count is not monotonic in text length.
+"""
+
+from types import SimpleNamespace
+from unittest.mock import patch
+
+import pytest
+
+from lightrag.llm.openai import (
+    _get_tiktoken_encoding_for_model,
+    openai_embed,
+)
+from lightrag.utils import TokenBudgetError
+
+
+pytestmark = pytest.mark.offline
+
+MODEL = "text-embedding-3-small"
+
+
+class _FakeEmbeddingClient:
+    """Captures the ``input`` list that actually reaches the embeddings API."""
+
+    def __init__(self, captured):
+        self._captured = captured
+        self.embeddings = SimpleNamespace(create=self._create)
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, exc_type, exc, tb):
+        return False
+
+    async def _create(self, **params):
+        self._captured.append(params)
+        return SimpleNamespace(
+            data=[SimpleNamespace(embedding=[0.0, 1.0]) for _ in params["input"]]
+        )
+
+
+async def _embed(monkeypatch, texts, *, max_token_size, model=MODEL, **kwargs):
+    """Drive the real ``openai_embed`` body; return the texts sent to the API."""
+    captured = []
+    monkeypatch.setattr(
+        "lightrag.llm.openai.create_openai_async_client",
+        lambda **_: _FakeEmbeddingClient(captured),
+    )
+    # ``.func`` unwraps @wrap_embedding_func_with_attrs, ``__wrapped__`` the
+    # tenacity retry, so a raised error surfaces once instead of three times.
+    await openai_embed.func.__wrapped__(
+        texts, model=model, api_key="test-key", max_token_size=max_token_size, **kwargs
+    )
+    assert len(captured) == 1
+    return captured[0]["input"]
+
+
+@pytest.mark.asyncio
+async def test_literal_special_token_does_not_fail_the_batch(monkeypatch):
+    """A chunk containing "<|endoftext|>" must embed like any other text."""
+    text = "an example of the <|endoftext|> marker in prose"
+
+    sent = await _embed(monkeypatch, [text], max_token_size=8192)
+
+    assert sent == [text]
+
+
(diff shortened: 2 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
Thanks for the thorough diagnosis here — the two defects are real, the scope analysis (which paths actually receive `max_token_size`) is accurate and saved us the work of re-deriving it, and the three-state + mutation verification is exactly the standard we want. #3714 stands as written.

We're taking a different fix, so I'm closing this in favour of #3718, which lands the minimal change (`encoding.encode(text, disallowed_special=())`) and keeps the existing token-slice truncation. The reasoning, since it's a judgement call rather than a defect in your patch:

**1. The base `Tokenizer` regresses harder than the bug it fixes.** Measured end to end through `openai_embed` on this branch, 192 000 characters (16 000-char contiguous CJK head + English tail) at the decorator's 8192 budget:

| | Sent to the API | Wall time |
|---|---|---|
| `main` | 8192 tokens | 0.99 s |
| this PR | **1 character** (`'语'`) | **13.1 s** |

Two independent causes, both in the generic implementation rather than in your patch:

- `_bounded_prefix_span` seeds from one whole-text `chars_per_token` average and retreats strictly one-directionally. A head *sparser* than the document average is accepted immediately at whatever length the average suggested — 4411/8192 tokens for an English head with a CJK body, ≈45 % of the budget silently dropped. A head much *denser* overshoots, and `max(1, candidate_len - step)` clamps the final exponential step straight to 1 character. Note its docstring promises the "**longest** safe prefix", so this is a contract gap in `Tokenizer`, not a matter of taste.
- The retreat re-encodes 15 candidates that all still contain the same CJK head, and `tiktoken` is superlinear in the length of a single regex pretoken — contiguous CJK has no whitespace to split on. 16 000 CJK chars: 783 ms; 32 000: 3106 ms; the same characters with a space every 20: 2.9 ms. The 13.1 s is 16 × 0.78 s of blocking CPU on the event loop. (The one-encode cost is pre-existing and hits `main` too — only the ×16 is new.) This is also where the "~1.3× the old cost" note comes apart: on CJK-mixed input it measures ~13×.

Your fuzz couldn't have caught either one — the inputs are homogeneous in density (pure ASCII / pure CJK / pure emoji / pure combining marks), and both failure modes need the *head* to differ in density from the document average.

**2. The two guarantees the contract buys aren't needed at this site.** The truncated string is sent over HTTP and discarded, so nothing consumes character offsets into the original. And the U+FFFD artefact is bounded to one replacement character at the very end (every token before the cut decodes to complete characters), which does not measurably move a 1536-dim vector. On the re-verification point specifically, I fuzzed the property the old code doesn't prove by construction — `encode(decode(tokens[:n])) <= n` — over 7097 random cases plus every cut point on CJK/emoji/ZWJ/mixed text (1103 of which produce a U+FFFD): worst deviation `+0`. Meanwhile the token slice fills the budget *exactly*, which no character-prefix scheme can do.

**3. `disallowed_special=()` beats `Tokenizer.encode` on the special-token axis itself.** That wrapper retries only *after* catching the `ValueError`, so affected texts pay a second full encode (two 0.78 s encodes on CJK content), and its fallback is gated on `"special token" in str(e)` — a `tiktoken` message change breaks it silently. Passing the argument up front has neither problem.

Two of your tests carry over to #3718 in spirit (`test_literal_special_token_does_not_fail_the_batch`, `test_oversized_text_containing_a_special_token`) — credited there.

Separately, and worth its own issue rather than a fix in either PR: the clamp-to-1 in `_bounded_prefix_span` also affects `split_by_token_limit`, so anyone injecting a custom `tokenizer_func` (no `decode_with_offsets` fast path) can get 1-character chunks on density-heterogeneous documents. If you'd like to take that one, it'd be a welcome follow-up — a bisect between the last-known-good and last-known-bad candidate instead of the exponential clamp. The `max_token_size` gaps you flagged (`optimized_embedding_function` and `azure_openai_embed` not forwarding it, so the server truncates nothing locally) are also still open and worth an issue.

PR #3716, opened 2026-08-24
Add DocumentDB as an all-in-one LightRAG storage backend 

sandeepsnairms:
## Description

Adds DocumentDB as an all-in-one LightRAG storage backend using its MongoDB-compatible wire protocol. The integration supports key-value, vector, graph, and document-status storage while maintaining separate configuration and connection management from MongoDB.

DocumentDB-specific behavior is handled through capability overrides, including native HNSW vector indexes, graph traversal support, regex-based entity search fallback, and binary file-path sorting when query collation is unavailable.

## Related Issues

No related issue.

## Changes Made

- Added the following storage implementations:
  - DocumentDBKVStorage
  - DocumentDBVectorDBStorage
  - DocumentDBGraphStorage
  - DocumentDBDocStatusStorage
- Added isolated DocumentDB client and connection-pool management.
- Added DOCUMENTDB_URI, DOCUMENTDB_DATABASE, and DOCUMENTDB_WORKSPACE configuration.
- Added native DocumentDB HNSW vector index creation and vector search support.
- Added graph storage using MongoDB-compatible graph operations.
- Added DocumentDB support to storage registries and factory resolution.
- Added DocumentDB options to the interactive setup wizard and validation flow.
- Added DocumentDB support to cache migration, cache cleanup, and vector-database rebuild tools.
- Added capability handling for unsupported collation indexes and query collation.
- Updated English, Chinese, and Japanese documentation.
- Added offline unit tests and a live DocumentDB integration smoke test.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

The live integration test requires a running DocumentDB instance and the following environment variables:
```
DOCUMENTDB_URI=mongodb://user:password@localhost:10260/?tls=true&tlsAllowInvalidCertificates=true
DOCUMENTDB_DATABASE=LightRAG
```

tlsAllowInvalidCertificates=true is intended only for local development with the DocumentDB container. Production deployments should use a trusted certificate with certificate validation enabled.


danielaskdd:
Thanks for the thorough work here — the refactor of `mongo_impl.py` into `_Mongo*StorageBase` + capability `ClassVar`s is a clean shape, and the fact that the new backend needs only 135 lines is a good sign that the wire-protocol compatibility story holds up.

Three items before this can move forward. The first is blocking.

---

### 1. (Blocking) Dropping `partialFilterExpression` changes the edge unique-index semantics, and the premise is unverified

`DocumentDBGraphStorage` sets `edge_index_partial_filter = None`, which removes the partial filter from the `(edge_lo, edge_hi)` unique index:

```python
edge_index_partial_filter: ClassVar[dict | None] = None
```

That filter is not decorative. Per the docstring on `create_edge_indexes_and_migrate_if_not_exists`, the index doubles as the **migration completion flag**, and the upsert filter relies on every doc having `edge_lo`/`edge_hi`. The partial filter is what confines the unique constraint to *already-migrated* docs:

- **With** the filter — endpoint-less legacy stragglers are excluded from the index, so the build succeeds and the migration completes.
- **Without** it — every endpoint-less doc indexes as the same `(null, null)` key. Two or more of them produce a duplicate-key error, `create_index` raises, and **startup aborts**. The same collection state that migrates cleanly on MongoDB hard-fails on DocumentDB.

Two things needed here:

**(a) Please verify the premise.** The PR asserts by omission that DocumentDB cannot do partial indexes, but no evidence is given — and since the DocumentDB engine is PostgreSQL-based, partial indexes are a native capability of the storage layer underneath. Please confirm what actually fails: partial indexes altogether, `$exists`, `$type`, or the combination? If only `$type` is unsupported, an `$exists`-only filter preserves the property that matters here. A capability override that silently widens a uniqueness constraint needs a documented reason, not a `None`.

**(b) If it genuinely is unsupported, the divergence needs a test and a documented consequence.** `edge_index_partial_filter` currently appears nowhere in `test_documentdb_storage.py`, `test_documentdb_integration.py`, or `test_documentdb_cache_tools.py` — the one behavioral change in this PR that can abort startup is the one with no coverage. A test that seeds two endpoint-less edge docs and asserts the intended behavior on both flavors would pin it.

### 2. (Question) `$vectorSearch` vs. the documented `$search: {cosmosSearch: ...}` — what is the minimum supported version?

Index *creation* correctly uses `cosmosSearchOptions` with `kind: "vector-hnsw"`, but the query path still emits the Atlas-style stage, only dropping the `index` field:

```python
vector_query_uses_index_name: ClassVar[bool] = False
```

so the pipeline becomes `{"$vectorSearch": {path, queryVector, numCandidates, limit}}` with `{"$meta": "vectorSearchScore"}`.

Microsoft's documented operator is `$search: {cosmosSearch: {vector, path, k}}`. Your live integration test does assert on `vectors.query()`, so `$vectorSearch` evidently works on the build you tested against — but that leaves an undocumented version floor. Could you state which DocumentDB / Azure DocumentDB versions accept `$vectorSearch` and `vectorSearchScore`, and add that to the docs? Users on an older engine would otherwise get a working ingest and a silently broken retrieval path, which is the worst possible failure mode here.

### 3. (Design) These are dialect differences, not a distinct backend

Stepping back: the diff adds 4 storage classes, 3 environment variables, a separate `ClientManager`, registry + factory entries, setup-wizard branches, and trilingual documentation — and all of it exists to carry **four boolean/ClassVar overrides** plus one index-creation method. Everything else is `class DocumentDBKVStorage(_DocumentDBStorageMixin, _MongoKVStorageBase): pass`.

A flavor switch expresses the same thing at a fraction of the surface:

```
MONGO_FLAVOR=mongodb|atlas|documentdb
```

Users keep the existing `MongoKVStorage` / `MongoVectorDBStorage` class names and `MONGO_*` variables; the flavor selects the `ClientManager` and the capability `ClassVar`s. This drops the registry, factory, wizard, `env.example`, docs, and cache/rebuild-tool changes entirely, and it scales to the next MongoDB-compatible engine without another copy of the same six touch points.

Worth noting that the capability overrides here are not all DocumentDB-specific. **The regex entity-search fallback is pre-existing behavior on every non-Atlas deployment** — `mongo_impl.py` says so directly ("This is normal if not using MongoDB Atlas - search will use regex fallback"), and it only affects the WebUI graph-explorer label search (`GET /graph/label/search`), not retrieval recall, which goes through `entities_vdb.query()` / `relationships_vdb.query()`. So `create_search_index_if_not_exists` being a no-op is really an `atlas` vs. `non-atlas` axis that self-hosted MongoDB shares — further evidence that flavor capabilities, not a parallel backend, is the right factoring.

**Minor, related:** please disambiguate the name in the docs. Bare "DocumentDB" collides with Amazon DocumentDB, which is a different engine with different vector-search syntax and a different default port. "Azure DocumentDB" or "DocumentDB (Microsoft/Linux Foundation)" at first mention in each README would save users a lot of confusion.

---

Also: `lightrag/kg/documentdb_impl.py` is missing a trailing newline.

sandeepsnairms:
Thanks for the detailed review. I addressed the blocking issue and vector-query compatibility concerns 

## 1. Partial unique edge index
Confirmed that DocumentDB supports partialFilterExpression, including both $exists and $type.

I also cross-checked the DocumentDB migration utility implementation in IndexCopier.cs, which explicitly lists these operators as supported:

$eq, $gt, $gte, $lt, $lte
$type
$exists
$and
The DocumentDB graph backend now inherits the same partial filter used by MongoDB:

{
    "edge_lo": {"$exists": True, "$type": "string"},
    "edge_hi": {"$exists": True, "$type": "string"},
}
This restores the intended migration-completion semantics and excludes endpoint-less legacy documents from the compound unique index.

A regression test now verifies the exact partial unique-index command generated during DocumentDB edge migration.

## 2. Vector query compatibility
I changed the DocumentDB query path to use the documented syntax:

{
  $search: {
    cosmosSearch: {
      path: "vector",
      vector: queryVector,
      k: topK
    }
  }
}
Scoring uses $meta: "searchScore".

The documentation now states a minimum supported version of v0.100-0 and explains that LightRAG does not require the newer $vectorSearch operator.

## 3. Separate backend design
I retained the separate DocumentDB*Storage classes and configuration namespace intentionally.

A process can use MongoDB and DocumentDB for different LightRAG storage roles simultaneously. Separate client managers and environment variables keep those connections isolated. A process-wide MONGO_FLAVOR switch would prevent that configuration and could route all Mongo-backed roles through only one engine.

The shared implementation remains in the _Mongo*StorageBase classes, while the DocumentDB classes carry only the dialect and capability differences.

## Documentation
The docs now identify the backend as DocumentDB (Linux Foundation) and explicitly distinguish it from Amazon DocumentDB.

danielaskdd:
@codex review

danielaskdd:
@codex  review

sandeepsnairms:
@danielaskdd : Can you help merge?

danielaskdd:
Thanks for the thorough work here, @sandeepsnairms. This is a substantial, well-tested implementation, and the quality of the contribution is not in question.

I want to be explicit about why this PR targets `dev-documentdb` instead of going directly into `main`, and what happens from here.

LightRAG already supports more than ten storage backends across KV, vector, graph, and document-status storage. Every backend added to `main` creates recurring maintenance cost: each storage-interface or schema change must be adapted, tested, performance-tuned, and migration-validated across every supported implementation. Because DocumentDB is new to this project and has not yet received broad community validation, we do not want to make that long-term compatibility commitment in `main` immediately.

At the same time, DocumentDB (Linux Foundation) is clearly useful to teams that want a MongoDB-wire-compatible backend covering all four storage roles, so we do not want to close the contribution. We have created `dev-documentdb` as a longer-lived integration branch where you and other interested users can continue testing it, report deployment experience, keep it aligned with evolving storage contracts, and build a track record of real-world adoption. Once the branch demonstrates sustained maintenance, usage, and stability, we can revisit promoting it into `main` as a first-class backend.

Practically, this PR will merge into `dev-documentdb`, whose base is already configured correctly, not into `main`. Keeping the backend synchronized with future storage-interface changes on `main` will be a community responsibility. Core maintainers will continue to review contributions, but will not carry DocumentDB-specific migration work as part of every core storage change during this validation period.

I have also updated `docs/LightRAG-API-Server.md` in this PR to point users to the `dev-documentdb` branch and explain its community-validation status. Thank you again for the contribution and for continuing to help mature this backend.

PR #3717, opened 2026-08-24
fix(embeddings): keep provider API keys request-local

pengpengyi92:
## Description

Keep explicit Jina and NVIDIA embedding API keys local to the current call instead of writing them into process-wide environment variables.

Previously, calling either embedding function with `api_key=...` permanently replaced `JINA_API_KEY` or `OPENAI_API_KEY`. A later call that relied on the process default could therefore inherit credentials from an unrelated earlier request. This is especially surprising for library users that construct provider functions with different credentials in the same process.

## Related Issues

No linked issue. This was found while auditing provider credential and client-lifecycle behavior.

## Changes Made

- Resolve the Jina request key from the explicit argument first, then the existing environment fallback, without mutating `os.environ`.
- Pass an explicit NVIDIA key directly to `AsyncOpenAI`, preserving SDK environment fallback when no key is supplied.
- Add offline regression coverage proving that explicit keys reach each provider while the process-wide defaults remain unchanged.
- Preserve the existing client cleanup, retry, base URL, and missing-key behavior.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (not necessary; public configuration behavior is unchanged)
- [x] Unit tests added

## Additional Notes

Validation:

- `python -m pytest tests/llm/jina_impl tests/llm/nvidia_impl -q -p no:cacheprovider` (4 passed)
- LightRAG pre-commit hooks on all four changed files (all passed)
- `git diff --check` (passed)

Boundary: this PR does not change environment variable names, API server configuration, credential storage, retry policy, or any other provider.

### DIFF
--- a/lightrag/llm/jina.py
+++ b/lightrag/llm/jina.py
@@ -116,16 +116,14 @@ async def jina_embed(
         aiohttp.ClientError: If there is a connection error with the Jina API.
         aiohttp.ClientResponseError: If the Jina API returns an error response.
     """
-    if api_key:
-        os.environ["JINA_API_KEY"] = api_key
-
-    if "JINA_API_KEY" not in os.environ:
+    effective_api_key = api_key or os.getenv("JINA_API_KEY")
+    if not effective_api_key:
         raise ValueError("JINA_API_KEY environment variable is required")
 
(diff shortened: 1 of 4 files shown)


danielaskdd:
@codex review

PR #3719, opened 2026-08-25
chore(deps): allow pandas 3.x

### LINKED ISSUE #3711: [Question]: bump the pandas version to 3.x in pyproject.toml
### Do you need to ask a question?

- [x] I have searched the existing question and discussions and this question is not already answered.
- [ ] I believe this is a legitimate question, not just a bug or feature request.

### Your Question

The pandas version in pyproject.toml was bumped to 2.4.0 in this pull: 
https://github.com/HKUDS/LightRAG/pull/2239

pandas 2.3.3 was the last 2.x release at that time, and now it moves to 3.x.  I see only file lightrag/utils.py imports pandas and only pd.DataFrame and pd.ExcelWriter are referenced, so it is possible to bump the pandas version to 3.x in pyproject.toml?

### Additional Context

_No response_

danielaskdd:
## Summary

Allow pandas 3.x by widening the dependency ceiling from `<2.4.0` to `<4.0.0` (core deps and the `api` extra), regenerate `uv.lock` so CI actually installs pandas 3, raise the `xlsxwriter` floor to the minimum pandas 3.x declares, and add the regression coverage the Excel export path never had.

Closes #3711.

## Motivation

`pandas>=2.0.0,<2.4.0` blocks pandas 3.x for everyone installing LightRAG, and the cap carries no technical rationale: it came from a mechanical Dependabot "permit the latest version" bump (#2239, `<2.3.0` → `<2.4.0`). Dependabot cannot lift it either — the `data-processing` group in `.github/dependabot.yml` only allows `minor`/`patch` updates, so the major has to be opened by hand.

As #3711 observes, pandas is used in exactly one place: the Excel branch of `lightrag/utils.py::aexport_data`, touching only `pd.DataFrame`, `DataFrame.empty`, `pd.ExcelWriter(engine="xlsxwriter")` and `DataFrame.to_excel` — all unchanged in 3.x. (`eval_rag_quality.py`'s `to_pandas()` is a ragas method, not a pandas API.)

## What's changed

**Widened, not floor-bumped.** pandas 3.0 requires Python >= 3.11 while this project's `requires-python` is `>=3.10`. Pinning `pandas>=3.0` would break installation on 3.10 outright, so the ceiling moves to `<4.0.0` and resolvers fork: pandas 2.x on 3.10, 3.x from 3.11 up. This also matches the next-major-cap style used elsewhere (`numpy<3.0.0`, `openai<4.0.0`).

**`uv.lock` regenerated with `--upgrade-package`.** This is the non-obvious half. CI runs `uv sync --frozen`, and a plain `uv lock` is conservative — it keeps 2.3.3 pinned because it still satisfies the new range, so the widened ceiling would have shipped completely untested. The lock now forks explicitly:

```
pandas 2.3.3  →  python_full_version < '3.11'
pandas 3.0.5  →  python_full_version >= '3.11'
```

CI's 3.12 and 3.14 lanes therefore run against pandas 3.0.5. The rest of the lock diff is only marker refinement on pandas' own transitive deps; no other package version moved. pandas 3 dropping its `pytz` dependency is absorbed by the `api` extra listing `pytz` directly.

**`xlsxwriter>=3.2.0`.** That is the minimum pandas 3.x declares for its xlsxwriter Excel engine (`pandas.compat._optional.VERSIONS`). pandas does not enforce it at runtime today — 3.1.9 still writes fine — so this is alignment rather than a fix; the lock already resolved 3.2.9.

**New test: `tests/utils/test_aexport_data_excel.py`.** The Excel branch had *zero* coverage. Two cases pin the observable workbook: one sheet per non-empty section (empty sections are skipped), `index=False` so the first column is a real field, and `None` cells left empty rather than rendered as text — the last one matters because pandas 3.0 switched text columns from `object` to the new `str` dtype.

## Verification

- New test passes on **both** pandas 2.3.3 and 3.0.5 — it is a cross-major compatibility guard, not a fix-proof for a bug (nothing was broken; the point is that nothing breaks).
- Full suite on pandas 3.0.5: **7041 passed, 200 skipped, 44 errors**. All 44 errors are `tests/parser/docx/test_smart_heading_*` requiring the spaCy models, which are not installed in this environment and fail identically on `main` — unrelated to this change.
- No other pandas pin exists in the repo (no `requirements.txt`, Docker, or k8s constraint), and `ragas` / `datasets` in the `evaluation` extra declare no pandas upper bound, so nothing conflicts.

## What could break

Users on Python >= 3.11 who run a fresh `uv sync` get pandas 3.0.5 instead of 2.3.3. Since LightRAG only exercises the four API calls listed above, the exposure is limited to the Excel export path, which this PR now covers with tests. Users on Python 3.10 see no change at all.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,12 @@ dependencies = [
     "networkx",
     "numpy>=1.24.0,<3.0.0",
     "packaging",
-    "pandas>=2.0.0,<2.4.0",
+    # pandas 3.0 requires Python >= 3.11 while this project still supports
+    # 3.10, so the ceiling is widened instead of the floor being raised:
+    # resolvers pick 2.x on 3.10 and 3.x from 3.11 up. Only the Excel branch
+    # of aexport_data uses pandas (DataFrame/ExcelWriter/to_excel), all of
+    # which are unchanged in 3.x.
+    "pandas>=2.0.0,<4.0.0",
     "pipmaster",
     "pydantic",
     "pypinyin",
@@ -42,7 +47,9 @@ dependencies = [
     # oldest release whose encode_batch/decode_batch fan encode()/decode() of
     # one Encoding across threads — the thread safety Tokenizer relies on.
     "tiktoken>=0.7.0",
-    "xlsxwriter>=3.1.0",
+    # 3.2.0 is the minimum pandas 3.x declares for its xlsxwriter Excel
+    # engine (pandas.compat._optional.VERSIONS).
+    "xlsxwriter>=3.2.0",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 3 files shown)


danielaskdd:
@codex review

Issue #3720, opened 2026-08-25
[Bug]: max_token_size never reaches azure_openai_embed or the API server binding


JessYanCoding:
## Description

Both `openai_embed` and `azure_openai_embed` are decorated with `max_token_size=8192`, but only one of them can ever receive it. `EmbeddingFunc.__call__` injects the value only when the wrapped function's own signature declares the parameter:

```python
# lightrag/utils.py:697-701
if self.max_token_size is not None and "max_token_size" not in kwargs:
    sig = inspect.signature(self.func)
    if "max_token_size" in sig.parameters:
        kwargs["max_token_size"] = self.max_token_size
```

`azure_openai_embed`'s signature (`lightrag/llm/openai.py:1237-1249`) does not declare it, so the injection is silently skipped and `openai_embed`'s truncation block never runs. Its own decorator says `max_token_size=8192`; that number is dead.

```
openai_embed        decorator=8192  signature declares it: yes  -> truncation runs
azure_openai_embed  decorator=8192  signature declares it: NO   -> truncation never runs
```

Measured on `main` @ 7ecd8a05, feeding a 20 001-token text with the network boundary mocked and reading the exact `input` that reaches the embeddings API:

| Path | Sent to the API |
|---|---|
| `LightRAG(embedding_func=openai_embed)` | 8192 tokens (truncated) |
| `LightRAG(embedding_func=azure_openai_embed)` | **20 001 tokens** (untouched) |
| API server, `EMBEDDING_TOKEN_LIMIT=8192` | **20 001 tokens** (untouched) |

The server case has the same root cause one layer up: `create_embedding_function` reads `EMBEDDING_TOKEN_LIMIT` into `EmbeddingFunc(max_token_size=final_max_token_size)` (`lightrag/api/lightrag_server.py:769, 987`), but the function it wraps is `optimized_embedding_function(texts, embedding_dim=None, context="document")` (`:789-791`), which does not declare the parameter either — so the configured limit stops at the wrapper.

The docstring at `lightrag/llm/openai.py:1289` describes `azure_openai_embed` as exposing "All parameters from the underlying `openai_embed` … to ensure full feature parity and API consistency". This is the one parameter where that is not true.

### ⚠️ How much this actually matters

Being precise about severity, because the ingestion path is already protected and this is a missing layer rather than a guaranteed failure:

| Layer | Covered? |
|---|---|
[...]

PR #3722, opened 2026-08-25
fix(tokenizer): bisect the prefix bracket instead of clamping to one char

### LINKED ISSUE #3721: [Bug]: _bounded_prefix_span returns a 1-character span on density-heterogeneous text
## Description

`Tokenizer._bounded_prefix_span` (`lightrag/utils.py`) seeds a candidate length from a whole-content `chars_per_token` average, returns the **first** candidate that fits, and on overflow steps down by exponentially growing amounts:

```python
candidate_len = min(remaining_len, max(1, int(max_tokens * chars_per_token)))
step = 1
while True:
    count = len(self.encode(content[start : start + candidate_len]))
    if count <= max_tokens:
        return TokenSpan(start, start + candidate_len, count)   # first fit wins
    if candidate_len <= 1:
        break
    candidate_len = max(1, candidate_len - step)                # clamps to 1
    step *= 2
```

The estimate is a whole-content average, so wherever token density varies it is wrong in one of two directions at `start`, and each direction breaks differently:

**A head denser than the average collapses the span to one character.** The seed overflows, the step-down doubles (1, 2, 4, 8, …) and overshoots the whole remainder within a few iterations, and `max(1, candidate_len - step)` clamps straight to 1.

**A head sparser than the average silently spends a fraction of the budget.** The shortened seed fits on the first try and is returned unexamined.

The method's own docstring promises the *longest* safe prefix, so both are contract violations rather than tuning choices. `danielaskdd` named this as a follow-up when closing #3715:

> the clamp-to-1 in `_bounded_prefix_span` also affects `split_by_token_limit`, so anyone injecting a custom `tokenizer_func` (no `decode_with_offsets` fast path) can get 1-character chunks on density-heterogeneous documents […] a bisect between the last-known-good and last-known-bad candidate instead of the exponential clamp.

### Measured

`truncate_by_token_limit`, budget 8192, `text-embedding-3-small` encoding driven through the base `Tokenizer` (what a `tokenizer_func` injection reaches — `TiktokenTokenizer` overrides both methods and is unaffected):

| Input | Returned | Budget spent |
|---|---|---|
| 192 000 chars, contiguous CJK head + Latin tail | **1 character** (`'语'`) | **0.0 %** |
| 119 600 chars, Latin head + CJK body | 7 926 chars | **19.8 %** |
| homogeneous Latin | 40 049 chars | 100 % |
| homogeneous CJK | 6 963 chars | 100 % |

**`split_by_token_limit` is where it actually hurts** — the clamp fires once per window, so the document is split per character. A 58 920-character document alternating CJK and Latin blocks, chunk budget 512:

```
spans            : 4633            (a correct split needs ~36)
median span      : 1 character
spans <= 4 chars : 4596 / 4633     (99.2 %)
first spans      : '语'  '言'  '模'  '型'  '的'
wall time        : 41.1 s
```

Every one of those single characters is then embedded as its own chunk.

## Steps to reproduce

```python
from lightrag.utils import Tokenizer
import tiktoken

tk = Tokenizer(model_name="x", tokenizer=enc)      # base class = tokenizer_func injection

CJK = "语言模型的知识图谱检索增强生成技术"
ENG = "the quick brown fox jumps over the lazy dog "

span = tk.truncate_by_token_limit(CJK * 941 + ENG * 4000, 8192)
print(span.end, "chars /", span.token_count, "tokens")      # 1 chars / 1 tokens

spans = tk.split_by_token_limit((CJK * 60 + ENG * 200) * 6, 512)
print(len(spans), "spans, median length",
      sorted(s.end - s.start for s in spans)[len(spans) // 2])   # 4633 spans, median length 1
```

## Expected Behavior

The returned span is the longest prefix that fits, whatever the local density: bracket the answer between a length verified to fit and one verified to overflow, then bisect. Windows in `split_by_token_limit` fill their budget instead of degenerating to single characters.

## LightRAG Config Used

Any `tokenizer_func` injection — a HuggingFace tokenizer, sentencepiece, or any object implementing `encode`/`decode` — over a document mixing scripts of different token density (CJK/Latin, emoji/Latin, whitespace-padded regions). `TiktokenTokenizer` users are not affected: it overrides both methods with a `decode_with_offsets` fast path.

## Logs and screenshots

Nothing is logged; both failures are silent. The per-character split surfaces only as an unexpected chunk count and embedding bill.

## Additional Information

- LightRAG Version: main @ 7ecd8a05
- Operating System: macOS 15 (arm64); platform-independent
- Python Version: 3.13.5, tiktoken 0.12.0
- Related Issues: #3565 introduced this method; #3715 is where the maintainer named this follow-up

JessYanCoding:
## Description

`_bounded_prefix_span` seeds a candidate length from a whole-content `chars_per_token` average, returns the first candidate that fits, and steps down from an over-long one by exponentially growing amounts — a step-down that overshoots the whole remainder within a few iterations, where `max(1, candidate_len - step)` clamps it to a single character.

Wherever token density varies, that average is wrong at `start` in one of two directions, and each breaks differently: a denser-than-average head collapses the span to one character; a sparser-than-average head fits on the first try and returns with most of the budget unspent. The docstring promises the *longest* safe prefix, so both are contract violations.

This is the follow-up @danielaskdd named when closing #3715:

> a bisect between the last-known-good and last-known-bad candidate instead of the exponential clamp

Fixes #3721

## Changes Made

- `_bounded_prefix_span` now brackets the answer between a length verified to fit (`good`) and one verified to overflow (`bad`) — reached by doubling up from the seed or halving down from it — and bisects that bracket. Plain bisection, no interpolation: the bracket halves every iteration, so the probe count is logarithmic in the remaining length and cannot stall.
- Updated the class-level "Safe split/truncate contract" comment block, whose three claims (strictly one-directional retreat, exponential character steps 1-2-4-8, `O(log max_tokens)`) all described the removed algorithm.
- The docstring now states the limit honestly: bracketing is an assumption about the *sampled* lengths, not a proof about every length between them, so under a sufficiently non-monotonic tokenizer a longer fitting prefix may exist that bisection never sampled. The span returned is always verified to fit — that part is unconditional.

## Verification

**Effect** (base `Tokenizer` over the `text-embedding-3-small` encoding — a `tokenizer_func` injection; `TiktokenTokenizer` overrides both methods and is untouched):

| `truncate_by_token_limit`, budget 8192 | before | after |
|---|---|---|
| 192 000 chars, CJK head + Latin tail | 1 char / **0.0 %** of budget / 5.56 s | 6 963 chars / **100 %** / 1.15 s |
| 119 600 chars, Latin head + CJK body | 7 926 chars / **19.8 %** / 14.69 s | 21 502 chars / **100 %** / 14.98 s |
| homogeneous Latin | 100 % / 0.01 s | 100 % / 0.01 s |
| homogeneous CJK | 100 % / 14.84 s | 100 % / 14.84 s |

| `split_by_token_limit`, 58 920 chars alternating CJK/Latin, budget 512 | before | after |
|---|---|---|
| spans | 4 633 | **36** |
| median span length | **1 character** | 2 070 chars |
| spans of ≤ 4 chars | 4 596 (99.2 %) | **0** |
| wall time | 41.1 s | **0.06 s** |

**Three-state**, same working tree, reverting only `lightrag/utils.py`:

| State | Result |
|---|---|
| impl + tests | `60 passed` |
| revert impl only | **`4 failed, 56 passed`** — genuine assertion failures |
| restored | `60 passed` |

**Mutations** — each applied to the implementation, suite re-run, then reverted:

| Mutation | Caught by |
|---|---|
| close the bracket by interpolating between the measurements | `test_probe_count_stays_logarithmic_on_heterogeneous_input` |
| return the seed as soon as it fits (no upward search) | `test_sparse_head_does_not_silently_drop_budget` |
| restore the exponential step-down and its clamp-to-1 | `test_dense_head_…`, `test_split_does_not_emit_single_character_spans`, `test_real_tokenizer_…` |

The first of those is worth spelling out, because I tried it first and it is wrong: interpolating converges in fewer probes on gentle density gradients, but a guard that only excludes the bracket endpoints lets a guess of `good + 1` be accepted forever — the overflowing bound stays pinned while the fitting one creeps, each probe re-encoding a near-document-length slice. On a 10 000:1 density cliff that costs **145 encodes where bisection costs 15**. `_DensityCliffTokenizer` in the new tests exists specifically to pin that; the gentler `_DensityVaryingTokenizer` (16:1) is enough to expose a wrong *answer* but not a wrong *probe count*.

- Full offline suite (`pytest tests/ -m offline`), Python 3.13.5: **5903 passed, 48 skipped, 1304 deselected, 0 failed** — the `main` baseline (5889 passed, 0 failed) plus this PR's 14 new tests.
- `pre-commit run --files lightrag/utils.py tests/llm/test_tokenizer_split_truncate_contract.py`: clean.

**Cost.** Bisection replaces "first candidate that fits" with a real search, so on homogeneous documents — where the average estimate was already right — a span costs ~13-17 encodes instead of 1-2. That is the price of the fix: the old single probe was cheap precisely because it was not looking for the longest prefix. Absolute cost stays small on Latin text (0.01 s for a 264 000-character document) and is dominated by tiktoken's own superlinear cost on contiguous CJK, which is unchanged. On the heterogeneous inputs this PR is about, it is strictly faster (5.56 s → 1.15 s; 41.1 s → 0.06 s) because it stops re-encoding the same over-long candidate.

Not verified: real third-party tokenizers other than tiktoken. Non-monotonicity is modelled with the existing `_WordCountingTokenizer` (a longer prefix costing fewer tokens) rather than measured on HuggingFace or sentencepiece.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (the contract comment block and the affected docstrings)
- [x] Unit tests added

## Additional Notes

The two `max_token_size`-forwarding gaps flagged in #3715 (`optimized_embedding_function` and `azure_openai_embed` not declaring the parameter, so `EMBEDDING_TOKEN_LIMIT` never reaches the binding) are filed separately as #3720.

### DIFF
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -3099,12 +3099,15 @@ class Tokenizer:
         return TokenSpan(start, start + 1, first_cp_tokens)
 
     def truncate_by_token_limit(self, content: str, max_tokens: int) -> TokenSpan:
-        """Return the longest safe prefix of ``content`` that fits ``max_tokens``.
+        """Return a verified safe prefix of ``content`` that fits ``max_tokens``.
 
         The result always starts at character offset 0. If the whole string
-        already fits, the returned span covers it entirely. Raises
-        ``ValueError`` for a non-positive budget, and ``TokenBudgetError`` if
-        not even the first complete Unicode code point fits.
+        already fits, the returned span covers it entirely. Otherwise the
+        bracketed search aims to use the available budget efficiently, but a
+        non-monotonic tokenizer may have a longer fitting prefix that was not
+        sampled. Raises ``ValueError`` for a non-positive budget, and
+        ``TokenBudgetError`` if not even the first complete Unicode code point
+        fits.
         """
         if max_tokens <= 0:
             raise ValueError(f"max_tokens must be positive, got {max_tokens}")
@@ -3214,7 +3217,7 @@ class TiktokenTokenizer(Tokenizer):
             import tiktoken
         except ImportError:
             raise ImportError(
-                "tiktoken is not installed. Please install it with `pip install tiktoken` or define custom `tokenizer_func`."
+                "tiktoken is not installed. Please install it with `pip install tiktoken` or pass a custom `tokenizer`."
             )
 
         try:
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -3005,12 +3005,19 @@ class Tokenizer:
     # the cost of a handful of extra ``encode()`` calls per span compared to
     # the tiktoken fast path.
     #
-    # BPE token count is not monotonic in character-prefix length, so this is
-    # a bounded, STRICTLY ONE-DIRECTIONAL retreat: candidate length only ever
-    # decreases (by exponential character steps: 1, 2, 4, 8, ...), never
-    # re-expands. That rules out oscillation and bounds the number of
-    # re-encodes to O(log max_tokens) regardless of how good the initial
-    # char/token ratio estimate turns out to be for a given region of text.
+    # BPE token count is not monotonic in character-prefix length, so no
+    # candidate is ever trusted without encoding it. The search brackets the
+    # answer between a length verified to fit and one verified to overflow --
+    # reached by doubling or halving from the ratio estimate -- and then
+    # bisects that bracket. The bracket only ever narrows, which rules out
+    # oscillation, and both phases are logarithmic in the remaining character
+    # count, so the number of re-encodes stays bounded however wrong the
+    # initial char/token ratio estimate is for a given region of text.
+    #
+    # Bracketing is an assumption about the sampled lengths, not a proof about
+    # every length between them: the span returned is always verified to fit,
+    # but under a sufficiently non-monotonic tokenizer a longer one may exist
+    # that the bisection never sampled.
     #
     # No cross-call state: nothing here is written back onto ``self``. An
     # injected tokenizer must remain safe under concurrent calls from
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3726, opened 2026-08-26
Allow Bedrock default credential providers

a-bruhn:
### Description

- Bedrock startup currently rejects deployments without static AWS credentials or `AWS_BEARER_TOKEN_BEDROCK`.
- This prevents the use of IAM Roles for Service Accounts (IRSA), even though the Bedrock client already uses botocore's default credential chain.

### Changes

- Allow the default chain when no explicit credentials are configured; still reject incomplete explicit key pairs.
- Add coverage for IRSA configuration on base and role-specific Bedrock bindings.


danielaskdd:
@codex review

PR #3735, opened 2026-08-28
fix(pipeline): keep generated failure summary within varchar(255)

danielaskdd:
## Summary

A parse-stage failure could not record its own `FAILED` status on PostgreSQL when the error message was long. The upsert was rejected, so the document stayed in `PARSING` with no error recorded — and then auto-resumed and failed the same way, invisibly.

## Motivation

Reported from a MinerU run that hit a CUDA OOM. The backend log shows both halves of the failure:

```
ERROR: Parse worker failed (mineru): MinerU local parse failed for task ...: CUDA out of memory. Tried to allocate 360.00 MiB. ...
ERROR: Failed to record FAILED status for doc-953716db7881ca3197956f8820c219cb: value too long for type character varying(255)
```

The document's `doc_status` still read `PARSING`, with no `error_msg` and no `metadata.error_type` / `error_stage`.

### Root cause

`doc_status_parse_failure_fields` builds the display summary as:

```python
FILE_EXTRACTION_SUMMARY_PREFIX + get_content_summary(error_text)
```

- `FILE_EXTRACTION_SUMMARY_PREFIX` = `"[File Extraction]"` — 17 chars
- `get_content_summary` caps at `max_length=250` and appends `"..."` — 253 chars

That is **270 chars**, while `LIGHTRAG_DOC_STATUS.content_summary` is declared `varchar(255)`. Any error message over 250 characters overflows — a CUDA OOM message is far past that.

`status`, `error_msg`, `metadata` and `content_summary` all ride the **same single-row** `INSERT ... ON CONFLICT DO UPDATE`, so one column overflowing discards the entire FAILED transition. The parse worker deliberately swallows the upsert exception (re-raising would kill the worker) and only logs, so the document silently keeps its previous status.

The consequence is worse than a missing message: `PARSING` is in `_AUTO_RESUME_DOC_STATUSES`, so the next pipeline run treats the document as a dead-process orphan and re-parses it. If the underlying condition persists (GPU still out of memory), the FAILED record fails to write again — the document loops between `PARSING` and re-parse, and the WebUI never shows why.

Only PostgreSQL declares a length on this column. JSON / Redis / MongoDB / OpenSearch `doc_status` are unconstrained and were writing the FAILED row correctly, which is why this only surfaces on PG deployments.

## What's changed

- **`lightrag/constants.py`** — new `DOC_STATUS_CONTENT_SUMMARY_MAX_LENGTH = 255`, documenting that PostgreSQL's `varchar(255)` is the narrowest storage a `content_summary` must fit, and that producers building one from unbounded text must budget against it.
- **`lightrag/utils_pipeline.py`** — the generated `[File Extraction]` summary now budgets the description as `255 - len(prefix) - 3`, so prefix + description + ellipsis lands exactly on 255 for an error message of any length. `error_msg` is `TEXT` and still carries the full text; only the display summary is shortened.
- **`tests/pipeline/test_parse_failure_doc_status_fields.py`** — a parametrized regression test over three inputs: a 1000-char message, the verbatim CUDA-OOM message from the report, and an exact-boundary string. The existing truncation test is kept, with its brittle `<= 17 + 253` bound replaced by an ellipsis assertion.

Other `content_summary` producers were audited and are bounded already: `get_content_summary(body_text)` carries no prefix (≤ 253), the duplicate-document summaries are fixed-shape (`[DUPLICATE:...] Original document: doc-<32 hex>`), and `apipeline_enqueue_error_documents` receives literal descriptions. This was the only site concatenating a prefix onto unbounded text.

## Why not widen the column

Migrating `content_summary` to `TEXT` is the more thorough fix and remains reasonable later. This PR takes the smaller change: no schema migration, no behavior difference on the unconstrained backends, and the value is a display summary that was already being truncated — 253 vs 235 characters of an error message is not a meaningful loss. The new constant makes the ceiling explicit for future producers either way.

## Verification

New regression test fails on the pre-fix code with the behavioral assertion (`AssertionError: assert 270 <= 255`) on both the synthetic and the real CUDA-OOM input, and passes after.

- `tests/pipeline/` — 684 passed

### DIFF
--- a/lightrag/constants.py
+++ b/lightrag/constants.py
@@ -425,6 +425,14 @@ DUPLICATE_DEMOTION_METADATA_KEYS: tuple[str, ...] = (
 # a stale generated summary while real raw-document summaries are preserved —
 # keep every producer on this constant so the match never drifts.
 FILE_EXTRACTION_SUMMARY_PREFIX = "[File Extraction]"
+# Hard ceiling for a doc_status ``content_summary``. PostgreSQL declares the
+# column as ``varchar(255)``; the other backends are unconstrained, so this is
+# the narrowest storage the value must fit. Producers that BUILD a summary from
+# unbounded text (an error message, a document body) must budget against this
+# rather than against ``get_content_summary``'s own default, or a long enough
+# input makes the whole upsert fail on PostgreSQL — including the FAILED
+# transition itself, which then leaves the document stuck in PARSING.
+DOC_STATUS_CONTENT_SUMMARY_MAX_LENGTH = 255
 
 # Suffixes for parser artifact subdirectories under ``<input>/__parsed__/``.
 # Centralising them here keeps the sidecar writer, engine cache modules and
--- a/lightrag/utils_pipeline.py
+++ b/lightrag/utils_pipeline.py
@@ -28,6 +28,7 @@ from lightrag.base import (
 )
 from lightrag.constants import (
     CUSTOM_CHUNK_PATCH_METADATA_KEY,
+    DOC_STATUS_CONTENT_SUMMARY_MAX_LENGTH,
     DUPLICATE_DEMOTION_METADATA_KEYS,
     FILE_EXTRACTION_SUMMARY_PREFIX,
     FULL_DOCS_FORMAT_LIGHTRAG,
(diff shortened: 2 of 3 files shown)


danielaskdd:
@codex review

Issue #3738, opened 2026-08-28
[Feature Request]: Display sidecar drawing originals in WebUI Retrieval after Query


aump-JUNJUN:
### Do you need to file a feature request?

- [x] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

## Summary

LightRAG indexes sidecar drawings (`im-id`, `drawings.json`, on-disk jpg), however Query / WebUI only return text. We propose enriching query results with `attachments` metadata, an authenticated asset download endpoint, and WebUI rendering **so users can view source figures under retrieval answers**. By default, attachments cover **all retrieved figures**; an **optional** LLM-driven subset mode (`ATTACHMENT_PRESENT_MODE=llm_select`, **off by default**) can follow in a later PR. Image bytes stay out of the LLM prompt(**the optional mode adds a compact text menu only**).

## Background

LightRAG already ships a strong multimodal indexing stack—MinerU / Docling parsing, the sidecar format, `drawings.json` with on-disk originals, stable `im-id`s, VLM analysis, graph + vector retrieval, FastAPI, and the WebUI Retrieval panel.

For engineering drawings and technical documentation, the figure is often the primary source—form, dimensions, and layout are hard to verify from text alone. Query today typically returns prose or pointers like “see Figures 1–4”, but the WebUI cannot render sidecar originals, so users fall back to the PDF. Others have raised the same need ([#149](https://github.com/HKUDS/RAG-Anything/issues/149), [#3405](https://github.com/HKUDS/LightRAG/issues/3405)); nothing in upstream ties retrieval hits back to figure assets yet.

## Current state

LightRAG already handles **drawings** as follows—Indexing persists and structures content; Query retrieves and answers in text:

| Stage | What exists today |
|-------|-------------------|
| **Indexing** | MinerU / Docling emit sidecars (`drawings.json`, jpg assets)<br>Each figure has a stable ID: `im-{doc_hash}-{seq}`; chunks may include `<drawing id="…"/>`<br>VLM analysis fills in `name` and `description`<br>`full_docs.sidecar_location` points at the parsed tree; chunks and the KG can reference `im-id` |
| **Query** | Retrieval returns entities, relations, and chunks (drawing tags or VLM summaries)<br>The LLM composes a Markdown answer from text context<br>The WebUI shows the answer and references |

## Goal

[...]

PR #3739, opened 2026-08-28
feat(sidecar): enrich query results with drawing attachments

aump-JUNJUN:
## Description

Library-only change (PR-1 of 4): after retrieval, collect stable drawing im-ids, join sidecar metadata from drawings.json, and populate **attachments** on query result data. No changes to prompts, LLM output, HTTP routes, or the WebUI.

Both `aquery_data` and `aquery_llm` now expose `data.attachments` alongside existing entities, chunks, and references. Missing sidecars, missing files, or zero hits degrade to an empty list without failing the query.

> PR-4 hooks exist in tree; with ATTACHMENT_PRESENT_MODE=off (default), PR-1 enriches the full retrieved set and stands alone for review.

## Related Issues

Part of #3738

**Stack:** PR-1 → PR-2 → PR-3 → (optional) PR-4  
**Depends on:** none

## Changes Made

| Type | File | Notes |
|------|------|-------|
| **New** | `lightrag/sidecar/query_attachments.py` | Collect IDs, read `drawings.json`, build records; `enrich_raw_data_attachments` (errors → `[]`) |
| **New** | `tests/sidecar/test_query_attachments_*.py`, `conftest_query_attachments.py` | Layered tests with GB/T 70.3 fixture (23 cases) |
| **New** | `scripts/print_attachments.py` | Optional debug helper |
| **Modified** | `lightrag/lightrag.py` (`aquery_data`, `aquery_llm`) | Enrich before return |

**Per-item `attachments` fields:**

| Field | Description |
|-------|-------------|
| `im_id` | Stable ID, e.g. `im-{doc_hash}-{seq}` |
| `doc_id` | Source document |
| `type` | Always `drawing` |
| `format` | Asset format, e.g. `jpg` |
| `path` | Relative path under sidecar (server-side; clients fetch via `doc_id` + `im_id`, not `path`) |
| `title` | From VLM `llm_analyze_result.name` |
| `heading` | Section heading from `drawings.json` (optional) |

**Example:**

```json
"attachments": [{
  "im_id": "im-116da71bde652ed78cf28934d0712aa6-0002",
  "doc_id": "doc-116da71bde652ed78cf28934d0712aa6",
  "type": "drawing",
  "format": "jpg",
  "path": "….blocks.assets/image-ffda008de9ce.jpg",
  "title": "Countersunk hex socket screw head dimensions",
  "heading": "4.1 Form and dimensions"
}]
```

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Additional Notes

**Test plan:**

```bash
python -m pytest tests/sidecar/test_query_attachments_collect.py \
                 tests/sidecar/test_query_attachments_resolve.py \
                 tests/sidecar/test_query_attachments_enrich.py -q
```

| File | # | Focus |
|------|:-:|-------|
| `test_query_attachments_collect.py` | 10 | ID parsing, multi-source collection, dedup, doc mapping |
| `test_query_attachments_resolve.py` | 8 | `drawings.json`, path resolution, title/heading |
| `test_query_attachments_enrich.py` | 5 | Writes to `data.attachments`, empty fallback, idempotency |

Follow-up PRs (same issue): PR-2 HTTP API, PR-3 WebUI, PR-4 optional LLM subset selection.

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -145,6 +145,10 @@ from lightrag.operate import (
     naive_query,
     rebuild_knowledge_from_chunks,
 )
+from lightrag.sidecar.query_attachments import (
+    enrich_raw_data_attachments,
+    resolve_include_im_ids_for_enrich,
+)
 from lightrag.utils_pipeline import (
     CUSTOM_CHUNK_PATCH_METADATA_KEY,
     KG_RECOVERY_WARNINGS_METADATA_KEY,
@@ -160,7 +164,7 @@ from lightrag.utils_pipeline import (
     normalize_document_file_path,
     require_doc_status_record,
 )
-from lightrag.constants import GRAPH_FIELD_SEP, RELATION_NO_EVIDENCE_SOURCE_IDS
+from lightrag.constants import GRAPH_FIELD_SEP
 from lightrag.exceptions import (
     IndexFlushError,
     KGPurgeOperationConflictError,
@@ -391,7 +395,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
 
     # Directory
     # ---
-
+    # Dataclass defaults below; field() attaches per-attribute metadata/docstrings.
     working_dir: str = field(default="./rag_storage")
     """Directory where cache and temporary files are stored."""
 
@@ -1362,6 +1366,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
             )
         self.embedding_token_limit = embedding_max_token_size
 
+
         # Fix global_config now
         global_config = self._build_global_config()
         # Restore original EmbeddingFunc object (asdict converts it to dict)
@@ -1369,6 +1374,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
 
         _print_config = ",\n  ".join([f"{k} = {v}" for k, v in global_config.items()])
         logger.debug(f"LightRAG init with param:\n  {_print_config}\n")
+
 
         # Step 2: Apply priority wrapper decorator to EmbeddingFunc's inner func
         # Create a NEW EmbeddingFunc instance with the wrapped func to avoid mutating the caller's object
@@ -1383,6 +1389,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
             )(self.embedding_func.func)
             # Use dataclasses.replace() to create a new instance, leaving the original unchanged
             self.embedding_func = replace(self.embedding_func, func=wrapped_func)
+
 
         # Initialize all storages
         self.key_string_value_json_storage_cls: type[BaseKVStorage] = get_storage_class(
@@ -1403,6 +1410,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
         self.graph_storage_cls = partial(  # type: ignore
             self.graph_storage_cls, global_config=global_config
         )
+
 
         # Initialize document status storage
         self.doc_status_storage_cls = get_storage_class(self.doc_status_storage)
@@ -1566,6 +1574,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
         self._storages_status = StoragesStatus.CREATED
 
     async def initialize_storages(self):
+        
         """Storage initialization must be called one by one to prevent deadlock"""
         if self._storages_status == StoragesStatus.CREATED:
             # Record the loop the storages (and their shared_storage locks) bind
@@ -1813,7 +1822,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
         Returns:
             str: tracking ID for monitoring processing status
         """
-        # Generate track_id if not provided
+        # Generate track_id if not provided.
         if track_id is None:
             track_id = generate_track_id("insert")
 
@@ -1824,11 +1833,14 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
         # carrier; runtime split args are an ainsert-only concern.
         from lightrag.parser.routing import resolve_chunk_options
 
+        # resolve_chunk_options returns a per-document chunk_options dict snapshot.
         chunk_opts = resolve_chunk_options(
             self.addon_params,
             split_by_character=split_by_character,
             split_by_character_only=split_by_character_only,
         )
+
+        # Enqueue documents; track_id lets callers poll ingest status while processing runs.
         await self.apipeline_enqueue_documents(
             input,
             ids,
@@ -1836,9 +1848,14 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
             track_id,
             chunk_options=chunk_opts,
         )
+
+
+        # Drain the queue through the full ingest pipeline (chunk, embed, graph, etc.).
         await self.apipeline_process_enqueue_documents()
 
         return track_id
+
+
 
     # TODO: deprecated, use insert instead
     def insert_custom_chunks(
@@ -3470,9 +3487,6 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
 
         Entity names and relationship endpoints are normalized with the same
         contract used by document extraction before any storage write begins.
-        A relationship weight is bounded below by its number of distinct real
-        evidence sources. Explicit weights may boost that baseline; omit the
-        relationship ``source_id`` to use a fractional source-less weight.
 
         .. warning:: (issue #3400 Phase 5 — direct-writer audit)
            This path is OUTSIDE the document-level recovery guarantee. It has
@@ -3491,10 +3505,6 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
 
         update_storage = False
         try:
-            from lightrag.utils_graph import (
-                relation_evidence_source_ids,
-                validate_relation_weight,
-            )
 
             def _normalize_custom_kg_entity_name(value: Any, *, field: str) -> str:
                 if not isinstance(value, str):
@@ -3552,17 +3562,6 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
                         f"'{normalized_relationship_data['src_id']}': src_id and "
                         "tgt_id must be different entities"
                     )
-                source_id = relationship_data.get("source_id", "")
-                # This is still a custom-KG alias rather than the graph edge's
-                # persisted source_id. Validate its shape now, but defer the
-                # evidence floor until chunk_to_source_map resolves the alias.
-                relation_evidence_source_ids(source_id)
-                normalized_relationship_data["source_id"] = source_id
-                normalized_relationship_data["weight"] = validate_relation_weight(
-                    relationship_data.get("weight", 1.0),
-                    "",
-                    context=f"Custom KG relationships[{index}]",
-                )
                 normalized_relationships.append(normalized_relationship_data)
 
             # Insert chunks into vector storage
@@ -3577,6 +3576,7 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
                 pipeline busy flag, so leaving the loop here would let it encode
                 concurrently with a document being chunked.
                 """
+                nonlocal update_storage
                 for chunk_data in custom_kg.get("chunks", []):
                     chunk_content = sanitize_text_for_encoding(chunk_data["content"])
                     source_id = chunk_data["source_id"]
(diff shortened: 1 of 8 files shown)


danielaskdd:
Thanks for this, and for the thorough T1–T3 measurement work in #3738 — the design discussion there has converged nicely. Before we can review this PR on its merits, though, the branch needs repair: CI is red, it no longer merges cleanly onto `main`, and the diff carries changes that are unrelated to attachments. Details below, roughly in the order I'd fix them.

## Blocking

### 1. `full_docs_db=` is passed to functions that do not accept it — every query raises `TypeError`

`aquery_llm` now calls:

```python
query_result = await kg_query(..., full_docs_db=self.full_docs)
...
query_result = await naive_query(..., full_docs_db=self.full_docs)
```

but `operate.py` is not in this diff, and neither `kg_query` nor `naive_query` takes that parameter (nor `**kwargs`) — checked on this branch:

```
kg_query    accepts full_docs_db: False | has **kwargs: False
naive_query accepts full_docs_db: False | has **kwargs: False
```

So `aquery_llm` raises `TypeError: kg_query() got an unexpected keyword argument 'full_docs_db'` for every `local` / `global` / `hybrid` / `mix` / `naive` query. This is the whole primary query path, and the PR's tests don't catch it because they exercise the sidecar module in isolation against plain dicts. It looks like a companion `operate.py` change was lost — please either restore it or drop the kwarg. Whatever the fix, please add one test that actually goes through `aquery_llm`, so a missing wiring change can't pass again.

### 2. The diff reverts unrelated committed work (relation weight contract + purge tracking)

The three-dot diff against the merge base (`d403abc`) shows deletions that are genuine branch changes, not a stale base — `main` still carries all of this:

- `lightrag/lightrag.py` `ainsert_custom_kg`: removes the `relation_evidence_source_ids` / `validate_relation_weight` imports, the pre-mutation shape validation, and the second pass that validates the *persisted* `source_id` before any chunk upsert; `weight` reverts to `relationship_data.get("weight", 1.0)`. This drops the [Relation weight contract](https://github.com/HKUDS/LightRAG/blob/main/docs/ProgramingWithCore.md#relation-weight-contract) requirement that public ingress paths validate the complete relation *before the first storage mutation*.
- `_purge_kg_contributions`: removes the `RELATION_NO_EVIDENCE_SOURCE_IDS` filter, so historical no-evidence placeholders get written back into the authoritative `relation_chunks` rows.
- Removes the `NOTE(#3609)` residue comments in the entity/relation purge branches, and the weight-contract paragraphs from the `edit_relation` / `create_relation` / `merge_entities` docstrings. `AGENTS.md` treats an undocumented accepted residue as a defect, and asks that those docstrings stay synchronized with the contract.
- Adds unrelated cosmetic churn to `lightrag.py` (stray blank lines, trailing whitespace at `initialize_storages`, filler comments like `# Generate track_id if not provided.`).

This is almost certainly the same bad merge that lost the `operate.py` change in (1). Please rebase onto current `main` and make sure `lightrag/lightrag.py` contains **only** the attachment enrichment hook — the enrichment itself is ~15 lines, so the file's diff should be tiny.

### 3. CI — Offline Tests: the new conftest aborts collection of the entire suite

```
ERROR tests/sidecar - Failed: Defining 'pytest_plugins' in a non-top-level conftest is no longer supported
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
================ 1 skipped, 1304 deselected, 1 error in 22.21s =================
```

`tests/sidecar/conftest.py` declares `pytest_plugins = ["tests.sidecar.conftest_query_attachments"]`, which modern pytest rejects outside the rootdir conftest — and it takes the whole run down, not just `tests/sidecar`. Simplest fix: delete `tests/sidecar/conftest.py` and move the fixtures directly into `tests/sidecar/conftest.py` (or import the helpers explicitly in each test module) rather than registering a plugin.

### 4. CI — Linting and Formatting

`ruff format` wants 5 files:

```
lightrag/lightrag.py
lightrag/sidecar/query_attachments.py
tests/sidecar/conftest_query_attachments.py
tests/sidecar/test_query_attachments_collect.py
tests/sidecar/test_query_attachments_enrich.py
```

There's also a dead statement left after `return clean` in `apply_present_drawings_to_raw_data`. `pre-commit run --all-files` clears these. (`ruff check` itself passes.)

### 5. Merge conflict with `main`

`git merge-tree origin/main <branch>` conflicts in `lightrag/lightrag.py`. Resolving (2) by rebasing onto current `main` should take care of this too.

## Test coverage — 11 of the 23 cases can never run

Running the three test files on this branch gives **12 passed, 11 skipped**, all skips from one cause:

```
SKIPPED - GB/T sidecar fixture missing: <…>/inputs/__parsed__/MinerU_markdown_GB_T_70.3-2023_3374_2089563653659709440.md.parsed
```

`conftest_query_attachments.py` resolves the fixture as `Path(__file__).resolve().parents[3]` — that is the directory *containing* the repository (the comment says `contributor/`), so the path depends on your local machine layout and can never resolve on CI or for any other contributor. The effect is that the entire resolve layer — `drawings.json` parsing, path resolution, `title` / `heading` extraction, `resolve_drawing_asset_file` — plus 4 of the 5 enrich cases are untested everywhere it matters. Please build a small synthetic sidecar tree in a `tmp_path` fixture (a minimal `drawings.json` plus a few-byte jpg) so the golden cases run unconditionally. That also matches the repo's rule of not depending on external artifacts in unit tests.

## Scope — this is not the PR-1 the issue describes

The description says "library-only change (PR-1 of 4) … no changes to prompts, LLM output, HTTP routes, or the WebUI", but `query_attachments.py` (883 lines) is mostly PR-2 and PR-4 code:

| Concern | Symbols |
|---|---|
| PR-1 (in scope) | `collect_im_ids*`, `build_im_to_doc_map`, `load_drawings_index`, `resolve_single_drawing`, `resolve_drawing_attachments`, `enrich_raw_data_attachments` |
| PR-4 (LLM select) | `build_drawing_candidates`, `format_drawing_menu_for_prompt`, `append_drawing_menu_to_sys_prompt`, `parse_present_drawings`, `apply_present_drawings_to_raw_data`, `filter_im_ids_by_whitelist`, `PresentDrawingsStreamWrapper`, `wrap_present_drawings_stream`, `get_attachment_present_mode`, `is_llm_select_present_mode`, `resolve_include_im_ids_for_enrich` |
| PR-2 (HTTP) | `resolve_asset_path`, `_path_within_root`, `media_type_for_drawing_format`, `resolve_drawing_asset_file` |

Grepping the tree, none of the PR-4 or PR-2 helpers has a caller outside this module (`resolve_drawing_asset_file` is referenced only by an always-skipped test). `resolve_include_im_ids_for_enrich` is called from `lightrag.py` but always returns `None`, since the mode defaults to off. So roughly two thirds of the new module is dead code shipped ahead of the PRs that would use it, including a prompt-appending helper in a PR that states it does not touch prompts.

I'd much rather review the small version: keep collect + resolve + enrich, and move the menu / `<present_drawings>` / stream-wrapper code to PR-4 and the asset-path / media-type code to PR-2, where each arrives with its caller and its tests. It also keeps the schema-freeze promise honest — PR-2 and PR-3 are supposed to consume nothing but `data.attachments`.

Two related notes:

- `ATTACHMENT_PRESENT_MODE` and `ATTACHMENT_CANDIDATE_TOP_K` are read via `get_env_value` but appear nowhere in `env.example` or the docs, and `attachment_present_mode` is not a `LightRAG` field, so the `global_config` branch is unreachable. Whichever PR actually uses them should introduce and document them.
- `collect_im_ids_from_entities` scans KG entity names — the approach your 2026-09-06 comment says you have already replaced ("we no longer scan KG entity names alone", with `figure_ids` + Figure List in PR-5). Worth deciding now whether PR-1 should land the superseded collection path at all, or whether the `figure_chunk` / `text_chunk` attribution from PR-5 belongs here instead.

## Smaller points, for whenever the branch is rebuilt

1. **`enrich_raw_data_attachments` does not honor its own contract.** The docstring says "Never raises — failures degrade to an empty attachment list", and the acceptance criterion in #3738 is "missing sidecar or missing files → `attachments: []`; the query still succeeds". But there is no `try`/`except` in the function: `text_chunks_db.get_by_ids` and `full_docs_db.get_by_id` propagate any backend error straight out of `aquery_data` / `aquery_llm`. Only `load_drawings_index` and `_path_within_root` are guarded. Note that a blanket `except Exception` isn't the answer either — `AGENTS.md` is explicit that nothing here licenses silent failure. My suggestion: catch narrowly around the storage reads, log at warning, degrade to `[]`, and say so in the docstring — then test it with a KV stub that raises.
2. **`_drawings_index_cache` is process-global, unbounded, and never invalidated.** It also caches negative results, so a query issued while a document is still parsing pins `{}` for that sidecar URI for the life of the process, and re-parsing a document to the same URI keeps serving the old index. `clear_drawings_index_cache()` exists but only tests call it. Either key on something that changes when the sidecar is rewritten, bound the size, or drop the cache and let the OS page cache do the work.
3. **Sync file I/O on the async path.** `load_drawings_index` does a blocking `read_text()` inside `resolve_drawing_attachments`, once per retrieved document. Worth pushing through the existing executor helper.
4. **`aquery_data` and `aquery_llm` diverge.** `aquery_data` calls `enrich_raw_data_attachments` without `include_im_ids`, so it always full-enriches even under `llm_select`, while `aquery_llm` respects the mode. Also, `aquery_data` only enriches `if query_result.raw_data`, so an empty result has no `attachments` key at all — clients then have to handle both "absent" and "empty list".
5. **`collect_im_ids` runs and is thrown away** when `include_im_ids` is not `None`.
6. **`scripts/print_attachments.py`** hardcodes Windows paths (`.venv\Scripts\...`), a Chinese sample query, and `--asset-urls` against HTTP endpoints that don't exist until PR-2; it also pulls in `requests` and `python-dotenv`. This reads like a personal debug script rather than repo tooling — please drop it, or move it to PR-2 as something portable.

## Summary

The core idea is right and the shape of `data.attachments` matches what #3738 specifies, so this is very fixable. What I need before a real review:

- [ ] Rebase onto current `main`; drop every unrelated revert and cosmetic change so `lightrag/lightrag.py` shows only the enrichment hook
- [ ] Restore or remove the `full_docs_db` wiring, with a test that goes through `aquery_llm`
- [ ] Fix the `pytest_plugins` conftest so collection isn't aborted; run `pre-commit run --all-files`
- [ ] Make the sidecar fixtures synthetic so all cases run in CI
- [ ] Narrow the PR to PR-1, moving PR-2 / PR-4 helpers to their own PRs

Please also run the mirrored subset locally before pushing (`./scripts/test.sh tests/sidecar`, plus `tests/pipeline` given the `lightrag.py` touch) — CI runs the full suite, but the collection abort above would have shown up immediately.

On the open question from #3738 — the trigger for *automatic* fallback when the model emits no `<present_drawings>` block — I'll reply in the issue thread rather than here, since it's a PR-4/PR-5 concern and shouldn't hold up PR-1.

---

danielaskdd:
Closing this one — not because the feature is unwanted, but because this branch has drifted too far from the PR-1 it describes to be repaired in place. To be clear about the reasoning:

**The branch state is not recoverable by review iteration.** Beyond the red CI and the merge conflict, the diff reverts committed work that has nothing to do with attachments — the relation weight validation in `ainsert_custom_kg`, the `RELATION_NO_EVIDENCE_SOURCE_IDS` filter in `_purge_kg_contributions`, the `NOTE(#3609)` residue comments, and the weight-contract paragraphs in three docstrings. In the same event, the `operate.py` change that would have accepted `full_docs_db=` was lost, which is why every `local`/`global`/`hybrid`/`mix`/`naive` query on this branch raises `TypeError`. Untangling a bad merge through review comments costs both of us more than rebuilding the commit on top of current `main` will.

**The scope is the larger issue.** #3738 defines PR-1 as one thing: resolve `im-id` → join sidecar → populate `data.attachments`. Roughly two thirds of the 883-line module here is PR-2 and PR-4 code with no caller anywhere in the tree — the candidate menu, `<present_drawings>` parsing, the stream wrapper, the asset-path and media-type helpers, including a function that appends to the system prompt in a PR whose description says it does not touch prompts. Dead code can't be reviewed meaningfully: there's no call site to check it against and no test that exercises it. And with 11 of the 23 test cases skipping unconditionally (the fixture path resolves outside the repository), even the in-scope third isn't covered where CI can see it.

**What I'd like instead:** a fresh PR-1 branched from current `main`, containing only `collect_im_ids*` / `build_im_to_doc_map` / `load_drawings_index` / `resolve_single_drawing` / `resolve_drawing_attachments` / `enrich_raw_data_attachments`, the ~15-line hook in `lightrag.py`, and synthetic `tmp_path` fixtures so every case runs in CI. That should be a few hundred lines and genuinely quick to review and merge. The PR-4 and PR-2 helpers are worth keeping — please hold them for the PRs that introduce their callers.

One design question to settle before you rebuild: `collect_im_ids_from_entities` scans KG entity names, which your 2026-09-06 comment on #3738 says you've already moved away from ("we no longer scan KG entity names alone", in favour of `figure_ids` + the Figure List). It may be that the `figure_chunk` / `text_chunk` attribution you built for PR-5 belongs in PR-1 from the start, rather than landing a collection path you plan to replace. Worth deciding in the issue thread first — it changes what PR-1 should contain.

The full detail is in [my earlier review](https://github.com/HKUDS/LightRAG/pull/3739#issuecomment-5582812628), including the smaller points (the `enrich_raw_data_attachments` "never raises" contract that has no `try`/`except`, the unbounded `_drawings_index_cache` that caches negative results, blocking file I/O on the async path, the `aquery_data`/`aquery_llm` divergence) — those are all still worth applying to the new branch.

Thank you for the T1–T3 measurement work; that was real effort and it moved the design forward more than any of the code did. The `data.attachments` shape you landed on is right, and I'd like to merge it. Reopen this or open a new PR whenever the narrow version is ready, and I'll review promptly. Tracking stays on #3738.

---

PR #3740, opened 2026-08-28
fix: move file-backend persistence off the event loop

danielaskdd:
## Summary

The file-backed storage backends rewrite their whole file on every commit, **synchronously, on the event loop**. This moves those writes to a dedicated single-worker thread pool so the loop keeps serving HTTP while a commit is in flight.

## Motivation

Reported symptom: with a large corpus, the WebUI freezes while a batch of failed documents is repeatedly retried, with the backend log sitting on `[purge] doc-...: deleted 369 chunk(s) from storage`.

The mechanism:

- `index_done_callback` serializes the entire store inline in an `async def` — `write_json` for the JSON KV / doc-status backends, `nx.write_graphml` for NetworkX, `NanoVectorDB.save` for nano.
- `_insert_done` flushes **all 12 storages** (`_index_storages`), including `llm_response_cache` and the GraphML file — usually the two largest files in `rag_storage`.
- A single document purge calls `_insert_done` **twice**, plus one `doc_status` flush for each of the four purge-journal phases. Frequent pipeline restarts make those windows dense.
- `lightrag-server` forces `workers=1` in uvicorn mode, so a blocked loop takes every endpoint down with it: `/health`, `/documents/pipeline_status` and `/documents/paginated` all hang together.

Offloading is effective here because none of these encoders holds the GIL monolithically: `write_json` uses `json.dump(fp)` with `indent=2` (the C one-shot encoder is excluded both by `indent` and by not being `_one_shot`), and `write_graphml_lxml` streams through a pure-Python per-element loop.

## What's changed

**`lightrag/utils.py`** — `bounded_submit`'s body moves to a private `_bounded_submit_impl` that takes a `wait_for_completion` flag. `bounded_submit`'s public signature and its "all `**kwargs` are forwarded verbatim to `fn`" contract are unchanged: a control keyword on that signature would either be positional-only (before the `/`, where passing it by keyword silently forwards it to `fn`) or steal a legitimate argument of `fn` (after `*args`). Adds `get_storage_io_executor()` (one worker, `lightrag-storage-io`) and `run_in_storage_io()`.

The uncancellable region starts at `executor.submit()`, **not** at the permit wait. A caller still queued for a permit has nothing in flight, so it stays cancellable — otherwise a cancelled flush would hold its namespace lock for the whole queue ahead of it and then still write to disk after the cancellation.

**`json_kv_impl` / `json_doc_status_impl`** — only the write moves. The `self._data.copy()` snapshot stays on the loop inside the lock, so the worker reads a private dict and the Manager RPC count is unchanged. The sanitize-reload branch also stays on the loop: it writes back into the shared dict.

**`nano_vector_db_impl`** — `_save_to_disk_locked` becomes a coroutine. The lock is held for the whole call because `_save_atomic` points `client.storage_file` at the tmp sibling for the duration of the save.

**`networkx_impl`** — the commit hands `self._graph` to a worker thread that iterates `graph._node` / `graph._adj`, which invariant (3) in the class docstring no longer covers. A new `_commit_gate` (`asyncio.Event`, cleared only while this process serializes) restores the exclusion.

Holding `_storage_lock` across the write is **not** sufficient by itself: releasing a `NamespaceLock` runs the release on a fresh task and awaits a shield, so `__aexit__` always suspends at least once, and a mutator suspended there is already past the lock body. The gate is therefore checked **after** `_get_graph`'s lock block — after the last suspension point — so from `is_set()` through the caller's synchronous mutation there is no `await`. It sits at `_get_graph`, the single choke point, rather than in the seven mutators.

Both classes' concurrency-invariant docstrings are rewritten; the NetworkX one explicitly named this change as what would break it.

## What this does and does not fix

Every callback takes its own `_storage_lock` **before** entering the single-worker queue, so a namespace can be holding its lock while queued.

- `/health` and `/documents/pipeline_status` keep responding during a commit — they depend on no committing namespace's lock.
- `/documents/paginated` is no longer frozen by unrelated namespaces (GraphML, LLM cache) writing on the loop. It still waits on the doc-status lock when a purge journal, status transition or upload commits doc-status itself. That short wait is accepted behaviour.
- Graph query endpoints can still wait while a GraphML commit is queued and writing.
- **nano is a partial win.** `NanoVectorDB.save()` base64-encodes the whole matrix via `tobytes()` + `b64encode()` — two single C calls that hold the GIL wherever they run — so a large matrix still stalls the loop for that stretch. Only the `json.dump` that follows is cooperative. A chunked or binary format is a separate change.

Not in this PR: faiss (same shape, but needs an audit of its lock-free `_id_to_meta` reads and a measurement of whether `faiss.write_index` releases the GIL); the read/parse paths (`_reload_client_from_disk_locked`, `_get_graph`'s reload, `initialize()`'s `load_json`), which only ever run in reader processes; and reducing the number of purge flushes, which touches the #3400 fail-closed ordering contract.

## Tests

New:

- `tests/utils/test_storage_io_offload.py` — cancellation waits for the submitted worker; survives repeated cancellation (each cancel proven to land in a distinct `asyncio.shield` round via a spy, since sleeping cannot prove the scheduler did not collapse them); cancel + IO failure prefers `CancelledError` while still logging the failure and producing no "never retrieved" callback; a caller waiting for a permit cancels **without** submitting; `bounded_submit` still forwards a `wait_for_completion` keyword to `fn`.
- `tests/kg/json_impl/test_json_write_off_event_loop.py`, `tests/kg/nano_impl/test_nano_save_off_event_loop.py`, `tests/kg/networkx_impl/test_networkx_commit_gate.py` — a heartbeat coroutine samples the loop from inside a blocking write. Against the pre-change code these fail with `assert 0 > 0`: the loop cannot advance at all. Also: a concurrent `upsert` stays pending while a commit is in flight (proving the offload stayed inside the lock), and a cancelled nano commit leaves `storage_file` restored.
- The gate tests are vacuous against the pre-change code (no thread, no window). What they pin is the intermediate mistake of offloading **without** the gate; removing `gate.clear()` / `finally: gate.set()` was verified by hand to fail two of them, and the record is in the module docstring.

Converting `_save_to_disk_locked` to a coroutine silently degrades synchronous test doubles — the caller awaits `None`, so the fault-injection tests would pass on a `TypeError` instead of the `OSError` they are about. All 25 references across the two nano test modules were converted (five were lambdas, expanded to `async def`). One counting double called the coroutine without awaiting it, so the save silently did not run while the test stayed green.

Full suite: **7162 passed, 200 skipped** (`./scripts/test.sh tests`), run because `lightrag/utils.py` is cross-cutting.

### DIFF
--- a/lightrag/constants.py
+++ b/lightrag/constants.py
@@ -615,7 +615,8 @@ MAX_RESPONSE_TYPE_CHARS = 256
 MAX_QUERY_TOP_K = 1000
 MAX_QUERY_TOKEN_BUDGET = 1_000_000
 
-# Submission ceilings for the two single-worker CPU pools (tokenizer, chunking).
+# Submission ceilings for the three single-worker pools (tokenizer, chunking,
+# storage IO).
 # A ``ThreadPoolExecutor`` wait queue is unbounded, and freeing the event loop
 # means more requests can be in flight at once, so submissions need their own
 # limit. Deliberately fixed process-wide constants rather than anything derived
@@ -628,6 +629,13 @@ MAX_QUERY_TOKEN_BUDGET = 1_000_000
 # not refusal.
 TOKENIZER_SUBMIT_LIMIT = 8
 CHUNKING_SUBMIT_LIMIT = 8
+# Storage IO is not CPU-bound like the other two, but it needs the same ceiling
+# for a sharper reason: ``_flush_storages`` gathers ``index_done_callback`` over
+# every storage at once (a dozen for a default deployment), and a purge issues
+# several such flushes per document. Each waiter here holds its namespace lock,
+# so the ceiling also bounds how many namespaces can be locked waiting for one
+# worker.
+STORAGE_IO_SUBMIT_LIMIT = 8
 
 # Per-workspace ceiling on manual retry requests that have been published but
 # not yet ACKed (LR2 §10.1). The channel is sticky — a request survives until an
--- a/lightrag/kg/faiss_impl.py
+++ b/lightrag/kg/faiss_impl.py
@@ -3,13 +3,18 @@ import os
 import time
 import asyncio
 from hashlib import md5
-from typing import Any, final
+from typing import Any, Awaitable, Callable, final
 import json
 import numpy as np
 from dataclasses import dataclass
 
 from lightrag.file_atomic import atomic_write, reap_orphan_tmp_files
-from lightrag.utils import logger, compute_mdhash_id, validate_workspace
+from lightrag.utils import (
+    logger,
+    compute_mdhash_id,
+    commit_in_storage_io,
+    validate_workspace,
+)
 from lightrag.base import BaseVectorStorage
 from lightrag.constants import DEFAULT_QUERY_PRIORITY
 
@@ -91,13 +96,22 @@ class FaissVectorDBStorage(BaseVectorStorage):
            ``index_done_callback`` completes. Reads in the gap between a
            writer's in-memory mutation and its commit may legitimately
            return the pre-update snapshot.
-        3. **Faiss + dict mutations are synchronous.** Under a
-           single-threaded asyncio event loop, ``index.add`` /
-           ``index.search`` / ``self._id_to_meta`` mutations cannot be
-           preempted by another coroutine, which gives them implicit
-           mutual exclusion. This is why most methods don't hold
-           ``_storage_lock`` while touching ``self._index`` /
-           ``self._id_to_meta``.
+        3. **Faiss + dict MUTATIONS are synchronous and stay on the
+           event loop.** ``index.add`` / ``index.remove_ids`` /
+           ``self._id_to_meta`` mutations cannot be preempted by another
+           coroutine, which gives them implicit mutual exclusion. This is
+           why most methods don't hold ``_storage_lock`` while touching
+           ``self._index`` / ``self._id_to_meta``.
+
+           ``_save_faiss_index`` is the one part that runs in a worker
+           thread, and it is compatible with this invariant because it
+           only READS: it takes its ``self._id_to_meta`` snapshot on the
+           loop before offloading, and it reads ``self._index`` while
+           holding ``_storage_lock``, which excludes every mutation
+           above. The only unlocked Faiss access, ``query``'s
+           ``index.search``, is a read as well. Moving a MUTATION (or an
+           unsnapshotted dict iteration) into the pool would break the
+           invariant and would require widening the lock scope instead.
 
     Cross-process sync protocol:
         Writer side (``index_done_callback``):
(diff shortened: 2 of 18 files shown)


danielaskdd:
@codex review

danielaskdd:
Thanks — the finding is correct, and it's now fixed in `bedf632`.

One refinement to the diagnosis: this is **pre-existing**, not introduced by the offload. With the loop frozen for the whole synchronous write, cancellation could only be delivered at the first suspension point *after* it — and in multiprocess mode that is the lock acquire inside `set_all_update_flags`, before any flag is flipped. So "file written, zero flags set" was reachable on `main` for the entire duration of a write too. What this branch changes is the delivery point and the absolute frequency (the loop is alive, so more can happen during a write). It also happens to be the branch that builds the uncancellable region needed to fix it properly, so it's closed here rather than left to grow.

**Fix**: `commit_in_storage_io(fn, on_committed)` runs the bookkeeping inside the same uncancellable region as the write, under three rules:

- write succeeded → the hook runs to completion, and only then is a deferred cancellation re-raised;
- `fn` raised → the hook does **not** run; the write did not land, so its bookkeeping must not either;
- cancelled while waiting for a permit → the hook does **not** run, because nothing was submitted.

That last rule is load-bearing rather than tidiness, and it is why the fix is not simply "run the bookkeeping unconditionally": Nano's hook retires the redo logs, so running it without a write would discard rows that were never persisted — a worse bug than the one being fixed.

The deferring wait is applied **twice** — to the worker future and again to the hook. The second application is not decoration: a cancel arriving after the first was absorbed would land on the `await` inside `set_all_update_flags`, moving the same gap one layer down.

**Scope is narrower than the comment suggests.** Only the backends with cross-process notification need this: Nano and NetworkX here, Faiss in #3742 (pushed). The two JSON backends are deliberately untouched — their commit has no `set_all_update_flags` at all, because multiprocess mode shares `self._data` as a `DictProxy` and there is nothing to tell other processes to reload. The `clear_all_update_flags` that does follow their write fails safe: skipping it costs one redundant write next time.

**Tests**: 5 new helper-layer cases (hook completes before the deferred cancel; never runs when the permit wait is cancelled / the write raises / `wait_for_completion` is off; its own failure surfaces) plus one per affected backend asserting a cancelled commit still flagged the namespace. Each backend test fails against the previous shape — offload the write, inline the bookkeeping after it — verified by hand.

danielaskdd:
@codex review

danielaskdd:
Both correct — fixed in `b1d2a6c`. They differ in provenance, which is worth recording:

### 1. JSON sanitize reconciliation — a **real regression from this branch**

Unlike the earlier bookkeeping finding, this one was genuinely introduced here. Before the offload, the write and the `needs_reload` read-back were **one unpreemptable synchronous block** — cancellation could only be delivered at the `clear_all_update_flags` await that follows, never between them. Offloading the write put a suspension point in the middle, so a cancel there discards `needs_reload`: the sanitized file is published while the shared dict keeps the rows that failed to encode.

Both JSON backends now run the reconciliation through `commit_in_storage_io`, restoring the original all-or-nothing property. The read-back itself stays on the loop (it writes into the shared dict, and only runs for payloads that fail to encode).

Worth noting this is a *different* mechanism from the one that made the JSON backends exempt from the previous finding. That exemption was about `set_all_update_flags`, which their commit genuinely never calls; the `needs_reload` branch is theirs alone and does need the guarantee. Thanks for catching that the exemption did not generalize.

### 2. NetworkX commit gate bound to a closed loop

Confirmed by measurement rather than inference — `asyncio.Event` binds to a loop only when a `wait()` **actually suspends**:

```
after construct:                  _loop = None
after set():                      _loop = None
after is_set():                   _loop = None
after wait() on an already-SET event: _loop = None
after a wait() that SUSPENDED:    _loop = <_UnixSelectorEventLoop ...>
then reused from a new loop:      RuntimeError: <Event> is bound to a different event loop
```

And the reuse path you point at is real: `_run_sync` deliberately lets calls through on a fresh loop once the original `owning_loop` has closed (documented there as the common `rag = asyncio.run(initialize_rag())` shape). So the sequence you describe — bind during one commit, close that loop, overlap a commit with a graph op on the next one — does raise.

`_gate()` now tracks the loop it built the Event on and rebuilds when the running loop differs. Rebinding is safe rather than racy: a commit in flight holds `_storage_lock` and runs on the loop being replaced, so reaching `_gate()` from a different loop means no commit of ours is outstanding and an open gate is the correct state. Lazy rebuild rather than eager re-creation in `initialize()`, since that is not where the loop change shows up.

### Tests

4 new cases for the JSON backends (parametrized over KV and doc-status: a cancelled commit still reconciles; a failed write reconciles nothing) and one for the gate, which uses two real `asyncio.run` calls because the failure only appears on the second loop. Each fails against the shape it guards — verified by reverting each fix in turn.

#3742 has been rebased on top; Faiss needs neither fix (no `needs_reload` semantics, no gate).

danielaskdd:
@codex review

PR #3742, opened 2026-08-28
fix(kg): save FaissVectorDBStorage off the event loop

danielaskdd:
## Summary

Moves `FaissVectorDBStorage._save_faiss_index` into the storage-IO worker pool, so committing the Faiss index no longer freezes the event loop.

**Stacked on #3740** — base is `fix/storage-io-off-event-loop`, which adds the `run_in_storage_io` / `commit_in_storage_io` pool this PR consumes. Review/merge that one first; the diff here is only the Faiss backend.

## Motivation

#3740 moved the JSON, NanoVectorDB and NetworkX backends off the loop and deferred Faiss, because two premises needed checking first: whether `faiss.write_index` releases the GIL (a C call that holds it would make the offload pointless), and whether the unlocked `_id_to_meta` read paths tolerate a worker thread. Both check out.

Measured on faiss 1.14.2, 60k × 512 `IndexFlatIP` (123 MB index, 15 MB meta) — a worker thread does the work while a busy main thread counts how often it gets scheduled:

| operation | duration | main-thread ticks/s |
|---|---|---|
| `faiss.write_index` | 0.05s | 1,222,265 |
| `json.dump` (meta.json) | 0.25s | 159,890 |
| control: pure-Python loop | 0.14s | 1,544 |

`write_index` releases the GIL outright; `json.dump` is the same cooperative pure-Python encoder the JSON backends use. Neither half holds the GIL the way a monolithic C call would, so unlike Nano's base64 step there is no residual stall here. Note the metadata file, not the index, is the slower half.

## What's changed

- `_save_faiss_index` becomes a coroutine and submits both `atomic_write` calls as **one** unit of work to the pool. One submission, not two: two would take two permits and could interleave another namespace's commit between the halves. The write order (index first, then meta) is preserved — it is what keeps the crash skew in the direction `_load_faiss_index` tolerates.
- The `_id_to_meta` snapshot is built **on the loop, before the offload**. The synchronous `client_storage` property reads that dict without the lock, so the worker must serialize a snapshot rather than the live mapping.
- `self._index` is handed over by reference. Sound because the worker only READS it while `_storage_lock` — held across the whole call — excludes every mutation (`add` / `remove_ids` / reload), and the one unlocked Faiss access, `query`'s `index.search`, is a read as well.
- Post-write bookkeeping (retire the redo logs, flag the other processes, clear the dirty bit) goes through `commit_in_storage_io`, so it cannot be separated from a write that already landed. This mirrors the fix in #3740 for the Codex finding there: inlined after the offload, a cancel would leave both files published with the other processes never told to reload them.
- Class docstring invariant 3 named this exact change as its breaking condition ("if either premise is broken (e.g. Faiss calls moved to a thread pool), the lock scope must be widened"). Rewritten to state what actually holds now: mutations stay on the loop, only the read-only save is offloaded.
- Every test double for `_save_faiss_index` became `async def` and takes the commit hook. A synchronous double wrapping the original would call the new coroutine without awaiting it — the save would silently not happen, and the fault-injection tests would degrade from asserting `OSError` to passing on a `TypeError`.

## What could break

The offload runs inside `_storage_lock`, so callers that go through `_get_index` still wait for the commit — this frees the event loop, not the Faiss namespace lock. Queries against *this* vector namespace can still queue behind a commit; every other endpoint keeps being served.

Uncancellable once submitted, matching the synchronous write it replaces (which could not be cancelled at all). Cancelling the caller mid-write defers `CancelledError` until both files have landed **and** the bookkeeping has run, so the lock is never released with a half-written pair or an unnotified namespace behind it.

## Tests

4 new cases in `tests/kg/faiss_impl/test_faiss_write_off_event_loop.py`:
- the loop keeps advancing while the index is written;
- a row added to `_id_to_meta` mid-write does not reach the file (proves the snapshot precedes the offload);
- a cancelled caller still waits for both files, leaving no tmp residue;
- a cancelled commit still flagged the namespace for the other processes.

Each fails against the shape it guards — verified by hand.

Ran `tests/kg/faiss_impl` + `tests/utils`: **294 passed**. Full suite green locally (CI does not trigger on a non-`main` base).

### DIFF
--- a/lightrag/kg/faiss_impl.py
+++ b/lightrag/kg/faiss_impl.py
@@ -3,13 +3,18 @@ import os
 import time
 import asyncio
 from hashlib import md5
-from typing import Any, final
+from typing import Any, Awaitable, Callable, final
 import json
 import numpy as np
 from dataclasses import dataclass
 
 from lightrag.file_atomic import atomic_write, reap_orphan_tmp_files
-from lightrag.utils import logger, compute_mdhash_id, validate_workspace
+from lightrag.utils import (
+    logger,
+    compute_mdhash_id,
+    commit_in_storage_io,
+    validate_workspace,
+)
 from lightrag.base import BaseVectorStorage
 from lightrag.constants import DEFAULT_QUERY_PRIORITY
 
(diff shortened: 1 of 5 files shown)


danielaskdd:
@codex review

PR #3745, opened 2026-08-29
fix: accept valid empty JSON extraction results in cache rebuild

### LINKED ISSUE #3744: [Bug]: Valid empty JSON extraction cache entries emit misleading delimiter warnings during rebuild

### Describe the bug

During extraction-cache rebuild, LightRAG emits a large volume of:

```text
WARNING: doc-...-chunk-...: Complete delimiter can not be found in extraction result
```

Cached extraction results are expected to be JSON in JSON extraction mode. A valid response may legitimately contain empty `entities` and `relationships` arrays when no reliable facts are present in a chunk.

However, the cache rebuild path treats this valid empty result as if JSON parsing had failed, then invokes the legacy delimiter parser. Since JSON responses do not contain `<|COMPLETE|>`, that fallback emits a misleading delimiter warning.

The warning therefore does not, by itself, indicate malformed JSON, truncation, or lost extraction data. Those cases should be reported separately when JSON parsing or schema validation actually fails.

### Related work

This is distinct from #3601 / PR #3607, which address genuine token-limit truncation and empty responses caused by `done_reason=length`. This report concerns syntactically valid JSON with the expected empty arrays during extraction-cache rebuild.

### Steps to reproduce

1. Use JSON-mode extraction with `ENTITY_EXTRACTION_USE_JSON=true`.
2. Ingest a chunk for which the model returns valid empty JSON:
   ```json
   {"entities": [], "relationships": []}
   ```
3. Trigger an extraction-cache rebuild.
4. Observe the legacy delimiter warning for that chunk.

### Expected Behavior

A syntactically valid JSON extraction result with the required `entities` and `relationships` arrays should be accepted as JSON even when both arrays are empty. It should return empty node/edge dictionaries without invoking the delimiter parser.

The delimiter fallback should be used only when the response is not valid JSON, or cannot be repaired into a valid JSON extraction shape.

### LightRAG Config Used

```env
ENTITY_EXTRACTION_USE_JSON=true
LLM_BINDING=ollama
```

LightRAG v1.5.7rc2 with a PostgreSQL-backed LLM cache.

### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

AmirF194:
`_rebuild_from_extraction_result` (lightrag/operate.py) only accepts a JSON-parsed extraction result when `nodes or edges` is truthy. A syntactically valid extraction that legitimately found nothing, `{"entities": [], "relationships": []}`, yields two empty (falsy) dicts, so the code falls through to the delimiter-based parser on the same JSON string. That string was never going to contain the `<|COMPLETE|>` marker, so the fallback logs a warning that reads like the extraction failed when it actually succeeded.

`tolerant_load_json_dict`, which the JSON branch already calls internally, documents a non-empty dict as its own signal that parsing succeeded, independent of what the extracted entities/relationships turn out to be. This PR gates the fallback on that signal instead of on the emptiness of the results, so a genuinely empty extraction is accepted and a genuinely unparseable one still falls back exactly as before.

Scope: this path is only reachable from cache rebuild. The live extraction path (`use_json_extraction` branch a few hundred lines up) picks a parser once by mode and never falls back between them, so it is unaffected.

Verification:
- New test `tests/extraction/test_rebuild_json_empty_fallback.py`: fails on `main` with the exact misleading warning, passes on this branch; also pins that a non-empty JSON result and a genuinely unparseable one keep behaving as before.
- Full offline suite (`pytest tests/ -m offline`): 5991 passed, 45 skipped (pre-existing spaCy model gaps), 0 failed.
- `pre-commit run --files lightrag/operate.py tests/extraction/test_rebuild_json_empty_fallback.py`: clean.
- Not verified: the live (non-cached) extraction path, since it does not exercise this fallback at all.

Fixes #3744

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -1665,8 +1665,9 @@ async def _rebuild_from_extraction_result(
             timestamp,
             file_path,
         )
-        # If JSON parsing yielded results, use them
-        if nodes or edges:
+        # A successful parse (tolerant_load_json_dict's own non-empty-dict
+        # signal) is accepted even when it legitimately yields no nodes/edges.
+        if nodes or edges or tolerant_load_json_dict(extraction_result):
             return nodes, edges
         # Otherwise fall through to text-based parsing
 
--- a/tests/extraction/test_rebuild_json_empty_fallback.py
+++ b/tests/extraction/test_rebuild_json_empty_fallback.py
@@ -1,0 +1,76 @@
+"""A cached extraction result that is valid, empty JSON must not fall through
+to the delimiter parser: that parser has no chance of finding a
+"<|COMPLETE|>" marker in a JSON payload and would log a misleading warning
+about it, even though the JSON was parsed successfully and simply carried no
+entities or relationships.
+"""
+
+from __future__ import annotations
+
+import logging
+
+import pytest
+
+from lightrag.operate import _rebuild_from_extraction_result
+
+pytestmark = pytest.mark.offline
+
+
+class DummyKV:
+    async def get_by_id(self, key):
+        return {"file_path": "doc.pdf"}
+
+
+@pytest.fixture
+def _propagate_lightrag_logger(monkeypatch):
+    """``lightrag.utils.logger`` sets ``propagate = False``; restore
+    propagation locally so ``caplog`` can capture WARNING records."""
+    monkeypatch.setattr(logging.getLogger("lightrag"), "propagate", True)
+
+
+@pytest.mark.asyncio
+async def test_rebuild_accepts_valid_empty_json_without_delimiter_warning(
+    caplog, _propagate_lightrag_logger
+) -> None:
+    with caplog.at_level(logging.WARNING, logger="lightrag"):
+        nodes, edges = await _rebuild_from_extraction_result(
+            DummyKV(),
+            '{"entities": [], "relationships": []}',
+            chunk_id="chunk-1",
+            timestamp=1,
+        )
+    assert nodes == {}
+    assert edges == {}
+    assert "Complete delimiter can not be found" not in caplog.text
+
+
+@pytest.mark.asyncio
+async def test_rebuild_still_returns_entities_from_nonempty_json() -> None:
+    nodes, edges = await _rebuild_from_extraction_result(
+        DummyKV(),
+        '{"entities":[{"name":"Alice","type":"person",'
+        '"description":"an engineer"}],"relationships":[]}',
+        chunk_id="chunk-2",
+        timestamp=1,
+    )
+    assert "Alice" in nodes
+
+
+@pytest.mark.asyncio
+async def test_rebuild_still_falls_back_on_unparseable_json_looking_text(
+    caplog, _propagate_lightrag_logger
+) -> None:
+    """Text that starts with '{' but is not recoverable JSON, and also has no
+    completion marker, is the pre-existing garbage-input case: falling
+    through to the delimiter parser (and its warning) is correct there,
+    unlike the valid-empty-JSON case above."""
+    with caplog.at_level(logging.WARNING, logger="lightrag"):
+        nodes, edges = await _rebuild_from_extraction_result(
+            DummyKV(),
+            "{not json at all, no delimiter either",
+            chunk_id="chunk-3",
+            timestamp=1,
+        )
+    assert nodes == {}
+    assert edges == {}
+    assert "Complete delimiter can not be found" in caplog.text


YashvantHange:
I duplicated this work in #3746 before noticing yours was already open — mine is closed now, yours was first. Two things I ran into while writing the same fix, offered in case they are useful; both are small and neither is a blocker.

**1. The parse runs twice.** `tolerant_load_json_dict(extraction_result)` in the condition re-parses a string `_process_json_extraction_result` has already parsed and thrown away. On a large cached extraction during a full rebuild that is two `json_repair` passes per chunk. Splitting the parse out of the parser so the dispatcher can reuse its result avoids it, though that is a bigger diff than this fix needs.

**2. "Parses as an object" is wider than "is an extraction result."** The condition accepts any non-empty JSON object, so a cached response like `{"summary": "no entities here"}` now returns empty instead of falling through. That is almost certainly harmless — the delimiter parser cannot extract anything from a JSON document either — but if you want the fallback to keep its old reach for genuinely non-extraction payloads, testing for the schema rather than for objecthood does it:

```python
def _has_json_extraction_shape(payload) -> bool:
    return any(isinstance(payload.get(f), list) for f in ("entities", "relationships"))
```

The presence of the fields is the test, never the presence of items in them — which is the same distinction your fix is making, just applied one level up.

One edge case worth a test either way: a cached response that is an empty JSON object with delimiter records appended after it. `tolerant_load_json_dict` accepts the records as trailing prose, so the condition passes and the records are dropped, where the old `if nodes or edges` recovered them. The JSON prompt never asks for that shape, so it may be out of scope — but if you want it covered, gating on `PROMPTS["DEFAULT_COMPLETION_DELIMITER"] not in extraction_result` when the JSON yielded nothing keeps the rescue and still kills the misleading warning, since a response carrying the terminator does not trigger it.

Happy to leave all of this; the reported bug is fixed either way.

danielaskdd:
I reviewed this fix and @YashvantHange's three points, and checked each of them empirically by running `_rebuild_from_extraction_result` on both `main` and this branch. Summary up front: **the fix is correct and I think it can merge as-is — none of the three points need to be acted on.** Details below, since two of them are true-but-inconsequential and the third is real in substance but not for the example given.

### The fix

Judging "did the parse succeed?" by `nodes or edges` conflates *"nothing was extracted"* with *"nothing could be parsed"*. Moving the test onto `tolerant_load_json_dict`'s own documented signal puts it back on the parser's contract instead of on downstream results. The scope claim checks out too: the live path picks a parser by mode with an `if`/`else` at `operate.py:4126-4134` and `4192-4200` and never crosses between them, so only cache rebuild is affected.

Observed behaviour, `main` vs. this branch:

| input | `main` | this PR |
|---|---|---|
| `{"entities":[],"relationships":[]}` | misleading delimiter warning, empty result | silent, empty result |
| `{"summary":"no entities here"}` | misleading delimiter warning, empty result | silent, empty result (same nodes/edges) |
| `{}` + delimiter records | recovers the record | recovers the record (**unchanged**) |
| `{"entities":[],"relationships":[]}` + delimiter records | recovers the record | returns empty |

### On the three review points

**1. Double parse — true, but the cost is much smaller than stated.** `or` short-circuits, so the second `tolerant_load_json_dict` only runs when `nodes` *and* `edges` are both empty. A large cached extraction never reaches it, so it is not "two `json_repair` passes per chunk" on a full rebuild — the extra parse lands only on chunks that yielded nothing, where the payload is a few dozen bytes. Splitting the parse out of the parser is a reasonable design change on its own merits, but it buys close to nothing here.

**2. Objecthood vs. schema — true, but with no observable effect.** For `{"summary": "no entities here"}` the delimiter fallback on `main` recovers nothing either; its entire contribution is the misleading warning this PR exists to remove. Nodes and edges are identical before and after. `_has_json_extraction_shape` is the more precise predicate and the "the presence of the fields is the test, never the presence of items in them" framing is a nice generalisation of the bug — but adopting it changes no behaviour in the current code.

**3. The edge case — right instinct, wrong example.** `{}` followed by delimiter records does *not* regress: `tolerant_load_json_dict` returns `{}` there (falsy), the condition fails, the fallback runs, and the record is recovered exactly as on `main`. Trailing prose being tolerated means the object still parses — but the object that parses is empty, which is precisely what keeps it falling through.

The shape that does regress is `{"entities": [], "relationships": []}` followed by delimiter records (last row above). Since the JSON prompt never asks for a valid-empty-schema-plus-delimiter-records hybrid, I agree with the "may be out of scope" read. If a maintainer wants the belt-and-braces version, the suggested `PROMPTS["DEFAULT_COMPLETION_DELIMITER"] not in extraction_result` gate does cover that real variant at the cost of one substring scan, and would not reintroduce the warning. Worth noting the two suggestions are not interchangeable: a schema check alone does not catch this case, since the hybrid payload has the schema.

### Tests

`tests/extraction` on this branch: 261 passed, 7 failed; on `main`: 258 passed, 7 failed. The 7 failures are identical on both sides and come from missing spaCy models in my environment (`test_keyword_extraction_drivers.py` also fails to collect for the same reason, on both). The three new tests pass.

### Unrelated observation

On genuinely unparseable JSON-looking input, two warnings are emitted for one failure — `_process_json_extraction_result`'s "JSON extraction result is empty or unrecoverable", then the fallback's "Complete delimiter can not be found". That predates this PR and is present on `main`; I'm opening a separate PR for it rather than widening this one.

---

danielaskdd:
@codex review

AmirF194:
Thanks for the deep empirical check on all three points, especially confirming the double-parse cost is bounded by the short circuit. Thanks to YashvantHange too for the {} plus delimiter edge case, good to have it on record even after spotting the overlap.

PR #3746, opened 2026-08-29
fix(extraction): read an empty JSON extraction as an answer, not a parse failure

### LINKED ISSUE #3744: [Bug]: Valid empty JSON extraction cache entries emit misleading delimiter warnings during rebuild

YashvantHange:
## Description

In `ENTITY_EXTRACTION_USE_JSON` mode a chunk that holds no reliable facts legitimately returns `{"entities": [], "relationships": []}`. During an extraction-cache rebuild, `_rebuild_from_extraction_result` decided whether the JSON parse had worked from the entity count:

```python
nodes, edges = await _process_json_extraction_result(...)
if nodes or edges:
    return nodes, edges
# Otherwise fall through to text-based parsing
```

So every empty-but-valid cached entry fell through to the delimiter parser, which found no `<|COMPLETE|>` in a JSON document and warned about it — once per chunk, for the whole rebuild. The warning does not, by itself, indicate malformed JSON, truncation, or lost extraction data, which is exactly what an operator reads it as.

The fallback now turns on the response's own format rather than on how much it yielded.

## Related Issues

Closes #3744.

## Changes Made

`_process_json_extraction_result` splits into two halves so the rebuild path can ask "did this parse as an extraction response?" before it has any entities to look at:

- `_load_json_extraction_payload` — the `tolerant_load_json_dict` call plus the LaTeX escape repair. Returns `None` when the response is not recoverable as a single JSON object.
- `_extract_from_json_payload` — the entity and relationship record loops, unchanged.
- `_has_json_extraction_shape` — whether a parsed object answered in the extraction schema, decided by the presence of the `entities` / `relationships` list fields and never by their contents. That distinction is the fix.

`_process_json_extraction_result` keeps its signature and semantics and is now those two halves in sequence, so initial extraction and gleaning are untouched. Neither has a fallback anyway — they know which format they asked for, and the empty-vs-failure question exists only in the rebuild dispatcher, which has to auto-detect the format of each cached entry.

### The fallback is narrowed, not removed

Delimiter parsing is still reached by a result that does not parse as an extraction object: a top-level array, unparseable text that happens to start with a brace, or an object answering some other question. Those all behave exactly as before.

One further shape kept working deliberately. A response that is an empty JSON object with delimiter records appended is not something the JSON prompt asks for, but `tolerant_load_json_dict` accepts the records as trailing prose, so the shape gate passes and a strict "JSON wins" rule would drop them. Its terminator is what identifies it, and a response carrying one reaches the delimiter parser *without* the spurious warning — the parser finds the terminator it was looking for. So the rescue costs nothing to keep, and the narrowing loses no cached shape at all rather than only no shape I could think of.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

New `tests/extraction/test_empty_json_extraction_rebuild.py` drives `_rebuild_from_extraction_result` end to end. It pins the reported case raw and fenced, asserts the rebuild logs *nothing* for it — not the delimiter warning and not the JSON parser's own "empty or unrecoverable", which would be just as wrong about a response that parsed perfectly well — and covers a payload whose records are all dropped by validation, which lands in the same place as an empty array and must be read the same way. Three tests pin the surviving fallback paths and one pins the appended-records rescue; those four pass on `main` as well, which is the point of them.

Run subset per AGENTS.md: `tests/extraction`.

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -902,51 +902,64 @@ def _looks_like_json_extraction_result(result: str) -> bool:
     return False
 
 
-async def _process_json_extraction_result(
-    result: str,
-    chunk_key: str,
-    timestamp: int,
-    file_path: str = "unknown_source",
-) -> tuple[dict, dict]:
-    """Process a JSON-formatted extraction result from LLM.
+def _load_json_extraction_payload(result: str, chunk_key: str) -> dict[str, Any] | None:
+    """Recover the JSON object from an extraction response, or None.
 
-    This function parses the LLM response as JSON and extracts entities and relationships.
-    It uses :func:`tolerant_load_json_dict` to recover JSON from fenced,
-    prose-wrapped, or slightly malformed responses from weaker models.
+    ``tolerant_load_json_dict`` strips fences, tolerates leading/trailing prose
+    (including trailing braces), and repairs the malformed-object slips
+    ``json_repair`` fixes. It never raises and returns ``{}`` for unparseable
+    input or a top-level array (callers require exactly one object). Note:
+    because the helper absorbs the underlying parse exception, only this single
+    "empty or unrecoverable" warning is logged — the low-level decode error is
+    no longer surfaced.
 
-    Args:
-        result: The JSON extraction result from LLM
-        chunk_key: The chunk key for source tracking
-        timestamp: The timestamp for the extraction
-        file_path: The file path for citation
-
-    Returns:
-        tuple: (nodes_dict, edges_dict) containing the extracted entities and relationships
+    ``None`` means the response was not recoverable as a single JSON object.
     """
-    maybe_nodes = defaultdict(list)
-    maybe_edges = defaultdict(list)
-
-    # tolerant_load_json_dict strips fences, tolerates leading/trailing prose
-    # (including trailing braces), and repairs the malformed-object slips
-    # json_repair fixes. It never raises and returns {} for unparseable input or
-    # a top-level array (callers require exactly one object). Note: because the
-    # helper absorbs the underlying parse exception, only this single "empty or
-    # unrecoverable" warning is logged — the low-level decode error is no longer
-    # surfaced.
     parsed = tolerant_load_json_dict(result)
     if not parsed:
         logger.warning(
             f"{chunk_key}: JSON extraction result is empty or unrecoverable"
             f"{_truncation_cause_suffix(result)}"
         )
-        return dict(maybe_nodes), dict(maybe_edges)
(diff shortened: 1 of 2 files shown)


YashvantHange:
Closing this in favour of #3745, which fixes the same defect and was opened about ten hours earlier. I did not check for an existing PR on #3744 before writing this one — my mistake, and the duplicated review effort is on me.

Two differences from #3745 that may be worth folding in there, or may not be — leaving that to @AmirF194 and the maintainers rather than pressing them here. I have left them as a review comment on #3745 so they are next to the code they concern.

Thanks, and sorry for the noise.

Issue #2572, opened 2026-01-06
[Feature Request]: RESTFul Query Token usage 


spoonbobo:
### Do you need to file a feature request?

- [x] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

### Feature Request Description

https://github.com/HKUDS/LightRAG?tab=readme-ov-file#token-usage-tracking

Will there be token usage field exposed when using `/query` in webui RESTFul API? Thanks

YashvantHange:
I'm taking this — implementation is ready, opening a PR shortly.

PR #3747, opened 2026-08-29
feat(ollama): report token usage through the TokenTracker contract

YashvantHange:
## Description

`TokenTracker` (`lightrag/utils.py`) is the binding-facing usage contract, but only `lightrag/llm/openai.py` and `lightrag/llm/gemini.py` implement it. Ollama — the most common local deployment in this tracker — reported nothing, even though the server returns `prompt_eval_count` / `eval_count` on every completed call and this module already reads both to decorate a truncation error message:

```python
diagnostics = format_response_diagnostics(
    done_reason="length",
    eval_count=response.get("eval_count"),
    prompt_eval_count=response.get("prompt_eval_count"),
    ...
)
```

The numbers were already in hand and already accessible; nothing surfaced them as usage.

On the chat path this was worse than a no-op. Everything left in `**kwargs` is forwarded verbatim to `AsyncClient.chat()`, which declares no `**kwargs`:

```
AsyncClient.chat(self, model='', messages=None, *, tools=None, stream=False,
                 think=None, logprobs=None, top_logprobs=None, format=None,
                 options=None, keep_alive=None)
```

So `LightRAG(llm_model_kwargs={"token_tracker": tracker})` — which works on the OpenAI binding, since `llm_model_kwargs` is splatted into the bound role function in `lightrag/llm_roles.py` — raised `TypeError` on every Ollama call. `ollama_embed` builds its client arguments explicitly rather than forwarding `**kwargs`, so there the same kwarg was silently dropped instead.

## Related Issues

Refs #2572, which asks for token usage on the REST surface. This is the layer under that: usage cannot be exposed for a binding that never reports it. It does not close the issue.

## Changes Made

- `token_tracker` is an explicit parameter on `_ollama_model_if_cache`, `ollama_model_complete` and `ollama_embed`. Explicit rather than a `kwargs` key because on the chat path an unconsumed one reaches the wire call. Appended after the existing parameters, so positional callers are unaffected.
- `_ollama_usage_counts()` maps Ollama's counters onto the tracker's key names and sums the total Ollama does not report. It reads through `.get`, which covers both the raw dict of ollama<0.4 and the `SubscriptableBaseModel` `ChatResponse` of >=0.4 — the accessor the truncation diagnostics already rely on — and returns `None` for a payload with no `.get` rather than raising, for the reason `_response_message_field` gives next door: a shape this helper cannot read must not take down a response the caller could have used.
- Non-streaming, streaming and embedding paths all account.

### Two decisions worth a reviewer's eye

**Absent is not zero.** A payload reporting neither counter records nothing. An older or non-conforming server that reports no numbers is not evidence of a zero-token call, and recording it as a genuine zero would make a tracked total silently under-count instead of visibly skipping. A payload reporting one of the two is real usage, and the missing half counts as zero.

**Streaming accounts after the loop and inside the `try`**, so a consumer that disconnects mid-stream is not billed. That is the OpenAI binding's behaviour; Gemini accounts from a `finally` and does bill a disconnect. The two upstream bindings already disagree, so this follows the one whose usage semantics the rest of this file mirrors, and says so in a comment rather than leaving the next reader to discover the split. Happy to flip it if you would rather the bindings converge the other way.

Non-streaming accounts before the empty-truncation check raises: the request burned its budget whether or not the output was usable, and a thinking model that spent the whole `num_predict` on its reasoning trace is exactly the case that raises there. This matches the OpenAI binding, which counts usage before its own empty-response validation.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

New `tests/llm/ollama_impl/test_ollama_token_usage.py`. Beyond the happy paths it pins the decisions above: usage counted before the truncation raise, a counter-less response and a counter-less stream recording nothing, an abandoned stream not billed, `token_tracker` never reaching `AsyncClient.chat()`, and the mapper against a real `ChatResponse` so the ollama>=0.4 claim is not left to the raw-dict fixtures.

One test is not about this feature. `tests/llm/ollama_impl/` pins no retry policy, and the `@retry(...)` block sits immediately above `async def _ollama_model_if_cache` — so a helper defined between the two silently takes the decorator and leaves every Ollama call unretried, with both lines showing as context in the diff. I did exactly that in an earlier revision of this branch and caught it in review, so `test_the_retry_policy_still_guards_the_call` now asserts the decorator is where it belongs.

The change is additive: 0 deleted lines in `lightrag/llm/ollama.py`. Run subset per AGENTS.md: `tests/llm/ollama_impl`.

### DIFF
--- a/lightrag/llm/ollama.py
+++ b/lightrag/llm/ollama.py
@@ -167,6 +167,38 @@ def _normalize_ollama_response_format(kwargs: dict) -> None:
     kwargs["format"] = response_format
 
 
+def _ollama_usage_counts(payload: Any) -> dict[str, int] | None:
+    """Map Ollama's eval counters onto the TokenTracker's key names, or None.
+
+    Ollama reports ``prompt_eval_count`` / ``eval_count`` and has no combined
+    total, so the total is their sum. ``.get`` works on both the raw dict
+    (ollama<0.4) and the ``SubscriptableBaseModel`` ChatResponse (ollama>=0.4)
+    -- the same accessor the truncation diagnostics below already rely on --
+    and a payload that has no ``.get`` at all degrades to ``None`` rather than
+    raising, for the reason :func:`_response_message_field` gives: a shape this
+    helper cannot read must not take down a response the caller could have used.
+
+    ``None`` means the payload reported no counters *at all*, which is an older
+    or non-conforming server rather than a call that consumed nothing. A
+    payload reporting one of the two is real usage, and the absent half counts
+    as zero.
+    """
+    get = getattr(payload, "get", None)
+    if not callable(get):
+        return None
+    prompt_tokens = get("prompt_eval_count")
+    completion_tokens = get("eval_count")
+    if prompt_tokens is None and completion_tokens is None:
+        return None
+    prompt_tokens = prompt_tokens or 0
+    completion_tokens = completion_tokens or 0
+    return {
+        "prompt_tokens": prompt_tokens,
+        "completion_tokens": completion_tokens,
+        "total_tokens": prompt_tokens + completion_tokens,
+    }
+
+
 @retry(
     stop=stop_after_attempt(3),
     wait=wait_exponential(multiplier=1, min=4, max=10),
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3754, opened 2026-08-30
feat(api): add GET /graph/export to download the full knowledge graph

### LINKED ISSUE #1937: [Feature Request]: Export graph to files via webui

This is a great project, thank you for your excellent work.
We hope to:
1. It can support users to import (append or overwrite) and export data through the webui for manual maintenance.

### Additional Context

_No response_

YashvantHange:
## Description

Adds `GET /graph/export` so the full knowledge graph (entities, relations, relationships) can be
downloaded through the REST API. Previously the only way to get this data out of a running
instance was direct filesystem access to internal storage files.

This is the **export half only**. Import (writing into the graph) and a WebUI button are
deliberately out of scope -- see *Additional Notes*.

## Related Issues

Fixes #1937

## Changes Made

- New route `GET /graph/export` in `lightrag/api/routers/graph_routes.py`, guarded by the same
  `combined_auth` dependency as every other route in the file. Accepts `file_format`
  (csv/excel/md/txt) and `include_vector_data` query params and wires up the existing
  `LightRAG.aexport_data()`, which already builds this export but was never exposed over HTTP.
- Writes to a server-managed temp file via `tempfile.mkstemp` (never a client-supplied path),
  streams it back as a `FileResponse` attachment, and removes the temp file once the response has
  been fully sent (via a `BackgroundTask`), or immediately on error.
- Fixes a portability bug in `lightrag/utils.py::aexport_data` found while wiring this up: the
  "Relationships (from VectorDB)" section unconditionally accessed
  `relationships_vdb.client_storage`. That attribute is a debug/snapshot escape hatch implemented
  by exactly two backends, which disagree on sync vs. async (`NanoVectorDBStorage`: async
  property; `FaissVectorDBStorage`: plain sync property) -- every other backend (Milvus, Qdrant,
  Postgres, Redis, MongoDB, OpenSearch) has no such attribute at all. Left as-is, exporting a
  graph on any of those backends would have crashed the whole export. The fix checks for the
  attribute on the class (so testing for it doesn't trigger an async property's getter) and
  inspects the value itself before deciding whether to await it, so unsupported backends now just
  skip that one supplementary section instead of failing outright.
- Tests: `tests/api/routes/test_graph_export_route.py` (auth, format validation via FastAPI's
  `Literal` query-param typing, temp-file cleanup on both success and failure, forwarding of
  `file_format`/`include_vector_data`) and
  `tests/utils/test_aexport_data_portable_relationships.py` (pins the portability fix for both the
  async-property and sync-property backend shapes, plus the no-`client_storage` case).

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Import and a WebUI download button are intentionally not part of this PR. Import means writing
into the graph -- validation, conflict handling, pipeline coordination -- which is a materially
larger and separate piece of work better reviewed on its own; a WebUI button is a natural,
additive follow-up once the endpoint exists. Noted on the issue as deliberate scoping, not an
oversight.

`pre-commit run --all-files` and the full `pytest` suite were run locally; the only failures are
pre-existing and environment-specific to this Windows dev machine (symlink tests requiring admin,
`psutil`/`fcntl`-dependent process checks, missing `libcairo` for SVG rasterization, gunicorn not
running on Windows), unrelated to this change.

### DIFF
--- a/lightrag/api/routers/graph_routes.py
+++ b/lightrag/api/routers/graph_routes.py
@@ -2,10 +2,15 @@
 This module contains all graph-related routes for the LightRAG API.
 """
 
-from typing import Optional, Dict, Any
+import os
+import tempfile
 import traceback
+from pathlib import Path
+from typing import Optional, Dict, Any, Literal
 from fastapi import APIRouter, Depends, Query, HTTPException
+from fastapi.responses import FileResponse
 from pydantic import BaseModel, Field, field_validator, model_validator
+from starlette.background import BackgroundTask
 
 from lightrag.base import DeletionResult
 from lightrag.utils import logger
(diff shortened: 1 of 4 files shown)


danielaskdd:
Thanks for the thorough write-up and for the tests — the `client_storage` portability issue you found is a real bug, and the endpoint itself (server-managed temp file, auth, `Literal`-validated format, cleanup on both paths) is cleanly implemented.

That said, we're not going to take this one, for a reason that sits underneath the PR rather than in it: **`aexport_data()` is legacy code that was never designed for large graphs, and we plan to mark it deprecated.** Exposing it over HTTP would lock in its cost model at exactly the point where it hurts most.

Concretely, the relation section discovers edges with an O(N²) scan — one `has_edge` await per ordered entity pair (`lightrag/utils.py:4463-4470`). At 50k entities that is ~2.5 billion calls; on an in-memory NetworkX graph that is tens of minutes, and on Neo4j or PostgreSQL, where each call is a round trip, it is days. It also materializes the entire export in memory before writing a byte, and the write path is synchronous inside an async route, so a single request would stall the event loop for the whole duration. `include_vector_data=True` compounds it further by stringifying every embedding. None of this is your doing — it is the function's existing shape — but a well-meaning click on `/graph/export` would take an instance down, and `GET /graphs` is deliberately bounded (`max_nodes`, `MAX_GRAPH_NODES`) precisely to avoid that.

Our direction for full-graph export is **direct access to the backing store** (an offline/ops-side tool that talks to the graph and vector backends and streams in batches — every backend already implements `get_all_nodes()` / `get_all_edges()`), rather than a request-scoped REST endpoint that must hold the whole graph in one process's memory and one HTTP response.

The portability fix is worth keeping on its own, though: it is an unconditional crash today for anyone calling `export_data()` on Faiss (`TypeError: object dict can't be used in 'await' expression`) or on any backend without `client_storage` at all (`AttributeError`), independent of the endpoint. We've carried that hunk and your `tests/utils/` test over into a separate PR so it lands with attribution to you, rather than being lost with this one.

Closing this as a scoping decision, not a quality one. If you'd like to keep working in this area, an export tool built on the batch APIs with streaming output would be very welcome — happy to discuss the shape on #1937 first.

---

PR #3758, opened 2026-08-30
perf(chunker): defer whole-document encode to the fixed-window path

### LINKED ISSUE #3757: [Bug]:[Performance] chunking_by_token_size() tokenizes the whole document even when split_by_character discards the result

### Describe the bug

**Every document pays for a whole-document tokenization whose result is thrown away whenever `chunk_split_by_character` is configured.**

```python
# lightrag/chunker/token_size.py:161 — runs unconditionally at the top
tokens = tokenizer.encode(content)
...
if split_by_character:
    raw_chunks = content.split(split_by_character)   # never touches `tokens`
    ...
    _tokens = tokenizer.encode(chunk)                # each segment re-encoded anyway
else:
    for index, start in enumerate(range(0, len(tokens), ...)):  # the ONLY reader
```

The whole-document token list is consumed exclusively by the fixed-window `else` branch. On the `split_by_character` path (a documented common configuration, e.g. `"\n\n"`), every split segment is encoded individually — necessarily, because BPE boundaries differ on substrings — and the up-front full encode is never read. The codebase's own comment puts a full encode at roughly 0.5 s per MiB (`lightrag/utils.py` tokenizer-executor notes), paid once per document for zero use.

**Measured** (cl100k_base, 2.27 MB document, 2,600 paragraphs, `split_by_character="\n\n"`, counting tokenizer wrapper):

| | encode calls | encoded text volume | median wall time (5 runs) |
|---|---|---|---|
| current | 2,601 | 4.53 MB | ~0.93 s |
| fix (encode deferred to the fixed-window branch) | 2,600 | 2.26 MB | ~0.46 s |

Roughly half the chunker's tokenizer work on the split path is the discarded full-document encode, and the encoded-text volume is doubled (every character passes through the tokenizer twice).

### Expected Behavior

The whole-document encode should happen only on the path that reads it.

### Steps to Reproduce

1. Configure `chunk_split_by_character` (e.g. `"\n\n"`).
2. Ingest a multi-MiB document.
3. Profile `chunking_by_token_size` / count `tokenizer.encode` calls: one call receives the entire document and its result is never used on this path.

### LightRAG Config Used

Default chunking with `CHUNK_SPLIT_BY_CHARACTER="\n\n"`.

### Logs and screenshots

_N/A_

### Additional Information

- Python Version: 3.12
- Related Issues: -

hata33:
Fixes #3757

## Summary

Moves the whole-document `tokenizer.encode()` from the top of `chunking_by_token_size()` into the fixed-window `else` branch, its only consumer. The `split_by_character` path re-encodes each split segment individually (BPE boundaries differ on substrings) and never reads the whole-document token list, so the up-front full encode was paid once per document for zero use.

Measured on a 2.27 MB document with `split_by_character="\n\n"` (cl100k_base): encoded-text volume drops from 4.53 MB to 2.26 MB and median chunker wall time from ~0.93 s to ~0.46 s (~50%).

## What changed

- `lightrag/chunker/token_size.py`: `tokens = tokenizer.encode(content)` moved into the `else` (fixed-window) branch, with a comment stating why the encode is deferred.
- `tests/chunker/test_chunking.py`: three regression tests using a counting tokenizer — the `split_by_character` and `split_by_character_only` paths issue exactly one encode per segment (never the whole document), and the fixed-window path still encodes the whole document.

## Side effect analysis

- **Chunk output is byte-identical**: no code on the split path reads `tokens` (verified by reading every reference in the function and pinned by the new regression tests).
- No API, config, or default-value change; negative-overlap validation still runs before any branching, so its behavior is unchanged on both paths.

## Tests

- `pytest tests/chunker -q` → **302 passed**
- `ruff check lightrag/chunker/token_size.py tests/chunker/test_chunking.py` → clean
- `ruff format --check` on both files → clean


danielaskdd:
@codex review

PR #3760, opened 2026-08-30
perf(operate): encode chunk-invariant extraction prompts once per run

### LINKED ISSUE #3759: [Bug]:[Performance] gleaning token guard re-encodes the chunk-invariant extraction system prompt for every chunk

### Describe the bug

**With gleaning enabled, every chunk of an ingestion re-encodes (and re-formats) a multi-thousand-token prompt whose content never changes across the run.**

```python
# lightrag/operate.py, gleaning token guard inside the per-chunk _process_single_content
gleaning_token_count = (
    len(extract_tokenizer.encode(entity_extraction_system_prompt))   # re-encoded per chunk
    + sum(len(extract_tokenizer.encode(msg.get("content", "") or "")) for msg in history)
    + len(extract_tokenizer.encode(entity_continue_extraction_user_prompt))
)
```

`entity_extraction_system_prompt` is built as `PROMPTS[...].format(**context_base)`, and `context_base` is constructed **once** at the top of `extract_entities` — the system prompt (which embeds the examples block) is byte-identical for every chunk. In JSON mode the continue prompt is also built from `context_base` alone. Yet both are re-formatted and re-encoded inside the per-chunk closure, and these are **synchronous `tokenizer.encode` calls** while the rest of the codebase deliberately routes bulk encodes through the tokenizer executor.

**Measured** (cl100k_base, default text-mode system prompt with the examples block: 12,020 chars / 2,031 tokens; median of 20 encodes on Windows 11 / Python 3.12):

| | cost |
|---|---|
| single encode of the system prompt | ~115 ms |
| wasted across a 1,000-chunk ingest (gleaning on) | ~116 s |
| wasted across a 10,000-chunk ingest | ~19 min |

That is pure tokenizer CPU spent re-deriving a constant, on top of the redundant `PROMPTS[...].format()` per chunk.

### Expected Behavior

Chunk-invariant prompt strings and their token counts should be computed once per extraction run.

### Steps to Reproduce

1. Configure `entity_extract_max_gleaning >= 1` (and any `MAX_EXTRACT_INPUT_TOKENS > 0` so the guard runs).
2. Ingest a corpus; count `tokenizer.encode` calls receiving the system prompt — one per chunk instead of one per run.

### LightRAG Config Used

Default extraction with `ENTITY_EXTRACT_MAX_GLEANING=1`.

### Logs and screenshots

_N/A_

### Additional Information

- Python Version: 3.12
- Related Issues: -

hata33:
Fixes #3759

## Summary

Hoists the chunk-invariant extraction prompts out of the per-chunk `_process_single_content` closure:

- The **system prompt** (both text and JSON mode) is now formatted once per `extract_entities` run instead of once per chunk — it is built solely from `context_base`, which is fixed for the run.
- The **JSON-mode continue prompt** is likewise hoisted (it also depends only on `context_base`).
- Their combined token count is precomputed once as `gleaning_invariant_tokens` and reused by the gleaning token guard. The text-mode continue prompt still embeds the chunk content, so its token count stays per-chunk.

Guard arithmetic is unchanged — same tokenizer, same strings, same counts — so gleaning skip/run decisions are bit-identical. Measured with cl100k_base, a single encode of the default system prompt (~2,031 tokens) costs ~115 ms on this machine; a 10,000-chunk ingest with gleaning on stops paying that per chunk (~19 min of redundant tokenization removed).

## Side effect analysis

- **No behavior change**: the hoisted strings are byte-identical to what the closure built per chunk; only construction count changes.
- `extract_tokenizer is None` is still handled (`gleaning_invariant_tokens = 0`, and the guard's existing None check still gates use).
- No API, config, or default-value change.

## Tests

- New regression test in `tests/extraction/test_extract_entities.py`: with gleaning enabled and 3 chunks, the longest encoded string (the system prompt) is encoded exactly once per run while all 3 chunks still run both LLM rounds.
- `pytest tests/extraction -q` → 281 passed (1 pre-existing failure on Windows, `test_prompt_file_rejects_absolute_paths`, fails identically on clean `main`)
- `ruff check` / `ruff format --check` clean


danielaskdd:
The optimization is sound — the system prompt really is chunk-invariant (I checked the placeholders: `entity_extraction_system_prompt`, `entity_extraction_json_system_prompt` and `entity_continue_extraction_json_user_prompt` draw only from `context_base`), and the guard arithmetic in this diff is item-by-item equivalent in both modes.

But there is a blocking regression: **JSON mode with gleaning enabled now raises `UnboundLocalError`.**

The JSON branch of `_process_single_content` no longer assigns `entity_continue_extraction_user_prompt`, but the text-mode branch still does. Python resolves scope statically per function, so an assignment anywhere in the closure makes that name a local for the *whole* function body — it no longer falls back to the enclosing scope. The new guard short-circuits (`else 0`) and so hides it, but the gleaning LLM call a few lines later reads the unbound local:

```python
glean_result, timestamp = await use_llm_func_with_cache(
    entity_continue_extraction_user_prompt,   # unbound in JSON mode
```

Reproduced on `da36c479` with a 1-chunk JSON-mode run (`entity_extraction_use_json=True`, `entity_extract_max_gleaning=1`):

```
UnboundLocalError: C[1/1]: chunk-001: cannot access local variable
'entity_continue_extraction_user_prompt' where it is not associated with a value
lightrag/operate.py:4514
```

The same probe passes on this PR's base (`18f99b61`), so it is introduced here. Trigger conditions are both reachable defaults for anyone who sets `ENTITY_EXTRACTION_USE_JSON=true` (`lightrag/lightrag.py:743`), since `DEFAULT_MAX_GLEANING = 1` — i.e. extraction fails outright for those users.

**Why CI stayed green:** the test matrix has a blind spot at exactly this intersection. In `tests/extraction/test_entity_extraction_stability.py` every `use_json=True` case runs with `max_gleaning=0`, and every gleaning case runs with `use_json=False`; `entity_extraction_use_json` appears in only three files under `tests/`. The new regression test here is text-mode only, so it cannot see it either.

Minimal fix — rename the hoisted string and bind the local explicitly in the JSON branch:

```python
if use_json_extraction:
    entity_continue_extraction_user_prompt = json_continue_extraction_user_prompt
```

Two smaller points:

1. **The precompute should stay lazy.** `gleaning_invariant_tokens` is computed unconditionally, but on `main` those encodes only happen when `gleaning > 0 and extract_tokenizer is not None and max_extract_input_tokens > 0`. As written, a run with gleaning off now pays an encode of the ~2k-token system prompt per `extract_entities` call that `main` never paid. Gating the precompute on the same condition as the guard keeps it a strict improvement.

2. **The text-mode comment is inaccurate.** `entity_continue_extraction_user_prompt`'s placeholders are `completion_delimiter / language / max_total_records / max_entity_records` — no `input_text`; the `input_text=content` kwarg passed at the call site is simply unused by the default template. Keeping it per-chunk is still the right call, because a custom prompt file *may* reference `{input_text}`, but that is the reason to state rather than "it embeds the chunk content".

Separately, the ~115 ms figure for a single 2,031-token `cl100k_base` encode (and the ~19 min extrapolation) looks off by roughly two orders of magnitude — tiktoken's throughput puts a 12 KB string in the single-digit-millisecond range. I could not measure it here (this environment cannot reach the BPE vocab download), so I am flagging it rather than asserting it; worth re-measuring with the registry load excluded. The optimization is worth doing either way — only the stated magnitude is in question.

I have pushed a corrected version of this change (fix + lazy precompute + the missing JSON-mode coverage, each new test verified against its mutation) as [#3763](https://github.com/HKUDS/LightRAG/pull/3763) — happy for it to be folded into this PR instead if you prefer to keep authorship here.

---

danielaskdd:
Correction to the last line above: the corrected version is #3787, not #3763 — I wrote the number before the PR was opened, and #3763 is an unrelated (already merged) WebUI change. Sorry for the noise.

---

hata33:
Thank you for the thorough review — all three points are addressed in `c0a3f3bd`, folded into this PR as you offered (so #3787 can be dropped if that's easier on your side):

1. **UnboundLocalError** — confirmed, excellent catch on the static scoping. The hoisted constant is renamed to `json_continue_extraction_user_prompt` and the JSON branch of the closure now binds `entity_continue_extraction_user_prompt` explicitly, with a comment explaining why the binding is load-bearing. I added the missing JSON-mode + gleaning regression test (`test_gleaning_json_mode_binds_continue_prompt`) and verified it reproduces the exact `UnboundLocalError` on the previous commit (`da36c479`) before the fix — the test-matrix blind spot you identified is now covered.

2. **Lazy precompute** — `gleaning_invariant_tokens` is now computed only when the guard's own condition holds (`gleaning > 0 and tokenizer present and MAX_EXTRACT_INPUT_TOKENS > 0`), so gleaning-off runs pay nothing, as on `main`.

3. **Comment corrected** — the text-mode continue prompt stays per chunk because a custom prompt file *may* reference `{input_text}` (the default template's placeholders don't include it); the comment now states that instead of the inaccurate claim.

On the timing figure: you're right — re-measured with the registry load excluded and warm caches, a single encode of the ~12 KB prompt is **~0.66 ms** (median of 50), so the earlier ~115 ms was a measurement artifact on my machine and the honest extrapolation is ~7 s of redundant tokenization per 10,000 chunks, plus the per-chunk `PROMPTS[...].format()` that this also removes. The optimization is unchanged; only the stated magnitude was wrong. I've noted the same correction on #3759.

CI on the new commit: lint/format clean, `tests/extraction` 282 passed (1 pre-existing Windows-only failure, `test_prompt_file_rejects_absolute_paths`, identical on clean `main`).

danielaskdd:
Confirmed fixed — `c0a3f3b` resolves the `UnboundLocalError`, and it also picks up the lazy precompute and the comment correction. I checked out your head and ran my own test set against it: all pass, including the JSON-mode gleaning path that failed before.

I have accordingly gutted #3787: it no longer carries any implementation of its own. It is now rebased onto your head, `lightrag/operate.py` is byte-identical to yours, and it adds only tests — one commit, additive, no existing test modified.

What it covers, and why I did not just drop it: I mutated your head and watched the suite stay green in three cases.

| Mutation | Your suite | With #3787 |
|---|---|---|
| JSON continue prompt dropped from `gleaning_invariant_tokens` | 🟢 6 passed | 🔴 2 failed |
| `gleaning_precheck` reverted to unconditional | 🟢 6 passed | 🔴 1 failed |
| Text-mode continue prompt dropped from the guard | 🟢 6 passed | 🔴 1 failed |
| JSON local binding removed | 🔴 caught | 🔴 caught |

Your binding test is solid. The first three are the "did hoisting *move* those tokens or silently *drop* them" question — dropping them makes the guard under-count, which lets an oversized gleaning payload reach the provider, i.e. exactly the `context_length_exceeded` the guard exists to prevent.

A cap-based skip assertion can't see it: at a 10-token cap the system prompt alone busts the budget, so the guard trips regardless of what else is counted. My own first draft had that flaw and two mutations survived it. The tests instead measure a delta — lengthen a continue prompt template by N characters, and the projected token count the guard logs must move by exactly N (the test tokenizer is 1 char = 1 token). That isolates each prompt's contribution.

The commit cherry-picks onto your branch cleanly if you would rather carry it in #3760 under your own authorship — say so and I will close #3787.

---

danielaskdd:
@codex review

hata33:
Cherry-picked as `696821ca` — the delta-measuring tests now ride in this PR under their own commit, so #3787 can be closed whenever convenient.

Verified after the pick (on top of your `main` merge): `tests/extraction` → 286 passed, with the single failure being the pre-existing Windows-only `test_prompt_file_rejects_absolute_paths` (identical on clean `main`); ruff check/format clean.

The mutation table convinced me — "did hoisting *move* those tokens or silently *drop* them" is exactly the question my cap-based assertions couldn't see, and the delta-of-N-characters design is a much sharper instrument than anything I had. Thanks for both the catch and the craftsmanship.

PR #3762, opened 2026-08-30
perf(api): read document statuses in one batched storage call

### LINKED ISSUE #3761: [Bug]:[Performance] GET /documents issues seven doc-status scans per request instead of one batched read

### Describe the bug

**The (deprecated but still served) `GET /documents` endpoint scans the doc-status storage seven times per request, although a one-pass batched API for exactly this pattern already exists.**

```python
# lightrag/api/routers/document_routes.py, GET /documents
statuses = (PENDING, PARSING, ANALYZING, PROCESSING, PREPROCESSED, PROCESSED, FAILED)
tasks = [rag.get_docs_by_status(status) for status in statuses]
results = await asyncio.gather(*tasks)   # 7 separate storage reads
```

Each `get_docs_by_status` call acquires the doc-status lock and scans the storage (the JSON backend's own docstring for `get_docs_by_statuses` notes that N separate calls "acquire the lock N times and scan the data N times"). The WebUI polls this endpoint, so a workspace with many rows pays 7 scans + 7 lock acquisitions per poll. The round-robin below also re-runs a 7-element `has_remaining` scan for every emitted document — O(docs × statuses), up to ~7,000 iterations at the 1,000-document cap — where a decrementing counter would do.

`DocStatusStorage.get_docs_by_statuses` (`lightrag/base.py:1281`) is implemented by every doc-status backend (JSON, PostgreSQL, MongoDB, OpenSearch, Redis) and is already used by the pipeline and the paginated endpoint's backing reads.

### Expected Behavior

One batched read per request.

### Steps to Reproduce

1. Populate a workspace with documents in several statuses (JSON doc-status storage makes the repeated scans obvious).
2. Poll `GET /documents` (the WebUI does this).
3. Observe 7 `get_docs_by_status` storage calls per poll instead of one `get_docs_by_statuses` call.

### LightRAG Config Used

Default (JSON doc-status storage).

### Logs and screenshots

_N/A_

### Additional Information

- Python Version: 3.12
- Related Issues: -

hata33:
Fixes #3761

## Summary

The deprecated-but-still-served `GET /documents` endpoint issued seven separate `get_docs_by_status` storage reads (one per status, each acquiring the doc-status lock and scanning the storage) on every poll. This replaces them with a single `get_docs_by_statuses` batched read — the API whose JSON implementation docstring explicitly calls out the N-separate-calls cost — and buckets the result by each document's own status. The fair round-robin distribution across statuses is unchanged.

Also replaces the per-emitted-document `has_remaining` rescan (O(docs × statuses), up to ~7,000 iterations at the 1,000-document cap) with a decrementing `remaining` counter.

## Side effect analysis

- **Response payload is equivalent**: same statuses tuple, same 1,000-document cap, same round-robin ordering; documents are bucketed by their own stored status, which is exactly what the per-status queries selected on.
- No API, config, or default-value change.
- `get_docs_by_statuses` is implemented by every doc-status backend (base.py abstract; JSON/PG/Mongo/OpenSearch/Redis) and is already the read path used by the pipeline supervisor.

## Tests

- New `tests/api/routes/test_document_routes_list_batched.py`: the endpoint performs exactly one `get_docs_by_statuses` call and zero `get_docs_by_status` calls, groups documents by their own status, omits empty statuses, and returns all documents under the cap.
- `pytest tests/api/routes -q` → 516 passed; the 8 failures present (docx symlink/scan-unsafe-source cases) fail identically on clean `main` on Windows and are unrelated.
- `ruff check` / `ruff format` clean.


danielaskdd:
Thanks for the careful analysis here — the 7-scans-per-request finding and the `has_remaining` O(docs × statuses) rescan are both real, and #3761 documents them well.

We're going a step further instead of optimizing: `GET /documents` has no callers left. The WebUI moved its document list to `POST /documents/paginated` and takes status counts from that same response, so the endpoint is served but never polled — the "the WebUI polls this endpoint" premise in #3761 is no longer true on `main`. Since all seven statuses are the complete `DocStatus` set, the batched read here would also still scan every doc-status row; it turns 7 scans into 1 on a path that should not be scanning at all.

So the endpoint is being removed rather than made faster, in #3788, which also drops the frontend's dead `getDocuments`/`getDocumentStatusCounts` clients and the `get_docs_by_status` read chain behind it (abstract + 5 backend impls + the `LightRAG` facade) — `get_docs_by_statuses` is the single remaining read. That PR closes #3761.

Closing this one as superseded. Your `get_docs_by_statuses` batching argument is exactly right and is what the remaining code path uses everywhere.

hata33:
Understood — removing the endpoint entirely in #3788 is strictly better than making an uncalled path faster. Thanks for keeping the analysis from #3761 (the batched-read argument) and for the pointer; happy to review #3788 if an extra pair of eyes helps.

PR #3764, opened 2026-08-30
feat(query): expose LLM token usage through the /query REST API

### LINKED ISSUE #2572: [Feature Request]: RESTFul Query Token usage

YashvantHange:
## Summary

Closes #2572. Exposes LLM token usage through the `/query` REST endpoint.

`kg_query`/`naive_query` now accumulate a `TokenTracker` across the keyword-extraction and
synthesis LLM calls made on that request, and attach the total to
`QueryResult.raw_data['metadata']['token_usage']`. `QueryResponse` gains an additive, optional
`token_usage` field populated from that metadata.

## Why this needed a capability check, not a straight kwarg pass

Only `lightrag/llm/openai.py`, `gemini.py`, and `ollama.py` accept a `token_tracker` kwarg on
their `*_complete_if_cache` functions. Every other binding (anthropic, bedrock, hf, lmdeploy,
lollms, zhipu, jina, voyageai, azure_openai, nvidia_openai) does not declare it, and several
forward unrecognized kwargs straight into the underlying SDK client call — passing
`token_tracker=` unconditionally would raise `TypeError` at the wire level for those.

`llm_roles.py::_wrap_llm_role_func` now inspects each role's bound raw function's real signature
once per (re)build and records whether it accepts `token_tracker`, exposed as
`role_supports_token_tracker`. `operate.py`'s call sites only add the kwarg when the bound role
supports it — the key is omitted entirely when unsupported, never passed as `token_tracker=None`.

## What's included

- `llm_roles.py` — per-role capability detection (`role_supports_token_tracker`)
- `lightrag.py` — threads the capability map into `global_config`; bypass-mode query path gated
  the same way
- `operate.py` — `_role_supports_token_tracker`/`_attach_token_usage` helpers; `kg_query`,
  `naive_query`, `extract_keywords_only` build a `TokenTracker` and attach usage at every point a
  caller can legitimately stop early (`only_need_context`, a cache hit, the no-keywords fail
  path, synthesis complete) — absent always means "unknown", never a zeroed-out figure
- `api/routers/query_routes.py` — `TokenUsage` model, additive `QueryResponse.token_usage`,
  populated on `/query` (not `/query/stream` — left out of scope; the streaming bypass path
  can't observe usage until the generator is drained by the caller, after the response has
  already been returned)
- `tests/llm/test_query_token_usage.py` — capability detection against real provider
  signatures, the `_attach_token_usage` helper directly, and `naive_query`/`kg_query`/
  `extract_keywords_only` end to end (supported binding, unsupported binding safety, cache-hit
  omission, keyword+synthesis aggregation, and the no-keywords fail-path edge case)

## What's not in this PR

`/query/stream` does not get a `token_usage` line. The bypass-mode streaming path only knows
usage once the response generator is fully consumed by the caller, which happens after this
endpoint has already returned its NDJSON — there's nowhere left to attach it without changing
the streaming response shape. Left for a follow-up if wanted.

## Test plan

- `ruff check` / `ruff format` clean on all touched files
- `tests/llm/test_query_token_usage.py` (17 tests) plus the existing suites it borrows fixture
  conventions from and the REST route tests over the same `QueryResponse` model

### DIFF
--- a/lightrag/api/routers/query_routes.py
+++ b/lightrag/api/routers/query_routes.py
@@ -244,6 +244,17 @@ class ReferenceItem(BaseModel):
     )
 
 
+class TokenUsage(BaseModel):
+    prompt_tokens: int = Field(
+        description="Tokens consumed by the prompt(s) sent to the LLM"
+    )
+    completion_tokens: int = Field(
+        description="Tokens consumed by the LLM's generated output"
+    )
+    total_tokens: int = Field(description="prompt_tokens + completion_tokens")
+    call_count: int = Field(description="Number of LLM calls this figure aggregates")
+
+
 class QueryResponse(BaseModel):
     response: str = Field(
         description="The generated response",
@@ -264,6 +275,17 @@ class QueryResponse(BaseModel):
             "no-context reply, or the only_need_context / only_need_prompt debug "
             "output. Clients that label machine-written text cannot recover this "
             "from the response text alone."
+        ),
+    )
+    token_usage: Optional[TokenUsage] = Field(
+        default=None,
+        description=(
+            "LLM token usage for this query (keyword extraction plus the "
+            "answer-generation call, when the mode does both). Omitted "
+            "-- not zero -- when the answer was served entirely from cache, "
+            "when only_need_context/only_need_prompt short-circuited before "
+            "any LLM call, or when the configured LLM binding does not "
+            "report usage (currently: openai, gemini, ollama)."
         ),
     )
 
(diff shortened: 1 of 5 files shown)


danielaskdd:
@codex review

danielaskdd:
Please resolve the merge conflict before proceeding with the review.

PR #3767, opened 2026-08-30
feat(prompt): deliver GRC taxonomy by configuration instead of code

devsoftx:
## Description

Delivers the GRC entity taxonomy through **configuration** instead of code.

`082f3807` added `lightrag/prompt_v1.py` and repointed the two production import sites at it. That coupled the domain vocabulary to the code: changing domains meant editing imports, the module diverged from upstream `prompt.py`, and 10 tests failed because they assert the pipeline emits the stock taxonomy.

LightRAG already ships the mechanism for this. `ENTITY_TYPE_PROMPT_FILE` resolves a YAML profile under `PROMPT_DIR/entity_type` and merges it via `resolve_entity_extraction_prompt_profile()` — the function `operate.py` calls on every extraction. This PR moves the taxonomy there and drops the forked module.

## Related Issues

None. Follow-up to `082f3807` on this branch.

## Changes Made

- **Add `prompts/entity_type/grc.yml`** — generated verbatim from `prompt_v1`, carrying 22 GRC entity types (`Framework`, `Regulation`, `Requirement`, `Control`, `Risk`, `Asset`, `Finding`, `RemediationAction`, …) plus both few-shot example sets. Force-added: `prompts/entity_type/` is gitignored (`.gitignore:62`) as a dropbox for user-supplied profiles, but this one is a project artifact and must travel with the repo.
- **Revert both import sites** in `lightrag/lightrag.py:38` and `lightrag/operate.py:77` back to stock `lightrag.prompt`.
- **Delete `lightrag/prompt_v1.py`** (909 lines), now unreferenced.

Net: this reverts `082f3807`'s code coupling and replaces it with configuration.

## Activation

```
ENTITY_TYPE_PROMPT_FILE=grc.yml
```

Documented at `env.example:270`. `.env` is gitignored, so this line is not in the diff.

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed
- [x] Documentation updated (if necessary) — `env.example:270` already documents the knob; profile is self-documenting
- [x] Unit tests added (if applicable) — no new tests; this restores 10 pre-existing tests that `082f3807` broke

## Additional Notes

**Verification.** End-to-end ingest through the config path produced `regulation`, `requirement`, `framework`, `control`, `role`, `finding`, `remediationaction` — including types that exist *only* in the GRC profile, confirming the YAML is actually driving extraction. All five query modes (`local`/`global`/`hybrid`/`naive`/`mix`) return correct graph traversals against the ingested corpus.

`tests/extraction`: **240 passed** (was 230 passed / 10 failed on this branch before the change).

Full suite: **6427 passed, 7 failed**. The 7 failures are pre-existing and unrelated — verified identical with these changes stashed. They are SVG-rasterization tests (missing optional render dependency) and two auth path-prefix tests. A further 272 `tests/setup` failures on this machine are environmental only: `scripts/setup/setup.sh` requires Bash 4+ and macOS ships Bash 3.2.

**Known gap — the reason this is not a full replacement.** `prompt_v1` also defined `default_relationship_types_guidance`, a 24-verb controlled vocabulary (`implements`, `part_of`, `maps_to`, `complies_with`, …). That key has **no config path**: the profile schema has no field for it and the extraction templates have no slot to render it. Measured on a live ingest, 7/10 relationship keywords still land inside that vocabulary by inference from the GRC entity types alone — but nothing constrains the other 3 (`adopts`, `has_control`, `has_remediation` instead of `complies_with`, `applies_to`, `remediates`). Free-form keywords fragment the graph, which matters for exactly the queries a GRC graph exists to answer.

The vocabulary is preserved in git history:

```bash
git show 082f3807:lightrag/prompt_v1.py
```

Widening `EntityExtractionPromptProfile` with a `relationship_types_guidance` field, plus a matching template slot, is the concrete follow-up. That change is **not** purely additive — the stock templates have no slot — so it belongs in its own PR. Any such override should be placeholder-validated at load time, since a template with a missing or unknown `{...}` raises `KeyError` deep inside the extraction worker rather than at startup.

Of the 12 `PROMPTS` keys `prompt_v1` changed, this PR covers the 3 that determine the taxonomy. The other 9 are wording refinements (GRC role framing, user/continue prompts, `keywords_extraction`, `summarize_entity_descriptions`) that can be evaluated separately.


devsoftx:
Opened against the wrong repository by mistake — this is a fork-local configuration change for devsoftx/LightRAG and is not intended for upstream. Apologies for the noise.

PR #3775, opened 2026-08-31
fix(parser/mineru): keep MinerU chart items instead of dropping them

### LINKED ISSUE #3774: MinerU parser silently drops 'chart' items (images + captions lost)
### Summary

The MinerU importer drops every `content_list.json` item of `"type": "chart"` — the picture is never copied as an asset, the caption/footnote text is lost, and **no log entry is produced**. On a real 342-page book all 5 charts vanished while 587 images / 25 tables / 21 equations were imported correctly.

### Where it happens

**LightRAG side** — `lightrag/parser/external/mineru/ir_builder.py` (main @ `89f1f1d`, identical in v1.5.6):

```python
# line 371
if item_type in {"image", "picture", "drawing"}:
    drawing, asset = self._build_ir_drawing(item, raw_dir, seen_assets)
    ...
```

`chart` is not in that set, so the item reaches the generic text fallback at line 387:

```python
if _append_text(_coerce_text(item)):
    _record_position(item)
```

and `_coerce_text` (line 607) only looks at `text` / `content` / `body` / `code_body`:

```python
def _coerce_text(item: dict) -> str:
    for key in ("text", "content", "body", "code_body"):
```

For a chart item `content` is `""`, so `_append_text` returns `False` and the item is discarded silently — no `IRDrawing`, no `AssetSpec`, no warning. Compare with `table`, where an empty body is at least logged (`_build_ir_table` "already logged the drop").

**MinerU side** — `mineru/utils/enum_class.py` defines `CHART = 'chart'` on `BlockType`, `ContentType` and `ContentTypeV2`, plus `CHART_CAPTION = 'chart_caption'` / `CHART_FOOTNOTE = 'chart_footnote'`. So MinerU >= 3.x emits charts as a first-class content type with its own caption fields (`chart_caption` / `chart_footnote`), not `image_caption` / `image_footnote`.

### Example item (real output, abridged)

```json
{
  "type": "chart",
  "img_path": "images/552c4d95012a5a67d41bb379ffe41d59d8c117e8795f1e13a908cb1af73184af.jpg",
  "content": "",
  "chart_caption": ["Taster unterschiedlich lang drücken"],
  "chart_footnote": ["Abbildung 4-17. Das obere Diagramm zeigt die Spannung am Trigger (Pin 2) …"],
  "bbox": [146, 240, 601, 466],
  "page_idx": 166
}
```

### Reproduction / impact

Book "Make: Elektronik" (342 pages), `content_list.json` type histogram: `text 3089, image 587, page_number 324, footer 319, header 316, table 25, equation 21, code 6, chart 5, ref_text 2`.

Running `MinerUIRBuilder().normalize_from_workdir(raw_dir, ...)` on that bundle:

| | drawings in IRDoc | assets | from chart items |
|---|---|---|---|
| main (`89f1f1d`) | 587 | 587 | 0 |
| with the fix | 592 | 592 | 5 (4 with caption, 1 with footnote) |

The lost items are exactly the diagrams / characteristic curves / oscillograms — i.e. the figures whose captions carry the most retrieval-relevant text.

### Suggested fix

Treat `chart` like `image` in the dispatch set and let `_build_ir_drawing` also read `chart_caption` / `chart_footnote`. PR follows.

boeschbenjamin-jpg:
Fixes #3774

### Problem

MinerU (>= 3.x) emits `content_list.json` items with `"type": "chart"` (`img_path`, `chart_caption`, `chart_footnote`; `content` often empty). `ir_builder.py:371` only dispatched `image` / `picture` / `drawing` to `_build_ir_drawing`, so chart items fell through to the text fallback. With an empty `content` `_coerce_text` finds nothing and the item was discarded with no drawing, no asset and no log entry; with a non-empty one only the text survived, while picture, caption and footnote were lost.

Real impact: a 342-page book lost all 5 charts (plots, characteristic curves, oscillograms) while 587 images / 25 tables / 21 equations were imported fine.

### Fix

- `chart` added to the drawing dispatch set → `IRDrawing` + `{{IMG:k}}` placeholder + `AssetSpec` for `img_path`, exactly like an image item.
- `_build_ir_drawing` also reads MinerU's chart-specific `chart_caption` / `chart_footnote` (image fields keep precedence, so nothing changes for existing items).
- The item's text payload — MinerU's recognised picture body, e.g. a chart's data table or OCR'd labels — is appended after the placeholder so it stays in the retrievable chunk text. It is read through `_coerce_text`, the module's own contract for "this item's text payload": that covers the keys the text fallback used to reach (`text` / `content` / `body` / `code_body`) and its `isinstance(val, str)` guard keeps a non-string payload (MinerU's v2 `content_list` nests the path under `content.image_source.path`) from entering the body as a Python repr.
- This last point deliberately changes `image` items too: they carry `content` as well, and it had always been discarded for them.
- Module docstring updated.

No other behaviour touched; the change is confined to `lightrag/parser/external/mineru/ir_builder.py` and its test file.

### Test

Four new tests in `tests/parser/external/mineru/test_ir_builder.py`:

- `test_adapter_chart_item_becomes_drawing` — built from real MinerU output (chart with caption, chart with footnote): asserts 2 drawings, caption/footnote/`fmt`/`self_ref`, the `{{IMG:…}}` placeholder in the block body, and asset source present/absent. Fails on `main` with `assert 0 == 2`.
- `test_adapter_drawing_body_content_kept_next_to_placeholder` — pins the exact block body for a chart with a markdown data table plus an image with OCR text.
- `test_adapter_drawing_body_text_read_from_all_payload_keys` — pins that a picture body carried under `text` / `body` is kept next to the placeholder.
- `test_adapter_drawing_non_string_body_never_enters_block_text` — pins that a non-string `content` never reaches the block text as a repr.

```
$ ./scripts/test.sh tests/parser/external/mineru
106 passed in 2.50s
```

Additionally verified offline against the stored 342-page bundle: drawings 587 → 592, assets 587 → 592, 5 of them from chart items.

https://claude.ai/code/session_01GqSL71EiBLCnvv7VnvKCDj

### DIFF
--- a/lightrag/parser/external/mineru/ir_builder.py
+++ b/lightrag/parser/external/mineru/ir_builder.py
@@ -27,8 +27,10 @@ Conversion rules (informed by spec §3-§六):
   is reserved for explicit 2D-array / non-HTML compatibility inputs. A real
   HTML ``<thead>`` populates ``table_header`` (per spec §5); otherwise the
   adapter does not guess a header row.
-- ``image`` / ``picture`` / ``drawing`` → IRDrawing + ``{{IMG:k}}`` placeholder.
-  Asset bytes are referenced via ``img_path`` relative to the raw dir.
+- ``image`` / ``picture`` / ``drawing`` / ``chart`` → IRDrawing + ``{{IMG:k}}``
+  placeholder. Asset bytes are referenced via ``img_path`` relative to the raw
+  dir. ``chart`` items (MinerU >= 3.x) carry their text in ``chart_caption`` /
+  ``chart_footnote`` instead of ``image_caption`` / ``image_footnote``.
 - ``equation`` → IREquation. ``is_block`` is decided by whether
   ``text_format=="block"`` (MinerU explicit flag) OR ``text_level==0`` with
   no inline neighbours; otherwise inline. The latex string is preserved
@@ -368,7 +370,7 @@ class MinerUIRBuilder:
                 _record_position(item)
                 continue
 
-            if item_type in {"image", "picture", "drawing"}:
+            if item_type in {"image", "picture", "drawing", "chart"}:
                 drawing, asset = self._build_ir_drawing(item, raw_dir, seen_assets)
                 placeholder = _next_key("im")
                 drawing.placeholder_key = placeholder
@@ -521,7 +523,11 @@ class MinerUIRBuilder:
     ) -> tuple[IRDrawing, AssetSpec | None]:
         img_path = str(item.get("img_path") or item.get("path") or "")
         src_val = str(item.get("src") or "")
-        captions = item.get("image_caption") or item.get("captions")
+        captions = (
+            item.get("image_caption")
+            or item.get("chart_caption")
+            or item.get("captions")
+        )
         caption = str(item.get("caption") or "")
         if not caption and isinstance(captions, list) and captions:
             caption = str(captions[0])
@@ -563,7 +569,11 @@ class MinerUIRBuilder:
             asset_ref=ref,
             fmt=fmt,
             caption=caption,
-            footnotes=_as_str_list(item.get("image_footnote") or item.get("footnotes")),
+            footnotes=_as_str_list(
+                item.get("image_footnote")
+                or item.get("chart_footnote")
+                or item.get("footnotes")
+            ),
             src=src_val,
         )
         return drawing, asset
--- a/lightrag/parser/external/mineru/ir_builder.py
+++ b/lightrag/parser/external/mineru/ir_builder.py
@@ -30,7 +30,9 @@ Conversion rules (informed by spec §3-§六):
 - ``image`` / ``picture`` / ``drawing`` / ``chart`` → IRDrawing + ``{{IMG:k}}``
   placeholder. Asset bytes are referenced via ``img_path`` relative to the raw
   dir. ``chart`` items (MinerU >= 3.x) carry their text in ``chart_caption`` /
-  ``chart_footnote`` instead of ``image_caption`` / ``image_footnote``.
+  ``chart_footnote`` instead of ``image_caption`` / ``image_footnote``. A
+  non-empty ``content`` (MinerU's recognised picture body — often empty, but
+  populated e.g. with a chart's data table) is appended after the placeholder.
 - ``equation`` → IREquation. ``is_block`` is decided by whether
   ``text_format=="block"`` (MinerU explicit flag) OR ``text_level==0`` with
   no inline neighbours; otherwise inline. The latex string is preserved
@@ -379,6 +381,13 @@ class MinerUIRBuilder:
                     assets.append(asset)
                 cb_drawings.append(drawing)
                 cb_lines.append(f"{{{{IMG:{placeholder}}}}}")
+                # MinerU fills ``content`` when the model recognised the
+                # picture's data (a chart's data table, OCR'd labels). Keep it
+                # in the block body — it is the only retrievable form of that
+                # text when no VLM analysis runs.
+                body_text = str(item.get("content") or "").strip()
+                if body_text:
+                    cb_lines.append(body_text)
                 _record_position(item)
                 continue
 
--- a/tests/parser/external/mineru/test_ir_builder.py
+++ b/tests/parser/external/mineru/test_ir_builder.py
@@ -801,7 +801,7 @@ def test_adapter_drawing_asset_source_only_when_file_exists(
 @pytest.mark.offline
 def test_adapter_chart_item_becomes_drawing(tmp_path: Path) -> None:
     """MinerU emits ``chart`` items (diagrams, plots, oscillograms) with
-    ``img_path`` + ``chart_caption`` / ``chart_footnote`` and an empty
+    ``img_path`` + ``chart_caption`` / ``chart_footnote`` and an often empty
     ``content``. They must become drawings like ``image`` items — otherwise
     the text fallback silently drops picture, caption and footnote.
     """
@@ -848,6 +848,49 @@ def test_adapter_chart_item_becomes_drawing(tmp_path: Path) -> None:
     by_ref = {a.ref: a for a in ir.assets}
     assert by_ref["images/chart_001.jpg"].source is not None
     assert by_ref["images/chart_002.jpg"].source is None
+
+
+@pytest.mark.offline
+def test_adapter_drawing_body_content_kept_next_to_placeholder(
+    tmp_path: Path,
+) -> None:
+    """MinerU fills ``content`` on ``chart`` / ``image`` items whenever the
+    model recognised the picture's data (e.g. a chart's data table). That
+    text must stay in the block body next to the ``{{IMG:k}}`` placeholder —
+    it is the only retrievable form of it when no VLM analysis runs.
+    """
+    raw = _write_bundle(
+        tmp_path,
+        [
+            {"type": "text", "text": "Results", "text_level": 1},
+            {
+                "type": "chart",
+                "img_path": "images/c1.jpg",
+                "content": "| year | value |\n|---|---|\n| 2024 | 42 |",
+                "chart_caption": ["Abb. 3-58"],
+                "chart_footnote": ["src: X"],
+            },
+            {
+                "type": "image",
+                "img_path": "images/i1.jpg",
+                "content": "OCR'd label text",
+            },
+        ],
+    )
+    ir = MinerUIRBuilder().normalize_from_workdir(raw, document_name="c.pdf")
+
+    block = ir.blocks[0]
+    chart, image = block.drawings
+    assert block.content_template == (
+        "# Results\n"
+        f"{{{{IMG:{chart.placeholder_key}}}}}\n"
+        "| year | value |\n|---|---|\n| 2024 | 42 |\n"
+        f"{{{{IMG:{image.placeholder_key}}}}}\n"
+        "OCR'd label text"
+    )
+    # Caption / footnote still travel on the drawing, not in the body text.
+    assert chart.caption == "Abb. 3-58"
+    assert chart.footnotes == ["src: X"]
 
 
 @pytest.mark.offline
--- a/tests/parser/external/mineru/test_ir_builder.py
+++ b/tests/parser/external/mineru/test_ir_builder.py
@@ -799,6 +799,58 @@ def test_adapter_drawing_asset_source_only_when_file_exists(
 
 
 @pytest.mark.offline
+def test_adapter_chart_item_becomes_drawing(tmp_path: Path) -> None:
+    """MinerU emits ``chart`` items (diagrams, plots, oscillograms) with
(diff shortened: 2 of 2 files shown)


danielaskdd:
## Review of `6e0730d`

Verdict: the fix is correct and I'd merge it. #3777 (same root cause, dispatch only, no `chart_caption` / `chart_footnote` mapping) has been closed in favour of this one. Two follow-ups below — the first is worth folding into this PR, the second is optional hardening.

### Verified independently

- **Defect reproduced on `main` (`89f1f1d`)**, and it has two shapes, not one:
  - chart *without* `content` → dropped entirely: no `IRDrawing`, no `AssetSpec`, and not even a `page_idx` position (matches the histogram in #3774);
  - chart *with* non-empty `content` → the text survives via `_coerce_text` (`ir_builder.py:607` reads `content`), but the image, caption and footnote are lost.
- **With the patch**, against the exact item shape from #3774 plus a footnote: one drawing, `caption='Abb. 3-58'`, `footnotes=['src: X']`, asset declared, `{{IMG:im1}}` in the block body, bbox position recorded.
- `./scripts/test.sh tests/parser/external/mineru` → **103 passed** at this head; `test_adapter_chart_item_becomes_drawing` fails on `main` as claimed. CI is 4/4 green.
- **Upstream cross-check**: `CHART` is absent from `mineru/utils/enum_class.py` at `release-2.0.0` and `release-2.5.4`, and present on master (3.4.5) — the "MinerU >= 3.x" note in the docstring is accurate. Upstream `vlm_middle_json_mkcontent.py` confirms `chart` carries `img_path` / `content` / `chart_caption` / `chart_footnote` / `sub_type`.
- **Dispatch completeness**: in MinerU's v1 `content_list`, the only item types carrying `img_path` are `image`, `table` and `chart`; `table` has its own branch, so this closes the last gap rather than one instance of a wider class.
- **No new hard-failure path**: a chart drawing whose asset file is absent resolves to `_skipped_result("image file not found: …")` in `_analyze_drawing` (`pipeline.py:6813`) — a benign skip, not a document failure. The test's `chart_002.jpg` covers exactly that case.

### Finding 1 (medium) — the chart's `content` is now dropped

`_build_ir_drawing` reads `img_path`, captions and footnotes, but not `content`. On the chart path that is a net loss relative to `main`, because the text fallback used to surface it:

```
input:  {"type": "chart", "img_path": "images/c1.jpg",
         "content": "| year | value |\n|---|---|\n| 2024 | 42 |",
         "chart_caption": ["Abb. 3-58"], "chart_footnote": ["src: X"]}

main    → block: '# Results\n| year | value |\n|---|---|\n| 2024 | 42 |'   (text kept, image lost)
6e0730d → block: '# Results\n{{IMG:im1}}'                                  (image kept, text lost)
```

Impact is bounded but real: in the #3774 bundle `content` is `""` throughout, so nothing regresses there, and when it *is* populated the VLM can re-derive a description from the image. But that only holds if `process_options` enables image analysis and a VLM role is configured — otherwise the chart contributes only its caption via `<drawing caption="…" />` (`sidecar/writer.py:535`), and MinerU's own recognised data table is gone. Upstream keeps both: its markdown renders the image *plus* a `<details><summary>chart content</summary>…</details>` block (`_build_visual_details_block`).

Suggestion: keep the placeholder and re-append the body text in the `chart` branch, so it stays in the retrievable chunk text:

```python
if item_type in {"image", "picture", "drawing", "chart"}:
    drawing, asset = self._build_ir_drawing(item, raw_dir, seen_assets)
    ...
    cb_lines.append(f"{{{{IMG:{placeholder}}}}}")
    body_text = str(item.get("content") or "").strip()
    if body_text:
        cb_lines.append(body_text)      # MinerU's recognised chart/image body
    _record_position(item)
```

(`IRDrawing.extras` is an alternative — `writer.py:640` persists it into `drawings.json` — but extras never reaches the block text, so it buys traceability only, not retrieval.)

Worth noting this is not chart-specific in principle: v1 `image` items carry `content` too (`para_content['content'] = image_content` in upstream `mkcontent`), and it has always been discarded for them. Charts are just the path where this PR changes the outcome, so handling both here would be consistent.

### Finding 2 (low) — the silent-drop hole is still open

The fallback at `ir_builder.py:389` discards any item whose `_coerce_text` is empty without a single log line — which is precisely why this bug stayed invisible until a 342-page book was audited by hand (#3774 makes the same point: an empty `table` body at least gets a `logger.debug`). `chart` is fixed, but the next image-format type MinerU adds will vanish the same way. MinerU's v2 `content_list` schema, for instance, nests the path under `content.image_source.path`, so its `image` / `chart` items would produce empty text and disappear silently as well.

A one-liner would have surfaced this in the first ingest:

```python
if _append_text(_coerce_text(item)):
    _record_position(item)
elif item_type not in _KNOWN_EMPTY_TYPES:
    logger.debug(
        "[mineru_ir_builder] dropping item with no usable text "
        "(type=%s, page_idx=%s)", item_type, item.get("page_idx")
    )
```

Fine as a separate PR if you'd rather keep this one at 20 lines.

### Nit

The module docstring and the test docstring both describe chart `content` as empty (`"``content`` is empty"` / `"an empty ``content``"`). It is empty in the sampled bundle, but upstream populates it whenever the model recognises the chart's data, so it reads as a schema guarantee that doesn't hold. Reword to "often empty" and the docstrings match Finding 1's handling.

---

boeschbenjamin-jpg:
Thanks for the thorough review — especially for reproducing both shapes of the defect and for the upstream cross-check.

Finding 1 and the nit are addressed in `f94d34c`: a non-empty `content` is now appended right after the `{{IMG:k}}` placeholder, for `image` items as well as `chart` ones, so the recognised picture body stays in the retrievable chunk text. Caption and footnote keep travelling on the drawing. New test `test_adapter_drawing_body_content_kept_next_to_placeholder` uses your example item (chart with a markdown data table plus an image with OCR text) and asserts the exact block body; the docstrings now say `content` is *often* empty rather than empty.

Finding 2 (the silent-drop log in the fallback) comes as a separate PR so this one stays small.

danielaskdd:
## Review of `f94d34c`

Thanks for the follow-up commit. Re-checked the previous round's three items against the new head: the **nit is fully addressed** (module docstring and the test docstring both read "often empty" now), **Finding 2 is correctly handed off** to #3789, but **Finding 1 is only half closed** — and the way it was closed introduces one problem that `main` does not have.

### Re-verified at this head

- `tests/parser/external/mineru` → **104 passed**. Reverting `ir_builder.py` to `main` turns both new tests red (`test_adapter_chart_item_becomes_drawing` → `assert 0 == 2`; `test_adapter_drawing_body_content_kept_next_to_placeholder` → `not enough values to unpack`), so the tests do discriminate.
- `ruff check` / `ruff format --check` clean; CI 4/4 green.
- Merged this branch with #3789 locally: `ort` auto-merges with no conflict, the combined dispatch is coherent (`chart` no longer reaches the fallback once it has its own branch), and the merged tree is green. Either merge order works.

### Finding 1 is not fully closed, and the fix regresses one case

My Finding 1 was that the chart path lost text *relative to `main`*, **because the text fallback used to surface it**. That fallback goes through `_coerce_text`, which reads four keys in order — `("text", "content", "body", "code_body")` — **and guards each with `isinstance(val, str)`**. What landed reads one key and drops the guard:

```python
body_text = str(item.get("content") or "").strip()
```

(That is my own suggested snippet from the last round — the omission originates in my suggestion, not in your reading of it. Sorry for the misdirection.)

Same chart items run on `main` (`89f1f1d`) and on this head:

| chart item | `main` | `f94d34c` | |
|---|---|---|---|
| text under `text` | `['chart data table: 2024 = 42']` | `['{{IMG:im1}}']` | text still lost |
| text under `body` | `['oscillogram legend']` | `['{{IMG:im1}}']` | text still lost |
| non-string `content` (v2 shape) | `[]` | `["{{IMG:im1}}\n{'image_source': {'path': 'images/a.png'}}"]` | **new: repr injected** |
| string `content` | `['\| y \| v \|']` | `['{{IMG:im1}}\n\| y \| v \|']` | correct |

Rows 1–2 are the original net loss, just reached through a different key. Row 3 is new: a non-string `content` is `str()`-ed into the block body, so a Python repr ends up in the retrievable chunk text and gets chunked, embedded and returned by retrieval. `main` cannot produce that, because `_coerce_text`'s `isinstance` check filters it. And that payload shape is exactly the v2 `content_list` one I named in Finding 2 (`content.image_source.path`), so it is reachable, not hypothetical.

**One line fixes both**, by reusing the module's own contract for "this item's text payload":

```diff
-                body_text = str(item.get("content") or "").strip()
+                body_text = _coerce_text(item).strip()
```

I applied that on your branch and re-ran all four cases: rows 1–2 recover their text (`'{{IMG:im1}}\nchart data table: 2024 = 42'`), row 3 goes back to a clean `'{{IMG:im1}}'`, row 4 is unchanged — `tests/parser/external/mineru` still **104 passed**. Worth one extra case pinning "a non-string `content` must not enter the block body", since that is the part no current test would catch.

### Minor — the PR description no longer matches the code

`f94d34c` invalidated three statements in the description, and they will travel into the merge commit:

- "`content` is empty" — the docstrings were corrected, the description was not;
- "No other behaviour touched" — `image` items changed behaviour too (deliberately, per Finding 1's "handling both here would be consistent");
- "20 lines in `lightrag/parser/external/mineru/ir_builder.py`" — now +119 / −5 across 2 files.

### Notes, not blocking

- `_build_ir_drawing` still takes only `captions[0]`, so a chart with several `chart_caption` entries keeps the first. That is pre-existing `image` behaviour and this PR merely extends the same rule to charts — please leave it here.
- When image analysis is enabled, MinerU's recognised `content` and the VLM description of the same picture can both land in the body. Upstream's markdown does the same (image + chart content block), so this is acceptable; a sentence in the docstring saying so would save the next reader the question.

With the one-line change above (and ideally the extra test), this is good to merge from my side.

---

danielaskdd:
Pushed `c523c02` to this branch myself (maintainer edit) so the round can close here.

**The one-line change from my last review**, plus the extra test I asked for:

- `body_text = _coerce_text(item).strip()` instead of `str(item.get("content") or "")`. Recovers the text of a chart carrying it under `text` / `body`, and stops a non-string `content` (the v2 `content_list` shape) from being `str()`-ed into the block body as a Python repr.
- `test_adapter_drawing_body_text_read_from_all_payload_keys` and `test_adapter_drawing_non_string_body_never_enters_block_text`. Both fail on the previous line with an assertion mismatch (not an error), so they discriminate.
- Module docstring and the inline comment now describe the payload keys and the `isinstance` guard rather than `content` alone, and note that MinerU's recognised text and a VLM description can coexist in the body (as upstream's markdown does).
- PR description refreshed — the three statements `f94d34c` had invalidated ("content is empty", "no other behaviour touched", "20 lines") now match the code.

`main` is merged into the branch (`04075b3`) to keep it current; the diff against `main` is unchanged in scope. `./scripts/test.sh tests/parser/external/mineru` → **106 passed**.

Good to merge from my side once CI is green.

PR #3776, opened 2026-08-31
fix(query): enforce meaningful RAG query length

danielaskdd:
## Description

Enforce a meaningful minimum length for retrieval-augmented queries while keeping direct LLM paths unrestricted. RAG queries now require an English-equivalent weight of at least 3 after trimming outer whitespace, with each Han character counting as 2 and every other Unicode code point counting as 1.

The WebUI now validates query input before submission and shows a localized, persistent floating error notice above the shared query composer without reserving layout space.

## Related Issues

N/A (no linked issue).

## Changes Made

- Add a shared backend RAG query validator and apply it to the core query APIs and all `/query*` routes.
- Validate the cleaned query after Ollama mode-prefix parsing for RAG branches of `/api/chat`.
- Keep `bypass`, Open WebUI metadata tasks forwarded directly to the LLM, and `/api/generate` exempt from the minimum.
- Share `/mode query` prefix parsing between `/webui` and `/workspace`, including malformed-prefix and unsupported-mode feedback.
- Show all client-side query validation errors in an accessible floating notice above the composer on both entries; keep the notice while editing, then replace or dismiss it on the next send.
- Reduce the admin retrieval page's bottom clearance from 48px to 32px, matching the document panel while still clearing the fixed connection-status indicator.
- Localize the validation message and dismiss action in all supported locales.
- Document the weighted minimum and direct-LLM exemptions in the English and Chinese API server guides.
- Add regression coverage for ASCII, Han, mixed-language, mode-prefix, entry alignment, floating feedback, bottom clearance, streaming, bypass, and generate paths.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Validation completed locally:

- `./scripts/test.sh tests/api/routes`: 545 passed, 5 skipped
- Focused backend query validation suite: 55 passed
- `bun test src/features src/utils/queryValidation.test.ts`: 125 passed
- `uv run ruff check` and `uv run ruff format --check`: passed
- `bun run lint`: passed
- `bun run build`: passed

A live mobile-browser screenshot check was not available in the current environment; responsive behavior is covered structurally by the upward, absolutely positioned notice and targeted source-level regression tests.


danielaskdd:
@codex review

PR #3777, opened 2026-08-31
Fix MinerU chart items dropped from IR

pyy-hotyouth:
MinerU emits type="chart" items for charts rendered as images, but _normalize_content_list only mapped image/picture/drawing to IRDrawing. chart fell through to the plain-text fallback, which reads no text field from an image-format item, so the chart silently vanished from the sidecar (blocks.jsonl / drawings.json).

Treat chart as a drawing so it survives into the IR and sidecar.

## Description

 MinerU emits `type="chart"` items for charts rendered as images, but
  `MinerUIRBuilder._normalize_content_list` only mapped `image` / `picture` /
  `drawing` to `IRDrawing`. A `chart` item fell through to the plain-text
  fallback, which reads no text field from an image-format item, so the chart
  silently vanished from the sidecar (`blocks.jsonl` / `drawings.json`).

  This change treats `chart` as a drawing so charts survive into the IR and
  sidecar.

## Related Issues

None

## Changes Made

 - `lightrag/parser/external/mineru/ir_builder.py`: added `chart` to the
    drawing branch so chart items are routed through `_build_ir_drawing`
    (producing an `IRDrawing` + `AssetSpec` and a `{{IMG:k}}` placeholder), and
    updated the module docstring.
  - `tests/parser/external/mineru/test_ir_builder.py`: added
    `test_adapter_chart_treated_as_drawing` regression test.

## Checklist

  - [x] Changes tested locally
  - [x] Code reviewed

## Additional Notes

  `./scripts/test.sh tests/parser/external/mineru` → 103 passed; `ruff check`
  on both changed files passed.

### DIFF
--- a/lightrag/parser/external/mineru/ir_builder.py
+++ b/lightrag/parser/external/mineru/ir_builder.py
@@ -27,8 +27,10 @@ Conversion rules (informed by spec §3-§六):
   is reserved for explicit 2D-array / non-HTML compatibility inputs. A real
   HTML ``<thead>`` populates ``table_header`` (per spec §5); otherwise the
   adapter does not guess a header row.
-- ``image`` / ``picture`` / ``drawing`` → IRDrawing + ``{{IMG:k}}`` placeholder.
-  Asset bytes are referenced via ``img_path`` relative to the raw dir.
+- ``image`` / ``picture`` / ``drawing`` / ``chart`` → IRDrawing + ``{{IMG:k}}``
+  placeholder. Asset bytes are referenced via ``img_path`` relative to the raw
+  dir. ``chart`` is a chart rendered as an image by MinerU, so it is treated as
+  a drawing.
 - ``equation`` → IREquation. ``is_block`` is decided by whether
   ``text_format=="block"`` (MinerU explicit flag) OR ``text_level==0`` with
   no inline neighbours; otherwise inline. The latex string is preserved
@@ -368,7 +370,7 @@ class MinerUIRBuilder:
                 _record_position(item)
                 continue
 
-            if item_type in {"image", "picture", "drawing"}:
+            if item_type in {"image", "picture", "drawing", "chart"}:
                 drawing, asset = self._build_ir_drawing(item, raw_dir, seen_assets)
                 placeholder = _next_key("im")
                 drawing.placeholder_key = placeholder
--- a/tests/parser/external/mineru/test_ir_builder.py
+++ b/tests/parser/external/mineru/test_ir_builder.py
@@ -179,6 +179,44 @@ def test_adapter_table_and_drawing_and_equation(tmp_path: Path) -> None:
 
 
 @pytest.mark.offline
+def test_adapter_chart_treated_as_drawing(tmp_path: Path) -> None:
+    """MinerU emits ``type="chart"`` items — a chart rendered as an image.
+
+    The adapter must treat them as drawings so the chart survives into the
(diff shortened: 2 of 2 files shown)


lightrag/parser/external/mineru/ir_builder.py
```diff
@@ -368,7 +370,7 @@ def _append_text(text: str) -> bool:
                 _record_position(item)
                 continue
 
-            if item_type in {"image", "picture", "drawing"}:
+            if item_type in {"image", "picture", "drawing", "chart"}:
```
AmirF194, line 373: I ran a repro against this PR's head (`b9eddcb`) in a clean container.

The dispatch fix is correct (a `chart` item now becomes a drawing instead of being dropped), but `_build_ir_drawing` at line 526 still reads only `image_caption`/`captions` for the caption and `image_footnote`/`footnotes` for the footnote. MinerU's `chart` items carry that text under `chart_caption`/`chart_footnote` instead, so the chart survives but its caption and footnote are silently dropped.

```
$ docker run --rm python:3.12-slim ... pytest -s -k chart_caption_field_repro
# input: {"type": "chart", "chart_caption": ["Abbildung 3-58"], "chart_footnote": ["Note text"], ...}
num drawings: 1
caption: ''
footnotes: []
```

Same input against #3775's head (`6e0730d`) gives `caption: 'Abbildung 3-58'` and `footnotes: ['Note text']`, because that PR also adds `chart_caption`/`chart_footnote` as fallback keys at lines 526 and 568. It's open since 10:37Z today (this one opened 12:16Z), targets the same root cause, and already has green CI. Worth checking before merging this one, since taking both would just re-add the same dispatch line twice.

danielaskdd:
Closing as a duplicate of #3775, which targets the same root cause and is a strict superset of this change.

Both PRs add `chart` to the drawing dispatch set with a byte-identical line. #3775 additionally teaches `_build_ir_drawing` to read MinerU's `chart_caption` / `chart_footnote`, which this PR does not. I ran both heads against the item shape MinerU actually emits (`{"type": "chart", "img_path": …, "content": …, "chart_caption": ["Abb. 3-58"], "chart_footnote": ["src: X"]}`):

| head | drawing emitted | caption | footnotes |
|---|---|---|---|
| this PR (`b9eddcb`) | yes | `''` | `[]` |
| #3775 (`6e0730d`) | yes | `'Abb. 3-58'` | `['src: X']` |

The caption is not cosmetic downstream: it is persisted to `drawings.json` (`sidecar/writer.py:635`), rendered into the block body as the `<drawing caption="…">` attribute (`writer.py:535`), and passed to the VLM as `captions` during multimodal analysis (`pipeline.py:6872` / `:7011`, where an empty value degrades to `"n/a"`). For a chart, the caption is usually the most retrieval-relevant text on the figure.

The regression test here passes only because the fixture puts `image_caption` on the chart item — a key real MinerU chart items never carry (upstream `vlm_middle_json_mkcontent.py` emits `chart_caption` / `chart_footnote` for `BlockType.CHART`), so the missing mapping stays invisible.

#3775 also links a reproduction (#3774: 342-page book, 5 charts dropped, drawings 587 → 592 with the fix) and has 4/4 green checks, while this PR has no check runs yet.

The diagnosis in the description was correct and matches what I reproduced independently — thanks for catching the same bug. Since the dispatch line is identical, merging both would just re-apply it, so this one is being closed in favour of #3775.

---

PR #3783, opened 2026-09-01
Add React component tests and test harness isolation checks

danielaskdd:
## Description

Establishes a comprehensive testing infrastructure for WebUI React components with proper DOM isolation and bundle safety guarantees. Converts the `TouchDescriptionPopover` test from source-text assertions to actual rendered behavior testing, and adds automated checks to prevent test dependencies from leaking into the production bundle.

## Related Issues

Addresses the need for reliable component testing that verifies rendered behavior rather than source code structure, and ensures the test harness remains isolated from production code.

### Test Infrastructure
- **`src/test/happydom.ts`**: Registers happy-dom globally as the first preload, providing a DOM for all test files with correct `window.location` and viewport dimensions
- **`src/test/setup.ts`**: Second preload that wires Testing Library matchers and cleanup into bun:test
- **`src/test/render.tsx`**: Shared render helper that provides a fixed English i18n instance, preventing tests from depending on ambient state like localStorage or settings migrations
- **`src/test/domGlobals.ts`**: Helpers for tests that need to assert DOM-less code paths (`withoutDomGlobals`, `restoreDomGlobals`), capturing real descriptors at import time to restore them after test cleanup
- **`src/test/matchers.d.ts`**: TypeScript declaration merge to register jest-dom matchers with bun:test's type surface

### Test Harness Isolation
- **`src/test/harnessIsolation.test.ts`**: Four automated checks ensuring:
  1. No production module imports test harness or test-only packages
  2. Test infrastructure is reached only through bunfig.toml preloads, never imports
  3. No test file deletes DOM globals outside the domGlobals helper
  4. DOM packages are devDependencies, not bundled

### Component Tests
- **`src/components/ui/TouchDescriptionPopover.test.tsx`**: Replaces source-text assertions with rendered behavior tests:
  - Verifies the trigger renders only on touch devices
  - Confirms accessible naming (button labeled "Deployment information", not the description text)
  - Tests that description is hidden until triggered
  - Asserts visibility after user interaction
  - Validates i18n key presence in all locales
- **Updated existing tests** (`documentTitle.test.ts`, `documentTitleSync.test.ts`, `brandTitle.test.ts`, `workspaceEntry.test.ts`): Migrate from bare `delete globalThis.window/document` to `restoreDomGlobals()` to work with process-wide DOM preload

### Configuration & CI
- **`bunfig.toml`**: Configures test preloads with correct ordering (DOM before Testing Library)
[...]

PR #3784, opened 2026-09-01
Fail legibly when bun test runs without the DOM preload

danielaskdd:
## Description

Follow-up to #3783. That PR gave `bun test` a DOM through a `bunfig.toml` preload; this one closes the two ways that setup wastes someone's time, both hit in practice right after it merged.

**1. Running `bun test` from the repository root silently loads no DOM.** Bun resolves `bunfig.toml` — and the preload paths inside it — against the working directory, so `bun test lightrag_webui/src/...` from the root gets no preload. Nothing reports that. Pure logic tests still pass; only the component tests break, deep inside user-event:

```
TypeError: undefined is not an object (evaluating 'document[isPrepared]')
    at prepareDocument (@testing-library/user-event/.../prepareDocument.js:10:9)
```

which names neither the cause nor the fix, and reads like a broken dependency. Passing a path to `bun test --config` does not work around it either — verified — because the preload paths inside the file are still resolved against the CWD.

**2. A branch switch across a `package.json` change leaves `node_modules` stale.** #3783 was the first change ever to add WebUI dependencies, so `git checkout` followed by `bunx tsc --noEmit` produced 10 errors that look like broken code:

```
error TS2307: Cannot find module '@testing-library/react' or its corresponding type declarations.
error TS2339: Property 'toBeInTheDocument' does not exist on type 'Matchers'.
...
```

Only 4 are the real symptom. The other 6 are downstream: `src/test/matchers.d.ts` merges jest-dom's matchers into `bun:test` through an `import type`, which resolves to nothing when the package is absent, so the declaration merge silently fails and every matcher reports as missing. `bun install --frozen-lockfile` clears all 10 at once.

## Related Issues

Follow-up to #3783 (merged). No open issue.

## Changes Made

- **`src/test/domGlobals.ts`**: adds `assertDomAvailable()`, which throws an error naming both the cause and the fix when no DOM is present.
- **`src/test/render.tsx`**: calls it at import time. Import time rather than inside `renderWithProviders` because a test may call `userEvent.setup()` before it renders anything — `TouchDescriptionPopover.test.tsx` does exactly that, and a guard inside the render helper would fire too late for it.
- **`src/test/domGlobals.test.ts`** (new): the tests this helper never had — the guard in both directions, plus `withoutDomGlobals`' restore-on-throw, its `keys` narrowing, and its return value.
- **`AGENTS.md`**: states that every WebUI command runs from `lightrag_webui/` and why; marks `bun install --frozen-lockfile` REQUIRED after a branch switch across a `package.json` change; documents the CWD-sensitivity of the preload under *React component tests*.

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Test Plan

Run from `lightrag_webui/`:

- `bun test` — **560 pass / 0 fail** across 62 files (was 553/61 on main; +7 from the new `domGlobals.test.ts`)
- `bunx tsc --noEmit` — clean
- `bun run lint` — clean
- `bun run build` — succeeds, first-load budget `2380078 bytes` unchanged

The guard was verified against the real failure, not just in a unit test. From the repository root, before this change the run failed at test 4 with `document[isPrepared]`; after it, the message arrives at module load, before any test body:

```
# Unhandled error between tests
error: No DOM: bun test started without the bunfig.toml preload. Run it from the
lightrag_webui/ directory (`cd lightrag_webui && bun test ...`) — bun resolves
bunfig.toml and its preload paths relative to the working directory, so running
from the repository root loads no DOM.
    at assertDomAvailable (src/test/domGlobals.ts:80:13)
    at src/test/render.tsx:24:1
```

Each new test was mutation-checked — a test that has not been seen to fail proves nothing:

| Mutation | Result |
|---|---|
| guard body replaced with `return` | 6 pass / **1 fail** |
| `withoutDomGlobals`' `finally` restore removed | 3 pass / **4 fail** |
| `keys` parameter ignored, always removes both | 6 pass / **1 fail** |
| (restored) | 7 pass / 0 fail |

## Additional Notes

Both problems are documentation-and-diagnostics only — no production module changes, and the build output is unaffected.

The branch reuses the name from #3783, restarted from `main` after that PR merged (its remote branch was deleted on merge); this is a new pull request, not a reopening.

One deliberate omission: the repository still has 17 tests that `readFileSync` a `.tsx` and assert on substrings. Converting them to real renders is worth doing now that the harness exists, but it is a separate change of a different size and is not bundled here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JCWY6MFANEXGX2h7JzUgDX)_

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -190,9 +190,19 @@ lightrag-gunicorn                                         # Multi-worker (gunico
 ```
 
 ### WebUI
+
+Every command below is run **from `lightrag_webui/`**, not the repository root.
+`bun test` in particular resolves `bunfig.toml` — and the preload paths inside
+it — relative to the working directory, so running it from the root silently
+loads no DOM (see *React component tests*).
+
 ```bash
 cd lightrag_webui
-bun install --frozen-lockfile      # Install dependencies
+bun install --frozen-lockfile      # REQUIRED after any change to package.json /
+                                   # bun.lock, including a branch switch across
+                                   # one: `git checkout` does not update
+                                   # node_modules, and a stale tree surfaces as
+                                   # a wall of TS2307 "Cannot find module".
 bun run dev                        # Dev server (Node + Vite)
 bun run dev:bun                    # Dev server (Bun native)
 bun run build                      # Production build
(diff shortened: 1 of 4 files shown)


danielaskdd:
@codex review

PR #3787, opened 2026-09-01
test(extraction): pin the gleaning guard arithmetic and lazy invariant precompute

danielaskdd:
## Description

Follow-up test coverage for #3760, which is now merged. This PR was previously stacked on it and carried @hata33's two commits so the diff would apply; with #3760 in `main` those became duplicates and conflicted, so the branch was rebased onto `main` and now holds **one commit touching one file** — `tests/extraction/test_extract_entities.py`, additions only. No production code, no existing test modified.

#3760's implementation is correct as merged. This PR only closes gaps in what the suite *proves*.

## Related Issues

- Follow-up test coverage for #3759 / #3760

## Changes Made

Two properties of the invariant-hoisting optimization were unpinned. I verified that by mutating `lightrag/operate.py` **as merged on `main`** and watching the existing suite stay green:

| Mutation of the merged implementation | Suite on `main` | With this PR |
|---|---|---|
| JSON continue prompt dropped from `gleaning_invariant_tokens` | 🟢 passes | 🔴 2 failed |
| `gleaning_precheck` reverted to unconditional | 🟢 passes | 🔴 1 failed |
| Text-mode continue prompt dropped from the guard | 🟢 passes | 🔴 1 failed |

The first and third matter because a guard that under-counts lets an oversized gleaning payload reach the provider — the `context_length_exceeded` error the guard exists to prevent. The second silently reintroduces a per-call encode of the ~2k-token system prompt on runs with gleaning off.

A cap-based skip assertion cannot detect any of them: at a 10-token cap the system prompt alone busts the budget, so the guard trips regardless of what else is counted. (My own first draft of these tests had exactly that flaw, and two mutations survived it.) The tests therefore measure a **delta** — lengthening a continue prompt template by a known number of characters must move the projected token count the guard reports by exactly that many, since the test tokenizer maps one character to one token. That isolates the prompt's contribution in each mode.

Added:

- **`test_json_mode_encodes_both_invariants_once_per_run`** — in JSON mode *both* hoisted strings are chunk-invariant, so each is encoded once across a 3-chunk run. Matches on each template's leading literal text rather than on rendered output, so it does not restate the rendering logic. Also catches the `UnboundLocalError` case that `test_gleaning_json_mode_binds_continue_prompt` covers.
- **`test_guard_still_counts_the_continue_prompt[text_mode, json_mode]`** — the delta measurement, covering the hoisted JSON prompt and the still-per-chunk text-mode one.
- **`test_invariant_tokens_not_encoded_when_guard_inactive`** — with gleaning off, the system prompt never reaches the tokenizer, so the lazy gate cannot silently regress.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [ ] Documentation updated (if necessary) — test-only change

## Additional Notes

Against the rebased head:

- `tests/extraction/test_extract_entities.py` → 10 passed
- `tests/extraction` (excluding `test_keyword_extraction_drivers.py`, uncollectable here without the `ollama` package) → 271 passed, 7 failed. Those 7 fail identically on clean `main`: my sandbox cannot reach the tiktoken BPE vocab download. Unrelated to this diff, and green in CI.
- `ruff check .` and `ruff format --check` clean
- All three mutations above re-verified against the merged `main` implementation after the rebase.

Every added test was confirmed to fail against the mutation it targets before being called done.

https://claude.ai/code/session_01S5U8CojutuGG5PB5U7hX2d

### DIFF
--- a/tests/extraction/test_extract_entities.py
+++ b/tests/extraction/test_extract_entities.py
@@ -1,6 +1,8 @@
 """Tests for entity extraction gleaning token limit guard."""
 
+import json
 import logging
+import re
 from unittest.mock import AsyncMock
 
 import pytest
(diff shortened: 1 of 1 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3791, opened 2026-09-01
refactor(webui): drop the buildLegacyDocs compatibility layer left over from GET /documents

### LINKED ISSUE #3790: refactor(webui): drop the buildLegacyDocs compatibility layer left over from GET /documents
### Summary

`DocumentManager`'s `docs` state and the `buildLegacyDocs` compatibility layer existed to hold the response shape of `GET /documents` — the deprecated full-listing endpoint removed in #3788. With that endpoint (and its `DocsStatusesResponse` model) gone, the layer no longer compatibility-bridges anything: it rebuilds a grouped structure *locally* from the paginated response, and every one of its four consumers has a strictly better source already in state.

It should be removed outright rather than repaired.

Note: `POST /documents/paginated` is **not** deprecated — it is the current and only document-list read path. What was removed is the legacy `GET /documents`.

### Why it is dead weight

**It carries no information `currentPageDocs` does not.** Its input is `response.documents`, the same array `setCurrentPageDocs` receives. The three `setDocs` calls and two `setCurrentPageDocs` calls are pairwise from one response (`DocumentManager.tsx` 774+777 and 893+907/909), so the two states can never disagree.

**The fallback branch is provably unreachable-as-meaningful.** `DocumentManager.tsx:628` uses `docs` only when `currentPageDocs` is empty — i.e. only when `response.documents` was empty. In that case `docs` is either `null` (`total_count === 0`) or a set of empty buckets, so the flatten yields `[]`, exactly what returning `currentPageDocs` would yield.

**Its key space is wrong in two different ways, and the type lies about it.** The declared TS type is `Partial<Record<DocStatus, DocStatusResponse[]>>`, but:

- `buildLegacyDocs` (`DocumentManager.tsx:95`, the main paginated/polling path) keys by `StatusBucket` — `'completed' | 'parse' | 'analyze' | 'process' | 'failed'` (`documentStatusFilters.ts:3`). `completed` is not a `DocStatus` value at all (the DocStatus is `processed`), and `getStatusBucket` returns `null` for `pending` / `preprocessed`, so `if (bucket)` at `:103` silently drops those documents. `tsc` accepts this only because `:99` asserts through `{} as Record<StatusBucket, ...>`.
- The manual-refresh branch's inline `legacyDocs` (`DocumentManager.tsx:896`) keys by real DocStatus values but omits `parsing` and `analyzing`.

The code already documents the divergence at `:724-726` and `:1352-1354` ("may be keyed by raw DocStatus (backend) or by filter bucket"), which is why `documentCounts` and the health-check fingerprint were written key-agnostic. That is the workaround; removing the layer is the fix.

**It actively loses information.** Because bucket keys drop `pending` / `preprocessed`, the `:1355` fingerprint is blind to changes in those statuses, so a page consisting only of pending documents never triggers the health check that fingerprint drives. `statusCounts` (backend `get_all_status_counts`, full-corpus, all seven statuses) does not have that blind spot.

### Proposed change

| Current consumer | Replacement |
| --- | --- |
| `:628-646` legacy flatten fallback | Delete — `currentPageDocs` is the single source |
| `:704-721` `documentCounts` fallback chain | Read `statusCounts` directly |
| `:1350-1369` health-check fingerprint | Digest `statusCounts` (covers all 7 statuses — strictly better) |
| `:1717/:1725` empty-state sentinel | `pagination.total_count > 0` |

Then delete: the `docs` state, `buildLegacyDocs`, the inline `legacyDocs` block at `:896`, and the `DocsStatusesResponse` type in `src/api/lightrag.ts` (it is no longer a wire type, and leaving it in the API module implies a backend counterpart that no longer exists).

### Pitfall to avoid in the rewrite

The existing counts use `||` fallback chains (`:717-721`, and `:1594`'s `statusCounts.all || documentCounts.all`). A genuine count of `0` is falsy and currently falls through to the `documentCounts` side, which happens to also yield `0` — so today the behaviour is correct by coincidence. When reading `statusCounts` directly, use `?? 0`, not `|| 0`. Keep `getCountValue`, which handles the upper/lower-case key duality (`'PROCESSED'` / `'processed'`).

### Verification

Per `AGENTS.md`, a WebUI change runs the full frontend set from `lightrag_webui/`: `bun install --frozen-lockfile`, `bun test`, `bunx tsc --noEmit`, `bun run lint`. Baseline on current `main` in a local environment is 447 pass / 28 pre-existing fail — compare against it rather than expecting zero failures.

Worth adding a component test that pins the property this cleanup protects: a page containing only `pending` documents must still drive the status-count-change effect (the case the bucket keys silently dropped).

### Context

- #3788 removed `GET /documents`, `DocsStatusesResponse` (backend), `get_docs_by_status`, and the dead `getDocuments` / `getDocumentStatusCounts` frontend clients.
- This issue covers the remaining frontend-internal residue that #3788 deliberately left untouched to keep that PR's scope to dead-code removal.

danielaskdd:
## Summary

Removes the `buildLegacyDocs` compatibility layer and the `docs` state it fed in `DocumentManager`, plus the code that only existed to serve them. Implements #3790.

## Motivation

`docs` / `buildLegacyDocs` existed to hold the response shape of `GET /documents` — the deprecated full-listing endpoint removed in #3788. With that endpoint gone the layer bridged nothing: it rebuilt a grouped structure **locally** from the paginated response, and every consumer already had a strictly better source in state.

It was also actively lossy. Its keys were filter buckets (`completed | parse | analyze | process | failed`), which have no entry for `pending` or the deprecated `preprocessed`:

- `buildLegacyDocs` silently dropped documents in those two statuses (`if (bucket)`).
- The digest driving the health-check effect was therefore blind to them — a page consisting only of pending documents produced a **constant** digest, so the check never fired.
- The declared type `Partial<Record<DocStatus, DocStatusResponse[]>>` did not describe those keys at all (`completed` is not a `DocStatus`; the DocStatus is `processed`). `tsc` accepted it only because the accumulator was asserted through `{} as Record<StatusBucket, …>`.
- The manual-refresh branch built a *second*, differently-keyed structure for the same state — which is why `documentCounts` and the fingerprint had to be written key-agnostic, with comments documenting the divergence. That was the workaround; this is the fix.

Note for anyone arriving from #3790's title: `POST /documents/paginated` is **not** deprecated — it is the current and only document-list read path. What #3788 removed was the legacy `GET /documents`.

## What's changed

| Old consumer | Replacement |
| --- | --- |
| legacy flatten fallback in `filteredAndSortedDocs` | `currentPageDocs` — the same array `buildLegacyDocs` consumed |
| `documentCounts` fallback chain for the filter tabs | `statusCounts` read directly (backend whole-corpus tally) |
| health-check fingerprint over `docs.statuses` | `computeStatusCountsFingerprint(statusCounts)` — covers all seven statuses, including the two with no filter bucket |
| `hasLoadedDocuments && docs` / `&& !docs` empty-state sentinel | `pagination.total_count > 0` / `=== 0` |

New: `src/features/documentStatusCounts.ts` — one pure exported helper, `computeStatusCountsFingerprint`, following this file's existing extracted-logic pattern (`documentRefreshCircuitBreaker`, `documentRefreshQueue`, `documentStatusFilters`). It lower-cases keys before sorting so a backend reporting `PROCESSED` rather than `processed` does not read as a count change — the same key-case duality `getCountValue` handles.

Counts now read `getCountValue(statusCounts, 'PROCESSED', 'processed')` with **no `|| 0` tail**: `getCountValue` already returns `0` for an unreported status, and a `||` tail is exactly the pitfall #3790 flagged — a genuine count of `0` is falsy and used to fall through to the `documentCounts` side (which also yielded `0`, so today's correctness was coincidental). `statusCounts.all` uses `?? 0`.

**Cascade deletions** — dead once the fallback branch is gone, and lint would flag most of them anyway:

- `DocumentManager.tsx`: `docs` state, `buildLegacyDocs`, the inline `legacyDocs` block in the manual-refresh branch, `STATUS_BUCKETS`, `documentCounts`, and the client-side `sortDocuments`. The backend query resolves sorting, filtering and pagination, so redoing either locally could only disagree with the pagination on screen.
- `documentStatusFilters.ts`: `STATUS_TO_BUCKET`, `getStatusBucket`, `matchesStatusFilter` (and its two describe-block tests). `BUCKET_TO_STATUS` / `getStatusRequestFilters` / the `StatusBucket`+`StatusFilter` types stay — they build the request.
- `api/lightrag.ts`: the `DocsStatusesResponse` type. It is no longer the shape of any endpoint's response, and leaving it in the API module implies a backend counterpart that #3788 deleted.

`filteredAndSortedDocs` becomes non-nullable, so the two stale nullable idioms guarding it (`?.map(…) || []`, `x && x.map(…)`) are simplified too.

## Behaviour changes

One deliberate change: the health check now also fires when the corpus **empties**. The old effect opened with `if (!docs) return`, and `docs` was set to `null` for `total_count === 0`, so a corpus going from N documents to 0 never signalled. A related consequence: `handleDocumentsCleared` zeroes `statusCounts` before its confirming refetch, so clearing all documents now fires one extra `/health` — benign for a user-initiated destructive action, and the surrounding code already schedules health work there.

The mount pass is explicitly preserved: `prevStatusFingerprint` is seeded `null` rather than `''`, and the effect fires only on a change *between two observations*. `statusCounts` starts at its `{ all: 0 }` seed, which is not an observation, so the first fire still lands after the first response — exactly where it was when the effect watched `docs`.

## What's broken / risk

Nothing knowingly. The equivalences were checked rather than assumed:

- `setDocs` and `setCurrentPageDocs` were pairwise from one response at both write sites, so the two states could never disagree.
- The flatten fallback ran only when `currentPageDocs` was empty — i.e. when `response.documents` was empty — where `docs` was either `null` or all-empty buckets, so it yielded `[]`: identical to returning `currentPageDocs`.
- Both consumers of `filteredAndSortedDocs` already treated `null` and `[]` identically.
- Every `setPagination` call site other than the two response handlers mutates only `page` / `page_size`, preserving `total_count`, so the new empty-state sentinel tracks the old `docs` nullability at every one of them.

## Tests

`src/features/documentStatusCounts.test.ts` (5 tests) pins the property this cleanup protects — the digest must see the two bucket-less statuses. Mutation-checked, not merely written against passing code:

- digest with `pending` filtered out → *"sees the pending count alone change"* goes red;
- digest with `pending` **and** `preprocessed` filtered out (reproducing the old bucket-key blind spot exactly) → 3 of 5 go red, including *"sees a document moving between the two bucket-less statuses"*.

Each test holds `all` and every bucketed count constant, so only genuine sensitivity to `pending` / `preprocessed` can make it pass — the first draft varied `all` alongside `pending` and the mutant survived it.

`usePageRestoreGeneration.test.ts` asserts the three mutually-exclusive render branches by source substring; its third assertion moved to the new sentinel.

Full WebUI check set from `lightrag_webui/`:

| | this branch | `main` baseline |
| --- | --- | --- |
| `bun test` | **450 pass / 28 fail** (478 tests) | 447 pass / 28 fail (475 tests) |
| `bunx tsc --noEmit` | clean | — |
| `bun run lint` | clean | — |

Net +3 tests (+5 new, −2 removed with `matchesStatusFilter`) and an **identical** failure count. The 28 failures are a pre-existing local-environment artifact of cross-file mock/module-load ordering (`spyOn` before initialization); every test file except `bootstrap/workspaceEntry.test.ts` passes when run in isolation, on this branch and on `main` alike. That baseline was re-measured on today's `main` (post-#3788), not carried over.

Closes #3790


danielaskdd:
@codex review

PR #3812, opened 2026-09-03
feat: support dynamic Langfuse configuration in WebUI and backend API

### LINKED ISSUE #2419: [Feature Request]:Allow setting Langfuse secrets from web UI

Hi all, I used LightRAG with Langfuse (docker compose way) and found no issues except that when we switch to a new Langfuse server instance, we have to create new secrets (public  key, secret key, endpoint), but LightRAG cannot simply reload new secrets (since it reads them from environment). So I had to do some `docker exec` to change the variables from inside the LightRAG container. Which is straightforward and works just fine but would also be great if we can set them in the UI? 

### Additional Context

_No response_

BLVCK-MAMBA-6:
## Description

Adds dynamic runtime configuration and UI support for Langfuse LLM observability. Previously, Langfuse integration was evaluated once at module import time, requiring process restarts to detect updated credentials. This change enables dynamic runtime activation, introduces a secure backend endpoint to persist credentials directly into the local environment and `.env` file, and provides an in-app settings interface within the WebUI.

## Related Issues

Closes #2419

## Changes Made

- **Backend Dynamic Client Hook**: Updated `lightrag/llm/openai.py` with `is_langfuse_enabled()` to check environment variables dynamically per client instantiation rather than fixing the wrapper at module import time.
- **Configuration API**: Added `lightrag/api/routers/config_routes.py` providing authenticated `GET` and `POST` handlers at `/api/config/langfuse` to retrieve current setup status (with masked/boolean secret key checks) and persist updates to `.env`.
- **Router Registration**: Mounted `create_config_routes` into FastAPI inside `lightrag/api/lightrag_server.py`.
- **WebUI Client & Settings UI**: 
  - Added `LangfuseConfig` interfaces, `getLangfuseConfig`, and `updateLangfuseConfig` in `lightrag_webui/src/api/lightrag.ts`.
  - Added dedicated Langfuse configuration fields (Public Key, Secret Key, Host URL) and live status indicators in `lightrag_webui/src/components/AppSettings.tsx`.
- **Testing**: Added integration tests in `tests/api/routes/test_config_routes.py` covering environment file synchronization, credential updates, and endpoint authorization.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

- All 56 OpenAI implementation unit tests (`tests/llm/openai_impl/`) continue to pass.
- Frontend assets compiled cleanly via `vite build` without bundle errors or type failures.


BLVCK-MAMBA-6:
Pushed fixes for the pre-commit/lint failures (formatting + unused variable). Should be green now

danielaskdd:
Thank you for taking the time to work on #2419 and for the thorough PR description. After reviewing this we have decided not to merge it, and I want to explain the reasoning in enough detail to be useful for any follow-up work.

## Why we are not taking this direction

**1. Changing the Langfuse server at runtime is a security and privacy risk, not just a convenience.**

`LANGFUSE_HOST` decides where every prompt and every LLM response gets shipped. Turning that into a value any client can rewrite over HTTP means the destination of the most sensitive data flowing through LightRAG becomes remotely controllable. That is a materially different kind of endpoint from the data endpoints we already expose.

This is made concrete by the route prefix chosen here. The default is `WHITELIST_PATHS=/health,/api/*` (`lightrag/api/config.py:847`), which exists so Open WebUI can reach the Ollama-compatible routes. The new router is mounted at `/api/config`, so it matches that whitelist and `get_combined_auth_dependency` returns at step 1 before any credential check runs. The `Depends(auth_dep)` on the handlers has no effect: **`POST /api/config/langfuse` is unauthenticated by default in every auth mode**, including deployments that have configured `AUTH_ACCOUNTS` or an API key. Anyone who can reach the port can redirect all traces to a server they control.

**2. Persisting backend configuration from a web handler, bypassing the operator, is a bigger problem than the one being solved.**

Credentials and endpoints are deployment configuration. They belong to the operator's toolchain — compose `environment:`, k8s secrets, the setup wizard — where they are versioned, reviewable and auditable. A web request that rewrites `.env` in place has no review, no audit trail, no atomicity (a failure mid-write truncates the operator's entire configuration file), and no locking.

It also creates configuration drift rather than removing it. Both `lightrag/api/config.py:61` and `lightrag/api/lightrag_server.py:100` call `load_dotenv(dotenv_path=".env", override=False)`, so process environment variables always win over `.env`. For the docker-compose user in #2419 — who supplies these values through `environment:` — the write succeeds, the UI reports success, and the setting silently reverts on the next restart. The user ends up worse off than with the documented restart path.

There is also no precedent for this in the codebase: no API route writes `.env` today, and `AppSettings.tsx` holds frontend preferences only. Opening this door for Langfuse invites the same request for LLM bindings, embedding models and storage connection strings, which carry far higher stakes.

**3. The mechanism does not actually work, so the issue would remain open even if the design were acceptable.**

Two things we verified empirically against `langfuse>=3.8.1`:

*The client-class switch is a no-op.* `langfuse/openai.py:53` re-exports `from openai import AsyncOpenAI` unchanged; the instrumentation is installed globally by `register_tracing()` (line 1449) via `wrap_function_wrapper` at module import.

```
from openai import AsyncOpenAI as A
from langfuse.openai import AsyncOpenAI as B
A is B   # True
```

So `if is_langfuse_enabled(): return LangfuseAsyncOpenAI(...) else: return StandardAsyncOpenAI(...)` selects between two names for the same class. Worse, the PR moves `import langfuse.openai` from "only when credentials are present" to "whenever the package is installed", so the global patch is always active. With the `observability` extra installed but unconfigured, every LLM call now constructs a fresh disabled `Langfuse()` and emits an authentication warning — three calls produce three warnings, and `LangfuseResourceManager._instances` stays empty.

*Rotating credentials still does not take effect,* which is precisely the scenario #2419 describes. `LangfuseResourceManager` caches a singleton per public key (`resource_manager.py:137-178`) and reads the environment only when that instance is first constructed. LightRAG never passes a public key to `get_client()`, so the cached instance is returned forever:

```
env = pk-A / sk-A / a.example.com   →  get_client() → pk-A, base=a.example.com
env = pk-B / sk-B / b.example.com   →  get_client() → pk-A, base=a.example.com
```

The change therefore only fixes the "unconfigured → configured" transition, not the "switch to a new Langfuse server" case the issue is about. A correct implementation would have to construct `Langfuse(public_key=..., secret_key=..., base_url=...)` explicitly, or reset the resource manager on change — environment reads alone cannot do it.

## What we would welcome instead

The genuinely useful half of #2419 is **a runtime toggle for tracing in the WebUI** — turn Langfuse tracing on or off without restarting the server, against credentials the operator has already provisioned. That is the high-frequency, low-risk operation: enable tracing to investigate something, disable it when done. It does not move secrets through the API, does not change where data is sent, and does not write to `.env`. Note that `env.example` already documents a `LANGFUSE_ENABLE_TRACE` flag that no code currently reads, so there is a natural place for it.

A PR along those lines would be very welcome. Switching the Langfuse server itself should stay an operator action: update compose `environment:` and `docker compose up -d`.

## Smaller notes, for whatever comes next

- `lightrag_webui/package-lock.json` (11366 lines) should not be committed — the WebUI uses Bun and `bun.lock`.
- The new UI strings are hard-coded English; the WebUI goes through i18next.
- `raise HTTPException(..., detail=f"...{str(e)}")` leaks internal error text; use `internal_server_error()` in `lightrag/api/utils_api.py:40`.
- `os.environ` mutation only affects the worker that served the request; under `lightrag-gunicorn` the other workers keep the old configuration while the UI reports success.
- The GET handler reads `LANGFUSE_BASEURL`; Langfuse 3.x uses `LANGFUSE_BASE_URL` (and treats `LANGFUSE_HOST` as deprecated with lower precedence), so a host set through the UI is silently ignored when `LANGFUSE_BASE_URL` is present.
- The tests do not cover the behavior the PR claims: both cases pass `api_key=None`, so no authorization path is exercised (and the whitelist bypass above goes undetected), and nothing asserts that a new client actually picks up rotated credentials.

Closing this one. Thanks again for the contribution and for engaging with the observability stack — please do consider the tracing-toggle version.

BLVCK-MAMBA-6:
Thank you @danielaskdd for the thorough and detailed review.

The security implications of mounting under `/api/*` with `WHITELIST_PATHS` and the singleton caching behavior in `LangfuseResourceManager` make complete sense. I also agree that treating credentials and host endpoints as operator-level configuration (via Docker, K8s, or environment variables) rather than persisting them through an HTTP handler is the right architectural boundary.

I'd be glad to take on the tracing-toggle PR. One detail regarding the mechanism: `LANGFUSE_ENABLE_TRACE` from `env.example` isn't wired to anything in the Langfuse SDK, and the SDK's own flag (`LANGFUSE_TRACING_ENABLED`) is evaluated once during client initialization into a private attribute with no public runtime toggle. Rather than reaching into SDK internals or attempting dynamic env-patching, gating this at LightRAG's own instrumentation call sites makes the most sense: an in-memory flag the WebUI can flip live, checked before LightRAG dispatches spans to Langfuse. No `.env` writes, no private API manipulation, and no secrets in transit—just a clean runtime gate on whether traces are emitted.

I will also ensure the smaller notes are addressed:
* Dropping the stray `package-lock.json` and sticking strictly to Bun/`bun.lock`.
* Routing all new WebUI text through `i18next`.
* Replacing raw exception leaks with `internal_server_error()`.
* Ensuring the new toggle endpoint is properly authenticated and outside any unauthenticated whitelist paths.

I'll open a fresh, focused PR for this. Thanks again for the breakdown.

BLVCK-MAMBA-6:
Thanks again for the detailed review. For the replacement PR, I’m planning to use LightRAG’s existing `get_namespace_data()` shared-storage mechanism—the same pattern used for pipeline status—to keep the runtime tracing toggle consistent across gunicorn workers. The shared state will be read by Langfuse’s `should_export_span` predicate for each span, while credentials and the destination host remain operator-managed. Flagging the approach before I begin in case there is any reason the shared-storage pattern would not be appropriate here.

PR #3813, opened 2026-09-04
test(utils): cover normalize_rerank_result

loseintwilight:
## Description

Adds direct tests for `lightrag/utils.py::normalize_rerank_result`, the single
validation boundary between a rerank provider's output and the chunk-scoring
pipeline. It had no test referencing it directly.

**Tests only — no production code is modified.** The diff is one new file, zero
deletions.

## Related Issues

None. `CONTRIBUTING.md` lists "Testing — add test coverage for untested code
paths" as a way to contribute.

### `normalize_rerank_result`

Every result a rerank provider emits — a hosted rerank model, a compatible proxy,
or a user-supplied `rerank_model_func` — passes through this helper in
`apply_rerank_if_enabled` before its `index`/`relevance_score` are trusted to
reorder chunks. A malformed result slipping through here would surface as a wrong
(or silently empty) ranking. The reject cases are therefore the ones that matter.

Covered:

- Accepted: float, string, and integral scores are normalized to float; the
  highest valid index is `max_index - 1`.
- Rejected result shape: non-dict inputs return `"not an object"`.
- Rejected index: missing / `None` / string / float / list → `"invalid index"`;
  negative, equal-to, and beyond `max_index` → `"index out of range"`.
- Two easy-to-break guards that look like hardening:
  - `bool` index is rejected as `"invalid index"` even though
    `isinstance(True, int)` holds — without the explicit bool check, `True`
    would be accepted as index `1`.
  - `bool` score is rejected as `"invalid relevance score"` even though
    `float(True)` would be `1.0` — without the explicit bool check, `True`
    would be accepted as a valid score.
- Rejected score: missing / `None` / non-numeric string / list / dict →
  `"invalid relevance score"`; `NaN` / `+inf` / `-inf` → `"non-finite
  relevance score"`.

The existing `tests/llm/test_apply_rerank_result_validation.py` exercises this
helper only indirectly through the happy path and a single bool-index case; this
file pins the remaining branches directly and exhaustively.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [ ] Documentation updated (if necessary) — not applicable, tests only
- [x] Unit tests added

## Additional Notes

**The tests were verified to have teeth.** Rather than assume, the file was run
against a deliberately mutated `lightrag/utils.py` before committing:

| Mutation | Result |
|---|---|
| `isinstance(index, bool)` guard removed | 1 test fails |
| `isinstance(score_value, bool)` guard removed | 1 test fails |
| `math.isfinite(score)` guard removed | 3 tests fail |

`lightrag/utils.py` was restored afterwards and is untouched in this PR.

**Verified locally** (per AGENTS.md, only the directory mirroring the module
under test):

- `pytest tests/utils/test_normalize_rerank_result.py -m offline` — 27 passed

### DIFF
--- a/tests/utils/test_normalize_rerank_result.py
+++ b/tests/utils/test_normalize_rerank_result.py
@@ -8,15 +8,22 @@ its ``index``/``relevance_score`` to reorder chunks. Two properties are therefor
 load-bearing rather than cosmetic:
 
 * **Malformed results are rejected, never coerced.** A result whose ``index`` or
-  ``relevance_score`` cannot be trusted must be dropped and the chunk left in its
-  original position; silently accepting it would surface as a wrong (or
-  unexpectedly empty) ranking. The ``bool`` cases matter specifically because
-  ``bool`` subclasses ``int`` and ``float(True)`` succeeds — both must be caught
-  explicitly or a ``True`` index/score leaks through.
-* **The reject reason string is stable.** Callers and future maintainers rely on
-  ``"not an object"`` / ``"invalid index"`` / ``"index out of range"`` /
-  ``"invalid relevance score"`` / ``"non-finite relevance score"`` as the single
-  vocabulary for diagnosing a misbehaving provider.
+  ``relevance_score`` cannot be trusted must be reported as an error rather than
+  normalized; silently accepting it would surface as a wrong ranking. Callers
+  drop the rejected result, which means the chunk it pointed at is absent from
+  the reranked output; a wholesale rejection is what makes
+  ``apply_rerank_if_enabled`` fall back to the original chunk list. The ``bool``
+  cases matter specifically because ``bool`` subclasses ``int`` and
+  ``float(True)`` succeeds — both must be caught explicitly or a ``True``
+  index/score leaks through.
+* **The reject reason string stays diagnosable.** Only one of the three call
+  sites consumes the reason (``lightrag/rerank.py`` aggregates it into a bounded
+  ``Counter`` log summary; the other two discard it), so ``"not an object"`` /
+  ``"invalid index"`` / ``"index out of range"`` / ``"invalid relevance score"``
+  / ``"non-finite relevance score"`` are a logging vocabulary for diagnosing a
+  misbehaving provider rather than a contract other code branches on. Pinning
+  them keeps that vocabulary from drifting silently, and keeps each reject branch
+  distinguishable in these tests.
 
 The function had no direct tests; ``tests/llm/test_apply_rerank_result_validation.py``
 exercises it only indirectly through the happy path and a single bool-index case.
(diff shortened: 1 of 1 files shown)


danielaskdd:
@codex review

danielaskdd:
Thanks for this — the helper genuinely had no direct coverage, and I confirmed the mutation table locally: removing the `isinstance(index, bool)` guard fails 1 test, removing `isinstance(score_value, bool)` fails 1, and removing `math.isfinite(score)` fails 3. Exactly as claimed, so the tests have teeth.

I pushed one follow-up commit (`0614c61f`) as a maintainer edit rather than asking you to round-trip it. Two small things:

**1. Two statements in the module docstring did not match the actual behavior.**

- It said a rejected result leaves *"the chunk left in its original position"*. It does not — callers drop the rejected result, so the chunk it pointed at is simply absent from the reranked output. Your own reference, `tests/llm/test_apply_rerank_result_validation.py::test_valid_results_survive_malformed_items`, shows this: `[None, {"index": 1, ...}, {"index": 0}]` returns only `second`, and `first` is gone. Only a wholesale rejection makes `apply_rerank_if_enabled` fall back to the original list (`lightrag/utils.py`, the `if not reranked_docs` branch).
- It said callers rely on the reject reason string as *"the single vocabulary"* they depend on. Of the three call sites, only `lightrag/rerank.py:406` consumes it, and only to aggregate a bounded `Counter` log summary; `lightrag/utils.py:6226` and `lightrag/rerank.py:190` discard it into `_`. So pinning the strings is worth doing for logging diagnosability and to keep the reject branches distinguishable in the tests — but it is not a contract other code branches on. Reworded to say that.

**2. Three branches were still unpinned.** Added, each verified against a targeted mutation:

| New case | Mutation | Result |
|---|---|---|
| score `10**400` → `"invalid relevance score"` | `OverflowError` dropped from the `except` clause | that 1 case fails (escapes as an exception) |
| `max_index == 0` → `"index out of range"` | range check widened to `max(max_index, 1)` | that 1 case fails |
| string `"inf"` / `"-Inf"` / `"nan"` → `"non-finite relevance score"` | `isfinite` guard narrowed to `isinstance(score_value, float)` | those 3 cases fail |

`10**400` matters because `float()` raises `OverflowError`, not `ValueError`, and the string spellings matter because `float()` accepts them — so a string score reaches the finiteness guard, not the conversion guard.

Now 32 passed. Ran `tests/utils` (298 passed) and `tests/llm/test_apply_rerank_result_validation.py` (3 passed) as well. `lightrag/utils.py` is still untouched by this PR. Note the PR description above still quotes the pre-follow-up numbers (27 passed, three mutation rows) — the table here supersedes it; no need to edit it.

PR #3817, opened 2026-09-04
fix: apply chunk transforms across ingestion paths

### LINKED ISSUE #3603: Per-chunk text "breadcrumb" / metadata hook — `chunking_func` is silently bypassed on `/documents/scan` & `/documents/upload`

Solaris-star:
## Description

Apply an optional per-chunk transform after chunking so metadata and text breadcrumbs are preserved consistently across direct insertion, document upload, and document scan ingestion paths.

## Related Issues

Fixes #3603

## Changes Made

- Add a sync or async `chunk_transform_func` hook to `LightRAG`.
- Apply the hook to all chunk-producing ingestion paths.
- Validate transformed chunk shape and preserve deterministic chunk identifiers.
- Add focused regression coverage for direct, upload, and scan selectors.

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Additional Notes

## Validation

- `python3.11 -m pytest -o 'addopts=' tests/pipeline/test_custom_chunking_selector.py -q` (14 passed)
- `pre-commit run --files lightrag/lightrag.py lightrag/pipeline.py tests/pipeline/test_custom_chunking_selector.py`
- `git diff --check`

### DIFF
--- a/lightrag/lightrag.py
+++ b/lightrag/lightrag.py
@@ -542,6 +542,25 @@ class LightRAG(_RoleLLMMixin, _StorageMigrationMixin, _PipelineMixin):
     tiktoken_model_name: str = field(default="gpt-4o-mini")
     """Model name used for tokenization when chunking text with tiktoken. Defaults to `gpt-4o-mini`."""
 
+    chunk_transform_func: (
+        Callable[
+            [list[dict[str, Any]], Any],
+            Union[
+                list[dict[str, Any]],
+                tuple[dict[str, Any], ...],
+                Awaitable[Union[list[dict[str, Any]], tuple[dict[str, Any], ...]]],
+            ],
+        ]
+        | None
+    ) = field(default=None)
+    """Optional instance-level transform applied after any base chunker completes.
+
+    The callback receives the chunk dictionaries and an immutable document
+    context. It may be synchronous or asynchronous and must return a list or
+    tuple containing only dictionaries. ``None`` preserves the existing
+    pipeline behavior exactly.
+    """
+
     chunking_func: Callable[
         [
             Tokenizer,
--- a/lightrag/pipeline.py
+++ b/lightrag/pipeline.py
@@ -457,6 +457,42 @@ _CUSTOM_CHUNKING_METHOD = "custom_chunking_func"
 _CUSTOM_CHUNKING_FALLBACK_METHOD = "custom_chunking_fallback_fixed_token"
 
 
+@dataclass(frozen=True)
+class ChunkTransformContext:
+    """Immutable document context supplied to ``chunk_transform_func``."""
+
+    doc_id: str
+    file_path: str | None
+    process_options: str
+    sidecar_location: str | None
+
+
+async def _apply_chunk_transform(
+    transform: Any,
+    chunks: list[dict[str, Any]] | tuple[dict[str, Any], ...],
+    context: ChunkTransformContext,
+) -> list[dict[str, Any]] | tuple[dict[str, Any], ...]:
+    """Apply and validate the optional post-chunking transform."""
+    if transform is None:
+        return chunks
+    try:
+        result = transform(chunks, context)
+        if inspect.isawaitable(result):
+            result = await result
+    except Exception as exc:
+        raise RuntimeError(
+            f"chunk_transform_func failed for d-id {context.doc_id}: {exc}"
+        ) from exc
+    if not isinstance(result, (list, tuple)) or not all(
+        isinstance(chunk, dict) for chunk in result
+    ):
+        raise TypeError(
+            "chunk_transform_func must return a list or tuple of dicts "
+            f"for d-id {context.doc_id}, got {type(result)}"
+        )
+    return result
+
+
 _CHUNK_LOG_KEY_ALIASES: dict[str, str] = {
     "chunk_overlap_token_size": "overlap",
     "breakpoint_threshold_type": "break",
@@ -5237,6 +5273,17 @@ class _PipelineMixin:
                         f"chunking_func must return a list or tuple of dicts, "
                         f"got {type(chunking_result)}"
                     )
+
+                chunking_result = await _apply_chunk_transform(
+                    self.chunk_transform_func,
+                    chunking_result,
+                    ChunkTransformContext(
+                        doc_id=doc_id,
+                        file_path=file_path,
+                        process_options=(content_data or {}).get("process_options", ""),
+                        sidecar_location=(content_data or {}).get("sidecar_location"),
+                    ),
+                )
 
                 # Reflect the format actually persisted in full_docs.
                 # Previously a structured-parse fallback always tagged
--- a/tests/pipeline/test_custom_chunking_selector.py
+++ b/tests/pipeline/test_custom_chunking_selector.py
@@ -442,3 +442,122 @@ def test_reprocess_persisted_c_after_callback_removal_uses_observable_fallback(
         assert len(warnings) == 1
 
     asyncio.run(_run())
+
+
+@pytest.mark.offline
+@pytest.mark.parametrize("process_options", ["F!", "!"])
+def test_chunk_transform_runs_after_builtin_and_legacy_chunkers(
+    tmp_path, process_options
+):
+    captured = {}
+
+    def _transform(chunks, context):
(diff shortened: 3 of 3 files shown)


danielaskdd:
Thank you for the contribution — the implementation is compact and the regression tests are focused. We are closing this PR nonetheless, for design reasons rather than code quality; the details are below so the decision is transparent.

## Why this is being closed

**The design this PR implements is deliberately deferred.** The discussion on #3603 concluded (see the latest scope update there) that the document *date* is the piece of envelope metadata that actually needs a channel, and that it belongs at the document level — stored once, never written into chunk text, rendered at both extraction and answer time. That is now being delivered by #3805 (the `document_date` field, `full_docs` persistence, extraction block, query-time attachment) and #3830 (filename-hint document-level block `(date=...)` and `LIGHTRAG_DOC_DATE_PATTERN` for scan / upload). With the date handled that way, the main use case for a content-rewriting transform is gone; source provenance already reaches the answer model through the `file_path` reference list, and section context through the `heading` channel.

What remains of the original #3603 ask — a strategy-agnostic per-chunk transform, and how arbitrary per-chunk / per-document metadata should be stored and consumed — is explicitly **not** being implemented until that design is settled. A chunk-level content-mutation hook carries costs the discussion agreed to avoid (embedding pollution, per-retrieval token cost, no way to correct without re-ingest), and landing one now would pre-empt the contract work on #3603 / #3664. `ChunkTransformContext` here is also a second, parallel shape for the same concept as the `ChunkingContext` proposed in RFC #3664; we would rather settle one.

Note also that `Fixes #3603` would auto-close an issue we are intentionally keeping open for that design.

## Code-level notes, for the record

These are secondary to the above, but would need addressing in any future version of this idea:

- **CI is red**: inserting the new dataclass field before `chunking_func` shifts the positional field order; `tests/test_dataclass_positional_compatibility.py::test_lightrag_positional_anchors_did_not_shift` fails. New fields must go after existing positional anchors.
- **Sidecar backfill**: the transform runs after chunking but before source-span backfill and does not clear `sidecar_backfill_eligible`. #3665 explicitly disabled backfill for `C` custom output because rewritten text cannot be located back in `blocks.jsonl`; a transform that rewrites `content` needs the same treatment.
- **Observability**: nothing in `doc_status.metadata` records that a transform ran (compare `chunk_method` = `custom_chunking_func` vs. `custom_chunking_fallback_fixed_token` for `C`). Silent customization is the failure mode #3603 was opened about.
- **Reachability**: `lightrag_server.py` never sets `chunk_transform_func`, so for server deployments the hook is unreachable — the same injection / registry question #3664 exists to answer.

## Where to take this next

If you are interested in the transform / context contract, #3664 is the right place — in particular the `ChunkingContext` shape and how `document_metadata` should be populated once the date channel exists. Design input there is very welcome.

---

PR #3824, opened 2026-09-04
fix(lmdeploy): handle truncated responses

Shizoqua:
## Problem

The LMDeploy adapter reads the legacy `response` field and does not mark token-limited output as truncated.

## Change

Read the current `text` field with legacy compatibility. Return `TruncatedResponse` when LMDeploy reports `finish_reason="length"`.

## Tests

- LMDeploy tests: 13 passed
- Full LLM tests: 528 passed
- Pre-commit: passed
- `git diff --check`: passed

## Related issue

None.

### DIFF
--- a/lightrag/llm/lmdeploy.py
+++ b/lightrag/llm/lmdeploy.py
@@ -18,6 +18,8 @@ from tenacity import (
     retry_if_exception_type,
 )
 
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3832, opened 2026-09-05
fix(bedrock): track token usage

Shizoqua:
## Problem

The Bedrock binding forwarded `token_tracker` to the AWS SDK instead of recording the token counts returned by the provider.

## Change

Capture input, output and total token usage from normal and streaming Bedrock responses. Streaming usage is recorded from the final metadata event.

## Tests

- Bedrock tests: 40 passed
- Pre-commit: passed
- `git diff --check`: passed


danielaskdd:
@codex review

PR #3839, opened 2026-09-06
feat: add shared runtime Langfuse tracing toggle

### LINKED ISSUE #2419: [Feature Request]:Allow setting Langfuse secrets from web UI

BLVCK-MAMBA-6:
## Description

Adds an authenticated runtime toggle for Langfuse trace export while keeping Langfuse credentials and host configuration under operator control.

The toggle is evaluated live for every span through the Langfuse client’s `should_export_span` predicate. Its state is backed by LightRAG’s existing shared namespace mechanism so all Gunicorn workers observe the same value.

This PR does not expose or modify credentials, change the Langfuse destination, or write to `.env`.

## Related Issues

Related to #2419.

Follow-up to the maintainer feedback on #3812. This PR implements only the runtime tracing-toggle portion; credential and host changes remain operator-managed and therefore this does not close #2419.

## Changes Made

* Added shared runtime tracing state using `get_namespace_data("langfuse_tracing", workspace="")`.
* Wired the shared state into the Langfuse client’s `should_export_span` predicate at construction so it is checked live for every span.
* Used the existing documented `LANGFUSE_ENABLE_TRACE` setting as the initial value while preserving tracing as enabled by default when it is unset.
* Added authenticated `GET` and `PUT` endpoints at `/observability/langfuse/tracing`.
* Kept the endpoint outside the default unauthenticated `/api/*` whitelist and explicitly disabled whitelist bypass for its authentication dependency.
* Added a WebUI tracing switch that reports unavailable states without exposing credentials.
* Routed all new UI text through i18next for every supported locale.
* Added tests covering:

  * endpoint authentication;
  * unavailable and error responses;
  * live per-span enable/disable behavior;
  * shared-state visibility across worker-local clients;
  * initial environment-default handling.
* Retained Bun as the WebUI package manager with no `package-lock.json` changes.

## Checklist

* [x] Changes tested locally
* [x] Code reviewed
* [x] Documentation updated (not required; this wires the existing documented `LANGFUSE_ENABLE_TRACE` setting)
* [x] Unit tests added

## Additional Notes

The shared namespace proxy is resolved during server initialization and cached so the synchronous `should_export_span` callback can read it without awaiting. Runtime updates use a single manager-dictionary key assignment, so no namespace lock or read-modify-write sequence is required.

The toggle is intentionally runtime state rather than persisted configuration. After a full server restart, it returns to the operator-configured `LANGFUSE_ENABLE_TRACE` default.

Validation completed:

* Backend targeted tests: 10 passed
* WebUI test suite: 626 passed
* TypeScript check: passed
* ESLint check for the modified component: passed
* Whitespace validation: passed

### DIFF
--- a/lightrag/api/lightrag_server.py
+++ b/lightrag/api/lightrag_server.py
@@ -71,6 +71,7 @@ from lightrag.parser.external.mineru.cache import MinerUParserOptions
 from lightrag.api.routers.query_routes import create_query_routes
 from lightrag.api.routers.graph_routes import create_graph_routes
 from lightrag.api.routers.ollama_api import OllamaAPI
+from lightrag.api.routers.observability_routes import create_observability_routes
 from lightrag.api.routers.ui_customization_routes import create_ui_customization_routes
 from lightrag.api.ui_customization import (
     WEBUI_CHROME_LOCALES,
@@ -79,6 +80,7 @@ from lightrag.api.ui_customization import (
 )
 
 from lightrag.utils import logger, set_verbose_debug
+from lightrag.llm.langfuse_tracing import initialize_langfuse_tracing
 from lightrag.kg.shared_storage import (
     get_namespace_data,
     get_default_workspace,
@@ -1593,6 +1595,7 @@ def create_app(args):
             # Initialize database connections
             # Note: initialize_storages() now auto-initializes pipeline_status for rag.workspace
             await rag.initialize_storages()
+            await initialize_langfuse_tracing()
 
             # Data migration regardless of storage implementation
             await rag.check_and_migrate_data()
@@ -2540,6 +2543,7 @@ def create_app(args):
     app.include_router(create_document_routes(rag, doc_manager, api_key))
     app.include_router(create_query_routes(rag, api_key, args.top_k))
     app.include_router(create_graph_routes(rag, api_key))
+    app.include_router(create_observability_routes(api_key))
     # Public read-only customization surface — registered unconditionally:
     # without a bundle it answers 200 {"customized": false, ...}.
     app.include_router(
--- a/lightrag/api/routers/observability_routes.py
+++ b/lightrag/api/routers/observability_routes.py
@@ -1,0 +1,80 @@
+"""Authenticated runtime observability controls."""
+
+from typing import Optional
+
+from fastapi import APIRouter, Depends, HTTPException, status
+from pydantic import BaseModel, Field
+
+from lightrag.api.utils_api import get_combined_auth_dependency
+from lightrag.llm.langfuse_tracing import (
+    get_langfuse_tracing_status,
+    set_langfuse_tracing_enabled,
+)
+
+
+class LangfuseTracingStatus(BaseModel):
+    installed: bool
+    configured: bool
+    enabled: bool
+    active: bool
+
+
+class LangfuseTracingUpdate(BaseModel):
+    enabled: bool = Field(
+        ...,
+        description="Enable or disable Langfuse trace export at runtime.",
+    )
+
+
+def create_observability_routes(api_key: Optional[str] = None) -> APIRouter:
+    """Create routes for runtime observability controls."""
+
+    router = APIRouter(
+        prefix="/observability",
+        tags=["observability"],
+    )
+    auth_dependency = get_combined_auth_dependency(
+        api_key,
+        respect_whitelist=False,
+    )
+
+    @router.get(
+        "/langfuse/tracing",
+        response_model=LangfuseTracingStatus,
+        dependencies=[Depends(auth_dependency)],
+    )
+    async def get_langfuse_tracing() -> LangfuseTracingStatus:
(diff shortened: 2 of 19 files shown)


BLVCK-MAMBA-6:
@danielaskdd Thanks again for the detailed review on #3812. I rebuilt this follow-up around the runtime-only tracing toggle you suggested.

The new implementation:

* Keeps Langfuse credentials and host entirely operator-managed—there are no credential/configuration endpoints, environment mutations, or `.env` writes.
* Places the endpoint at `/observability/langfuse/tracing`, outside the default `/api/*` whitelist, and explicitly requires authentication with `respect_whitelist=False`.
* Passes a live `should_export_span` predicate when constructing the Langfuse client instead of switching OpenAI client classes.
* Backs that predicate with LightRAG’s existing `get_namespace_data()` shared-storage mechanism. Each worker caches its shared namespace proxy, so an already-constructed client reads the current toggle value for every span without requiring client reconstruction.
* Updates the boolean with a single manager-dictionary key assignment; no namespace lock or read-modify-write sequence is involved.
* Uses `LANGFUSE_ENABLE_TRACE` only as the operator-configured startup default. Runtime changes intentionally reset to that value after a full restart.
* Routes all new UI text through i18next and leaves the Bun lockfile unchanged.
* Adds tests for authentication, unavailable configurations, live enable/disable behavior, and independently initialized worker-local predicates observing the same shared state.

The targeted backend tests pass (10/10), and the complete WebUI suite passes (626/626), along with TypeScript, ESLint, and whitespace checks.

I wanted to flag the `get_namespace_data()` choice explicitly since it addresses the multi-worker consistency concern from the previous review. I’d appreciate your feedback on whether this use of the existing shared-storage pattern fits what you had in mind.

danielaskdd:
Thanks for reworking this around the runtime-only toggle — the shared-storage approach and the care taken to keep credentials operator-managed are both a clear improvement over #3812. The reason we're not merging is about timing and architecture, not the quality of this implementation.

**1. The WebUI has no permission model yet.** The frontend currently has two entries — `/webui` (intended as the administrator entry) and `/workspace` — but there is no role/permission separation behind them. Any authenticated user, including a guest token in a fully-open deployment, would be able to flip a deployment-wide switch from the appearance popover. Before we open the door to changing backend parameters from the frontend at all, we need to decide that question once, for *every* parameter that should be runtime-mutable — not per parameter, one PR at a time.

**2. Runtime backend parameters need a single, unified control surface with cross-worker propagation.** That mechanism is still at the RFC stage — see #3847. The direction is to manage runtime-mutable backend parameters through `addon_params`, whose changes propagate automatically to the `LightRAG` instance in every worker. The Langfuse switch should become one parameter on that surface rather than a parallel one-off channel with its own endpoint and its own shared namespace. #3847's suggested phasing is explicit about this ordering: propagation layer first, REST control plane second.

So we'd rather revisit this once #3847 lands. Please keep an eye on that issue — input there is very welcome, and this toggle is a good concrete use case to validate the design against.

Two concrete defects found while reviewing, recorded here in case this work is revived later:

- **`should_export_span` is a langfuse 4.x-only constructor parameter**, but the `observability` extra pins `langfuse>=3.8.1`. Every 3.x release (verified 3.8.1 through 3.15.0, the last 3.x) lacks it, so `Langfuse(should_export_span=...)` raises `TypeError`. `initialize_langfuse_tracing()` is awaited inside the lifespan `try:`, which has only a `finally:` — no `except` — so the API server fails to start entirely for any deployment on langfuse 3.x with credentials configured. Reproduced locally. Observability initialization should never be able to break server startup.
- **`AppSettings` is rendered on the login page** (`LoginPage.tsx:260`) and on `WorkspaceWelcome`, and the new effect fires the authenticated `GET /observability/langfuse/tracing` whenever the popover opens, regardless of auth state. Unauthenticated, that returns 401 → the response interceptor calls `navigateToUnauthenticated()` and the component shows an error toast, so opening the language/theme menu on the login screen produces a spurious error.

---

BLVCK-MAMBA-6:
Thanks for reviewing this and for reproducing both issues. I understand the direction now — this should wait for #3847 and be added through `addon_params` instead of having its own shared namespace and endpoint.

Good catch on the Langfuse 3.x incompatibility. My test used a fake constructor that accepted any keyword argument, so it didn’t reveal that `should_export_span` only works with 4.x. I also missed that `AppSettings` is used on the login and workspace screens, which explains the unauthenticated request and error toast. I totally apologize for that.

I’ll stop work on this PR for now and follow #3847. Once that infrastructure and the permission model are ready, I’d be more than happy to revisit the toggle with proper version handling and startup failure protection.

PR #3843, opened 2026-09-06
fix(zhipu): support token_tracker instead of forwarding it to the API

Shizoqua:
**Problem**

Every other completion provider accepts token_tracker to record usage. Zhipu had no such parameter, so passing one forwarded it through kwargs straight into the client call, which rejects it as an unexpected argument.

**Change**

Add the parameter and record usage from response.usage before extracting content.

**Tests**

- Zhipu tests: 10 passed
- Pre-commit passed
- git diff --check passed

**Related Issue**

No related issue.

### DIFF
--- a/lightrag/llm/zhipu.py
+++ b/lightrag/llm/zhipu.py
@@ -30,7 +30,7 @@ from lightrag.utils import (
 )
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3844, opened 2026-09-06
fix(llama_index): fix llm_instance forwarding and blocking embed call

Shizoqua:
**Problem**

llama_index_complete used kwargs.get for llm_instance instead of pop, so it stayed in kwargs and got forwarded a second time into llama_index_complete_if_cache, which has no such parameter. Crashed on every call using the documented usage.

llama_index_embed also called the synchronous _get_text_embeddings, blocking the event loop for the whole request.

**Change**

Pop llm_instance instead of get. Use the async _aget_text_embeddings for the embed call.

**Tests**

- New llama_index tests: 4 passed
- Pre-commit passed
- git diff --check passed

**Related Issue**

No related issue.

### DIFF
--- a/lightrag/llm/llama_index_impl.py
+++ b/lightrag/llm/llama_index_impl.py
@@ -165,6 +165,9 @@ async def llama_index_complete(
         settings: Optional LlamaIndex settings
         **kwargs: Additional arguments. ``response_format`` is not supported by
             this adapter and is stripped before calling LlamaIndex.
+            ``max_tokens`` is folded into ``chat_kwargs`` (without overriding
+            an explicit ``chat_kwargs["max_tokens"]``) since llama_index has
+            no top-level parameter for it.
 
     Structured output note:
     - This adapter does not support OpenAI-style ``response_format`` JSON mode.
@@ -190,14 +193,21 @@ async def llama_index_complete(
             stacklevel=2,
         )
     kwargs.pop("response_format", None)
-    # hashing_kv is injected unconditionally by the role LLM wrapper;
-    # max_tokens is injected by use_llm_func_with_cache when configured.
-    # Neither is a llama_index_complete_if_cache parameter, and streaming
-    # is not supported by this adapter -- all three are dropped, matching
+    # hashing_kv is injected unconditionally by the role LLM wrapper.
+    # Neither it nor streaming is a llama_index_complete_if_cache
+    # parameter, or supported by this adapter -- both are dropped, matching
     # the explicit-pop pattern in openai.py / ollama.py / anthropic.py.
     kwargs.pop("hashing_kv", None)
-    kwargs.pop("max_tokens", None)
     kwargs.pop("stream", None)
+    # max_tokens is injected by use_llm_func_with_cache when configured.
+    # llama_index_complete_if_cache has no top-level max_tokens parameter
+    # of its own, but forwards chat_kwargs straight into the underlying
+    # LLM's achat(), so fold it in there instead of discarding it silently.
+    max_tokens = kwargs.pop("max_tokens", None)
+    if max_tokens is not None:
+        chat_kwargs = dict(kwargs.get("chat_kwargs") or {})
+        chat_kwargs.setdefault("max_tokens", max_tokens)
+        kwargs["chat_kwargs"] = chat_kwargs
     result = await llama_index_complete_if_cache(
         kwargs.pop("llm_instance", None),
         prompt,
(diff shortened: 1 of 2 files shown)


danielaskdd (changes requested):
Thanks for the PR — both defects you identified are real, and I reproduced them on `main`. The direction is right and there is no regression risk here. But the `llm_instance` fix is incomplete: after applying this patch, `llama_index_complete` still cannot be used as a `llm_model_func` in LightRAG. Please address the points below before this is merged.

## What I verified

**Defect 1 (`llm_instance`) — confirmed on `main`.** `llama_index_complete_if_cache` has no `llm_instance` parameter and no `**kwargs` catch-all, so every call raises:

```
TypeError: llama_index_complete_if_cache() got an unexpected keyword argument 'llm_instance'
```

**Defect 2 (blocking embed) — confirmed, and the fix is correct.** I checked `llama-index-core` 0.14.24: `BaseEmbedding._aget_text_embeddings` has a base-class default (`llama_index/core/base/embeddings/base.py:300`), so there is no `AttributeError` risk, and the three mainstream integrations all override it with genuine batched async — OpenAI (`base.py:474`), HuggingFace (`base.py:335`, via `asyncio.to_thread`), Ollama. This part is good to go.

## Required change: the kwargs leak is not fully closed

Popping only `llm_instance` fixes the isolated call but not the real wiring. With this patch applied:

```
bare llm_instance only:                     OK -> 'hello back'
llm_instance + hashing_kv (real wiring):    TypeError -> ... unexpected keyword argument 'hashing_kv'
llm_instance + hashing_kv + max_tokens:     TypeError -> ... unexpected keyword argument 'hashing_kv'
```

The role LLM wrapper injects `hashing_kv` **unconditionally** on every call — `lightrag/llm_roles.py:194-198`:

```python
partial(
    raw_func,
    hashing_kv=self.llm_response_cache,
    **model_kwargs,
)
```

So the moment `llama_index_complete` is actually used as `llm_model_func`, `hashing_kv` lands in `kwargs`, gets forwarded into `llama_index_complete_if_cache`, and the same `TypeError` fires under a different parameter name. Two more leak sources exist: `use_llm_func_with_cache` (`lightrag/utils.py`) forwards `max_tokens` when set, and the query path forwards `stream=` (`lightrag/operate.py:4853`, `lightrag/operate.py:6905`).

Every other provider in this repo already guards against this — `llama_index_impl.py` is the only one missing it:

```
lightrag/llm/openai.py:414      kwargs.pop("hashing_kv", None)
lightrag/llm/ollama.py:296      kwargs.pop("hashing_kv", None)
lightrag/llm/anthropic.py:91    kwargs.pop("hashing_kv", None)
```

Please follow that established pattern and strip `hashing_kv`, plus handle `max_tokens` and `stream` (note this adapter has no streaming support at all, so `stream` needs an explicit decision, not a silent drop). Explicit pops are preferred over widening `llama_index_complete_if_cache` with a `**kwargs` catch-all, to stay consistent with the other providers.

## Required change: the test does not pin the real failure path

`test_llama_index_complete_accepts_llm_instance_kwarg` calls `llama_index_complete("hi", llm_instance=llm)` with nothing else in `kwargs`. It passes while the actual integration still crashes. Adding a single `hashing_kv=object()` argument to that call would have caught the gap — please add a case that exercises the real calling convention.

## Please also correct the test docstring

The docstring states this is "the documented calling convention (see the repo's own unofficial-sample llama_index demo scripts)". That is not accurate: none of `examples/unofficial-sample/lightrag_llamaindex_direct_demo.py`, `lightrag_llamaindex_litellm_demo.py`, or `lightrag_llamaindex_litellm_opik_demo.py` ever calls `llama_index_complete` — they all call `llama_index_complete_if_cache` directly.

Worth noting: those demos carry the same bug this PR fixes. For example `lightrag_llamaindex_direct_demo.py:48-53` passes `kwargs["llm_instance"]` positionally *and* expands `**kwargs` still containing `llm_instance`, producing the identical `TypeError`. Fixing them in this PR would be a welcome addition, though I will not block on it.

## Optional

For the embed change, `await asyncio.to_thread(embed_model._get_text_embeddings, texts)` would be strictly safer for third-party embedders that implement batched sync but not async — the base-class `_aget_text_embeddings` default fans out via `asyncio.gather` into N single-text calls, each still blocking. Your choice is better for OpenAI/Ollama/HuggingFace (true async I/O), so I am fine with keeping it; just flagging the trade-off.

---

Shizoqua:
@danielaskdd Now pops hashing_kv, max_tokens, and stream, matching the other providers. Added a test with the real calling convention.

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd (approved):
all findings addressed

PR #3850, opened 2026-09-06
fix(retrieval): preserve candidate order for vector similarity ties

Anormalm:
## Description

Preserve first-occurrence order when deduplicating candidate chunk IDs in `pick_by_vector_similarity`.

The previous set-to-list conversion made tied-score selection and the storage-error fallback depend on Python's hash seed. With a result limit, identical ordered candidates could therefore yield different evidence across processes. An insertion-ordered dictionary keeps deduplication while giving the existing stable sort a deterministic tie-break.

Distinct-score ranking, missing-vector handling (`[]` for the caller's WEIGHT fallback), and caller inputs are unchanged. This guarantees repeatability for fixed ordered candidates, not invariance to arbitrary candidate permutations.

## Tests

Native editable installation on Ubuntu / Python 3.12.14, using the actual `lightrag.utils` import:

```bash
PYTHONHASHSEED=0 python -m pytest tests/utils/test_pick_by_vector_similarity_order.py -q
# unchanged source: 11 failed, 10 passed
# fixed source:     21 passed
```

The fixed suite also passes all 21 cases with `PYTHONHASHSEED=1,2,3,7,42`. It covers equal and distinct scores, duplicates, missing vectors, storage failures, empty inputs, and query-embedding generation.

Ruff 0.15.17 (the repository's pinned version) reports zero source/test lint findings; test formatting and `git diff --check` pass. No external model or database calls were used; storage responses are controlled test doubles. The full repository suite and live-model evaluation were not run.

[Executed validation and retained logs](https://github.com/Anormalm/LightRAG/actions/runs/34042440104). The validation workflow is fork-only and is not included in this PR.

---

## Maintainer follow-up (commit `d09252c`)

The regression tests were originally added at `tests/llm/test_vector_similarity_order.py` with no
`offline` marker. CI runs `pytest tests/ -m offline`, so all of them were deselected — the fix would
have shipped unguarded. Follow-up commit:

- Moved the file to `tests/utils/test_pick_by_vector_similarity_order.py`, mirroring `lightrag/utils.py`
  per the repository test layout and placing it beside the existing `test_pick_by_vector_similarity.py`.
- Added `pytestmark = pytest.mark.offline` so the CI gate selects it.
- Dropped `asyncio.run` in favour of the repository's `asyncio_mode = "auto"`, and added a module docstring.
- Removed the missing-vector case already covered by `test_pick_by_vector_similarity.py`.

Verification: `./scripts/test.sh tests/utils -m offline` → **307 passed, 15 deselected**, with 19 cases
from the new file now selected (previously 0). Reverting `lightrag/utils.py` to the pre-fix source fails
13 of those 19 on assertion, so the suite still proves the fix. `ruff check` and `pre-commit` clean.

### DIFF
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -6013,11 +6013,12 @@ async def pick_by_vector_similarity(
     if not entity_info or num_of_chunks <= 0:
         return []
 
-    # Collect all unique chunk IDs from entity info
-    all_chunk_ids = set()
-    for i, entity in enumerate(entity_info):
-        chunk_ids = entity.get("sorted_chunks", [])
-        all_chunk_ids.update(chunk_ids)
+    # Preserve first occurrence order for similarity ties and fallback selection.
+    all_chunk_ids = dict.fromkeys(
+        chunk_id
+        for entity in entity_info
+        for chunk_id in entity.get("sorted_chunks", [])
+    )
 
     if not all_chunk_ids:
         logger.warning(
--- a/tests/llm/test_vector_similarity_order.py
+++ b/tests/llm/test_vector_similarity_order.py
@@ -1,0 +1,134 @@
+import asyncio
+from copy import deepcopy
+from itertools import permutations
+from unittest.mock import AsyncMock
+
+import numpy as np
+import pytest
+
+from lightrag.utils import pick_by_vector_similarity
+
+
+class VectorStore:
+    def __init__(self, vectors):
+        self.vectors = vectors
+        self.requested_ids = None
+
+    async def get_vectors_by_ids(self, ids):
+        self.requested_ids = list(ids)
+        # The returned mapping's order must not control tie-breaking either.
+        return {key: self.vectors[key] for key in reversed(ids) if key in self.vectors}
+
+
+def _select(entities, vectors, count, *, embedding=None, query_embedding=None):
+    store = VectorStore(vectors)
+    result = asyncio.run(
+        pick_by_vector_similarity(
+            query="evidence",
+            text_chunks_storage=None,
+            chunks_vdb=store,
+            num_of_chunks=count,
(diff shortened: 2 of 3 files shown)


danielaskdd:
@codex review

danielaskdd:
Thanks for the fix — the root cause analysis is right, and the `dict.fromkeys` change is the minimal correct one: `sorted_chunks` is already ranked, so first-occurrence order is a meaningful tie-break rather than just an arbitrary stable one.

I pushed one follow-up commit (`d09252c`) to your branch rather than asking for another round, since it was a small mechanical fix.

**The blocker:** the tests were at `tests/llm/test_vector_similarity_order.py` with no `offline` marker. CI runs `pytest tests/ -m offline`, so every one of them was deselected:

```
$ pytest tests/llm/test_vector_similarity_order.py -m offline
21 deselected in 0.02s
```

The regression they guard would not have been caught by CI. The follow-up:

- moves the file to `tests/utils/test_pick_by_vector_similarity_order.py` (mirrors `lightrag/utils.py` per the layout in `AGENTS.md`, and sits next to the existing `test_pick_by_vector_similarity.py`)
- adds `pytestmark = pytest.mark.offline`
- drops `asyncio.run` in favour of the repo-wide `asyncio_mode = "auto"`
- removes the missing-vector case, already covered by `test_pick_by_vector_similarity.py`

`./scripts/test.sh tests/utils -m offline` → 307 passed, 15 deselected, with 19 cases from the new file now selected. Reverting `lightrag/utils.py` to the pre-fix source fails 13 of them on assertion, so they still prove the fix.

I have also merged `main` in and approved the workflow runs. Once CI is green this is good to merge.

Anormalm:
Thanks for catching the CI-selection gap, and thanks for merging!

PR #3851, opened 2026-09-06
feat(api): add delete_parsed_files flag to DELETE /documents

Shizoqua:
**Problem**

DELETE /documents (clear all) has no way to also remove the __parsed__ directory. A prior PR that deleted it unconditionally was rejected since parsed artifacts should stay recoverable by default.

**Change**

Add an opt in delete_parsed_files query flag, default false, matching the maintainer's suggested shape from that review.

**Tests**

- New tests: 3 passed
- Full api/routes suite: 571 passed (19 pre-existing unrelated failures, missing ollama package in sandbox)
- Pre-commit passed
- git diff --check passed

**Related Issue**

No related issue.

### DIFF
--- a/docs/FileProcessingPipeline-zh.md
+++ b/docs/FileProcessingPipeline-zh.md
@@ -832,9 +832,10 @@ __parsed__/<base>.docling_raw/
 | 首次解析 | 取回产物，然后原子写入 `_manifest.json`。docling 的取回过程是 `POST /v1/convert/file/async` → 长轮询 `/v1/status/poll/{task_id}?wait=N` → `GET /v1/result/{task_id}` → 安全解压 zip（拒绝绝对路径与 `..`）。 |
 | 重新解析（缓存命中） | 不调用外部服务，不重写产物；仅重跑 adapter + writer 重新生成 sidecar（这正是 adapter 升级代价很低的原因）。 |
 | 重新解析（缓存未命中） | 清空目录，重新取回并写 manifest。 |
-| `DELETE /documents` 且 `delete_file=True` | `*.parsed/`、原始产物包、源文件一并删除。 |
-| `DELETE /documents` 且 `delete_file=False` | 保留全部产物，仅删除 doc_status 与 KG 数据。 |
-| `clear_documents` / 整体清空 `__parsed__` | 随之一并清除。 |
+| `DELETE /documents/delete_document` 且 `delete_file=True` | `*.parsed/`、原始产物包、源文件一并删除。 |
+| `DELETE /documents/delete_document` 且 `delete_file=False` | 保留全部产物，仅删除 doc_status 与 KG 数据。 |
+| `DELETE /documents`（`clear_documents`）且 `delete_parsed_files=true` | 整体删除 `__parsed__`；顶层输入文件始终会被删除，与此参数无关。 |
+| `DELETE /documents`（`clear_documents`）且 `delete_parsed_files=false`（默认） | 保留 `__parsed__`；顶层输入文件仍会被删除。 |
 | scan 周期 | **不会**回收孤立的产物包——只有用户显式删除时才移除，避免误扫掉调试现场。 |
 
 强制重解析（完全绕过缓存）：`LIGHTRAG_FORCE_REPARSE_NATIVE` / `LIGHTRAG_FORCE_REPARSE_MINERU` / `LIGHTRAG_FORCE_REPARSE_DOCLING`（§3.7）。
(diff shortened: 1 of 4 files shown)


danielaskdd:
Thanks — this is exactly the opt-in shape I asked for when closing #3326: an explicit flag aligned with the per-document opt-in, defaulting to preserving `__parsed__/`. Direction approved. I checked out the head and ran the new tests locally: `tests/api/routes/test_clear_documents_parsed_files.py` → **3 passed**.

One note for future readers, since "why not just reuse the existing `delete_file` name?" is the obvious question: on the per-document endpoint, `delete_file` governs a free axis — keep or remove the user's uploaded file. On full clear that axis is already pinned to "remove": `DocumentManager.indexed_files` is an in-memory set (`lightrag/api/routers/document_routes.py:1441`) and `doc_status` has just been dropped, so anything left at the input-dir top level would be classified as brand new by the next `/documents/scan` and re-ingested — a full clear would partially undo itself. The only remaining free axis on this endpoint is `__parsed__/`, and `delete_parsed_files` names exactly that. Reusing `delete_file` would give two endpoints the same parameter name with different meanings (`delete_file=false` = "my originals are untouched" per-document, vs. "top-level originals are deleted anyway" here), which is a worse trap than a narrower, honest name. Keeping the name as proposed.

Findings to address before merge:

### 1. The lifecycle docs are already wrong, and this PR should fix them

`docs/FileProcessingPipeline.md:835-837` and its mirror `docs/FileProcessingPipeline-zh.md:835-837`:

```
| `DELETE /documents` with `delete_file=True`      | `*.parsed/`, the raw bundle, and the original file are all removed together. |
| `DELETE /documents` with `delete_file=False`     | All artifacts are preserved; only doc_status and KG data are deleted. |
| `clear_documents` / a full sweep of `__parsed__` | Naturally cleared together. |
```

Two defects:

- The first two rows label the **per-document** endpoint as `DELETE /documents`. The real path is `DELETE /documents/delete_document`; `DELETE /documents` is `clear_documents`. This mislabeling is very likely where the "full clear already handles `__parsed__`" impression comes from.
- The third row asserts that `clear_documents` clears `__parsed__` "naturally". It never has: the deletion loop is `input_dir.glob("*")` guarded by `if file_path.is_file()` (`document_routes.py:5995-6001`), so every subdirectory — `__parsed__/` included — is skipped. That row describes the behavior this PR introduces, and only under the new flag.

Please update both files in this PR: name the per-document path correctly, and make the `clear_documents` row read "top-level input files always; `__parsed__/` only with `delete_parsed_files=true`".

### 2. The default path stays silent about what it preserved

`parsed_dir_message` is only ever non-empty on the opt-in branch. A default clear still returns `All documents cleared successfully. Deleted N files.` while the entire raw corpus sits untouched in `__parsed__/`. Note that on a fully processed corpus that `N` is typically near zero — parsed source files are *moved* into `__parsed__/` (`move_file_to_parsed_dir`, `lightrag/utils.py:917`), so the top level only holds pending/failed leftovers. "Deleted 0 files. All documents cleared successfully." next to a full copy of every original document on disk is misleading enough to matter for anyone using this endpoint for compliance-driven deletion.

Please make the default path say so explicitly, e.g. append `__parsed__ preserved (pass delete_parsed_files=true to remove it).`

### 3. Record the asymmetry in the endpoint docstring

Three or four lines, covering: top-level input files are deleted unconditionally and this cannot become opt-out (the re-scan reasoning above — including the `SOURCE_CONFLICT` and `SOURCE_IDENTITY_UNKNOWN` files that scan deliberately leaves at the top level, `document_routes.py:3834-3841` and `:3881-3887`), whereas `__parsed__/` is opt-in for cache reuse and recovery. AGENTS.md asks for every accepted residue to be written down next to the code with its reason and recovery path; a preserved `__parsed__/` after a "clear all" is precisely such a residue.

### 4. Minor: partial `rmtree` failure

Worth one sentence in the same place: a failed `shutil.rmtree` can leave a partially removed tree, and the recovery is re-running the clear with the flag. The error handling itself is fine — it appends to `errors` and degrades to `partial_success`.

### WebUI — please take it as a follow-up PR

Keep this PR API-only; the UI belongs in its own change. But the flag is unreachable for end users until the UI exposes it, so I'd like you to do that follow-up as well:

- `lightrag_webui/src/api/lightrag.ts:1039` — `clearDocuments()` is a bare `delete('/documents')`; it needs to pass the query param.
- `lightrag_webui/src/components/documents/ClearDocumentsDialog.tsx` — add the checkbox, modeled on `DeleteDocumentsDialog.tsx:152-158`.
- **Do not reuse** the `documentPanel.deleteDocuments.deleteFileOption` copy. In the per-document dialog that checkbox means "keep or delete my uploaded files"; here the top-level files go regardless and the checkbox only governs `__parsed__/`. Identical wording for different behavior would mislead. Add a new key and translate it across every file under `src/locales/`.
- Per AGENTS.md, a WebUI change runs the whole frontend set from `lightrag_webui/`: `bun install --frozen-lockfile`, then `bun test`, `bunx tsc --noEmit`, `bun run lint` — please report the pass count.

---

Shizoqua:
@danielaskdd Docs and message updated. WebUI left as a separate PR per your note.

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3853, opened 2026-09-06
fix(storage): preserve successful NetworkX drops

### LINKED ISSUE #3849: Storage: NetworkXStorage.drop() reports a successful drop as an error when the reload notification fails
Sibling of the misclassification fixed in #3846 (bc7ed9e9), found by the storage-backend review of that change. Same shape, different call site and different failure surface, so it was deliberately left out of that PR.

## Summary

`NetworkXStorage.drop()` performs its destructive work first and notifies the other processes afterwards, inside one `try`:

```python
try:
    async with self._storage_lock:
        if os.path.exists(self._graphml_xml_file):
            os.remove(self._graphml_xml_file)          # durable
        self._graph = nx.Graph()                       # local state reset
        await set_all_update_flags(...)                # <-- can raise here
        self.storage_updated.value = False
    return {"status": "success", "message": "data dropped"}
except Exception as e:
    logger.error(...)
    return {"status": "error", "message": str(e)}
```

If `set_all_update_flags` raises, the file is already deleted and the in-memory graph is already empty — the drop **succeeded** — but the method reports `{"status": "error"}`. This is the same class of bug as the `index_done_callback` one fixed in #3846: a failure of the cross-process reload notification reported as a failure of the operation it follows.

## Why it is worth fixing separately

The consequence is not local. `/documents/clear` counts these status dicts:

- a non-`"success"` status increments `storage_error_count`;
- if **every** drop task reports non-success, the endpoint returns `ClearDocumentsResponse(status="fail", ...)` with "All storage drop operations failed. Aborting document clearing process." and skips the input-file deletion.

So on a deployment where the shared-storage notification is broken, a clear that actually destroyed the graph data can be reported to the caller as having destroyed nothing, and the input files are then left in place to be re-ingested against storage that no longer matches. Telling an operator their data survived when it did not is the opposite of the fail-loud posture the rest of this file holds.

Two smaller notes on the same path:

- `set_all_update_flags` flips one flag per registered process. A mid-loop failure leaves some processes notified and the remainder not, so any message here should not claim a total notification failure.
- On that path `self.storage_updated.value = False` is never reached, so this process's own flag can stay `True` until the next `_get_graph` triggers a redundant (harmless) reload.

## Scope

- separate the destructive work from the notification, the way `index_done_callback` now does: a notification failure is logged and does not change the reported status of a drop that already happened;
- decide what a *partial* notification failure should say, given the per-process loop;
- a regression test at the backend level (drop succeeds, hook raises, status is `"success"`, failure is in the log), and one at the `/documents/clear` level pinning that such a drop is not counted as a storage error.

Worth auditing the other backends' `drop()` for the same ordering while in there.

be-student:
## Description

A completed NetworkX GraphML deletion now remains a successful drop when the later cross-process reload notification fails. The failure is logged as potentially partial, and the writer reload flag is reset independently so `/documents/clear` can safely continue deleting input files.

## Related Issues

Closes #3849.

## Changes Made

- separate the durable file deletion from peer notification status
- add backend coverage for notification failure and destructive failure
- add endpoint coverage proving input deletion continues after a successful NetworkX drop
- document the destructive-status invariant in `AGENTS.md`

## Validation

- `./scripts/test.sh tests/kg/networkx_impl` — 94 passed
- focused backend and clear-route tests — 7 passed
- Ruff check and format check passed
- mutation returning an error for notification failure made both regressions fail

## Backend audit

JSON KV and doc-status drops notify before their empty state is persisted, so their notification errors still represent incomplete drops. FAISS and NanoVectorDB delete their files before notification and retain the same classification shape; their backend-specific reset and redo-log semantics should be handled in separate follow-ups rather than folded into this NetworkX issue.

AI-assisted tooling was used; I reviewed the change and ran the reported checks.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -51,6 +51,7 @@ LightRAG uses 4 storage types with pluggable backends:
 
 Each `LightRAG` instance can pass a `workspace` parameter for data isolation. Implementation differs per storage type:
 - **File-based**: subdirectories under `working_dir`.
+- **NetworkX destructive status**: `NetworkXStorage.drop()` reports the durable GraphML deletion. A later peer reload-notification failure is logged as partial propagation and must not turn that completed deletion into an error response; `/documents/clear` uses the status to decide whether input files are safe to delete.
 - **Collection-based**: collection name prefixes.
 - **Relational DB**: workspace column filtering.
 - **Qdrant**: payload-based partitioning.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -51,7 +51,6 @@ LightRAG uses 4 storage types with pluggable backends:
 
 Each `LightRAG` instance can pass a `workspace` parameter for data isolation. Implementation differs per storage type:
 - **File-based**: subdirectories under `working_dir`.
-- **NetworkX destructive status**: `NetworkXStorage.drop()` reports the durable GraphML deletion. A later peer reload-notification failure is logged as partial propagation and must not turn that completed deletion into an error response; `/documents/clear` uses the status to decide whether input files are safe to delete.
 - **Collection-based**: collection name prefixes.
 - **Relational DB**: workspace column filtering.
 - **Qdrant**: payload-based partitioning.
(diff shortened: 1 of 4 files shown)


danielaskdd:
@codex review

PR #3862, opened 2026-09-07
fix(milvus): run search/upsert/delete off the event loop thread

Shizoqua:
## Problem

MilvusVectorDBStorage.query() and _flush_pending_vector_ops() call MilvusClient's search/upsert/delete directly inside async methods. MilvusClient is a synchronous SDK (blocking gRPC), so every call blocks the whole event loop for the round trip, stalling every other concurrent task.

## Change

Wrap the search/upsert/delete calls (and the collection-load check before them) in asyncio.to_thread(). Scoped to the hot read/write path; startup/migration/admin methods in this file share the same limitation and are left for a follow-up.

## Tests

tests/kg/milvus_impl: 196 passed

## Related Issue

No related issue.

### DIFF
--- a/lightrag/constants.py
+++ b/lightrag/constants.py
@@ -636,6 +636,13 @@ CHUNKING_SUBMIT_LIMIT = 8
 # worker.
 STORAGE_IO_SUBMIT_LIMIT = 8
 
+# The Milvus pool is the exception to the single-worker shape above: its work is
+# blocking gRPC I/O, not CPU, so one worker would serialize every search in the
+# process behind one flush. This one constant is BOTH the worker count and the
+# submission ceiling, so a submission over the limit waits on the semaphore
+# instead of growing the executor's unbounded wait queue.
+MILVUS_SUBMIT_LIMIT = 8
+
 # Per-workspace ceiling on manual retry requests that have been published but
 # not yet ACKed (LR2 §10.1). The channel is sticky — a request survives until an
 # exclusive reset acknowledges it — so an operator hammering /reprocess_failed
--- a/lightrag/kg/milvus_impl.py
+++ b/lightrag/kg/milvus_impl.py
@@ -101,6 +101,19 @@ async def run_in_milvus_executor(
     caller releases the awaiter while the call itself runs to completion in the
     pool. Callers that must not return before the call has landed keep their own
     deferral (see `_flush_pending_vector_ops`).
+
+    Accepted residue: the semaphore is per event loop (it has to be --
+    `asyncio.Semaphore` binds to the first loop that waits on it, see
+    `get_loop_semaphore`) while the executor is per process, so N concurrently
+    active loops in one process could have N * MILVUS_SUBMIT_LIMIT submissions
+    outstanding. That bounds nothing worse than queue depth: concurrent Milvus
+    round trips stay capped at the pool width process-wide, because a blocking
+    SDK call in flight IS an occupied worker and `ThreadPoolExecutor` never
+    starts more than `max_workers` of them; and a queued submission holds only
+    arguments its awaiting caller already keeps alive, so the queue duplicates
+    no memory. The deployment shape has one active loop per process anyway
+    (gunicorn forks a worker per loop, each with its own gRPC channel), and the
+    three sibling pools carry the identical per-loop/per-process split.
     """
     semaphore = get_loop_semaphore("milvus", MILVUS_SUBMIT_LIMIT)
     return await bounded_submit(get_milvus_executor(), semaphore, fn, *args, **kwargs)
--- a/lightrag/kg/milvus_impl.py
+++ b/lightrag/kg/milvus_impl.py
@@ -1,13 +1,17 @@
 import asyncio
 import json
 import os
+import threading
 import time
-from typing import Any, final, Optional, Dict
+from concurrent.futures import ThreadPoolExecutor
+from typing import Any, Callable, final, Optional, Dict
 from dataclasses import dataclass, fields
 import numpy as np
 from lightrag.utils import (
     logger,
+    bounded_submit,
     compute_mdhash_id,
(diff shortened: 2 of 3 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
Pushed a follow-up commit to this branch (`8fd1adc`) — thanks for the fix, the direction is right and the cancellation deferral around the flush write is correct as written.

**What the commit changes.** Moving the blocking `MilvusClient` calls off the loop with `asyncio.to_thread` frees the loop, but it lands them on the process **default** executor — which this repo explicitly rules out for exactly this kind of work (see the warning block on `get_storage_io_executor` in `lightrag/utils.py`): that pool is shared with `UnifiedLock._acquire_mp_lock_in_executor`, login password hashing, and the document routes' `stat()` calls, and its wait queue is unbounded. A 64MB upsert batch — or a cold `load_collection`, which pymilvus polls to completion — would then delay `pipeline_status` / namespace lock acquisition under gunicorn and `/login`, i.e. trade an unbounded private resource (the loop) for a bounded shared one.

So the five call sites now go to a dedicated Milvus pool, mirroring the chunking / storage-IO pools:

- `MILVUS_SUBMIT_LIMIT` (in `constants.py`, next to its three siblings) is **both** the worker count and the semaphore capacity, so a submission over the limit waits on the semaphore instead of growing the executor's unbounded queue.
- Multi-worker, unlike those siblings, and deliberately **not** `run_in_storage_io`: this is blocking gRPC I/O, not CPU, and a single worker would serialize every search in the process behind one flush (mix mode alone fans out three searches per query).
- `bounded_submit` has the same cancellation semantics as `asyncio.to_thread` (thread runs to completion, awaiter released), so your `_wait_deferring_cancellation` wrapper is untouched and still covers the buffer bookkeeping.
- Two regression tests pin it: one asserts every blocking call runs on a `lightrag-milvus` thread rather than an `asyncio_*` one, the other that concurrent searches never exceed the ceiling. Both fail on the `to_thread` version (`12 <= 8` / `asyncio_0`).

One note on the knob: `8` is a judgement value, not a derived one. Versus the `to_thread` state it is a slight reduction in parallelism (the default pool is `min(32, cpu+4)`); versus `main` it is 8× rather than serialized-on-the-loop. Happy to change the number if you have a deployment that argues for more.

**Withdrawing my second review finding** (`query()` now able to interleave with `finalize()`'s `client.close()`) — no action needed, please don't spend time on it. On re-check: every other server-backed backend's `query()` awaits network I/O, so that interleave is an already-accepted residue across PG / Neo4j / Qdrant / Mongo / OpenSearch and Milvus was the outlier, not the baseline; the pre-change `query_embedding=None` path already awaited `embedding_func` anyway; and the direction is harmless — `query` is read-only and `finalize` flushes under `_flush_lock` before `close()`, so no write is at risk. An in-flight counter on `close()` would be machinery no sibling backend has.

`tests/kg/milvus_impl`: 199 passed.

danielaskdd:
@codex review

danielaskdd:
Re the P2 "Make the Milvus submission limit process-wide" — not taking this one; documented the residue instead in `c06f9fe` (`run_in_milvus_executor` docstring). Reasoning, since it is a judgement call:

**Scope first: this is cross-*loop*, not cross-process.** Cross-process admission is a non-question — each gunicorn worker forks its own pymilvus client and its own gRPC channel, so per-process is the natural unit and no in-process semaphore could span it anyway. Coordinating it would mean an mp-lock admission gate on the query hot path, which is precisely the cost this repo has spent several PRs removing from `pipeline_status`. A fleet-wide total belongs in Milvus server-side quota, or in lowering `MILVUS_SUBMIT_LIMIT` per worker.

**On the in-process claim: it conflates queue depth with concurrency.** Concurrent Milvus round trips are capped at the pool width **process-wide**, regardless of how many loops submit: a blocking SDK call in flight *is* an occupied worker, and `ThreadPoolExecutor._adjust_thread_count` starts a thread only `if num_threads < self._max_workers` (CPython `concurrent/futures/thread.py`). A second loop can enqueue more submissions; it cannot produce a ninth in-flight gRPC call. And a queued submission holds only argument objects its awaiting caller already keeps alive, so the queue duplicates no memory — including the upsert batches. What actually varies with N loops is ordering fairness between them, which is moot at N=1. So the backpressure the change intends to provide (do not hammer Milvus, do not spawn unbounded threads) is intact.

**The multi-loop shape does not exist here.** `utils.py:4593` / `utils.py:4993` are same-thread `run_until_complete` sync wrappers, and `parser/llm_bridge.py` / `chunker/semantic_vector.py` use `run_coroutine_threadsafe` to submit *into* the one loop from a worker thread, not to run a second one.

**And per-loop is forced, not chosen.** `get_loop_semaphore`'s docstring spells it out: `asyncio.Semaphore` binds to the first loop that waits on it and raises from any other (`asyncio.mixins._LoopBoundMixin`). The tokenizer / chunking / storage-IO pools all carry the identical per-loop-semaphore + per-process-executor split, so changing it is a repo-wide redesign of a decided pattern rather than a Milvus fix. The suggested alternative — an executor per loop — would be strictly worse: N × 8 threads is the unbounded concurrency the finding is trying to prevent.

**Driver layer cannot help, structurally.** pymilvus 3.0.1 has no client-side admission gate on the query/upsert path — the only `ThreadPool`/`Semaphore` in the package is `bulk_writer/volume_file_manager.py`. A synchronous SDK over one multiplexed gRPC channel has no notion of "too many outstanding requests": the number of blocked threads *is* the in-flight count. Which is the point — the executor width **is** the driver-level control, because it is the only one that exists. Server-side limits (`proxy.maxTaskNum`, quotas) reject rather than backpressure, and are per-server rather than per-client.

danielaskdd:
@codex review

Issue #3378, opened 2026-07-08
[Question]: Mineu has some problem


lyydd:
### Do you need to ask a question?

- [ ] I have searched the existing question and discussions and this question is not already answered.
- [ ] I believe this is a legitimate question, not just a bug or feature request.

### Your Question

ERROR: Parse worker failed (mineru): MinerU local task submission for 'xxx.pptx' failed: HTTP 400 {"detail": "Unsupported file type: unknown"}

.png/.jpg is ok. but .pptx/.doc/.pdf don't work.

danielaskdd:
Please submit the file to the official online parsing service to check if it can be successfully parsed: https://mineru.net/OpenSourceTools/Extractor

PR #3865, opened 2026-09-07
fix(parser): preserve text in nested PPTX groups

er-s-an:
## Description

The legacy PPTX extractor drops text inside grouped shapes. `slide.shapes` exposes the group container, but the container has no `text` property, so its text-bearing children are never visited. A deck with grouped content can be indexed incompletely; a deck containing only grouped text fails parsing with `extracted no usable text`.

This change visits nested group children in shape-tree order while preserving the existing handling of leaf text and newlines.

## Related Issues

Found and reproduced on current main (`7eb87008`). I did not find an existing issue or PR covering legacy PPTX group traversal in the related open/closed PR and issue searches. #3378 concerns remote MinerU parsing and is not addressed here.

## Changes Made

- Recurse into PPTX group shapes, including nested groups, using python-pptx's existing shape API.
- Preserve slide/shape order, paragraph and soft-break text, and empty-textbox newline behavior; ignore empty groups and non-text shapes as before.
- Add serialized PPTX regressions for ungrouped, grouped and nested text, plus a group-only document test through `LegacyParser.parse` that verifies persisted content and the archive handoff.

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed
- [x] Documentation updated (extractor docstring)
- [x] Unit tests added

## Additional Notes

Validation on Python 3.13.13 / python-pptx 1.0.2:

- Before the fix: **3 failed, 8 passed** across the two affected test files. The ungrouped control passed; grouped/nested content disappeared, and the group-only document was rejected as empty.
- After the fix: **11 passed** across those files.
- `tests/parser`: **1534 passed, 45 skipped**. Optional pinned spaCy models are not installed. Localhost fixtures and tokenizer cache downloads required network-enabled execution; macOS SVG tests additionally needed the installed Homebrew Cairo path (`DYLD_FALLBACK_LIBRARY_PATH=/opt/homebrew/lib`) with a direct Python invocation.
- `_extract_pptx` and its traversal helper: all 13 statements and all 6 measured branches covered.
- Repository-pinned pre-commit checks passed for the three changed files; `git diff --check` passed.

No new dependency, public API, ZIP-budget change, or parser-routing change. This does not add table/chart extraction, OCR, notes/master-slide text, or geometric reading-order inference. Tests generate actual PPTX files locally; external LLM services and desktop PowerPoint rendering were not exercised. Previously indexed documents are not automatically reprocessed.

### DIFF
--- a/lightrag/parser/legacy/extractors.py
+++ b/lightrag/parser/legacy/extractors.py
@@ -81,17 +81,25 @@ def _extract_docx(file_bytes: bytes) -> str:
 
 
 def _extract_pptx(file_bytes: bytes) -> str:
-    """Extract PPTX content (synchronous)."""
+    """Extract PPTX text, including nested groups, in shape order (synchronous)."""
+    from collections.abc import Iterator
+
     from pptx import Presentation  # type: ignore
+    from pptx.enum.shapes import MSO_SHAPE_TYPE
+    from pptx.shapes.shapetree import GroupShapes, SlideShapes
 
     pptx_file = BytesIO(file_bytes)
     prs = Presentation(pptx_file)
-    content = ""
-    for slide in prs.slides:
-        for shape in slide.shapes:
-            if hasattr(shape, "text"):
-                content += shape.text + "\n"
-    return content
+
+    def iter_text(shapes: SlideShapes | GroupShapes) -> Iterator[str]:
+        for shape in shapes:
+            # Groups have no text of their own; their children can include groups.
+            if shape.shape_type == MSO_SHAPE_TYPE.GROUP:
+                yield from iter_text(shape.shapes)
+            elif hasattr(shape, "text"):
+                yield shape.text + "\n"
+
+    return "".join(text for slide in prs.slides for text in iter_text(slide.shapes))
 
 
 def _extract_xlsx(file_bytes: bytes) -> str:
--- a/tests/parser/test_legacy_extractors.py
+++ b/tests/parser/test_legacy_extractors.py
@@ -8,6 +8,9 @@ from xml.etree import ElementTree as ET
 
 import pytest
 from openpyxl import Workbook
+from pptx import Presentation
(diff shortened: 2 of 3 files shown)


danielaskdd:
@codex review

er-s-an:
Thanks for merging this! I'd like to keep helping with parser/chunking reliability—reproducing issues, adding regression tests, and making focused fixes. Is there a particular area or set of issues where ongoing help would be useful? I'm happy to start with a scoped responsibility and follow your review process.

danielaskdd:
Thanks @er-s-an — and thanks for this fix in particular. Text silently vanishing from nested PPTX groups is exactly the kind of defect that is expensive to notice downstream: the document parses, ingests, and reaches `PROCESSED` with content missing and nothing anywhere saying so. Your `C`-selector work in #3665 had the same quality: you took the parts that were decided and left the undecided parts undecided, which made it reviewable.

On your question — yes, there is a scoped area, and it is the one you already asked about. The next slice on the chunking side is now tracked as **#3868: Server-startup discovery and registration of third-party chunkers**. That is the injection seam you proposed on #3664: discover third-party chunkers at startup, register them under allowlisted names, let an operator select one by name, and inject it as the callback `C` resolves to. It is yours if you want it.

#3868 is deliberately **requirements only**. The entry-point group name, the registry module placement and spec shape, the configuration key name, and the wiring are all yours to decide — I wrote down what the result has to do, not how. The last line of that issue is meant literally: if a requirement forces an ugly design, push back on the requirement rather than working around it.

Same boundaries you proposed and I agreed to before: `C(name=...)`, the context-aware callback contract, and identity/version *enforcement* stay out. #3868 names the constraint each of those leaves behind so the slice does not paint them into a corner.

Beyond that, parser/chunking reliability with regression coverage is where ongoing help is most useful, and the pattern you have been following — reproduce, pin with a test, fix narrowly — is the right one. Take #3868 first if you want it; comment there so it is visible you have it.

---

PR #3869, opened 2026-09-07
feat: select installed custom chunkers at server startup

### LINKED ISSUE #3868: RFC: Server-startup discovery and registration of third-party chunkers
## Summary

Define the functional requirements for **discovering, registering and selecting a third-party chunking implementation at LightRAG Server startup**, and injecting the selected one as the callback that the `C` process-option resolves to.

Spun out of #3664 so the registry work has its own discussion and tracking thread. #3664 remains open for the part it still owns — the strategy-agnostic transform that composes with `R`/`V`/`P` (see the discussion with @geraldtscharre there and in #3603).

This issue is **requirements only**. Module layout, the spec dataclass shape, where the registry lives, and how it is wired are the implementer's to decide.

## Context

#3665 landed the `C` selector: it is accepted and persisted everywhere `F/R/V/P` are, it dispatches to `LightRAG.chunking_func` through the legacy six-argument contract, it is rejected with a 4xx at caller-present ingress when the callback is still the built-in default, and it takes an observable fixed-token fallback (one warning per processing attempt, distinct `chunk_method`) on paths with no caller to receive a rejection.

What is still missing is the other half: on current `main`, `create_app()` constructs `LightRAG` without a custom `chunking_func`, so in a Server deployment `C` can only ever fall back. This issue closes that gap.

The mechanism to mirror is the parser plugin seam — `lightrag/parser/plugins.py` (entry-point discovery) plus `lightrag/parser/registry.py` (import-cheap spec table, lazy implementation import).

## Scope

**In scope.** Making a chunker that ships *outside* the LightRAG package selectable by an operator, discovered and registered at Server startup, and injected as the callback `C` resolves to.

**Out of scope, unchanged from #3664.** `C(name=...)`, the context-aware callback contract (§4 of that RFC), identity/version *enforcement*, more than one simultaneously active custom chunker, and the strategy-agnostic transform hook.

Nothing here changes what `C` means, how it is admitted, or how it falls back. It changes only **where the callback comes from**. Everything #3665 landed stays as written.

## 1. Discovery

- **FR-1.1** A third-party package MUST be able to publish one or more chunkers with no change to LightRAG source, and with no Python import path anywhere in operator configuration.
- **FR-1.2** Discovery MUST run once per process at Server startup, **before** configuration validation, so the selection key may name a third-party chunker.
- **FR-1.3** Discovery MUST be import-cheap: registering N chunkers imports 0 chunker implementations. Only the selected one is imported (FR-4.2).
- **FR-1.4** A plugin that raises during discovery MUST be logged with its entry-point name and origin and skipped; it MUST NOT abort startup. Built-in `F/R/V/P` are static and unaffected. One exception: FR-4.3.
- **FR-1.5** The discovery entry point MUST be callable by library/SDK users embedding LightRAG directly — not server-only.

## 2. Registration and identity

- **FR-2.1** Registration is by **name**, and the name MUST be constrained now: lowercase, `[a-z0-9]` first character, then `[a-z0-9_-]`, max 64. The name will appear in logs, in persisted observability metadata, and — once `C(name=...)` exists — inside the `C(...)` hint parameter block, whose grammar already reserves `,` `=` `(` `)` and whitespace (`lightrag/parser/param_schema.py`; `chunk_token_size` / `chunk_overlap_token_size` already list `C` in their `targets`). Fixing the character class in this slice is what keeps that evolution open.
- **FR-2.2** Reserved names that MUST be refused at registration: the built-in snapshot keys `fixed_token`, `recursive_character`, `semantic_vector`, `paragraph_semantic`, and the bare selector letters `f`/`r`/`v`/`p`/`c`.
- **FR-2.3** A registration MUST declare at least: name; a lazily-resolvable implementation reference; an opaque `version` string; a one-line human description for the startup log.
- **FR-2.4** `version` is **observability only**. Nothing may be gated on it, for the reason already recorded on #3664: a string the author controls cannot detect a changed implementation behind an unchanged name, and a fence that does not hold is worse than no fence.
- **FR-2.5** A duplicate name MUST be reported at ERROR with both origins. If the duplicated name is the **selected** one, startup MUST fail — silently shadowing the chunker the operator selected is exactly the class of silence this work exists to remove. If it is not selected, last-wins plus the ERROR line is enough.
- **FR-2.6** The declared metadata MUST be extensible with new optional fields (e.g. "accepts `ChunkingContext`") without changing the entry-point contract or the configuration key.

## 3. Operator selection

- **FR-3.1** At most one registered chunker is selected per Server process, through a single configuration key (`.env` plus CLI arg, consistent with the other `CHUNK_*` keys).
- **FR-3.2** The value is a bare registered name from the closed set. Module paths, import strings, file paths and Python code MUST NOT be accepted — not from `.env`, not from the CLI, and never from an HTTP payload.
- **FR-3.3** A value that *looks* like an import path (contains `:` `/` `\` or `.`) MUST be rejected with a message saying import paths are not supported and pointing at the entry-point mechanism — not with a generic "unknown chunker". An operator who tried `my_pkg.mod:func` needs to be told the mechanism, not the symptom.
- **FR-3.4** Unset — the default, and every existing deployment — MUST reproduce today's behavior exactly: `chunking_func` stays the built-in default, caller-present `C` still returns 4xx, no-caller `C` still takes the observable fixed-token fallback. A deployment with no plugin installed MUST gain no new warnings and at most one new INFO line.

## 4. Startup resolution and failure semantics

- **FR-4.1** A key naming an unregistered chunker MUST fail startup with a non-zero exit and list the registered names. Falling back to the default here would give a whole deployment whose `C` documents are chunked the wrong way and still reach `PROCESSED`.
- **FR-4.2** Import and attribute lookup of the selected chunker MUST happen at startup, not on first document. Unimportable or non-callable MUST fail startup, not fail a document hours later.
- **FR-4.3** Combining FR-1.4 and FR-4.1: a broken plugin that was the provider of the selected name fails startup through the unknown-name path, while a broken plugin nobody selected does not. The two messages MUST be distinguishable — an operator reading the second one needs to know their `C` documents are still fine.
- **FR-4.4** When the resolved callable is introspectable, it MUST be verified to accept the six positional arguments of the legacy contract, and startup MUST fail when it cannot. When it is not introspectable, resolution proceeds. A fence where one holds, and no false confidence where it does not.
- **FR-4.5** Resolution MUST be identical under `lightrag-server` (uvicorn) and `lightrag-gunicorn` (preload plus workers). Two workers running different chunkers is not an acceptable outcome of any code path.

## 5. Injection and invocation

- **FR-5.1** The resolved callable is injected as `LightRAG.chunking_func` at construction, and that MUST remain the only thing that flips `C` from "will fall back" to "will run custom". The `is_builtin_chunker` identity test in `lightrag/pipeline.py` and `_validate_custom_chunking_available` in `lightrag/api/routers/document_routes.py` stay exactly as they are.
- **FR-5.2** The invocation contract is unchanged from #3665: six positional arguments, synchronous or async, on the event loop, never receiving private built-in arguments such as `_emit_source_span`, and a raised exception fails the document with actionable context — never a silent fallback to `F`.
- **FR-5.3** Injection is **instance-wide, not `C`-only**: with a chunker selected, a no-selector `ainsert` also routes through it, exactly as a constructor-supplied `chunking_func` does today. That is the correct behavior — registry injection must be indistinguishable from passing `chunking_func=` — but it is a consequence operators will not guess, so it MUST be stated in the startup log line (FR-6.1) and in the documentation.
- **FR-5.4** A registration SHOULD be able to declare that it is safe to run in the chunking executor, with the default staying on the event loop. Once the Server injects code it did not author, a CPU-bound chunker stalls the whole loop, and `run_in_chunking_executor` is the bounded pool that already exists for this. Declaring executor offload MUST be rejected at startup for an async callable. If the implementation shows this declaration does not earn its keep — the author can offload themselves — say so on this issue and it will be dropped.

## 6. Observability

- **FR-6.1** One startup INFO line: registered chunkers (name, version, origin) and which one is selected. When one is selected, the line MUST also say that it serves no-selector inserts (FR-5.3).
- **FR-6.2** `chunk_method` MUST keep its current values (`custom_chunking_func`, `custom_chunking_fallback_fixed_token`) so existing consumers and string matching keep working. *Which* chunker ran belongs in separate metadata fields alongside `chunk_opts`, marked explicitly non-authoritative.
- **FR-6.3** On reprocess or restart of a persisted `C` document whose recorded chunker name/version differs from the currently configured one — including "none configured now" — emit exactly one WARNING per processing attempt naming both, then proceed under the **current** configuration. The recorded identity MUST NOT gate any decision. It is the forensic aid, not the fence.
- **FR-6.4** Fallback-warning cadence is unchanged: once per processing attempt, per document, never collapsed or rate-limited across a batch.
- **FR-6.5** The `C` rejection at caller-present ingress SHOULD name the configuration key, and list the registered names when chunkers are registered but none is selected. Give the operator the fix, not just the diagnosis.

## 7. Documentation

- A `docs/ThirdPartyChunker.md` + `-zh` authoring guide, mirroring `docs/ThirdPartyParser*.md`.
- `env.example`, next to the `CHUNK_*` block.
- `docs/FileProcessingPipeline.md` + `-zh`: what `C` resolves to once a chunker is selected, plus FR-5.3.
- The `LightRAG.chunking_func` docstring and the `custom` API schema description.

## Acceptance criteria / test matrix

- [ ] No key set: `C` admission, fallback, warning cadence, and the no-selector legacy path are all unchanged.
- [ ] Key set to a registered chunker: `C` invokes it with the correct six arguments; `F/R/V/P` never do; no bypass warning; `chunk_method` is `custom_chunking_func` with name/version recorded.
- [ ] Key set to an unknown name: startup fails, message lists the registered names.
- [ ] Key set to an import-path-looking value: the FR-3.3 message, not the unknown-name one.
- [ ] Selected chunker unimportable / non-callable / wrong arity: startup fails.
- [ ] A broken plugin that is **not** selected: startup succeeds, ERROR logged, the selected chunker still works.
- [ ] Duplicate name, selected and not selected: the two different outcomes of FR-2.5.
- [ ] Reserved-name registration refused.
- [ ] Async registered chunker; registered chunker raising (document FAILED, no fallback).
- [ ] Persisted `C` document reprocessed after the key changed to a different chunker, and after it was removed entirely: one warning per attempt, current configuration wins.
- [ ] Sidecar-backfill eligibility on the fallback path still keys on what actually ran — the #3665 invariant must not regress.

## Deferred items, and the constraint each leaves here

- `C(name=...)` — FR-2.1's character class keeps it open.
- The context-aware callback contract — FR-2.6 keeps the declared metadata extensible.
- Identity/version enforcement — FR-6.3 keeps identity observability-only.
- Multiple simultaneously active chunkers — FR-3.1 says one *selection*, not one *registration*.
- The strategy-agnostic transform composing with `R`/`V`/`P` (#3603, #3664) is not addressed here.

## Open questions for the implementer

- Entry-point group name (`lightrag.chunkers` seems the obvious mirror of `lightrag.parsers`).
- Registry module placement and spec shape.
- The configuration key name (`CUSTOM_CHUNKER`?).
- Whether FR-5.4 earns its keep.

If any requirement above forces an ugly design, push back on the requirement here rather than working around it.

@er-s-an you offered to take this slice on #3664 — it is yours if you still want it, and nobody else is working on it.

er-s-an:
## Description

Implements the startup-discovery slice assigned in #3868. Server construction currently leaves `chunking_func` at its built-in default, so installing an external chunker does not make it usable by Server operators. This adds the missing discovery, validation and constructor-injection seam without changing the six-argument callback contract or the C admission/fallback rules from #3665.

## Related Issues

Closes #3868. Scope confirmed in https://github.com/HKUDS/LightRAG/issues/3868#issuecomment-5574247557.

## Changes Made

- Mirror the parser entry-point pattern with `lightrag.chunkers`, lightweight `ChunkerSpec` registrations and selected-only implementation imports. Preserve the existing chunker package exports lazily so importing the registry itself does not load built-in implementations.
- Select one bare registered name with `CUSTOM_CHUNKER` / `--custom-chunker`. Validate names, reserved names, duplicates, imports, callability and introspectable six-argument signatures before the shared uvicorn/Gunicorn app factory constructs LightRAG. Operator/HTTP input never provides an import reference.
- Isolate failed providers, including partial registrations. In response to the FR-4.3 clarification, each Server failed-provider ERROR reports the validated selection outcome: selected chunker available for C/no-selector inserts, startup aborted, or unset/default behavior unchanged. SDK discovery without selection remains available and does not claim validated selection.
- Keep C, no-selector and explicit F/R/V/P dispatch behavior intact. The one startup INFO and both authoring guides explain the instance-wide injection. Keep `executor_safe=False` by default; optional synchronous offload reuses the existing bounded executor, with async offload rejected at startup. I retained FR-5.4 because it needs no new executor/lifecycle machinery and allows CPU-bound plugins to use the existing bound instead of creating an unbounded offload path. It remains an author declaration, not a thread-safety guarantee.
- Record `{name, version, authoritative: false}` separately in `doc_status.metadata.custom_chunker`. Retain it across reset only to diagnose drift once per C attempt; current configuration always determines execution. Existing `chunk_method` values, callback-failure behavior and fallback/backfill invariants remain unchanged.
- Add English/Chinese authoring and pipeline documentation, env guidance, callback/API descriptions and mirrored regression tests.

## Validation

Python 3.13.13; final upstream base `0ac2c5b0`. No dependency or lockfile changes.

```bash
python -m pytest tests/chunker/test_registry.py tests/pipeline/test_custom_chunking_selector.py tests/api/test_custom_chunker_startup.py tests/api/config/test_api_config_custom_chunker.py -q
# 97 passed

python -m pytest tests/chunker tests/pipeline tests/api -q
# 2392 passed, 15 skipped, 8 failed — baseline comparison below
```

- New configuration, persisted-identity, lazy-import and selection-aware failure-log cases were observed failing before the respective implementation changes.
- Coverage across the two new registry/plugin modules: 160 statements and 46 branches, 100% exercised by 74 registry/startup tests. This is not project-wide coverage.
- Pinned pre-commit checks passed for all 19 changed/new files; compileall and `git diff --check` passed.
- Tests include a temporary installed-distribution metadata/entry-point fixture, both shared app-factory entry points, actual executor thread identity, Task-returning callbacks, offline document processing/reset, explicit built-in bypass and existing C fallback/backfill cases.

### Baseline failures and limits

The wider suite is **not all green** in this local environment. All eight failing cases also reproduce on the untouched `25ed861e` base tree: running `tests/api/config/test_embedding_dimension_guard.py tests/api/test_path_prefixes.py` there gives 47 passed / 8 failed. The later base update only changes Zhipu implementation/tests.

- Four existing Bedrock/Voyage dimension-guard expectations (missing exception/default dimension).
- Two existing mount/root-path redirect expectations (307 vs 200), and two prefix-auth expectations (403 vs 401).

I have not changed or diagnosed those unrelated behaviors here. The local run used the locked optional `ollama` dependency and Cairo library path; some optional tests were skipped. No live LLM service, real multi-worker deployment or published external plugin has been exercised. Startup validation cannot establish later third-party runtime/output correctness.

## Checklist

- [x] Changes tested locally (scope and failures above)
- [ ] Code reviewed
- [x] Documentation updated
- [x] Unit tests added

## Additional Notes

AI-assisted implementation and local diff inspection; no independent human review is claimed. Commits include DCO sign-off. This does not implement named C selectors, context-aware callbacks, identity enforcement, multiple active custom chunkers or the transform hook remaining in #3664.

### DIFF
--- a/docs/FileProcessingPipeline-zh.md
+++ b/docs/FileProcessingPipeline-zh.md
@@ -249,6 +249,10 @@ LIGHTRAG_PARSER=pdf:legacy-R(chunk_ts=800,chunk_ol=80);*:legacy-R  # 规则
 
 文本 API 以 `chunking.strategy="custom"` 暴露同一路径；`params` 使用完整的 fixed-token/legacy 参数契约（`chunk_token_size`、`chunk_overlap_token_size`、`split_by_character`、`split_by_character_only`）。如果没有替换 `LightRAG.chunking_func`，`/documents/text` 与 `/documents/texts` 会返回 422。
 
+Server 部署可以通过 `CUSTOM_CHUNKER=<注册名>`（或 `--custom-chunker`）在启动时选择已安装的 `lightrag.chunkers` 插件，见 [ThirdPartyChunker-zh.md](./ThirdPartyChunker-zh.md)。注入作用于**整个实例**：不带分块 selector 的插入也调用它，不限于 `C`。显式 F/R/V/P 仍调用内置策略。未配置时保持现有准入和回退行为；配置错误则启动失败，不会悄悄换回默认回调。
+
+`doc_status.metadata.custom_chunker` 独立记录最近一次尝试的注册名/版本及 `authoritative: false`，不改变 `chunk_method` 与 `chunk_opts`。该观察值跨重试/重置保留，仅供诊断，不是执行指令。持久化 C 文档在配置身份变化或移除后重处理，每次尝试告警一次并按**当前配置**继续。已有的逐次 fallback 告警独立保留；作者提供的版本号无法检测同名同版本背后的实现变化。
+
 > `drop_references` 检测调参 `CHUNK_P_REFERENCES_TAIL_N`（默认 `0`：扫描全部内容块；正数表示只扫描文末最后 N 块）/ `CHUNK_P_REFERENCES_HEADINGS`（竖线分隔，默认 `References\|Bibliography\|参考文献`）仅经环境变量、运行时实时读取。drop_references可以通过环境变量 `CHUNK_P_DROP_REFERENCES` 设置为全局默认值.
 
 ### 2.7 校验、优先级与回退
@@ -1117,7 +1121,7 @@ PENDING ─►├─ parse_queues["mineru"]  ─► [mineru 池  × N2] ─┼�
 4. **task 上限与请求上限不同**：`MAX_ASYNC_LLM` 设置 N5 单文档 chunk 抽取的 task 上限，以及其两倍的合并 task 上限。实际抽取和合并摘要请求使用 Extract 角色的上限：设置了 `EXTRACT_MAX_ASYNC_LLM` 时使用它，否则使用 `MAX_ASYNC_LLM`。缓存、图的 keyed lock 和角色上限都会让观测到的实际请求并发低于 task 并发。
 5. **queue size 与背压**：`QUEUE_SIZE_INSERT=4` 这个偏小的默认值是有意为之——process 阶段慢且占内存，让 analyze 阶段在队列写满时阻塞、再反压到 parse 阶段，避免一次性把成千上万份解析结果堆在内存里。
 6. **改后生效方式**：所有参数通过 `.env`（或环境变量）传入，仅在 `LightRAG` 实例构造时读取一次；改完需要重启服务。
-7. **分块不随并发增长**：分块在一个专用的单 worker 线程池里执行，目的是不阻塞事件循环，并发度不随 `MAX_PARALLEL_INSERT` 提高，调大并发不会让分块更快。自定义 `chunking_func` 仍在事件循环上执行（它的契约允许触碰运行中的事件循环），CPU 密集的实现应自行 `asyncio.to_thread`。
+7. **分块不随并发增长**：分块在一个专用的单 worker 线程池里执行，目的是不阻塞事件循环，并发度不随 `MAX_PARALLEL_INSERT` 提高，调大并发不会让分块更快。自定义 `chunking_func` 默认仍在事件循环上执行（它的契约允许触碰运行中的事件循环）。CPU 密集的实现应自行 offload，或由同步注册插件声明 `executor_safe=True` 复用有界分块线程池；异步插件不能启用此选项。
 
 **典型调优场景：**
 
--- a/docs/FileProcessingPipeline.md
+++ b/docs/FileProcessingPipeline.md
@@ -249,6 +249,10 @@ Currently supported parameters (canonical name / short alias):
 
 The text APIs expose the same path as `chunking.strategy="custom"`. Its `params` object uses the complete fixed-token/legacy contract (`chunk_token_size`, `chunk_overlap_token_size`, `split_by_character`, and `split_by_character_only`). `/documents/text` and `/documents/texts` return 422 unless `LightRAG.chunking_func` was replaced with a non-default callback.
 
+For Server deployments, `CUSTOM_CHUNKER=<registered-name>` (or `--custom-chunker`) selects an installed `lightrag.chunkers` plugin at startup; see [ThirdPartyChunker.md](./ThirdPartyChunker.md). This injection is **instance-wide**: no-selector inserts also use the callback, not just `C`. Explicit F/R/V/P still use built-ins. With the key unset, existing admission and fallback behavior is unchanged. Invalid selections fail startup rather than silently reverting to the built-in callback.
+
+`doc_status.metadata.custom_chunker` records the last attempted registered name/version with `authoritative: false`, separately from the unchanged `chunk_method` and `chunk_opts`. It survives retry/reset for diagnosis, not as an instruction. Reprocessing a persisted `C` document after changing/removing the configured identity warns once per attempt and proceeds under the current configuration. The existing per-attempt fallback warning is separate and unchanged. Author-supplied versions cannot detect implementation changes behind unchanged identities.
+
 > `drop_references` detection knobs `CHUNK_P_REFERENCES_TAIL_N` (default `0`: scan all content blocks; a positive value scans only the last N) / `CHUNK_P_REFERENCES_HEADINGS` (pipe-separated, default `References\|Bibliography\|参考文献`) are env-only and read live at run time. Global default can be set via env var `CHUNK_P_DROP_REFERENCES`.
 
 ### 2.7 Validation, Priority, and Fallback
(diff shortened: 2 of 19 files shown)


danielaskdd:
@codex review

PR #3876, opened 2026-09-08
fix(milvus): handle concurrent database creation race on first boot

### LINKED ISSUE #3875: [Bug]: Milvus first-boot database-creation race

### Describe the bug

`MilvusVectorDBStorage._create_milvus_client` in
[`lightrag/kg/milvus_impl.py#L603-L630`](https://github.com/HKUDS/LightRAG/blob/ab2e08fb191179f28ae2ea4c207ee2fcd826f9cb/lightrag/kg/milvus_impl.py#L603-L630)
creates the configured Milvus database with an unguarded check-then-act:

```python
existing_databases = set(client.list_databases())
if db_name not in existing_databases:
    logger.warning(...)
    client.create_database(db_name)   # not wrapped in try/except
```

`list_databases()` then `create_database()` is not atomic on the Milvus
server side. Any deployment that starts more than one LightRAG process
against the **same fresh Milvus database name** — a Kubernetes deployment
with `replicas > 1`, a manually-started second instance, or any other
multi-process/multi-host setup — can have two processes both observe the
database as absent and both call `create_database()`. Milvus rejects the
loser with `MilvusException` (server error code `65535`,
`"database already exist: <name>"`), which propagates uncaught out of
`initialize_storages()` and crashes that process's startup.

Because the race window only exists while the database doesn't exist yet, it
only manifests on a genuinely fresh volume — a restart afterward always
succeeds, which makes it easy to miss in single-replica manual testing and
easy to hit for the first time in an actual multi-replica production
rollout.

**Proposed fix**: wrap the `create_database()` call in a `try/except
MilvusException`, and treat the "already exists" case as benign — the state
it is trying to reach (the database exists) has already been reached by the
other process — while still re-raising any other `MilvusException`
unchanged. The match is on the exception's message (a case-insensitive
`"already exist"` substring check), not on `MilvusException.code`: the
numeric code for this condition isn't consistently assigned across Milvus
server versions, and re-checking via a second `list_databases()` call risks
observing stale cluster metadata and re-raising even though the create
actually succeeded. A PR implementing this is incoming and will reference
this issue.

### Steps to reproduce

Reproduce with the official `ghcr.io/hkuds/lightrag:latest` image against a
genuinely fresh, official Milvus standalone stack (etcd + MinIO + Milvus)
started via Docker Compose 

1. Start a fresh Milvus (etcd + MinIO + Milvus standalone, official images,
   no prior volume).
2. Start **two** LightRAG instances (`ghcr.io/hkuds/lightrag:latest`) at the
   same time, both configured with `LIGHTRAG_VECTOR_STORAGE=MilvusVectorDBStorage`
   and the same `MILVUS_URI` / `MILVUS_DB_NAME` pointing at the same
   never-before-used database name.
3. Observe: one instance boots normally; the other exits with an uncaught
   `MilvusException`.

### Expected Behavior

Both LightRAG processes complete `initialize_storages()` successfully. The
process that loses the create-database race should treat "database already
exists" as success (the desired end state — the database exists — has been
reached) rather than crashing.

### LightRAG Config Used

```
LIGHTRAG_VECTOR_STORAGE=MilvusVectorDBStorage
MILVUS_URI=http://milvus:19530
MILVUS_DB_NAME=lightrag
LIGHTRAG_KV_STORAGE=JsonKVStorage
LIGHTRAG_DOC_STATUS_STORAGE=JsonDocStatusStorage
LIGHTRAG_GRAPH_STORAGE=NetworkXStorage
```

### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version: upstream `main` @ `ab2e08fb191179f28ae2ea4c207ee2fcd826f9cb`
- Operating System:
- Python Version:
- A PR fixing this (wrapping `create_database()` in a `try/except MilvusException`
  keyed on a case-insensitive "already exist" match against the exception
  message) is incoming and will reference this issue.

AdalbertoDam:
## Description

Fixes a crash-on-first-boot (`MilvusException: database already exist`,
Milvus server error code `65535`) that any multi-process LightRAG deployment
(Kubernetes `replicas > 1`, or any second instance started concurrently
against the same fresh Milvus) can hit when
[`MilvusVectorDBStorage._create_milvus_client`](https://github.com/HKUDS/LightRAG/blob/ab2e08fb191179f28ae2ea4c207ee2fcd826f9cb/lightrag/kg/milvus_impl.py#L603-L630)
races on Milvus's non-atomic `list_databases()`-then-`create_database()`
check.

## Related Issues

Fixes [#3875](https://github.com/HKUDS/LightRAG/issues/3875)

## Changes Made

- `MilvusVectorDBStorage._create_milvus_client` now wraps
  `client.create_database(db_name)` in a `try/except MilvusException`,
  treating a case-insensitive `"already exist"` substring match against the
  exception's message as benign, and re-raising any other `MilvusException`
  unchanged. The match is on the message rather than `MilvusException.code`
  because the numeric code for this condition isn't consistently assigned
  across Milvus server versions, and re-verifying via a second
  `list_databases()` call risks observing stale cluster metadata and
  re-raising even though the create actually succeeded.
- Added four regression tests to `tests/kg/milvus_impl/test_milvus_index_creation.py`:
  - `test_create_database_race_already_exists_is_swallowed` — a
    `create_database` call raising an "already exist" `MilvusException`
    must not fail `_create_milvus_client()`.
  - `test_create_database_race_message_match_is_case_insensitive` — the
    match tolerates the server's varying capitalization of the message.
  - `test_create_database_unrelated_milvus_exception_is_reraised` — any
    *other* `MilvusException` (e.g. a permission error) from
    `create_database` must still propagate, so the fix narrowly targets the
    known race and doesn't swallow genuine errors.
  - `test_create_database_non_milvus_exception_is_not_swallowed` — a
    non-`MilvusException` error (e.g. a connection reset) is never
    inspected for the race message and always propagates unmodified.

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed
- [x] Documentation updated (if necessary) — N/A, no behavior/API change on
      the non-race path

## Additional Notes

**Root cause**: `list_databases()` then `create_database()` is not atomic on
the Milvus server. Any deployment starting more than one LightRAG process
against the same fresh Milvus database name can have two processes both
observe the database as absent and both call `create_database()`; the loser
receives error code `65535` ("database already exist"), which currently
propagates uncaught through `initialize_storages()` and kills that process.
On any later restart the database already exists, so the race window only
appears on a genuinely fresh volume.

**Verification**: reproduced independently using the official 
`ghcr.io/hkuds/lightrag:latest` image against a fresh, official Milvus standalone 
stack (two replicas started at once) — one replica crashed with the exact traceback 
this PR fixes, while the surviving replica (identical image/config) booted cleanly, 
ruling out a configuration issue. Re-verified against the patched code (`uv run
lightrag-server`, two processes, same fresh Milvus): one process hit the
identical `MilvusException` the "before" run crashed on, and both processes
now complete `Application startup complete.` and return `200` from
`/health`. `ruff check` / `ruff format --check` pass on both changed files,
and `tests/kg/milvus_impl/test_milvus_index_creation.py` passes in full (46
tests, including the 4 new ones).

### DIFF
--- a/lightrag/kg/milvus_impl.py
+++ b/lightrag/kg/milvus_impl.py
@@ -615,7 +615,22 @@ class MilvusVectorDBStorage(BaseVectorStorage):
             logger.warning(
                 f"[{self.workspace}] Milvus database '{db_name}' not found, creating it"
             )
-            client.create_database(db_name)
+            try:
+                client.create_database(db_name)
+            except MilvusException as e:
+                # Two or more workers can both see the database missing and race to
+                # create it; the loser must treat "already exists" as
+                # success rather than failing startup. Matching on the
+                # message (not the error code, which isn't consistently
+                # assigned across server versions) avoids a second
+                # list_databases() call, which could itself observe stale
+                # cluster metadata and re-raise despite the create having
+                # actually succeeded.
+                if "already exist" not in str(e).lower():
+                    raise
+                logger.debug(
+                    f"[{self.workspace}] Milvus database '{db_name}' already exists — continuing"
+                )
 
         use_database = getattr(client, "use_database", None) or getattr(
             client, "using_database", None
--- a/tests/kg/milvus_impl/test_milvus_index_creation.py
+++ b/tests/kg/milvus_impl/test_milvus_index_creation.py
@@ -9,9 +9,9 @@ This test suite validates:
 import asyncio
 import pytest
 from unittest.mock import MagicMock, patch
-from pymilvus import MilvusException
 from lightrag.kg.milvus_impl import (
     MILVUS_MAX_VARCHAR_BYTES,
+    MilvusException,
     MilvusIndexConfig,
     MilvusVectorDBStorage,
 )
--- a/tests/kg/milvus_impl/test_milvus_index_creation.py
+++ b/tests/kg/milvus_impl/test_milvus_index_creation.py
@@ -9,6 +9,7 @@ This test suite validates:
 import asyncio
 import pytest
 from unittest.mock import MagicMock, patch
+from pymilvus import MilvusException
 from lightrag.kg.milvus_impl import (
     MILVUS_MAX_VARCHAR_BYTES,
     MilvusIndexConfig,
@@ -66,6 +67,7 @@ def _make_model_storage(namespace="entities", workspace="test_workspace", dim=12
         workspace=workspace,
         global_config={
             "embedding_batch_num": 100,
+            "working_dir": "/tmp/lightrag",
             "vector_db_storage_cls_kwargs": {
                 "cosine_better_than_threshold": 0.3,
             },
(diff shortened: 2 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
Thanks for the careful analysis and the regression tests — the fix itself is correct and we intend to take it. One thing to set straight about its scope, since the PR is framed around multi-replica Kubernetes deployments.

**LightRAG currently supports only ONE active server instance per workspace.** This PR should not be read as making `replicas > 1` a supported configuration; it turns one specific crash into a non-crash, which is a real improvement, but the underlying constraint is unchanged.

The reason is that LightRAG's coordination primitives — `get_data_init_lock()`, the namespace locks, and the whole `pipeline_status` contract — live in per-host shared memory (`lightrag/kg/shared_storage.py`). They serialize the worker processes of a single instance; they do not span pods. So beyond `create_database`, the rest of Milvus startup is still unguarded across instances: `_create_collection_if_not_exist` is another check-then-act whose failure path drops and recreates the collection (filed separately as #3891), and `_migrate_collection_schema_attempt` is worse — its recovery logic is written for "this process crashed and restarted", so a second live instance's in-flight migration gets mistaken for leftover state and can be dropped or promoted out from under the instance that is running it.

**What can run concurrently:** multiple instances serving read-only traffic are fine in principle, once the storages are already initialized and no instance is ingesting. What must not overlap is the *initialization* phase.

**So the practical rule is to start instances in sequence:** bring one up, wait for its `/health` to return 200, then start the next. That check is meaningful because `initialize_storages()` and `check_and_migrate_data()` both run in the FastAPI lifespan (`lightrag/api/lightrag_server.py:1587-1598`) before the app serves any request — a healthy instance has provably finished creating and migrating everything, so every later instance finds the storages already in place and never enters a create-or-migrate path.

In Kubernetes this is exactly what a **StatefulSet with `podManagementPolicy: OrderedReady`** gives you: pods are created one at a time in ordinal order, and each must be Running and Ready before the next is created. Combined with the readiness probe the chart already has on `/health`, ordered startup comes for free. (Note that the current chart ships a `Deployment` with `updateStrategy: Recreate`, so this would be a chart change rather than a values tweak.)

Nothing is required from you for this PR — we're documenting the single-active-instance constraint in the Helm chart and its README separately. Just wanted the framing on the record so the fix isn't mistaken for multi-instance support.

---

PR #3877, opened 2026-09-08
fix(cache): partition KG answers by chunk selection

Shizoqua:
## Summary

Partition KG answer-cache entries by chunk-selection configuration.

## Problem

Changing `related_chunk_number` or `kg_chunk_pick_method` can change the retrieved context while reusing an answer cached under the previous configuration.

## Changes

Include both settings in the KG answer-cache key and recorded cache metadata.

## Testing

47 relevant query-cache tests passed. Ruff, pre-commit and `git diff --check` passed.

## Compatibility and Risk

Low risk. Existing KG cache entries using the previous key will be bypassed.

## Related Issue

No related issue.

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -4804,6 +4804,19 @@ async def kg_query(
 
     # Handle cache
     answer_cache_kv = _answer_cache_kv(query_param, hashing_kv)
+    retrieval_config = (
+        text_chunks_db.global_config if text_chunks_db is not None else global_config
+    )
+    retrieval_cache_args = ()
+    if (
+        "related_chunk_number" in retrieval_config
+        or "kg_chunk_pick_method" in retrieval_config
+    ):
+        retrieval_cache_args = (
+            "\n<kg_chunk_selection>\n",
+            retrieval_config.get("related_chunk_number", DEFAULT_RELATED_CHUNK_NUMBER),
+            retrieval_config.get("kg_chunk_pick_method", DEFAULT_KG_CHUNK_PICK_METHOD),
+        )
     args_hash = compute_args_hash(
         _ANSWER_CACHE_POLICY_VERSION,
         query_param.mode,
@@ -4827,6 +4840,7 @@ async def kg_query(
         effective_user_prompt.text,
         query_param.enable_rerank,
         global_config.get("enable_content_headings", False),
+        *retrieval_cache_args,
         *(("\n<system_prompt>\n", system_prompt) if system_prompt else ()),
         "\n<llm_identity>\n",
         serialize_llm_cache_identity(llm_cache_identity),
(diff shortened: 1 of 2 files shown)


danielaskdd:
Thanks for the fix — the bug is real and the direction is right.

I verified it locally: reverting only the `lightrag/operate.py` hunk makes both new parametrized cases fail with `== LLM cache == Query cache hit`, so the test genuinely pins the defect. 251 related cache tests pass with the patch applied, and CI is green.

I also want to call out something you got right that is easy to miss: reading the two settings from `text_chunks_db.global_config` rather than from the `global_config` parameter is the correct choice. Storages capture an `asdict(self)` snapshot at construction time, while `aquery` builds a fresh config on every call, and the chunk-selection code paths (`_find_most_related_text_unit_from_entities` / `..._from_relationships`) read the storage snapshot. Keying on the same dict the consumer reads is what keeps the key and the behavior in sync.

Three things I'd like changed before this merges.

**1. The presence check is dead code in production, and it splits the key for no benefit**

```python
if (
    "related_chunk_number" in retrieval_config
    or "kg_chunk_pick_method" in retrieval_config
):
```

Both are `LightRAG` dataclass fields, so `asdict()` always contains them and this branch is unconditionally true on every real query. Its only effect is on stub dicts in tests. If the intent was backward compatibility for existing cache entries, it does not achieve that — production entries are invalidated either way, as your PR description already notes.

What it does add is a semantic wrinkle: a config that omits the keys and a config that sets them to the defaults behave identically but hash differently. That is miss-only, not a wrong answer, but it is complexity paying for nothing. Please make the two components unconditional.

**2. The key is conditional but the recorded metadata is not**

The `queryparam_dict` block calls `retrieval_config.get(..., DEFAULT_...)` with no guard. So in the "neither key present" case the stored metadata advertises a partition the hash does not actually have. `queryparam` is diagnostic only (it is written in `save_to_cache` and only ever copied verbatim by `tools/migrate_llm_cache.py`), so nothing breaks — but two branches of the same decision disagreeing is exactly the kind of thing that misleads the next reader. Dropping the guard from (1) resolves this too.

**3. The subtlest line in the patch has no comment**

The `text_chunks_db.global_config` vs `global_config` distinction explained above is not self-evident, and the patch makes the same `compute_args_hash(...)` call mix both sources — the line immediately above yours is `global_config.get("enable_content_headings", False)`. Every other component in that call carries a "why" comment (the policy-version rationale, why `disable_user_prompt_prefix` is deliberately not a separate component). Please add a short comment stating that the value must come from the storage snapshot because that is what the chunk-selection code reads.

Two notes that are **not** requests, just to close them out:

- No `_ANSWER_CACHE_POLICY_VERSION` bump is needed here. Adding key components already makes old entries unreachable, and the semantics of an entry have not moved, which is what that constant's docstring reserves a bump for. Your current handling is correct.
- Separately, `enable_content_headings` is *consumed* from `text_chunks_db.global_config` but keyed from the fresh `global_config` — the same source mismatch this PR avoids for its two new fields. That is pre-existing and out of scope here; I'll track it separately rather than widen this PR.

---

danielaskdd:
@codex review

PR #3881, opened 2026-09-08
fix: clarify worker timeout guidance

### LINKED ISSUE #2415: [Bug]:LLM func: Worker execution timeout after 360s

### Describe the bug

  File "/home/zhangkailin/.conda/envs/rag/lib/python3.12/site-packages/lightrag/operate.py", line 2978, in extract_entities
    raise prefixed_exception from first_exception
TimeoutError: C[1/3]: chunk-e91ad3a5c8b93996cd4d789b270fcaeb: LLM func: Worker execution timeout after 360s

ERROR: Failed to extract document 1/1: unknown_source
INFO: Enqueued document processing pipeline stopped

=====================
Query mode: naive
=====================
WARNING: LLM func: Worker timeout for task 140104065711744_3790223.878330866 after 360s
WARNING: LLM func: Worker timeout for task 140104065707712_3790223.890454997 after 360s
WARNING: Embedding func: Worker timeout for task 140104075381824_3790583.900605295 after 60s
ERROR: Error in _get_vector_context: Embedding func: Worker execution timeout after 60s
INFO: [naive_query] No relevant document chunks found; returning no-result.
Sorry, I'm not able to provide an answer to that question.[no-context]

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### LightRAG Config Used

# Paste your config here

### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

lorenzozanee:
## Summary

Worker timeout errors previously reported only elapsed duration, making the failure difficult to diagnose. The message now explains how to increase the configured LLM timeout or reduce provider work, with a regression test covering the public error.

## Motivation

This gives operators a supported next step when provider work exceeds the configured budget.

Fixes #2415

### DIFF
--- a/lightrag/utils.py
+++ b/lightrag/utils.py
@@ -1151,7 +1151,11 @@ class WorkerTimeoutError(Exception):
     def __init__(self, timeout_value: float, timeout_type: str = "execution"):
         self.timeout_value = timeout_value
         self.timeout_type = timeout_type
-        super().__init__(f"Worker {timeout_type} timeout after {timeout_value}s")
+        super().__init__(
+            f"Worker {timeout_type} timeout after {timeout_value}s; increase the "
+            "configured LLM timeout or reduce provider work if the call is "
+            "expected to take longer"
+        )
 
 
 class HealthCheckTimeoutError(Exception):
--- a/tests/llm/test_llm_role_runtime.py
+++ b/tests/llm/test_llm_role_runtime.py
@@ -119,6 +119,26 @@ ROLE_MAX_ASYNC_ENV_KEYS = (
 
 
 @pytest.mark.asyncio
+async def test_worker_timeout_explains_how_to_adjust_provider_budget():
+    async def slow_func(**_kwargs):
+        await asyncio.sleep(0.05)
+
+    wrapped = priority_limit_async_func_call(
+        1,
+        max_execution_timeout=0.01,
+        queue_name="extract LLM func",
+    )(slow_func)
+
+    with pytest.raises(
+        TimeoutError,
+        match="increase the configured LLM timeout or reduce provider work",
+    ):
+        await wrapped()
+
+    await wrapped.shutdown()
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


danielaskdd:
Thanks for the PR. We won't merge it in its current form. Three points, and the direction we'd like the change to take:

**1. `WorkerTimeoutError` is not LLM-specific.**
`priority_limit_async_func_call` wraps three kinds of queues: the LLM role queues (`lightrag/llm_roles.py`), the `Embedding func` queue and the `Rerank func` queue (`lightrag/lightrag.py`). Each has its own timeout setting (`LLM_TIMEOUT` / `*_LLM_TIMEOUT`, `EMBEDDING_TIMEOUT`, `RERANK_TIMEOUT`). With this change an embedding or rerank timeout would tell the operator to "increase the configured LLM timeout", which points at the wrong setting. #2415 itself contains an `Embedding func: Worker execution timeout after 60s` line.

**2. The exception message should describe the failure, not prescribe a fix.**
A worker-level timeout fires at `llm_timeout * 2`, i.e. after the provider-level timeout should already have stopped the call. Whether the right response is a longer timeout, a smaller output budget, lower concurrency, a provider-side problem or something else depends on the deployment, and the exception has no way to know. Raising the timeout or capping max tokens is not necessarily the correct action, so hard-coding either into the message will be wrong for a share of cases and would have to be revised repeatedly. Please keep the message to what failed and why: which queue, which timeout tier (worker execution), and the configured limit that was exceeded. Remediation guidance belongs in documentation and issue replies, not in the exception string.

**3. `Fixes #2415` should be changed.**
That issue is a live timeout problem, not a message-wording problem, and merging this PR must not auto-close it. Please change it to "Related to #2415" or drop the reference.

If you'd like to continue with this: keep `WorkerTimeoutError` generic. If you find the current message insufficient, the missing piece is identification (which tier and which limit tripped), not advice, and the regression test should then assert on that identification rather than on a recommendation string.

---

PR #3886, opened 2026-09-08
fix(parser): preserve alphabetic numbering after z

Shizoqua:
## Summary

Preserve Word-style alphabetic numbering after `z`.

## Problem

DOCX list numbering wrapped from `z` back to `a` instead of continuing to `aa`.

## Changes

Generate repeated-letter numbering for lower- and upper-case alphabetic lists.

## Testing

DOCX parser tests: 846 passed, 44 skipped. Ruff, pre-commit and `git diff --check` passed.

## Compatibility and Risk

Low risk. Other numbering formats are unchanged.

## Related Issue

No related issue.

### DIFF
--- a/lightrag/parser/docx/numbering_resolver.py
+++ b/lightrag/parser/docx/numbering_resolver.py
@@ -558,10 +558,19 @@ class NumberingResolver:
                 n -= value
         return result
 
+    #: Above this count, _to_letter degrades to the plain decimal string
+    #: instead of a repeated-letter label. Output length is n // 26, and n
+    #: comes from an attacker-controlled w:start (see _parse_numbering_xml,
+    #: parsed with no upper bound) -- without this cap a single crafted
+    #: numbering.xml value can force a multi-gigabyte allocation from a few
+    #: bytes of XML. Same order of magnitude as _to_roman's 4000 cutoff;
+    #: at the cap the output is still only ~385 characters.
+    _MAX_LETTER_COUNT = 10_000
+
     @staticmethod
     def _to_letter(n: int, *, upper: bool = False) -> str:
         """Convert an integer to the repeated-letter numbering used by Word."""
-        if n <= 0:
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
Thanks @Shizoqua — you identified the real bug, and the fix has landed in #3905 (which also covers the recognition half tracked in #3902). Closing this one as superseded.

For the record, the two PRs overlapped on the same two lines of `FORMAT_CONVERTERS`, and one concrete trade-off was decided differently here, so this is not a plain "already covered":

**The cap: 78 rather than 10 000.** Both PRs bound the converter before the string multiplication — that part of your change was right and necessary. #3905 sets the bound to 78 (three letters) and registers `lowerLetter` / `upperLetter` in `LIMITED_DOMAIN_FORMATS`, for two reasons:

- 78 is exactly the domain the heading classifier reads back (`parse_alpha_ordinal` accepts a homogeneous run of at most 3 letters). A 10 000 cap keeps rendering letters for counts 79–10 000, and every one of those labels is a 4-or-more-letter run that `smart_heading` then refuses to classify — the generator would be emitting labels nothing downstream can read.
- Registering the format routes the fallback through the existing `_note_out_of_range` path, so exceeding the domain logs a warning and increments `numbering_out_of_range_formats`. The `_MAX_LETTER_COUNT` approach degraded to decimal silently, which is what the `LIMITED_DOMAIN_FORMATS` docstring in that file explicitly argues against ("never silently: a wrong-looking-yet-plausible label is harder to notice than an outright error").

Everything else your PR asserted is covered and verified against the current #3905 head: `a`/`z`/`aa`/`zz`/`aaa` and `AA` all render as you specified, and out-of-domain starts (including `2147483647`, via both `w:start` and `w:startOverride`) fall back to decimal without allocating.

PR #3887, opened 2026-09-08
fix(query): preserve minimum vector chunk quota

Shizoqua:
## Summary

Preserve vector-based chunk selection when the configured quota is small.

## Problem

A positive `related_chunk_number` could produce a zero vector quota, causing selection to fall back to weight-based ranking.

## Changes

Ensure positive chunk limits allocate at least one vector-selected chunk.

## Testing

Relevant query tests: 28 passed. Ruff, pre-commit and `git diff --check` passed.

## Compatibility and Risk

Low risk. Non-positive limits and larger quotas remain unchanged.

## Related Issue

No related issue.

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -6203,6 +6203,12 @@ async def _find_most_related_edges_from_entities(
     return all_edges_data
 
 
+def _vector_chunk_quota(max_related_chunks: int, group_count: int) -> int:
+    if max_related_chunks <= 0 or group_count <= 0:
+        return 0
+    return max(1, int(max_related_chunks * group_count / 2))
+
+
 async def _find_related_text_unit_from_entities(
     node_datas: list[dict],
     query_param: QueryParam,
@@ -6287,7 +6293,9 @@ async def _find_related_text_unit_from_entities(
     #     The order of text chunks aligns with the naive retrieval's destination.
     #     When reranking is disabled, the text chunks delivered to the LLM tend to favor naive retrieval.
     if kg_chunk_pick_method == "VECTOR" and query and chunks_vdb:
-        num_of_chunks = int(max_related_chunks * len(entities_with_chunks) / 2)
+        num_of_chunks = _vector_chunk_quota(
+            max_related_chunks, len(entities_with_chunks)
+        )
 
         # Get embedding function from global config
         actual_embedding_func = text_chunks_db.embedding_func
@@ -6579,7 +6587,9 @@ async def _find_related_text_unit_from_relations(
     selected_chunk_ids = []  # Initialize to avoid UnboundLocalError
 
     if kg_chunk_pick_method == "VECTOR" and query and chunks_vdb:
-        num_of_chunks = int(max_related_chunks * len(relations_with_chunks) / 2)
+        num_of_chunks = _vector_chunk_quota(
+            max_related_chunks, len(relations_with_chunks)
+        )
 
         # Get embedding function from global config
         actual_embedding_func = text_chunks_db.embedding_func
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -6204,6 +6204,17 @@ async def _find_most_related_edges_from_entities(
 
 
 def _vector_chunk_quota(max_related_chunks: int, group_count: int) -> int:
+    """How many chunks VECTOR-mode selection may draw, scaled by group count.
+
+    Floored at 1 (once max_related_chunks and group_count are both positive)
+    so a query with very few related entities/relations still gets at least
+    one chunk instead of being rounded down to zero by the /2 scaling below.
+
+    The group_count <= 0 branch is defensive only: both call sites already
+    guard on their group list being non-empty (entities_with_chunks /
+    relations_with_chunks) before reaching here, so group_count is always
+    >= 1 in practice.
+    """
     if max_related_chunks <= 0 or group_count <= 0:
         return 0
     return max(1, int(max_related_chunks * group_count / 2))
--- a/tests/test_kg_vector_chunk_quota.py
+++ b/tests/test_kg_vector_chunk_quota.py
@@ -7,10 +7,26 @@ from lightrag.base import QueryParam
 from lightrag.operate import (
     _find_related_text_unit_from_entities,
     _find_related_text_unit_from_relations,
+    _vector_chunk_quota,
 )
 
 
 pytestmark = pytest.mark.offline
+
+
+@pytest.mark.parametrize(
+    ("max_related_chunks", "group_count", "expected"),
+    [
+        (0, 5, 0),  # related_chunk_number disabled entirely
+        (5, 0, 0),  # defensive branch: no call site reaches this in practice
+        (1, 1, 1),  # floor kicks in: 1 * 1 / 2 would round down to 0
+        (2, 1, 1),  # 2 * 1 / 2 == 1, floor is a no-op here
+        (4, 3, 6),  # normal scaling, no floor/rounding involved
+        (3, 1, 1),  # 3 * 1 / 2 == 1.5, int() truncates to 1
+    ],
+)
+def test_vector_chunk_quota(max_related_chunks, group_count, expected):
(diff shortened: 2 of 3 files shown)


tests/test_kg_vector_chunk_quota.py
danielaskdd, line 1: **Please move this file — `tests/llm/` is the wrong directory (blocking).**

`AGENTS.md` requires tests to mirror the module under test, and reserves `tests/llm/` for LLM/embedding provider bindings plus cross-provider concerns (embedding, VLM, cache, role). The module under test here is `lightrag/operate.py`, which has nothing to do with `lightrag/llm/`.

The existing precedent for these functions is:

- `tests/test_operate_vector_context.py` (repo root)
- `tests/utils/test_pick_by_vector_similarity.py`, `tests/utils/test_pick_by_vector_similarity_order.py`

Please move this to the `tests/` root, e.g. `tests/test_kg_vector_chunk_quota.py`, so `tests/llm/` stays scoped to provider code.

Two smaller nits while the file is being touched:

1. `@pytest.mark.asyncio` is redundant — `pyproject.toml` sets `asyncio_mode = "auto"`. The rest of the suite relies on that; dropping the three decorators keeps it consistent.
2. Consider adding a direct unit test for `_vector_chunk_quota` itself (a small `pytest.mark.parametrize` over `(max_related_chunks, group_count)`). It would cover the `group_count <= 0` branch — which the current three async tests cannot reach, see my other comment — and pin the "no other input combination changes" property that the PR description claims, at a fraction of the setup cost of driving the full async path through mocks.


---

lightrag/operate.py
```diff
@@ -6203,6 +6203,12 @@ async def _find_most_related_edges_from_entities(
     return all_edges_data
 
 
+def _vector_chunk_quota(max_related_chunks: int, group_count: int) -> int:
+    if max_related_chunks <= 0 or group_count <= 0:
+        return 0
+    return max(1, int(max_related_chunks * group_count / 2))
```
danielaskdd, line 6220: The clamp value itself is right, but please document *why* it is `1` and not something else.

`AGENTS.md` ("Consistency without transactions") asks that every accepted design choice be written down next to the code. Here the reason is a real invariant worth recording rather than a magic number: the WEIGHT path guarantees at least one chunk per group via `pick_by_weighted_polling(..., min_related_chunks=1)`, and the VECTOR quota `n * max / 2` is the same linear-gradient budget as WEIGHT's `n * (max + 1) / 2` minus the `n / 2` term — so `(max_related_chunks=1, group_count=1)` is the single point where VECTOR drops below WEIGHT's floor. `max(1, ...)` restores parity exactly there, and the surrounding `<= 0` guard is what keeps `related_chunk_number=0` a genuine kill switch rather than a silent 1.

Both neighbouring helpers (`pick_by_vector_similarity`, `pick_by_weighted_polling`) carry full Args/Returns docstrings; a short one here stating that contract would stop a future reader from "simplifying" the clamp away.

Two placement notes:

1. Consider putting this next to the two pickers in `lightrag/utils.py` instead of `operate.py`. It is a pure function about their shared quota contract, and the reason for the floor only makes sense when read against `min_related_chunks=1`, which lives there.
2. The `group_count <= 0` branch is currently unreachable from both call sites — `_find_related_text_unit_from_entities` returns early on `if not entities_with_chunks`, and `_find_related_text_unit_from_relations` on `if not relations_with_chunks`. Keeping it as a defensive guard is fine, but please say so in the docstring so it does not read as a live case, and cover it in the suggested parametrized test.


---

danielaskdd (changes requested):
Thanks for this — I reviewed it by checking the branch out, reverting the fix to confirm the new tests actually go red, and re-running the surrounding suites. **The diagnosis and the fix are both correct**; I am requesting changes only for the test file location, which violates the layout `AGENTS.md` mandates, plus a couple of small documentation/coverage items. No change to the fix logic is needed.

## What I verified

- **The bug is real, and the trigger is exactly one input combination.** Since `related_chunk_number` and the group count are both integers `>= 1`, `int(max * n / 2) == 0` requires `max * n < 2`, which brute-forcing over `max ∈ [0,5] × n ∈ [1,5]` confirms has the single solution `(related_chunk_number=1, group_count=1)`. Every other combination is byte-identical before and after, so the "larger quotas remain unchanged" claim in the description holds.
- **The impact is a silent strategy downgrade, not data loss.** `pick_by_vector_similarity` returns `[]` on `num_of_chunks <= 0` (`lightrag/utils.py:6207`), and the call site reads that `[]` as "vector selection failed" and flips to `WEIGHT`. So a deployment that explicitly configured `KG_CHUNK_PICK_METHOD=VECTOR` silently got weight-based ranking, plus a factually wrong `No entity-related chunks selected by vector similarity` warning in the log. The result was never empty — `pick_by_weighted_polling`'s `n == 1` branch still returns one chunk — it was just the highest-occurrence chunk rather than the most query-similar one.
- **The fix is at the right layer.** Clamping at the call site rather than inside `pick_by_vector_similarity` is the correct call: that function's `num_of_chunks <= 0 → []` contract is a legitimate "disabled" signal that `tests/utils/test_pick_by_vector_similarity.py` depends on, and clamping inside it would make `RELATED_CHUNK_NUMBER=0` impossible to honour.
- **The tests are genuine regression tests.** Reverting `_vector_chunk_quota` back to the inline expression turns `test_single_entity_uses_vector_picker_with_one_chunk` and `test_single_relation_uses_vector_picker_with_one_chunk` red; restoring it makes all three pass. Both the entity and relation paths are covered, and `test_zero_related_chunk_limit_remains_disabled` correctly pins the kill switch.
- **Local run**: `tests/llm/test_kg_vector_chunk_quota.py`, `tests/utils/test_pick_by_vector_similarity.py`, `tests/utils/test_pick_by_vector_similarity_order.py`, `tests/test_weighted_polling_nonpositive_max.py`, `tests/test_operate_vector_context.py` — **31 passed**. `ruff check` and `ruff format --check` clean on both files.

This also sits well alongside the existing `pick_by_weighted_polling` boundary fix (`tests/test_weighted_polling_nonpositive_max.py`), which guards the mirror-image degenerate input.

## Requested changes

1. **Move the test file out of `tests/llm/`** — the one blocking item. Details inline.
2. **Document why the floor is `1`** in `_vector_chunk_quota`, and note that the `group_count <= 0` branch is defensive/unreachable. Details inline.
3. **Optional but recommended**: a parametrized unit test on `_vector_chunk_quota` directly, which covers the unreachable branch and pins the compatibility property much more cheaply than the async mock tests.

## Out of scope — no action wanted here

While tracing the relation path I found a pre-existing issue this PR does not touch, and should not: `relations_with_chunks` is built *before* the dedup against `entity_chunks`, so a relation whose chunks are all removed by dedup stays in the list with an empty `sorted_chunks`. `len(relations_with_chunks)` therefore over-counts groups and over-estimates the vector quota. That errs in the opposite direction from this fix (granting too much rather than zero), and is harmless in direction. I will open a separate issue — please leave it out of this PR.

CI has not reported on `1cc0965` yet (`total_count: 0`); the full suite should run once you push the test move.


---

danielaskdd:
Correction to the "out of scope" note in my review above, and the promised issue link: HKUDS/LightRAG#3898.

I had the two paths backwards. The **relation** path already filters emptied groups after dedup (`lightrag/operate.py:6550-6554`), so its group count is correct; the **entity** path is the one missing that filter. Nothing changes for this PR — the finding is still out of scope and #3898 tracks it — but the earlier comment would have sent you looking in the wrong function.

Worth knowing while you address the review: the two bugs push in opposite directions on the entity path, so the phantom groups can mask the zero quota this PR fixes. With `related_chunk_number=1` and three duplicate entities the inflated count of 3 yields a quota of 1 rather than 0, which is part of why the defect only shows up with a genuinely single entity group. It does not affect your fix or your tests — just don't be surprised if adding duplicate entities to a test case makes the old bug disappear.

The requested changes on this PR are unchanged: move the test file out of `tests/llm/`, and document the reason for the floor of `1`.


---

PR #3888, opened 2026-09-08
fix(nvidia): enable asymmetric embedding support

Shizoqua:
## Problem

nvidia_openai_embed() had no context parameter, so every embedding used the hardcoded input_type="passage", even for search queries.

## Change

Added a context parameter that maps to NVIDIA's query/passage input_type, matching the asymmetric pattern used by voyageai_embed and jina_embed.

## Tests

tests/llm/nvidia_impl: 8 passed

## Related Issue

No related issue.


danielaskdd:
@codex review

Issue #3854, opened 2026-09-06
Storage: NetworkX cross-process fence depends solely on the shared-storage flag, so a failed notification can lose a write


danielaskdd:
Raised by Codex review on #3846 and confirmed against the code. The window predates that PR; what the review made visible is that the fence has no fallback when its only channel fails.

> **Revision note.** The *Summary* / *Scope* / *Why this is not solvable…* sections below are the original diagnosis, unchanged. Everything from *Settled design* onward replaces the earlier open-ended "Suggested direction / Points to settle" with the design as settled in discussion, including measurements that **reversed two of the original suggestions** (see *Two reversals* — `st_ino` is not usable, and the flag is not to be removed) and a **third reversal on sequencing** (see *Phasing and priority* — this does not wait on #3838 R3).

## Summary

`NetworkXStorage`'s cross-process fence rests entirely on the `storage_updated` flag distributed through `lightrag.kg.shared_storage`:

- `_get_graph` (`networkx_impl.py:351`) is the single entry point every read and write funnels through, and the **only** place a process picks up a fresher on-disk snapshot. Its sole test is `if self.storage_updated.value:`.
- `index_done_callback`'s first block — the "another process committed, decline this one" branch — reads the same flag.

The flag is set by `set_all_update_flags`, which walks the registered `Manager().Value` proxies one at a time. Every step of it is an RPC to the shared-storage manager process, so it can fail, and it can fail **partway through**, leaving some workers notified and the rest not.

A worker whose flag was never flipped therefore:

1. does not reload in `_get_graph`, so it mutates a stale in-memory graph;
2. does not decline in `index_done_callback`, because that branch reads the same flag;
3. writes the **whole** GraphML file (networkx has no incremental serialization), silently overwriting the durable update it never saw.

The *Single writer per workspace* invariant does not prevent this. It guarantees at most one writer **at a time**, not that the writer is always the same process — a later request legitimately makes another worker the next writer.

## Scope of the window

Reaching the lost write needs the notification to fail **and** another worker to become the next writer afterwards. Acquiring the right to write (the pipeline reservation, the keyed locks in `utils_graph`) also goes through the same manager, so while it is fully unavailable no other worker can take over. The reachable case is therefore a **transient** manager failure: the publication fails, the manager recovers, another worker writes from a snapshot that never learned about the first write.

[...]

PR #3878, opened 2026-09-08
Fix commit ordering for entity/relation creation to prevent data loss

danielaskdd:
## Description

This PR fixes a critical data loss bug in entity and relation creation by ensuring tracking rows are committed to durable storage before the graph objects they describe. The fix mirrors the existing deletion path ordering and prevents a forbidden state where a graph object is durable without its attribution tracking row.

## Related Issues

Issue #3838 - Commit ordering and residue handling for graph mutations

### Core Fix: `lightrag/utils_graph.py`
- **Reordered `_persist_graph_updates` to commit in two phases:**
  1. **Phase 1 (tracking rows first):** `entity_chunks_storage` and `relation_chunks_storage`
  2. **Phase 2 (objects second):** `entities_vdb`, `relationships_vdb`, and `chunk_entity_relation_graph`
- Within each phase, stores still flush in parallel; only the boundary between phases is ordered
- Phase 1 failure now skips Phase 2 entirely, preventing object publication without tracking attribution
- Added comprehensive documentation explaining the invariant: "a graph object must never be durable while the tracking row that carries its attribution is not"
- Added concurrency warnings to `acreate_entity`, `acreate_relation`, `aedit_entity`, `aedit_relation`, `adelete_by_entity`, `adelete_by_relation`, and `amerge_entities` docstrings

### Documentation: `docs/ProgramingWithCore.md`
- Documented the one-directional commit ordering rule for all admin paths (create, edit, delete, merge)
- Explained the accepted residue: a tracking row whose object never became durable (harmless, non-inheritable, repairable)
- Clarified that admin-vs-pipeline writes are guarded by `check_pipeline_busy_or_raise`, but admin-vs-admin writes are not serialized
- Documented why a workspace-wide admin lock was rejected: it wouldn't prevent the residue reachable with no concurrency
- Added guidance for direct API users: don't call admin functions concurrently on file-backed workspaces

### Storage Backend Documentation
- **`lightrag/kg/networkx_impl.py`:** Documented commit granularity (whole-namespace publishes), scope decision (testing/validation only), and the accepted residue for concurrent admin writes
- **`lightrag/kg/json_kv_impl.py`:** Documented whole-file rewrite semantics and co-tenant flush implications
- **`lightrag/kg/nano_vector_db_impl.py`:** Clarified that admin flows do exercise these paths and documented the timing implications
- **`lightrag/kg/faiss_impl.py`:** Similar clarifications about admin flow exposure and flush semantics

[...]

PR #3894, opened 2026-09-09
fix(graph): surface a declined graph commit through _persist_graph_updates

danielaskdd:
## Description

Rule 5 of #3854 requires a **DECLINED** commit to reach its caller as a failure: declining DISCARDS the in-memory mutation, so acknowledging it reports work that no longer exists.

Two of the three commit paths did that. The third did not:

| Commit path | Checks `False`? | Covers |
|---|---|---|
| `LightRAG._flush_storages` → `_flush_one` | ✅ raises `IndexFlushError` | pipeline writes |
| `utils_graph._commit_graph_or_raise` | ✅ raises `RuntimeError` | `adelete_by_entity`, `_edit_entity_impl`, `_merge_entities_impl` |
| `utils_graph._persist_graph_updates` | ❌ **return value discarded** | `aedit_relation`, `acreate_entity`, `acreate_relation` |

This is the quietest form of the defect the fence exists to prevent. `NetworkXStorage.index_done_callback` declines when the file on disk has moved past the snapshot it holds; the decline branch calls `_reload_locked`, which **destroys the pending mutation**, and returns `False`. The helper swallowed it, the caller logged `Entity Create: 'X' successfully created` and returned **200** — while the same operation's `entities_vdb` / `relationships_vdb` / chunk-tracking upserts had already been flushed by the same `asyncio.gather`.

The observable outcome: a vector row and a tracking row with **no graph object behind them**, reported as a success.

It needs no peer to reach, either. The failed-save recovery arm makes `index_done_callback` return `False` in single-process mode too — through `storage_updated` on `main`, through `_recovery_reload_pending` once #3874 lands.

## Related Issues

Third call site of rule 5 in #3854. Split out of #3874, whose class docstring previously asserted "Both callers do that" and now names this helper as the known gap; that PR corrects the text, this one closes the gap.

## Changes Made

**`lightrag/utils_graph.py`**

- `_flush` returns the store's namespace when `index_done_callback` answered an explicit `False`, and `None` otherwise.
- `_persist_graph_updates` collects those and raises `RuntimeError` naming the declining store(s), matching `_commit_graph_or_raise`'s message and semantics.
- **Only an explicit `False` counts.** `StorageNameSpace.index_done_callback` is declared `-> None`, so every backend that simply returns nothing — all of them except NetworkX / Nano / FAISS — is unaffected. `True` (Nano and FAISS on success) is not a decline either.
- `CommitBookkeepingError` handling is untouched: it is the *opposite* answer, and folding it in would report a durable write as one that did not happen.
- Docstring records the reasoning and the accepted residue below.

**`tests/pipeline/test_persist_graph_updates_decline.py`** — 8 cases.

## What could break, and why it does not

- **Other graph backends.** Neo4j, Memgraph, PostgreSQL and friends return `None`; the test is `is False`, so nothing changes for them. Pinned by a test.
- **`CommitBookkeepingError` is still swallowed.** Pinned by a test, including the case where one store raises it while another declines — the decline still wins.
- **Admin endpoints change from 200 to 500 in the racing case.** That is the point, and it is the same direction #3854 already accepted for the paths using `_commit_graph_or_raise`: "the fence makes a stale writer decline more often … where it previously lost the write silently. That is the correct direction."

## Accepted residue

Per *Consistency without transactions* in `AGENTS.md`, written down rather than left implicit, and pinned by `test_every_sibling_still_flushes_before_the_raise`:

**All flushes are awaited to completion before the decline is raised**, so a declined graph commit leaves that operation's vector and chunk-tracking rows durable with no graph object behind them.

Raising early to cancel the siblings would be worse, not better: `commit_in_storage_io` defers cancellation, so a cancelled flush may have written anyway, and *which* ones did would depend on scheduling. Awaiting them all makes the outcome deterministic.

The residue heals and is harmless in direction: the rows are keyed by deterministic ids (entity name / relation pair hashes), so the caller's retry rewrites them and a rebuild rewrites them; in the meantime a query may surface a row whose graph object is missing. Reporting a discarded mutation as a success is the direction that never heals.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (if necessary) — `_persist_graph_updates` docstring; no file under `docs/` describes this helper

## Additional Notes

**Tests.** 8 cases in `tests/pipeline/test_persist_graph_updates_decline.py`:

*The fix* — a declined graph commit raises; the error names the store that declined; a decline alongside another store's `CommitBookkeepingError` still raises.

*The no-regression half* — a backend returning `None` still succeeds; `True` is not a decline; a publication failure is still swallowed; no storages is still a no-op.

*The residue* — every sibling still flushed before the raise.

Verified against a mutation: with the `committed is False` branch removed, 4 of the 8 go red and the other 4 stay green (they pin the half that must not change). Restored afterwards.

**Test runs:** `tests/pipeline`, `tests/api` — **96 failed, 1916 passed, 14 skipped**. An unmodified checkout of `main` gives **96 failed, 1908 passed** on the same selection; the delta is exactly the 8 new tests. Every one of the 96 is `ModuleNotFoundError: No module named 'ollama'`, an optional binding that is not installable in this sandbox. CI covers those.

https://claude.ai/code/session_01HfwqnMzXtJmpoP5yaBQq7P

---
_Generated by [Claude Code](https://claude.ai/code/session_01HfwqnMzXtJmpoP5yaBQq7P)_

### DIFF
--- a/lightrag/utils_graph.py
+++ b/lightrag/utils_graph.py
@@ -289,6 +289,29 @@ async def _persist_graph_updates(
     tracking rows the retry would then believe it still owed. Every other
     exception still propagates: those mean the flush did not happen.
 
+    An explicit ``False`` is the third answer, and it is the opposite of a
+    ``CommitBookkeepingError``: the commit was **DECLINED**, so the in-memory
+    mutation was discarded rather than persisted (see
+    ``NetworkXStorage.index_done_callback``'s first block, and rule 5 of
+    #3854). This helper used to drop the return value, which made a decline
+    reaching ``aedit_relation`` / ``acreate_entity`` / ``acreate_relation``
+    silent: the graph mutation gone, the caller told it succeeded. It now
+    raises, matching :func:`_commit_graph_or_raise`, which every other admin
+    graph path already routes its commit through.
+
+    **Accepted residue — the sibling flushes still land.** All flushes are
+    awaited to completion before the decline is raised, so a declined graph
+    commit leaves this operation's vector and chunk-tracking rows durable with
+    no graph object behind them. Raising early to cancel the siblings would be
+    worse, not better: ``commit_in_storage_io`` defers cancellation, so a
+    cancelled flush may have written anyway, and which ones did would depend
+    on scheduling. Per *Consistency without transactions* in ``AGENTS.md``
+    this is the acceptable direction — the rows are keyed by deterministic ids
+    (entity name / relation pair hashes), so the caller's retry rewrites them,
+    a rebuild rewrites them, and in the meantime a query may surface a row
+    whose graph object is missing. Reporting a discarded mutation as a success
+    is the direction that is never acceptable.
+
     Args:
         entities_vdb: Entity vector database storage (optional)
         relationships_vdb: Relationship vector database storage (optional)
@@ -310,9 +333,10 @@ async def _persist_graph_updates(
     if relation_chunks_storage is not None:
         storages.append(relation_chunks_storage)
 
-    async def _flush(storage_inst) -> None:
+    async def _flush(storage_inst) -> str | None:
+        """Flush one store. Returns its namespace if it DECLINED, else None."""
         try:
-            await cast(StorageNameSpace, storage_inst).index_done_callback()
+            committed = await cast(StorageNameSpace, storage_inst).index_done_callback()
         except CommitBookkeepingError as e:
             log_without_raising(
                 logger.error,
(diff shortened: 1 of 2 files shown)


danielaskdd:
Closing as obsolete — `dac05a0` ("close the three ordering holes left by the create/edit rework", follow-up to #3878) landed the same fix on `main` while this was being written, and did it better.

Same diagnosis, hole 1 of three: `_persist_graph_updates` dropped `index_done_callback`'s return value, so a declined graph commit through `acreate_entity` / `acreate_relation` / `aedit_relation` was reported as a success.

Where `main`'s version is better than this one:

- **Only the graph store is checked** (`require_commit=True` on that entry alone). This PR raised on any store answering `False`, which is wrong in principle — the vector stores carry no such fence, and their staleness is the rebuildable window accepted elsewhere in the module.
- **Two-phase flush** (tracking rows, then the objects they describe) rather than one `asyncio.gather`. That addresses the residue this PR merely documented and accepted: the sibling writes landing anyway.
- **`return_exceptions=True` plus a decline that outranks a sibling failure**, so a stale vector store cannot mask the one answer that says the caller's mutation was thrown away. This PR did not consider that ordering.
- **Test coverage is a superset**: `test_entity_creation_raises_when_the_graph_declines`, `test_relation_creation_raises_when_the_graph_declines`, `test_relation_edit_raises_when_the_graph_declines`, and `test_a_declined_commit_outranks_a_failing_vector_flush` in `tests/pipeline/test_graph_deletion_tracking_ordering.py`.

Nothing here is worth porting. The related docstring note on #3874, which described this call site as an open defect, has been corrected to match — see `a609b14` on that branch.

---

PR #3897, opened 2026-09-09
Settle aedit_relation's tracking shrink inside a cancellation-deferring region

### LINKED ISSUE #3895: Graph: aedit_relation's tracking shrink can be skipped silently by a cancellation or a co-flush failure
## Summary

`aedit_relation`'s grow-then-shrink staging (`lightrag/utils_graph.py`) keeps the right *state*, but its final step can be **skipped entirely with nothing reported**. The shrink sits outside any cancellation-deferring region, and after the `_persist_graph_updates` call that flushes the graph **and** `relationships_vdb` together. So two things exit the edit before it runs:

1. a **cancellation** delivered during that step-5 commit — `commit_in_storage_io` defers it through the write and re-raises at the end, and `CancelledError` is a `BaseException`, so the `except Exception` around the shrink never sees it;
2. a **failure of the vector flush** inside that same call — it propagates before the shrink is reached.

Either way the edge is durably narrowed while the row still holds the staged superset, with no log line and no error naming the row.

## Why this matters even though the direction is benign

The residue is `rows ⊃ graph` — under-deletion, the direction this codebase accepts, and no data is lost: a purge reading the wider row is more conservative, not less.

The problem is the **reporting**, because this is the one residue in the function that **no retry heals**: once the edge carries the new `source_id`, the next edit's tracking block sees it unchanged and skips the staging entirely, so `lightrag-repair-chunk-tracking` is the only thing that can reconcile the row — and a silent success guarantees the operator never learns to run it. That is what [AGENTS.md](https://github.com/HKUDS/LightRAG/blob/main/AGENTS.md#consistency-without-transactions) separates out:

> This licenses nothing for **silent failure**. ... a failure must never be swallowed: fail loud, then let the documented residue heal.

Note the shrink already raises `VectorStorageConsistencyError` when it *runs and fails* (#3889 and its [review thread](https://github.com/HKUDS/LightRAG/pull/3889#discussion_r3964302442)). This issue is only about the paths where it does not run at all.

## Suggested fix

The same two moves that closed both routes for the entity path in #3892:

- run the shrink **inside** a cancellation-deferring region alongside the graph commit, so a deferred cancel cannot skip it (`_finish_deferring_cancellation` then logs a shrink failure when a cancel is pending, which its docstring names as the one place such a failure can still be seen);
- order it **before** the vector flush, so a vector failure cannot strand it. The reverse exposure is the acceptable one — a shrink failure skipping the vector flush leaves records a re-issued edit rewrites and `lightrag-rebuild-vdb` restores.

`aedit_relation` has no such region today, so one has to be introduced — which is why this is its own change rather than part of #3892. Note that its step-5 call flushes the graph and the vector store **together**, and #3889 deliberately ordered the declined-commit check *inside* phase 2 of `_persist_graph_updates` rather than around it; splitting that call needs to preserve that, so the declined-commit error still outranks a vector flush failure.

## Acceptance

- A cancel delivered past the graph commit still completes the shrink (mirroring `test_a_cancel_past_the_commit_still_completes_the_shrink`).
- A failing `relationships_vdb` flush cannot skip the shrink.
- A failed shrink still raises `VectorStorageConsistencyError` naming the row key and the repair tool.
- A declined graph commit still outranks a vector flush failure (the #3889 phase-2 ranking).
- The open-residue paragraph on the `aedit_relation` bullet in `docs/ProgramingWithCore.md` → *Concurrent admin writes* is retired, and the entity bullet's "goes further than the relation one's" wording updated, once parity holds.

Found while reviewing #3892; the residue is documented in that PR rather than left unrecorded, per AGENTS.md ("an undocumented residue is a defect; a documented one is a decision").

danielaskdd:
## Description

`aedit_relation`'s grow-then-shrink chunk-tracking staging kept the right *state*, but its final step could be **skipped entirely with nothing reported**. The shrink sat outside any cancellation-deferring region and after the `_persist_graph_updates` call that flushed the graph **and** `relationships_vdb` together, so two routes exited the edit before it ran:

1. a **cancellation** delivered during that commit — `commit_in_storage_io` defers it through the write and re-raises at the end, and `CancelledError` is a `BaseException`, so the `except Exception` around the shrink never saw it;
2. a **failure of the vector half** of that same call, which propagated before the shrink was reached.

Either way the edge was durably narrowed while the row still held the staged superset. The residue direction is the accepted one (`rows ⊃ graph`, under-deletion, no data lost), but it is the one residue in this function **no retry heals** — the next edit reads an unchanged `source_id` and skips the tracking block — so only `lightrag-repair-chunk-tracking` can reconcile it, and a silent success guarantees the operator never learns to run it. That is the *silent failure* AGENTS.md separates out from an accepted residue.

This applies the two moves #3892 made on the entity path, bringing the relation path to parity.

## Related Issues

Closes #3895. Follow-up to #3892 (entity path) and #3889 (the shrink's own failure reporting and the declined-commit ranking).

## Changes Made

`lightrag/utils_graph.py` — `aedit_relation`:

- The graph commit and the shrink now run together inside one `_finish_deferring_cancellation` region, so a deferred cancel cannot walk past the shrink. A shrink failure under a pending cancel is logged where `_finish_deferring_cancellation` documents it.
- The graph is committed on its own with `_commit_graph_or_raise` instead of the combined flush. This preserves the #3889 ranking without needing phase 2: a declined commit raises before any vector work is attempted, so it still outranks a vector failure, and the message wording is unchanged (both sites raise `_declined_commit_error`).
- The vector delete/upsert **and** its flush moved after that region, so neither can strand the shrink — on an immediate-write backend `upsert_edge` is already durable when they run, which made them the same hazard one call earlier. A failure there now raises `VectorStorageConsistencyError` (previously the raw backend error), matching the entity path; the shrink's own error additionally names `lightrag-rebuild-vdb`, since the vector records are no longer written when it fires.
- The one residue that cannot be closed — a teardown inside `upsert_edge`'s own await, where deferring has nothing to defer and settling blind would risk the over-deleting mirror — is now diagnosed with the row key and the repair tool, mirroring `_edit_entity_impl`.

`docs/ProgramingWithCore.md` → *Concurrent admin writes*: the open-residue paragraph on the `aedit_relation` bullet is retired and replaced with the deliberate residue that remains; the entity bullet's "goes further than the relation one's" wording is updated now that parity holds.

Tests:

- `tests/pipeline/test_graph_deletion_tracking_ordering.py` gains the relation counterparts of the entity cases: cancel past the commit still completes the shrink, a failing vector flush does not strand it, a failing vector write cannot skip it, a failed shrink skips the vector work and says so (naming both tools), a declined commit outranks a failing vector flush, and a failed edge write leaves a diagnosed wide row (cancelled + ordinary exception). Five of the six go red on the previous ordering; the declined-commit one is a stability pin for the split.
- `tests/extraction/test_vdb_content_truncation_coverage.py`: `test_edit_relation_vdb_delete_failure_still_leaves_graph_updated` now expects `VectorStorageConsistencyError` (still asserting the graph write landed).

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Additional Notes

Local runs (subset mirroring the changed modules):

- `tests/pipeline` — all green, including the 75 in `test_graph_deletion_tracking_ordering.py`.
- `tests/extraction` — green except 7 pre-existing failures in `test_entity_extraction_stability.py` (prompt/addon-params env), reproduced on a clean checkout of `main`; `test_keyword_extraction_drivers.py` cannot be collected in this sandbox (it tries to pip-install `ollama`).
- `tests/api` — 96 pre-existing failures (login/CORS/path-prefix/ollama config), likewise reproduced unchanged on `main` here.

`ruff check` and `ruff format --check` clean on the touched files.

https://claude.ai/code/session_013AC7HAHAN8N6E2E6aq1vzN

---
_Generated by [Claude Code](https://claude.ai/code/session_013AC7HAHAN8N6E2E6aq1vzN)_

### DIFF
--- a/docs/ProgramingWithCore.md
+++ b/docs/ProgramingWithCore.md
@@ -1667,25 +1667,32 @@ What that can leave behind, and why it is tolerated:
   because a retry cannot heal it — the second edit sees an unchanged `source_id`
   and skips the tracking update — so the operator is the only recovery path, and
   a 200 would guarantee they never learn to take it.
-  **Residue, not yet closed:** that last step can still be *skipped* rather than
-  fail, and then nothing is reported. It sits outside any
-  cancellation-deferring region and after the `_persist_graph_updates` call that
-  flushes the graph and the relation vector store together, so a cancellation
-  delivered during that commit, or a failure of the vector flush inside it,
-  exits the edit with the row still holding the superset and no line naming it.
-  The state stays the accepted direction, but the operator is not told to run
-  the repair — which is the part that cannot heal itself. Tracked in issue
-  #3895; `aedit_entity`'s non-rename path (below) does not share it.
+  Both ways that last step could be *skipped* rather than fail are closed
+  (issue #3895): it runs *inside* a cancellation-deferring region alongside the
+  graph commit, so a cancellation deferred through that commit cannot walk past
+  it, and *before* the relation's vector work, so no vector failure can strand
+  it. The reverse exposure is the acceptable one — a shrink failure skips the
+  vector write, leaving records a re-issued edit rewrites and
+  `lightrag-rebuild-vdb` restores, and the message names both. A declined graph
+  commit still outranks a vector failure, because the commit is confirmed on its
+  own before any vector call is made.
+  **One residue stays open, deliberately, the same one the entity path carries
+  below:** a cancellation — or an ordinary error, an acknowledgement lost after
+  the fact carries the same ambiguity — delivered inside `upsert_edge`'s own
+  await tears the edit down before the shrink, so the edge may be narrowed while
+  the row keeps the superset. It cannot be deferred (the exception originates in
+  that coroutine) and must not be settled blind (narrowing a row whose write the
+  backend never accepted is the over-deleting mirror), so the row stays wide and
+  the failure is *logged* with the row key and the repair tool.
 - `aedit_entity`'s non-rename path stages its row the same way, for the same
   reason: a **growing** edit there used to commit the node before flushing the
   row that attributes it, leaving `rows ⊂ graph` — reachable from `POST
   /graph/entity/edit`, whose `updated_data` accepts `source_id`. The superset
   row is now durable before `upsert_node` is called at all, and the removals are
-  applied after the graph commit. Its shrink goes further than the relation
+  applied after the graph commit. Its shrink is staged exactly like the relation
   one's: it runs *inside* the cancellation-deferring region and *before* the
-  vector flush, so neither a cancellation nor a vector failure can skip it —
-  the two ways the relation path still can. It either completes or raises with
-  the row key and the repair tool named.
+  vector write, so neither a cancellation nor a vector failure can skip it. It
+  either completes or raises with the row key and the repair tool named.
   **One residue there stays open, deliberately:** a cancellation delivered
   *inside* `upsert_node`'s own await — after an immediate-write backend accepted
   the row, before control returns — tears the edit down before the shrink, so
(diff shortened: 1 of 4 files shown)


danielaskdd:
@codex review

PR #3900, opened 2026-09-09
test(pipeline): pin duplicate-document track_id retrievability (#2468)

### LINKED ISSUE #2468: [Bug]: Track id is generated for duplicate documents

### Describe the bug

document/text API return track id for duplicated documents, using same track id if i call API .documents/status_track/{track_id} then got nothing, status can not be tracked on this id.

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### LightRAG Config Used

# Paste your config here

### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version:
- Operating System:
- Python Version:
- Related Issues:

TiSsaa:
## Summary

Issue #2468 reported that uploading a duplicate document returns a
`track_id`, but querying `/documents/track_status/{track_id}` for that
exact `track_id` came back empty instead of showing the FAILED
duplicate record.

Reading the current pipeline, `apipeline_enqueue_documents` already
stamps every duplicate record it creates with the **current** call's
`track_id` (the "Handle duplicate documents" block in
`lightrag/pipeline.py`), so the underlying bug already appears to be
fixed — but there was no test pinning this behavior, and the issue was
still open with no way to confirm it from the tracker.

## What's changed

Adds `tests/pipeline/test_duplicate_track_id_retrieval.py` with
end-to-end regression coverage through both duplicate-detection paths:

- an exact file-path repeat (`duplicate_kind == "filename"`)
- a different file_path with identical content (`duplicate_kind ==
  "content_hash"`)

Both assert the duplicate's record is retrievable via
`aget_docs_by_track_id` under its **own** `track_id`, and that the
original document keeps answering separately under its own `track_id`.

No production code changes.

## Verification

- `uv run python -m pytest tests/pipeline/test_duplicate_track_id_retrieval.py -v` — 2 passed
- `uv run python -m pytest tests/pipeline -q` — 819 passed, no regressions
- `ruff check` / `ruff format --check` — clean
- Fix-proof check: temporarily reverted the `track_id` assignment in
  `lightrag/pipeline.py` back to the pre-fix behavior and confirmed
  both new tests fail with exactly the reported symptom (empty
  result), then restored the original code and confirmed green again.

## Test plan

- [x] Changes tested locally
- [x] Pre-commit-equivalent checks pass (ruff check/format; pre-commit itself couldn't reach its hook repos from this sandbox, so ruff was run directly)
- [x] Test added (no applicable production code change)
- [x] N/A documentation update (test-only change)

Closes #2468

https://claude.ai/code/session_01KR4xANLugcEAXszMDUPjP8

### DIFF
--- a/tests/pipeline/test_duplicate_track_id_retrieval.py
+++ b/tests/pipeline/test_duplicate_track_id_retrieval.py
@@ -1,0 +1,158 @@
+"""A duplicate-document record must be retrievable by the CURRENT track_id
+used for the duplicate upload attempt, not only by the original document's
+track_id.
+
+Regression test for https://github.com/HKUDS/LightRAG/issues/2468: a client
+uploading a duplicate got back a track_id from the enqueue call, but querying
+``/documents/track_status/{track_id}`` (backed by ``aget_docs_by_track_id``)
+for that exact track_id returned nothing — the duplicate's FAILED record
+existed in ``doc_status``, but not under any track_id the client could ever
+query for, since a client only ever learns the track_id handed back from ITS
+OWN request.
+
+``apipeline_enqueue_documents`` stamps every duplicate record it creates with
+the CURRENT call's ``track_id`` (see the "Handle duplicate documents" block in
+``lightrag/pipeline.py``), so this pins the behavior end-to-end through both
+duplicate-detection paths: an exact file-path repeat (``duplicate_kind ==
+"filename"``) and a different file_path with identical content
+(``duplicate_kind == "content_hash"``).
+"""
+
+from __future__ import annotations
+
+import asyncio
+from uuid import uuid4
+
+import numpy as np
+import pytest
+
+from lightrag import LightRAG
+from lightrag.base import DocStatus
+from lightrag.kg.shared_storage import finalize_share_data, initialize_share_data
+from lightrag.utils import EmbeddingFunc, Tokenizer
+
+pytestmark = pytest.mark.offline
+
+
+class _SimpleTokenizerImpl:
+    def encode(self, content: str) -> list[int]:
+        return [ord(ch) for ch in content]
+
+    def decode(self, tokens: list[int]) -> str:
+        return "".join(chr(t) for t in tokens)
+
+
+async def _dummy_embedding(texts: list[str]) -> np.ndarray:
+    return np.ones((len(texts), 8), dtype=float)
+
+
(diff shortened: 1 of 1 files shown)


danielaskdd:
Reviewed and pushed a maintainer edit on top of your commit (`0514378`) — thanks for backfilling this, the coverage gap was real: `aget_docs_by_track_id` is exactly what `/documents/track_status/{track_id}` reads, `test_content_hash_dedup.py` never touches `track_id`, and `test_enqueue_basename_strict.py` asserts the duplicate record's shape without ever asserting its `track_id`. I reproduced your fix-proof locally and it fails as described.

Three things the follow-up commit changes:

1. **Pinned the other half of the contract.** The fix is a trade, not just an addition: the duplicate record's top-level `track_id` now answers the duplicate upload, which *displaces* the original's `track_id` into `metadata.original_track_id`. That key was asserted nowhere in the suite, so dropping it would trade one unanswerable lookup for another — the client reaches its own record but can no longer name the document it collided with — with every test still green. Both metadata keys are now asserted, with `original_doc_id` tied to the real doc id resolved from the original's own `track_id` rather than a literal (it had no coverage at all on the `content_hash` path).

2. **Both tests share one assertion helper.** This removes the asymmetry between them (only the filename test checked the original's status) and gives the original-side lookup a failure message, since under the pre-fix behavior the duplicate piles onto the original's `track_id` and that is the assertion which trips first.

3. **Dropped `chunking_func` / `_chunking` from the harness.** Enqueue never chunks — chunking happens in `apipeline_process_enqueue_documents`, which these tests don't call — so it was dead setup that implied chunking takes part in duplicate detection. Verified both tests still pass without it.

Fix-proof re-run on both halves, each failing behaviorally with the code restored afterwards: reverting the record's `track_id` to the original's trips the original-side lookup (`2 == 1`); deleting the `original_track_id` metadata line trips the new assertion on the missing key. `pre-commit` ruff/ruff-format clean.

One thing deliberately left out of scope: `/documents/track_status` itself still has no test anywhere in the suite. The route is a thin wrapper over `aget_docs_by_track_id`, so your core-level test is a faithful proxy for the reported symptom, but the route's `metadata=doc_status.metadata` passthrough — the field a client actually reads to learn "this was a duplicate" — is untested. That belongs in `tests/api/routes/`, as a separate change.

PR #3904, opened 2026-09-09
fix(query): drop empty entity chunk groups

### LINKED ISSUE #3898: [Bug]: entity chunk selection keeps emptied groups after dedup, inflating the related_chunk_number budget (relation path filters them, entity path does not)
## Summary

`_find_related_text_unit_from_entities` and `_find_related_text_unit_from_relations` (`lightrag/operate.py`) both deduplicate each group's chunk list in Step 2, and dedup can empty a group completely. The relation path then **drops** the emptied groups; the entity path **keeps** them.

Because both selection strategies size their budget from the *group count*, those zero-chunk entities still consume budget. The result is that `related_chunk_number` does not mean the same thing on the two paths, and on the entity path it silently over-delivers.

## Root cause

The relation path filters, at `lightrag/operate.py:6550-6554`:

```python
# Check if any relations still have chunks after deduplication
relations_with_chunks = [
    relation_info
    for relation_info in relations_with_chunks
    if relation_info["chunks"]
]
```

The entity path has no equivalent. After its Step 2 dedup (`lightrag/operate.py:6269`) it goes straight to Step 3 sorting, so an entity whose every chunk was already claimed by an earlier-positioned entity stays in `entities_with_chunks` with `sorted_chunks == []`.

The inflated `len(entities_with_chunks)` then feeds both strategies:

- **VECTOR** — `num_of_chunks` is computed from the group count (`lightrag/operate.py:6290`), so the quota is scaled by phantom groups.
- **WEIGHT** — `pick_by_weighted_polling` builds `expected_counts` as a linear gradient across `n` positions summing to roughly `n * (max_related_chunks + 1) / 2`. Phantom groups raise `n`, and their unusable quota is not discarded: the second-round redistribution loop (`lightrag/utils.py:6154-6172`) hands it to the groups that *do* have chunks. Extra `n` also pushes the function off its `n == 1` fast path (`lightrag/utils.py:6120`) onto the gradient path.

## Reproduction

Five entities all carrying the same 20 chunk IDs — so exactly one real group survives dedup — against five relations of the same shape, with `related_chunk_number=2`:

```
entities  WEIGHT n=2, 5 dup entities  (1 real group): 8 chunks
relations WEIGHT n=2, 5 dup relations (1 real group): 2 chunks

entities  VECTOR num_of_chunks handed to picker: 5
relations VECTOR num_of_chunks handed to picker: 1
```

Identical logical input, 4-5x the budget on the entity path. Both paths log `from 20 ... chunks`, so the pool is the same; only the divisor differs. The relation-path numbers are the intended ones.

The overshoot is bounded by the real chunk pool, so it is invisible whenever the surviving group has fewer chunks than the inflated quota — which is why this has gone unnoticed. It bites when a well-connected entity carries many chunks and its neighbours are duplicates of it, precisely the shape a dense KG produces.

## Impact

Extra KG chunks are pushed into the LLM context beyond what `related_chunk_number` configures, inflating prompt size and cost. Nothing is lost or corrupted, and the direction is over-fetch rather than under-fetch, so this is low severity — but the knob does not honour its documented meaning, and the two paths disagree, which makes the entity/relation budget split unpredictable when tuning retrieval.

## Suggested fix

Mirror the relation path: filter `entities_with_chunks` after Step 2 dedup and before Step 3 sorting, returning `[]` when nothing survives. That makes both paths compute their budget from real groups and removes the divergence, rather than special-casing the empty groups inside the two pickers.

Worth adding a regression test asserting the two paths agree on group count for equivalent input — the asymmetry is what makes this easy to reintroduce.

## Relationship to #3887

Independent bugs that happen to touch the same expression, and they push in opposite directions on the entity path — the phantom groups inflate the divisor and can therefore *mask* the zero-quota bug #3887 fixes (with `related_chunk_number=1` and three duplicate entities, the count of 3 yields a quota of 1 instead of 0). #3887 should land as-is; this is deliberately out of its scope.

## Correction to my earlier note

I raised this in my review on #3887 and attributed it to the **relation** path. That was backwards: the relation path is the one that already filters correctly, and the entity path is the defective one. Corrected here; the underlying finding stands.

Solaris-star:
## Summary

Drop entity groups emptied by chunk deduplication before calculating the per-group retrieval budget.

## Motivation

The entity retrieval path retained empty groups after deduplication, unlike the relation path. Those phantom groups inflated both WEIGHT and VECTOR chunk budgets and could over-deliver context beyond `related_chunk_number`.

On five duplicate entities carrying the same 20 chunks with `related_chunk_number=2`, the entity path delivered 8 chunks where the relation path delivered 2, and handed the vector picker a budget of 5 instead of 1 — identical logical input, 4-5x the budget, purely because the divisor counted groups that no longer carried anything.

Fixes #3898.

## What changed

- filter entity groups that have no chunks after deduplication, before the group count is read by either selection strategy
- return early when no entity chunks survive, and mark that guard as defensive only. It is unreachable on the entity path: deduplication drops a chunk solely because an earlier-positioned entity already claimed it, so the first group keeps everything it carries and always survives. The relation path's equivalent guard *is* reachable, because it additionally excludes the chunks the entity path already delivered.
- add offline regression tests over five duplicate entity groups collapsing to one real group, covering the arguments handed to the weighted picker, the chunk ids actually delivered under WEIGHT (2, not 8), the `num_of_chunks` handed to the vector picker (1, not 5), and a pair of tests asserting the entity and relation paths agree on both budgets for equivalent input.
- pin the interaction with #3887 at `related_chunk_number=1`, the only input where the two fixes can disagree. Collapsing the duplicates to one group makes the ungated `int(1 * 1 / 2)` truncate to 0, which downgrades a valid VECTOR configuration to WEIGHT; #3887's `_vector_chunk_quota` floor is what answers it. The zero predates this branch — a single entity at `related_chunk_number=1` hits it on the base commit too — but the phantom groups used to inflate the divisor past it, so dropping them is what exposes the input.

Nothing leaves the candidate pool — only the budget divisor changes. Empty groups contributed no chunk ids to begin with, so filtering them alters the quota and nothing else. `chunk_occurrence_count` is deliberately left untouched, so `chunk_tracking[...]["frequency"]` still reflects every entity a chunk appeared in, not just the surviving ones. Filtering preserves the relative order of the survivors, so they move up into the positions the linear-gradient allocation intended for them — the same thing the relation path already does.

One downstream effect is intended rather than incidental: the entity path now delivers fewer chunks, which shrinks the `entity_chunks` exclusion set passed to the relation path, so the relation side may pick up a few more. That re-balancing is the point of making both paths honour the same knob.

## Verification

- `./scripts/test.sh tests/test_operate_entity_chunk_selection.py` — 6 passed; all six fail against the base commit `d964d92`
- `./scripts/test.sh tests/test_operate_entity_chunk_selection.py tests/test_weighted_polling_nonpositive_max.py tests/utils tests/test_operate_vector_context.py` — 353 passed
- `main` is merged into this branch, so #3887's `_vector_chunk_quota()` now owns both call sites; the filter still sits ahead of the quota call
- `./scripts/test.sh tests/test_operate_entity_chunk_selection.py tests/test_kg_vector_chunk_quota.py tests/test_docstring_budget.py tests/test_weighted_polling_nonpositive_max.py tests/utils tests/test_operate_vector_context.py tests/test_operate_truncation_order.py` — 403 passed
- the `related_chunk_number=1` test tells all three states apart: `(2, 5)` on the base commit, `(0, 1)` on this branch before `main` was merged, `(1, 1)` only with both fixes present. Dropping the filter fails the `related_chunk_number=2` VECTOR assertion with `(5, 5) == (1, 1)`.
- `ruff check lightrag/operate.py tests/test_operate_entity_chunk_selection.py`
- `git diff --check`

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -6275,6 +6275,12 @@ async def _find_related_text_unit_from_entities(
         entity_info for entity_info in entities_with_chunks if entity_info["chunks"]
     ]
 
+    # Defensive only: unreachable on this path. Deduplication drops a chunk
+    # solely because an earlier-positioned entity already claimed it, so the
+    # first entity keeps every chunk it carries and at least one group always
+    # survives. The relation path's equivalent guard IS reachable, because it
+    # additionally excludes the chunks already delivered by the entity path,
+    # which can empty every relation group.
     if not entities_with_chunks:
         logger.info(
             f"Find no entity-related chunks from {len(node_datas)} entities after deduplication"
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -6269,6 +6269,18 @@ async def _find_related_text_unit_from_entities(
         # Update entity's chunks to deduplicated chunks
         entity_info["chunks"] = deduplicated_chunks
 
+    # Drop entities emptied by deduplication so they do not inflate the
+    # per-group chunk budget used by the selection strategies.
+    entities_with_chunks = [
+        entity_info for entity_info in entities_with_chunks if entity_info["chunks"]
+    ]
+
+    if not entities_with_chunks:
+        logger.info(
+            f"Find no entity-related chunks from {len(node_datas)} entities after deduplication"
+        )
+        return []
+
     # Step 3: Sort chunks for each entity by occurrence count (higher count = higher priority)
     total_entity_chunks = 0
     for entity_info in entities_with_chunks:
--- a/tests/test_operate_entity_chunk_selection.py
+++ b/tests/test_operate_entity_chunk_selection.py
@@ -1,0 +1,45 @@
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+from lightrag.base import QueryParam
+from lightrag.constants import GRAPH_FIELD_SEP
+from lightrag.operate import _find_related_text_unit_from_entities
+
+
+pytestmark = pytest.mark.offline
+
+
+@pytest.mark.asyncio
+async def test_entity_chunk_selection_drops_groups_emptied_by_deduplication():
+    chunk_ids = [f"chunk-{index}" for index in range(20)]
+    source_id = GRAPH_FIELD_SEP.join(chunk_ids)
+    node_datas = [
+        {"entity_name": f"entity-{index}", "source_id": source_id} for index in range(5)
+    ]
+
+    text_chunks_db = AsyncMock()
+    text_chunks_db.global_config = {
+        "kg_chunk_pick_method": "WEIGHT",
+        "related_chunk_number": 2,
+    }
+    text_chunks_db.get_by_ids.return_value = [
+        {"content": f"content-{chunk_id}"} for chunk_id in chunk_ids[:2]
+    ]
+
+    with patch(
+        "lightrag.operate.pick_by_weighted_polling",
+        return_value=chunk_ids[:2],
+    ) as weighted_picker:
+        result = await _find_related_text_unit_from_entities(
+            node_datas,
+            QueryParam(),
+            text_chunks_db,
+            AsyncMock(),
+        )
+
+    groups, max_related_chunks = weighted_picker.call_args.args[:2]
+    assert len(groups) == 1
+    assert groups[0]["sorted_chunks"] == chunk_ids
+    assert max_related_chunks == 2
+    assert [chunk["chunk_id"] for chunk in result] == chunk_ids[:2]
--- a/tests/test_operate_entity_chunk_selection.py
+++ b/tests/test_operate_entity_chunk_selection.py
@@ -1,45 +1,144 @@
+"""Budget parity between the entity and relation chunk-selection paths.
+
+Both paths deduplicate each group's chunk list, both can empty a group that
+way, and both size their chunk budget from the surviving group count. The
+entity path used to keep the emptied groups, so phantom groups inflated the
+WEIGHT and VECTOR budgets and over-delivered beyond ``related_chunk_number``.
+"""
+
 from unittest.mock import AsyncMock, patch
 
 import pytest
 
 from lightrag.base import QueryParam
 from lightrag.constants import GRAPH_FIELD_SEP
-from lightrag.operate import _find_related_text_unit_from_entities
+from lightrag.operate import (
+    _find_related_text_unit_from_entities,
+    _find_related_text_unit_from_relations,
+)
 
 
 pytestmark = pytest.mark.offline
 
+CHUNK_IDS = [f"chunk-{index}" for index in range(20)]
+SOURCE_ID = GRAPH_FIELD_SEP.join(CHUNK_IDS)
+RELATED_CHUNK_NUMBER = 2
+DUPLICATE_GROUPS = 5
 
-@pytest.mark.asyncio
-async def test_entity_chunk_selection_drops_groups_emptied_by_deduplication():
-    chunk_ids = [f"chunk-{index}" for index in range(20)]
-    source_id = GRAPH_FIELD_SEP.join(chunk_ids)
-    node_datas = [
-        {"entity_name": f"entity-{index}", "source_id": source_id} for index in range(5)
-    ]
 
+def _text_chunks_db(pick_method: str) -> AsyncMock:
     text_chunks_db = AsyncMock()
     text_chunks_db.global_config = {
-        "kg_chunk_pick_method": "WEIGHT",
-        "related_chunk_number": 2,
+        "kg_chunk_pick_method": pick_method,
+        "related_chunk_number": RELATED_CHUNK_NUMBER,
     }
-    text_chunks_db.get_by_ids.return_value = [
-        {"content": f"content-{chunk_id}"} for chunk_id in chunk_ids[:2]
+    text_chunks_db.get_by_ids.side_effect = lambda chunk_ids: [
+        {"content": f"content-{chunk_id}"} for chunk_id in chunk_ids
+    ]
+    return text_chunks_db
+
+
+def _duplicate_entities() -> list[dict]:
+    """Entities that all carry the same chunks, so one group survives dedup."""
+    return [
+        {"entity_name": f"entity-{index}", "source_id": SOURCE_ID}
+        for index in range(DUPLICATE_GROUPS)
     ]
 
+
+def _duplicate_relations() -> list[dict]:
+    """The relation-path twin of :func:`_duplicate_entities`."""
+    return [
+        {"src_tgt": (f"src-{index}", f"tgt-{index}"), "source_id": SOURCE_ID}
+        for index in range(DUPLICATE_GROUPS)
+    ]
+
+
+async def _select_from_entities(text_chunks_db: AsyncMock, **kwargs) -> list[dict]:
+    return await _find_related_text_unit_from_entities(
+        _duplicate_entities(),
+        QueryParam(),
+        text_chunks_db,
+        AsyncMock(),
+        **kwargs,
+    )
+
+
+async def _select_from_relations(text_chunks_db: AsyncMock, **kwargs) -> list[dict]:
+    return await _find_related_text_unit_from_relations(
+        _duplicate_relations(),
+        QueryParam(),
+        text_chunks_db,
+        entity_chunks=[],
+        **kwargs,
+    )
+
+
+def _vector_call_args(picker: AsyncMock) -> tuple[int, int]:
+    """Return the (budget, group count) a path handed to the vector picker."""
+    return (
+        picker.call_args.kwargs["num_of_chunks"],
+        len(picker.call_args.kwargs["entity_info"]),
+    )
+
+
+async def test_weighted_polling_sees_only_surviving_entity_groups():
     with patch(
         "lightrag.operate.pick_by_weighted_polling",
-        return_value=chunk_ids[:2],
+        return_value=CHUNK_IDS[:RELATED_CHUNK_NUMBER],
     ) as weighted_picker:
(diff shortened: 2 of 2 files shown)


danielaskdd:
@codex review

lightrag/operate.py
```diff
@@ -6269,6 +6269,18 @@ async def _find_related_text_unit_from_entities(
...
+    # per-group chunk budget used by the selection strategies.
+    entities_with_chunks = [
+        entity_info for entity_info in entities_with_chunks if entity_info["chunks"]
+    ]
+
+    if not entities_with_chunks:
+        logger.info(
+            f"Find no entity-related chunks from {len(node_datas)} entities after deduplication"
+        )
+        return []
```
danielaskdd, line 6407: This early return is unreachable on the entity path — please mark it as defensive rather than leaving it looking like a live case.

The filter above it is the real fix and is correct. But entity-side dedup only drops chunks already claimed by an *earlier* entity, and `chunk_occurrence_count` starts empty, so every chunk of the first group gets count 1 and is kept. Step 1 only admits groups with a non-empty chunk list, so the first surviving group is always non-empty and `entities_with_chunks` can never be empty here.

I fuzzed Step 1 + Step 2 over 200k randomized `source_id` combinations (1-5 entities, overlapping/empty/whitespace chunk IDs): 172,851 reached Step 2, and this branch fired **0** times.

This is where the entity path genuinely differs from the relation path you are mirroring. The relation version *is* reachable, because it dedups against `entity_chunk_ids` — an external set — so a relation whose every chunk was already returned by the entity path really can be emptied. Copying the guard across is fine for symmetry and cheap insurance, but the reachability argument does not come with it.

Suggest a brief comment saying so. There is a precedent to match in #3887, which documents its own unreachable `group_count <= 0` branch this way:

```
    The group_count <= 0 branch is defensive only: both call sites already
    guard on their group list being non-empty ...
```

Without that note, a future reader will assume it fires and may write a test trying to reach it.


---

tests/test_operate_entity_chunk_selection.py
danielaskdd, line 1: Good regression test — I confirmed it fails on unfixed code with `assert 5 == 1`, which is exactly the phantom-group count, so it pins the right thing.

One coverage gap worth closing: it exercises `WEIGHT` only, but #3898 documented the inflation hitting **both** strategies, and the VECTOR half is the one with a directly observable number. `num_of_chunks` is derived from the same `len(entities_with_chunks)`, so on this exact fixture it goes from 5 down to 1 — a 5x budget swing that nothing in this PR currently pins.

A second test reusing the same five-duplicate-entity fixture with `kg_chunk_pick_method: "VECTOR"`, patching `lightrag.operate.pick_by_vector_similarity` and asserting the `num_of_chunks` it receives, would cover it. That assertion is also stronger than the WEIGHT one: it reads the budget directly instead of inferring it from the group list.

Worth doing because that path is about to get a second owner — #3887 routes both call sites through a new `_vector_chunk_quota()` helper. Once both land, the VECTOR budget on this fixture depends on the two fixes agreeing, and nothing would catch a regression in that interaction.


---

danielaskdd (commented):
Thanks for picking up #3898 — I filed that issue, and this is the fix I had in mind: mirroring the relation path's filter rather than special-casing empty groups inside the two pickers. I checked the branch out and verified the fix works. Two non-blocking notes inline, plus one merge-ordering point below that does need a decision before this lands.

**Verified**

- The filter resolves the asymmetry. On the five-duplicate-entity fixture from #3898, the entity path goes from 8 chunks / VECTOR quota 5 to **2 chunks / quota 1**, matching what the relation path already produced for the same logical input.
- The regression test genuinely fails on unfixed code (`assert 5 == 1`).
- Merges cleanly with current `main`, and with #3887 applied as well; the combined tree passes **600 tests** across `tests/` root and `tests/utils/`.

## Merge ordering — please land #3887 first

This PR **alone** regresses the bug #3887 fixes, so the order matters.

The phantom groups this PR removes were incidentally *masking* the zero-quota defect: they inflated the divisor in `int(max_related_chunks * n / 2)`, which kept the result off zero. Remove the masking without the floor in place and the quota collapses.

Measured with `related_chunk_number=1` and two entities both referencing only `c1`:

| tree | VECTOR `num_of_chunks` | outcome |
| --- | --- | --- |
| `main` today | 1 | vector selection preserved |
| **this PR alone** | **0** | **silently degrades to WEIGHT** |
| #3887 alone | 1 | vector selection preserved |
| both | 1 | vector selection preserved |

On this PR alone the run also emits `WARNING: No entity-related chunks selected by vector similarity, falling back to WEIGHT method` — the misleading log #3887 exists to remove.

So merging this first opens a window where the zero-quota defect is *easier* to hit than it is on `main` today, for anyone running `RELATED_CHUNK_NUMBER=1`. Merging #3887 first leaves every point in history safe. No rebase needed either way — the two diffs touch different hunks and merge cleanly.

Nothing to change in this PR for that; it just should not go in ahead of #3887.


---

lightrag/operate.py
danielaskdd: Done in eeb77572d — the guard now carries a note in the shape you asked for:

```python
    # Defensive only: unreachable on this path. Deduplication drops a chunk
    # solely because an earlier-positioned entity already claimed it, so the
    # first entity keeps every chunk it carries and at least one group always
    # survives. The relation path's equivalent guard IS reachable, because it
    # additionally excludes the chunks already delivered by the entity path,
    # which can empty every relation group.
```

It states the unreachability, the reason it holds (the first surviving group is always non-empty), and the asymmetry with the relation path being mirrored — so the reachability argument no longer travels silently with the copied guard.

tests/test_operate_entity_chunk_selection.py
danielaskdd: Covered in eeb77572d. The file now holds five tests instead of one:

- `test_vector_budget_counts_only_surviving_entity_groups` — the one you asked for: same five-duplicate-entity fixture, `kg_chunk_pick_method: "VECTOR"`, patches `lightrag.operate.pick_by_vector_similarity` and asserts the `num_of_chunks` it receives (1, not 5).
- `test_both_paths_hand_the_same_budget_to_the_vector_picker` — the same budget read through both paths, asserted equal.
- `test_entity_path_honours_the_configured_weight_budget` — the delivered chunk ids under WEIGHT (2, not 8), the user-visible symptom the original test left unpinned.
- plus the original picker-argument test and its relation-path twin.

All five fail against the base commit.

On the #3887 interaction you flagged: it has since merged, so I verified it rather than leaving it hypothetical. Merging current `main` into this branch is conflict-free, and with `_vector_chunk_quota()` owning both call sites the subset still passes (402 tests, including #3887's own `tests/test_kg_vector_chunk_quota.py`). Dropping the filter from that merged tree fails the VECTOR assertion with `(5, 5) == (1, 1)`. So the two fixes do agree on this fixture, and the interaction is now pinned rather than unguarded.

danielaskdd:
@codex review

lightrag/operate.py
```diff
@@ -6269,6 +6269,24 @@ async def _find_related_text_unit_from_entities(
         # Update entity's chunks to deduplicated chunks
         entity_info["chunks"] = deduplicated_chunks
 
+    # Drop entities emptied by deduplication so they do not inflate the
+    # per-group chunk budget used by the selection strategies.
+    entities_with_chunks = [
+        entity_info for entity_info in entities_with_chunks if entity_info["chunks"]
```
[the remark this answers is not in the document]
danielaskdd: Accurate for the commit you reviewed (`2862c68d`), and now addressed — but the attribution needs one correction.

**The zero is real.** On `2862c68d`, `related_chunk_number=1` with five duplicate entities gives `groups=1, num_of_chunks=0`, and the path does downgrade to WEIGHT. Confirmed by probing the call site directly.

**It is not introduced by this filter, though.** On the base commit `d964d92`, before the filter exists, a *single* entity at `related_chunk_number=1` already yields `num_of_chunks=0` and downgrades. The defect predates this branch. What the filter changes is exposure: the phantom groups used to inflate the divisor past the truncation point (five duplicates gave `int(1 * 5 / 2) = 2`), so removing them stops masking a bug that was already there.

**The clamp you ask for already exists.** It is #3887's `_vector_chunk_quota()`, whose floor of 1 is documented for exactly this failure mode. That PR merged on 2026-09-13, and #3898 anticipated the interaction when it was filed: *"the phantom groups inflate the divisor and can therefore mask the zero-quota bug #3887 fixes... #3887 should land as-is; this is deliberately out of its scope."* Implementing a second clamp here would collide with it.

`main` is now merged into this branch, so the floor is in place. Measured across the three states, same fixture, `related_chunk_number=1`:

| state | groups | `num_of_chunks` | VECTOR |
| --- | --- | --- | --- |
| base `d964d92` (no filter, no floor) | 5 | 2 | ok — zero is masked |
| `2862c68d` (filter, no floor) | 1 | **0** | **downgrades** ← your finding |
| now (filter + floor) | 1 | 1 | ok |

Your finding did surface a genuine gap in the tests, so thanks for it: every VECTOR test here ran at `related_chunk_number=2`, where `int(2 * 1 / 2)` is 1 with or without the floor — so nothing could tell the two fixes apart. `5b9795a2` adds `test_collapsed_groups_keep_a_nonzero_vector_quota` at `related_chunk_number=1`, which fails `(2, 5)` on the base commit, fails `(0, 1)` on the commit you reviewed, and passes only with both fixes present. #3887's own `tests/test_kg_vector_chunk_quota.py` did not cover this, since it exercises a single entity rather than duplicates collapsing.

danielaskdd:
@codex review

Issue #3866, opened 2026-09-07
[Question]: Could connected subgraph optimization help LightRAG’s context selection?


berendmarkhorst:
### Do you need to ask a question?

- [x] I have searched the existing question and discussions and this question is not already answered.
- [x] I believe this is a legitimate question, not just a bug or feature request.

### Your Question

Hi! First of all, thanks for building LightRAG!
I maintain [SteinerPy](https://github.com/berendmarkhorst/SteinerPy), a Python package for Steiner tree problems and related graph optimization problems. While looking at LightRAG, I wondered whether these methods could help select the context passed to an LLM.
Suppose we assign relevance scores and token costs to information in the graph. We could then select a connected subgraph with the highest total relevance within a fixed token budget. Two relevant pieces of information may be connected through something that seems less useful on its own. Yet, that connection could help the LLM understand how they relate.
LightRAG already supports graph retrieval and token limits. Would explicitly optimizing this selection be worth exploring, or is something similar already covered?
I would be happy to try a small prototype and compare answer quality and retrieval time under the same token budget.

danielaskdd:
Thanks for the thoughtful question, and for offering to prototype it — any experiment that improves LightRAG is welcome.

**First, an important property of LightRAG's graph: connectivity is not a constraint anywhere in the pipeline.**

Entity and relation extraction has no connectivity requirement, so **isolated entities are very common** — that is expected, not a defect. And an isolated entity is still valuable for answering: LightRAG relies heavily on the LLM-generated **descriptions** of entities and relations (a pre-summarization step, merged and re-summarized across chunks in `_handle_entity_relation_summary`) rather than on graph topology. A single node with a rich description often carries most of what the query needs. Entities are retrieved directly from the entity vector DB on those descriptions (`_get_node_data`), so an isolated node competes on equal footing with a well-connected one.

This is the main thing to watch in your design: if connectivity becomes a **hard** constraint, a Steiner-style selector will systematically **crowd out isolated-but-highly-relevant entities** in favor of bridge nodes with low individual relevance. On our data that trade is not obviously a win. A soft formulation — connectivity as a bonus rather than a feasibility requirement — is probably the shape worth testing first.

**That said, the objective you state is exactly what LightRAG is after:** *select the content most relevant to the user's query within a fixed token budget.* Today that is done by rank-then-truncate, not by optimization, so there is real room to explore.

[...]

Issue #28, opened 2024-10-16
Some questions about the preprint manuscript


jeremyzhangsq:
Hi @LarFii ,

Thank you for releasing this interesting work.
Maybe I have missed something, but I am confused about the experimental results in the preprint manuscript. 

For example, in Table 1, do the LightRAG in the same column have the same configuration? Why do these performances vary a lot? If they are not in the same configuration, could you explain this?

![WX20241016-234449](https://github.com/user-attachments/assets/5a9238f0-f0a3-4ec5-af03-a38a2c94485f)

Thanks.

Jeremy

LarFii:
Hi @jeremyzhangsq,

Thank you for your question! Throughout the entire table, LightRAG uses the same configuration. The variation in values comes from our one-to-one evaluation method, where each baseline is compared individually with LightRAG. The percentage values represent the win rate, which explains the differences. For example, in the first row and first column, the value indicates that, in terms of comprehensiveness, LightRAG outperforms Naive RAG in 67.31% of the queries.

I hope this helps clarify the results!

Best regards,
Zirui

dannis97500:
Hi @LarFii,

Thank you for releasing this interesting work. Perhaps we need the test scripts in the paper to help us accurately reproduce the experimental results. Can you provide them in future version updates?

Thanks.

Dannis

PR #3914, opened 2026-09-10
Prototype: optional SteinerPy context selection with soft connectivity

### LINKED ISSUE #3917: [Bug]: `_apply_token_truncation` rebuilds `filtered_entities` / `filtered_relations` in stage-1 order, discarding stage-2 selection order
## Describe the bug

Stage 3 treats the order of `filtered_entities` / `filtered_relations` as an **importance ranking**, but stage 2 rebuilds those lists in the original stage-1 retrieval order rather than in the order it actually selected.

`lightrag/operate.py:5570-5598` (`_apply_token_truncation`) derives the filtered originals by membership test over the *pre-truncation* lists:

```python
final_entity_names = {e["entity"] for e in entities_context}
seen_nodes = set()
for entity in final_entities:            # <- stage-1 order
    name = entity.get("entity_name")
    if name in final_entity_names and name not in seen_nodes:
        filtered_entities.append(entity)
```

So `filtered_entities` preserves the relative order of `final_entities`, not the order of `entities_context` that stage 2 just produced. The same pattern applies to `filtered_relations`.

Downstream, that order is load-bearing in two places, for both entities and relations:

1. **Chunk attribution on duplicates.** `_find_related_text_unit_from_entities` step 2 (`operate.py:6255`) and `_find_related_text_unit_from_relations` step 2 (`:6514`) deduplicate by *first occurrence* — the comment says verbatim "keep chunks from earlier positioned entities". A chunk shared by two entities is attributed to whichever appears earlier in the list.

2. **Chunk quota allocation.** Both call `pick_by_weighted_polling` (`operate.py:6328`, `:6619`), whose contract is explicit (`lightrag/utils.py:6095-6133`):

   > `entities_or_relations`: List of entities or relations **sorted by importance (high to low)**

   It assigns a linearly decreasing quota by list position — position 0 gets `related_chunk_number` chunks, the last position gets `min_related_chunks=1`.

## Why it is latent today, and when it stops being latent

On `main` this is harmless, which is why it has gone unnoticed:

- Stage 2 is prefix truncation, so the surviving set is a prefix of `final_entities`.
- Filtering `final_entities` by membership therefore reproduces exactly the same order as `entities_context`.

The two lists agree **only because no current code path reorders records in stage 2**. Nothing enforces that; it is an accident of the current implementation, not an invariant.

It breaks as soon as stage 2 reorders. The concrete trigger already on the table is PR #3914: its `prize_source="rerank_score"` arms call `rerank_records`, which sorts entities and relations by reranker score before selection. In that configuration:

- the **prompt** lists entities in reranker order, but
- **chunk quota allocation** still ranks them by stage-1 order (vector order interleaved round-robin with the global branch; for local relations, by node degree — `operate.py:6200`).

The reranker's judgment reaches which records survive, but not which records get chunk budget. That systematically caps what a stage-2 reranking change can deliver and would contaminate the B-arm vs. A-arm comparison in that PR. Note that a plain non-prefix *subset* selector (packing without reordering) is still fine — only reordering triggers this.

## Expected behavior

Stage 2's output should be the single source of truth for downstream importance. `filtered_entities` / `filtered_relations` should follow the order of `entities_context` / `relations_context`, so that any future stage-2 selector — reranked, packed, or otherwise — propagates its ranking into chunk selection.

## Steps to reproduce

No user-facing reproduction on `main`, since no shipped path reorders in stage 2. To observe it directly:

1. In `_apply_token_truncation`, reverse `entities_context` immediately after truncation.
2. Log the `entity_name` order of `filtered_entities` and the `expected_counts` computed inside `pick_by_weighted_polling`.
3. The quota allocation is unchanged by the reversal — it still follows stage-1 order.

## Proposed fix

Build the filtered lists by walking `entities_context` / `relations_context` and looking up the originals through the existing `entity_id_to_original` / `relation_id_to_original` maps (both are already built earlier in the same function), instead of walking `final_entities` / `final_relations`.

The change is order-only and a no-op for the current default path, so it needs a regression test that pins the new invariant: given a stage-2 result whose order differs from stage-1, assert that `filtered_entities` matches `entities_context` order. `tests/pipeline/` or a focused test alongside the existing query-context tests.

## Additional context

Found while reviewing PR #3914 (https://github.com/HKUDS/LightRAG/pull/3914#issuecomment-5628720723). Filing separately because the invariant gap is ours and should be fixed regardless of whether that PR is adopted.

berendmarkhorst:
Related to HKUDS/LightRAG#3866.

LightRAG currently selects entity and relation context by taking a prefix of retrieved records within separate token limits. This opt-in prototype tests prize-collecting selection while keeping isolated descriptions eligible. It integrates at `_apply_token_truncation`, calls SteinerPy's directed heuristic with `exact=False`, and preserves the normal downstream chunk-selection path.

Connectivity earns a spanning-forest bonus through an artificial-root construction; it is not a feasibility requirement. This adapts the graph-retrieval idea discussed in [SteinerPy #28](https://github.com/berendmarkhorst/SteinerPy/issues/28). The original rank path remains the default, and SteinerPy remains optional.

Following [the adoption review](https://github.com/HKUDS/LightRAG/pull/3914#issuecomment-5628720723), the experiment now compares six matched arms:

| Prize source | Prefix truncation | Relevance packing | Steiner soft |
|---|---|---|---|
| Current ordering | A0 | A1 | A2 |
| KG reranker score | B0 | B1 | B2 |

B1 and B2 use numerical reranker scores, not their ranks. All six arms enable downstream chunk reranking with the same configured model, project-default minimum score, chunk limit, and entity/relation/total token caps. B0 is independently usable without the SteinerPy backend. Above the candidate cap, experimental arms report a fallback to the prefix control under the same prize source.

The runner exports contexts, answers, input/output token estimates, question-level paired uncertainty, serial latency, and retrieval-only load trials at 1/8/16/32 concurrency. Request-local accounting records both KG and all downstream chunk inputs to the reranker, including net load relative to A0. Logical adapter calls and token estimates are distinguished from GPU time and provider billing. Live defaults cover top_k 20/40/60 and report fallback frequency.

The update also replaces the query-path budget assertion with an explicit exception, separates optional SteinerPy CI dependencies into their own job, and reports per-proposal heuristic certificates without claiming a guarantee for budget repair or answer quality.

Validation:
- 138 focused tests pass, covering the evaluation directory, addon configuration, cache behavior, rerank validation and concurrent tokenization as documented in `docs/SteinerContextReview.md`.
- A six-arm smoke run completed 72 paired context rows and 720 load rows at concurrency 1/8/16/32. It uses four synthetic cases and a lexical test double, calls no answer model, and is not production evidence.
- Ruff checks for changed selector/evaluation code and tests pass; no whitespace errors.

**The acceptance study is still pending.** No fixed real index or configured answer/rerank models were available for this revision. There is no demonstrated real answer-quality gain, production latency result, or real-index fallback frequency. Adoption requires a significant useful gain over both same-prize controls; a B0/B2 or B1/B2 tie does not justify adding the solver.

See `docs/SteinerContextPrototype.md` for the formulation and runnable study, `docs/SteinerContextReview.md` for the response to each review request, and `docs/evaluation/steiner-review-smoke.json` for the diagnostic summary. The original reranking-disabled measurements remain explicitly labelled historical in `docs/SteinerContextResults.md`.

Following [the remaining-comments review](https://github.com/HKUDS/LightRAG/pull/3914#issuecomment-5630770491), query-level `enable_rerank=False` now skips KG and chunk reranking and explicitly reports the ordering control. KG rerank-score configuration requires a callable reranker at construction/config refresh. Capped KG responses retain only valid scored records and report the excluded tail; submitted-document load accounting remains intact. The guide documents pre-seeding `TIKTOKEN_CACHE_DIR` for offline fixtures. The formatter, final-newline and source-docstring CI failures are also corrected.

The upstream ordering fix HKUDS/LightRAG#3917 remains with the maintainers; this PR adds no workaround. Their proposed 30–60-question real-index study is pending internal confirmation and that fix. A real hosted-provider smoke check and the acceptance study remain outstanding; capped-provider behavior here is covered with test doubles.


danielaskdd:
Thanks for this — it is an unusually careful prototype. Before the requests, the parts that are clearly right:

- The exhaustive-subset oracle (`test_transformation_against_exhaustive_subset_oracle`) is real proof that the directed split-node construction optimizes `F(S) − λ·cost(S)`. That is the kind of validation we rarely get.
- You built controls that could falsify your own proposal (`relevance`, `soft_greedy`) and reported honestly that they tie with `steiner_soft` on all four diagnostics.
- `steiner_hard` losing the isolated evidence (75% vs 100%) empirically reproduces the failure mode predicted in #3866. That is a useful negative result on its own.
- Keeping the default path free of the import, gating on an optional extra, separating the answer cache, and moving the CPU work off the event loop are all the right engineering choices.

The prototype is sound. What it does not yet do is establish that the method is worth adopting. Below is what we would need, and why.

## 1. Every measurement was taken with reranking off

`steiner_ablation.py:453` makes `--rerank` opt-in (`action="store_true"`), the demo path passes `rerank=False` (`:276`), `min_rerank_score` is pinned to `0.0` (`:291`), and every test uses `enable_rerank=False`. Meanwhile `QueryParam.enable_rerank` defaults to **True** (`lightrag/base.py:166`).

So the configuration you measured is not the configuration LightRAG runs in. This matters in both directions:

- Stage 2's output determines the chunk candidate pool that stage 3's reranker scores and filters by `min_rerank_score`. A downstream reranker can partially repair a poor stage-2 selection, which compresses whatever gain stage 2 contributes. Your 50% → 100% gap is measured against a baseline we deliberately weakened.
- The opposite risk is equally invisible: if the connectivity bonus admits low-relevance bridge records, their source chunks dilute the pool and get filtered out downstream. That is a net loss that only appears with reranking on.

All quality claims need to be re-measured with reranking enabled and the project's default `min_rerank_score`.

## 2. The prizes are the weaker half of the design, and they are not what you think they are

`r_i = 1/sqrt(position_i)` proxies an ordering that is a round-robin interleave of two heterogeneous sources (`_perform_kg_search`). For relations in local and hybrid mode, the local branch arrives from `_find_most_related_edges_from_entities` sorted by `(rank, weight)` — and `rank` there is **node degree** (`operate.py:6200`), not query relevance. So for a large share of relation records the "relevance prize" is a monotone function of degree.

That undercuts the premise. Prize-collecting optimization over numbers uncorrelated with relevance cannot be evaluated as an optimization result.

This splits the problem into two independent axes, and the PR currently only varies the second:

| prize source \ selection | prefix truncation | budget packing | Steiner soft |
|---|---|---|---|
| current ordering | **A0** (today's baseline) | A1 | **A2** (your `steiner_soft`) |
| rerank score | **B0** | B1 | **B2** |

You measured A0/A1/A2 and found A1 ≈ A2. The interesting comparisons are B0 and B2. Note that if B0 is added, **A2 must be re-run as B2** — comparing "good scores + simple selection" against "degree-derived scores + sophisticated selection" says nothing about the selection algorithm.

## 3. The rerank arm is not the cheap option either — please measure both costs properly

We have a concrete datapoint: **bge-reranker-v2-m3 on a 2080Ti sorts roughly 50 text records per second**; a substantially faster GPU should roughly double that.

Entity and relation descriptions are not short. `DEFAULT_SUMMARY_LENGTH_RECOMMENDED = 600` and `DEFAULT_SUMMARY_MAX_TOKENS = 1200` (`lightrag/constants.py:32-34`) put them in the same size class as a text chunk (`CHUNK_SIZE` default 1200), so that throughput transfers with little discount.

At 120 candidate records that is roughly **2.4 s on a 2080Ti, ~1.2 s on a faster card**, against your measured **433 ms (p95 454 ms)** for `steiner_soft`. The KG-rerank arm is the *more* expensive option in wall-clock, and it consumes GPU — a scarce, shared resource with a hard throughput ceiling that concurrency cannot hide. Your 433 ms is CPU, which is easier to scale out.

So neither option is free, and the honest question is quality per unit of the scarce resource. Please report:

- **Net** rerank load, not gross. `process_chunks_unified` scores the entire `unique_chunks` set (`top_n=chunk_top_k` only caps what is returned, `utils.py:6537-6546`). If stage-2 reranking drops irrelevant entities and relations, stage 3 sees a smaller pool. The B arms may cost far less than the naive sum suggests — but it has to be measured.
- **Concurrency**, not just serial latency. p50/p95 at 8/16/32 concurrent queries. A CPU-bound 433 ms in a thread pool and a GPU-queued 2.4 s degrade very differently, and neither degradation is visible in the current sequential numbers.
- **Fallback rate at realistic `top_k`.** `DEFAULT_TOP_K = 40` in hybrid mode routinely produces more than `max_candidates = 120` entity-plus-relation rows, and above the cap every experimental arm silently reverts to rank truncation. If most production queries fall back, the quality discussion is moot. Please report the frequency at `top_k` 20/40/60 on a real index.

## 4. Theory that is still missing

The oracle test validates the transformation. It does not cover:

- Any approximation guarantee for the `exact=False` path that actually runs in production. The test only asserts `objective(heuristic) <= optimal`, which holds trivially.
- Any guarantee for the Lagrangian-proposal → `repair` → `grow` → refill pipeline that returns to the budget-feasible region. That chain is unanalyzed.
- **The central gap: why `F(S)` should correlate with answer quality at all.** You state this limitation yourself, and it is the one that decides adoption. `β = 0.15` is currently being compared against `1/sqrt(position)` values, so even its magnitude has no defensible scale.

One structural observation that may help you argue the case. With `DEFAULT_MAX_ENTITY_TOKENS = 6000`, `DEFAULT_MAX_RELATION_TOKENS = 8000` and ~600-token descriptions, stage 2 admits on the order of ten entities and a dozen relations out of many more candidates. Two consequences pull in opposite directions, and both are worth addressing explicitly:

- Picking 10 of 60 makes the packing quality far more consequential than picking 55 of 60 — a single oversized early record costs a whole slot. This is a genuine argument for replacing prefix truncation with *something*.
- But at ten selected records, a spanning-forest bonus has very little room to act, and which ten are chosen is dominated by prize quality. That argues the gain lives in axis one, not axis two.

If you believe the connectivity term earns its place at this selection size, we would like to see the argument.

## 5. What we would need to accept the method

- Six arms: {current ordering, rerank score} × {`rank`, `relevance`, `steiner_soft`}, at minimum B0 and B2.
- Reranking enabled throughout, project-default `min_rerank_score`, identical rerank model, `chunk_top_k`, `max_entity_tokens`, `max_relation_tokens` and `max_total_tokens` across all arms.
- B2's prizes taken from `rerank_score`, not from retrieval position.
- 30–60 held-out questions on a fixed real index with a configured answer model, including isolated-evidence, multi-hop and disconnected-evidence categories, with paired question-level uncertainty. Your own evaluation guide already specifies this correctly.
- Latency reported serially and under concurrency, plus net rerank calls and token usage per arm, plus fallback frequency.

**Decision rule.** `steiner_soft` is worth adopting only if it beats **both** `rank` and `relevance` *on the same prize source*, significantly, and the margin justifies its cost relative to simply reranking entity and relation descriptions. If B0 alone reaches B2's quality, we take the reranking change and not the solver. If B2 ≈ B0, the selection layer is not the bottleneck.

Worth saying plainly: the reranking arm is likely the change we want regardless of how the Steiner comparison resolves, because relation ordering by node degree is a defect in its own right. If you would rather not carry that experiment, we are happy to prototype B0 ourselves and hand you the arm so both are measured in your framework.

## 6. Two small code notes

- `lightrag/steiner_context.py` enforces the token budget with a bare `assert`. `python -O` strips it. A hard invariant on the query path should `raise`.
- Adding `--extra steiner` to the shared offline CI job pulls SteinerPy, HiGHS and SciPy into every run for a path that is off by default. Please scope it to a separate job, or mark those tests and install the extra only there.

None of this is a rejection. The experiment design is better than most, and the negative results you reported are genuinely useful to us. We just cannot adopt a 433 ms-per-query selector on the strength of four synthetic questions measured with reranking disabled, where the two solver-free controls tie with it.


---

berendmarkhorst:
Thanks for the detailed review, @danielaskdd. I agree that the comparison needs to separate prize quality from selection quality.

I've pushed [79cb5e9](https://github.com/HKUDS/LightRAG/commit/79cb5e9c840132517f94665c7c812abc7bf71516) with the following changes:

- **Six arms:** A0/A1/A2 use the current ordering; B0/B1/B2 use the same KG reranker. B1 and B2 consume numerical reranker scores directly. B0 is independently usable without the SteinerPy backend.
- **Reranking throughout:** all six arms enable downstream chunk reranking and share the same model and token/chunk limits. The evaluator reads `DEFAULT_MIN_RERANK_SCORE`, which is 0.0 at this revision; a separate positive-threshold test exercises filtering. Missing, failed or incomplete scoring cannot silently count as a valid comparison.
- **Measurement:** separate serial QA and retrieval-only load trials at 1/8/16/32 concurrency, net KG-plus-chunk rerank load, token usage, fallback reporting at top_k 20/40/60, and paired question-level uncertainty. Load requests perform actual stage-1 retrieval. Counters include all submitted chunks, not just those returned by top_n; adapter wall time and token estimates are distinguished from GPU time.
- **Theory and implementation:** beta now scales with the largest prize. Proposal diagnostics expose SteinerPy's instance-specific certificate, with explicit limits on what it says about budget repair or answer quality. The budget invariant raises an exception under `python -O`, and optional SteinerPy tests have their own CI job.

124 tests pass locally. The revised diagnostic also completed 72 paired context rows and 720 load rows. Those use four synthetic cases and a lexical test double, with no answer model; they validate the measurement path and do not satisfy the quality requirement. GitHub currently reports `action_required` for CI, with no jobs started.

**The real-index study remains outstanding.** I have not yet run the 30–60 held-out questions with configured answer/rerank models, so I am not claiming a real quality gain, production latency result, or real-index fallback rate. The runnable protocol and the status of each request are documented in [the review follow-up](https://github.com/HKUDS/LightRAG/blob/79cb5e9c840132517f94665c7c812abc7bf71516/docs/SteinerContextReview.md).

I agree with your decision rule: if B0 or B1 matches B2 on real answers, there is no case for adding the solver.

danielaskdd:
Thanks — I went through 79cb5e9 rather than taking the summary at face value, and the changes are real. Verified:

- `prize_source` is a genuine second axis, and B0 is wired into the **production** path, not just the evaluator (`_apply_token_truncation` now takes `query` and forwards `rerank_func=global_config.get("rerank_model_func")`). That was the part I most expected to find evaluator-only.
- `rerank_records` raises on invalid, duplicate, negative or partial scores instead of quietly falling back to retrieval order. That matches the "fail loud, never silently substitute" rule in our AGENTS.md.
- `check_budget` raises and now covers **both** return paths — the fallback path previously had no budget check at all, which I had missed.
- `beta = connectivity_bonus * max(prizes)` makes the bonus dimensionless against the prize scale, with `prize_max` / `effective_connectivity_bonus` exposed. That answers the "β has no defensible scale" point.
- The shared offline job no longer pulls the extra; `steiner-tests` is separate.

**One correction on my side:** my complaint about `min_rerank_score: 0.0` was wrong. `DEFAULT_MIN_RERANK_SCORE = 0.0` (`lightrag/constants.py:67`) — you were using the project default. Apologies for the noise.

Your framing of the synthetic six-arm run as measurement-path validation only, with the real-index study still outstanding, is accurate and appreciated.

Four things below. The first is ours, not yours.

## 1. Stage-2 ordering does not reach chunk selection — filed as #3917

This one caps what your B arms can show, and it is a defect in our code, so do not work around it.

`_apply_token_truncation` rebuilds `filtered_entities` / `filtered_relations` by walking the **pre-truncation** `final_entities` / `final_relations` and filtering by membership (`operate.py:5570-5598`). The result keeps stage-1 order, not the order stage 2 produced.

Stage 3 treats that order as an importance ranking in two places, for entities and relations alike:

- Duplicate chunks are attributed to whichever record appears **earlier** in the list (`operate.py:6255`, `:6514` — the comment says so verbatim).
- `pick_by_weighted_polling` (`operate.py:6328`, `:6619`) allocates a **linearly decreasing chunk quota by list position**; its docstring requires the input "sorted by importance (high to low)" (`utils.py:6095-6133`). Position 0 gets `related_chunk_number` chunks, the last gets 1.

On `main` the two lists happen to agree, because prefix truncation makes the survivors a prefix. A non-prefix *subset* selector is still fine. **Reordering is what breaks it** — and `rerank_records` sorts by score before selection. So in your B arms the prompt lists entities in reranker order while chunk budget is still allocated by stage-1 order (vector order round-robin'd with the global branch; node degree for local relations).

Net effect: the reranker decides which records survive but not which records get chunk budget. That will systematically understate B0/B1/B2 and contaminate the same-prize contrasts. I have filed #3917 with the analysis and a proposed fix; we will handle it. Worth re-running the comparison once it lands.

## 2. `query_param.enable_rerank` is ignored on the KG path

`_apply_token_truncation` passes `rerank_func` unconditionally, and `select_context` decides purely on `prize_source`. A caller who sets `enable_rerank=False` still pays a KG reranker call. Stage 3 honours the flag; stage 2 should too, and the two should not disagree about whether reranking is on for a given query.

## 3. Missing reranker fails per-query rather than at construction

`rerank_records` raises `ValueError` when `rerank_func` is not callable. Failing loud is the right instinct and consistent with our contract — but the wrong moment. `prize_source="rerank_score"` with no `rerank_model_func` is a static misconfiguration, so it should be rejected when the config is validated, not on every query in production. Compare `apply_rerank_if_enabled`, which warns and passes through; whichever contract we pick, stage 2 and stage 3 should not differ.

## 4. `top_n=len(documents)` plus "every candidate must be scored" is a portability risk

`rerank_records` requests a score for every candidate and raises when `len(scores) != len(documents)`. Several hosted rerankers cap `top_n` server-side or silently return fewer results; against those, a query with many candidates fails outright rather than degrading. Only the deterministic test double exercises this today. Please either relax to "every *returned* index must be valid, and the unscored tail is excluded from the candidate set" — documented as such — or validate the provider's cap up front. Worth testing against at least one real hosted provider before the acceptance run.

## On CI

You are right that nothing has started: as a first-time contributor the workflows need maintainer approval. I will approve the run so both jobs execute.

I could not independently confirm your 124-pass figure locally — `tests/evaluation` gives me 14 passed / 13 skipped (no SteinerPy) / 43 errors, and every error is the `cl100k_base` fixture trying to download its encoding through a blocked proxy, not a code fault. Minor note from that: these files are marked `@pytest.mark.offline` but need network on first use. Hosted runners have it, so CI is fine, but consider pre-seeding `TIKTOKEN_CACHE_DIR` or documenting the requirement.

## Where this leaves us

The implementation and measurement scaffolding now satisfy the review. What is outstanding is exactly what you say it is: the real-index study, which your environment cannot run.

Since you have already built the six-arm harness and the metrics, the efficient split is that we run it. We have a fixed index with configured answer, embedding and rerank models, and the 2080Ti figure quoted earlier is ours. Let me confirm internally, but plan on us producing the 30–60 held-out question run — after #3917 lands, so the B arms are not measured against a capped ceiling. If that works for you, the remaining asks are items 2–4 above.


---

berendmarkhorst:
@danielaskdd Thanks for the follow-up. Addressed the three implementation points in [c8a564b](https://github.com/HKUDS/LightRAG/pull/3914/commits/c8a564b7af36cdd761a55e1d3843fe11d1f03723):

- `enable_rerank=False` now skips KG reranking as well as downstream chunk reranking. A rerank-score selector uses the ordering control for that query, with the requested/effective prize source and opt-out recorded in diagnostics.
- `prize_source="rerank_score"` without a callable `rerank_model_func` is rejected at construction/config refresh, before storage creation. The existing chunk-only warning/pass-through behavior is preserved.
- Capped/partial KG responses now keep valid returned scores and exclude the unscored tail. Empty responses select no KG records. Invalid/duplicate scores still fail explicitly. Diagnostics report input and unscored counts, and evaluation accounting still counts every document submitted to the provider.

Added regression coverage for these paths: 138 focused tests pass, including addon configuration and the existing cache/rerank/tokenizer checks. Ruff 0.15.17 and whitespace checks pass. I also corrected the formatter, JSON final-newline and source-docstring issue-reference failures from the previous CI run, and documented pre-seeding `TIKTOKEN_CACHE_DIR` before offline tests.

Agreed on the split: #3917 stays upstream with you, with no workaround in this PR. Your proposed 30–60 held-out question study on a fixed real index/models/hardware after that fix lands would be very useful, pending your internal confirmation. The six-arm runner and equal-budget comparisons remain ready for that handoff.

The capped-provider cases here use test doubles; I have not run a real hosted-provider call in this environment. That smoke check should precede the acceptance run. No new real answer-quality or production-latency result is claimed, and the existing synthetic measurements remain labelled as diagnostics.

berendmarkhorst:
Any updates on this @danielaskdd ? 😄

PR #3918, opened 2026-09-11
Fix token truncation to preserve stage-2 reordering in filtered lists

### LINKED ISSUE #3917: [Bug]: `_apply_token_truncation` rebuilds `filtered_entities` / `filtered_relations` in stage-1 order, discarding stage-2 selection order

danielaskdd:
## Description

Fix a bug in `_apply_token_truncation` where the `filtered_entities` and `filtered_relations` lists were being rebuilt in stage-1 retrieval order instead of preserving the order from stage-2 truncation/selection. This caused any reordering performed by a stage-2 selector (e.g., a reranker) to be invisible to stage-3 chunk selection, which relies on list position as an importance ranking.

## Related Issues

Closes #3917

### lightrag/operate.py
- **Fixed filtered list reconstruction**: Changed from filtering `final_entities`/`final_relations` through membership tests (which reimposed stage-1 order) to walking the `*_context` lists (stage-2's output) and resolving each record through pre-truncation maps. This preserves any reordering a stage-2 selector performs.
- **Deduplication semantics**: Changed `entity_id_to_original` and `relation_id_to_original` to use `setdefault()` instead of direct assignment, ensuring the first-retrieved record wins in case of duplicates (preserving existing behavior).
- **Unresolvable records are reported, not dropped silently**: a `*_context` record that resolves to no original (a selector that invents or renames a record) reaches the prompt but cannot reach chunk selection. Those are now collected per list and reported in one `logger.warning`, instead of a bare `continue`.
- **Documentation**: Added detailed docstring explaining the ordering invariant, why `filtered_*` lists must follow stage-2 order rather than stage-1 order, and that a selector may reorder and drop but must not invent. The accepted residue of first-occurrence dedup is documented next to the `setdefault` calls: if a name ever repeats, the prompt shows the context row a reordering selector picked while `filtered_*` carries the first-retrieved original — unreachable today, since stage 1 dedups entities by name and relations by sorted pair.

### tests/test_operate_truncation_order.py (new file)
- Added comprehensive test suite with 9 tests covering:
  - Reordering selectors propagating into filtered lists
  - Reordering combined with record dropping
  - Prefix truncation (unchanged behavior)
  - Non-prefix subsets without reordering
  - Duplicate name deduplication (first-retrieved wins)
  - Alternative relation format (`src_tgt` vs `src_id`/`tgt_id`)
  - Missing tokenizer short-circuit path
  - An unresolvable (invented) context record being dropped and warned about
  - Stage-3 composition: `filtered_entities` piped into `_find_related_text_unit_from_entities`, pinning the chunk order `pick_by_weighted_polling` produces from the reranked ranking

## Checklist

- [x] Changes tested locally (comprehensive test suite added)
- [x] Code reviewed (logic verified against stage-3 usage patterns)
- [x] Documentation updated (docstring clarifies ordering invariant)
- [x] Unit tests added (9 new tests in test_operate_truncation_order.py)

## Additional Notes

The bug was invisible while stage-2 only performed prefix truncation (a prefix of stage-1 order filtered by membership reproduces stage-1 order exactly), but becomes critical when stage-2 reorders records. Stage-3 uses list position for two load-bearing operations: attributing shared chunks to earlier-positioned records and allocating chunk quota via `pick_by_weighted_polling`. The fix ensures stage-2's importance ranking is the single source of truth for downstream processing.

Because the change is order-only and a no-op on the current default path, the regression test observes the invariant directly: it monkeypatches the truncation step into a selector that reorders, then asserts `filtered_*` follows `*_context`. Verified it can fail — reverting `operate.py` to `main` turns the five order- and reporting-sensitive cases red (three reordering cases, the unresolvable-record case, and the stage-3 quota case) while the four order-insensitive ones (prefix, subset, duplicates, missing tokenizer) stay green, confirming the no-op claim for the shipped path.

Per `AGENTS.md`, no source line may cite a GitHub issue number (enforced by `tests/test_docstring_budget.py::test_no_source_file_cites_a_github_issue`), so the docstring and comments name the invariant instead of pointing at this issue.

https://claude.ai/code/session_012a9NmyeP5mBwZmcvXAG6tV

### DIFF
--- a/lightrag/operate.py
+++ b/lightrag/operate.py
@@ -5444,6 +5444,17 @@ async def _apply_token_truncation(
 ) -> dict[str, Any]:
     """
     Apply token-based truncation to entities and relations for LLM efficiency.
+
+    Returns ``entities_context`` / ``relations_context`` (the records handed to
+    the prompt) plus ``filtered_entities`` / ``filtered_relations`` (the matching
+    original records handed to chunk selection).
+
+    Ordering rule for any selector added here: the ``*_context`` lists are this
+    stage's output and the only importance ranking downstream sees. The
+    ``filtered_*`` lists MUST follow that same order -- stage 3 attributes a
+    shared chunk to the earlier-positioned record and allocates chunk quota by
+    list position -- so a selector that reorders must not leave the
+    ``filtered_*`` lists in stage-1 retrieval order.
     """
     tokenizer = global_config.get("tokenizer")
     if not tokenizer:
@@ -5484,8 +5495,11 @@ async def _apply_token_truncation(
         if isinstance(created_at, (int, float)):
             created_at = time.strftime("%Y-%m-%d %H:%M:%S", time.localtime(created_at))
 
-        # Store mapping from entity name to original data
-        entity_id_to_original[entity_name] = entity
+        # Store mapping from entity name to original data.
+        # First occurrence wins: the filtered rebuild below resolves records
+        # through this map, and duplicate names must keep the record that was
+        # retrieved first.
+        entity_id_to_original.setdefault(entity_name, entity)
 
         entities_context.append(
             {
@@ -5510,9 +5524,10 @@ async def _apply_token_truncation(
         else:
             entity1, entity2 = relation.get("src_id"), relation.get("tgt_id")
 
-        # Store mapping from relation pair to original data
+        # Store mapping from relation pair to original data.
+        # First occurrence wins, for the same reason as entities above.
         relation_key = (entity1, entity2)
-        relation_id_to_original[relation_key] = relation
+        relation_id_to_original.setdefault(relation_key, relation)
 
         relations_context.append(
             {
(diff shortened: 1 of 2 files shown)


danielaskdd:
@codex review

PR #3921, opened 2026-09-11
fix(kg): make the self-loop rule a data invariant, and batch the degree lookups

danielaskdd:
## Description

`BaseGraphStorage` never wrote down how a self-loop is counted, so each backend derived it from the shape of its own query and they disagreed — including *within* a single class (`node_degree` vs `node_degrees_batch` on Mongo and OpenSearch).

This PR settles it as a **data invariant, not a query-time rule**: a self-loop must not be in the store, so its degree is not contracted. Backends keep their cheapest natural degree query and may disagree on an edge no admissible graph holds.

`_reject_self_loop_relation` (`lightrag/utils_graph.py`) already states why when it refuses to create one:

> a self-loop carries no connectivity for graph retrieval

Degree is exactly that connectivity measure — it feeds relation `rank`, BFS truncation priority and `get_popular_labels` — so an edge reaching nothing has no meaningful degree to report. A store still holding one holds **invariant-violating data to be removed, not data to be ranked**.

### Query-time filtering was implemented across all seven backends, measured, and reverted

An earlier revision of this PR excluded self-loops inside every degree query. That is recorded in `base.py` so nobody re-derives it, and the reason is cost:

| | cost of filtering at the query |
|---|---|
| **Neo4j `node_degrees_batch`** | drops off the O(1) `GetDegree` plan (read from the relationship-chain header) onto an expand **per requested node** — and BFS levels plus `edge_degrees_batch` both route through it |
| **OpenSearch `get_popular_labels` / `get_knowledge_graph('*')`** | their query bodies carry **no other filter**, so a script clause becomes **one script evaluation per document in the edge index** — OpenSearch cannot compare two fields with a term query |
| pgtable / AGE / networkx / memgraph / Mongo | predicted near-zero, but **unprovable without EXPLAIN on a live server**, which is not available here |

Both expensive paths sit on the retrieval hot path. Paying on every query to describe data that must not exist is the wrong trade, and a half-rule ("filter where it's free") would re-derive wrong on the next backend while reintroducing scalar≠batch divergence exactly where it can't be fixed cheaply.

**Every degree query in every backend is now byte-identical to `main`.** `git diff origin/main -- lightrag/kg/` matches zero lines containing `count_documents`, `client.count`, `$group`, `aggs`, `terms`, `COUNT(` or `graph.degree`. Five of the seven backend files — `networkx_impl`, `neo4j_impl`, `postgres_impl`, `pgtable_impl`, `memgraph_impl` — do not appear in the diff at all. Zero added per-query cost anywhere.

## Enforcement moved to the boundary

`tools/migrate_graph_storage.py` was the one in-tree path that copied a self-loop into a target graph — `detect_reciprocal_pairs` skips them and `upsert_edges_batch` writes them through. It now drops them:

- `drop_self_loop_edges` runs at the **READ boundary**, so the refusal checks, the written-id derivation and the verification comparison all see one edge set — verification still passes, on what was actually written.
- `dropped_self_loop_edges` is in the run report, so `verified` is never read as "the graph came across whole" when rows were deliberately left behind.
- Dropped rather than refused, unlike reciprocals and duplicate node ids: a self-loop does not make the migration *ambiguous*, it is data the target must not hold.

A cleanup script for graphs that already contain self-loops (written before the custom-KG guard existed, or via direct storage-API use) is a **tracked follow-up**. It will go through `adelete_by_relation`, which does not reject self-loops, so VDB relation rows and chunk tracking are cleaned with them.

## Also fixes a cost defect this PR's review surfaced (predates the PR)

Mongo and OpenSearch never overrode `edge_degrees_batch`, so they inherited the base default: `edge_degree` per pair = two awaited `node_degree` calls. `_find_most_related_edges_from_entities` hands it the **whole incident-edge set** of the top entities (truncation happens after), so a `top_k=60` query over an average degree of 20 is ~1200 edges.

```
edge pairs = 1200   distinct node ids = 1260

main:   Mongo / OpenSearch  2400 SERIAL count round trips
after:  Mongo       0 counts,  2 aggregations
        OpenSearch  0 counts,  1 aggregation search
```

On Mongo every one of those counts was a **collection scan** — see below.

That an aggregation search is more expensive than a count (the trade `test_node_degree_uses_count_api` pins for the scalar) does not decide this: it runs once instead of thousands of times, and the scalar leaves the retrieval hot path entirely.

## Kept from the earlier revisions

- **`get_nodes_edges_batch` self-loop de-duplication** on Mongo and OpenSearch — one edge, one tuple, the guard `pgtable_impl` already carried. Free, and independent of the degree rule.
- **The listing rule on `get_node_edges`** — a self-loop is listed, never hidden. Not in tension with "the store must not hold one": this is *how a store that does is repaired*. Entity deletion, rename, merge and `_purge_kg_contributions` all resolve relation rows through these methods, so an edge hidden here becomes an orphan relation vector row behind a deleted entity.
- **`node_degrees_batch` answers every requested id** (0, not a missing key) on Mongo and OpenSearch, matching `pgtable`.

## Documentation consistency sweep

Changing what `BaseGraphStorage.node_degree` promises invalidated statements elsewhere, so the surrounding docstrings were swept rather than left to drift.

**Contradicted the new contract** — each asserted that a self-loop's degree IS defined, as two:
- `pgtable_impl.get_popular_labels` and the BFS candidate ranking both read *"self-loops count twice … consistent with node_degree"*. They now say there is no guard **because** the degree is not contracted, and that adding the predicate was measured and rejected.
- Two pgtable test comments used "self-loops count twice" as the justification for an unrelated assertion.

**Stated in the wrong file:**
- *"`node_degrees_batch` answers every requested id (0, not a missing key)"* was written on `node_degree`, where someone implementing the batch would never find it — and it is exactly the trap an aggregation-backed override falls into (no bucket, no key). Moved to the method it binds.
- `edge_degrees_batch`'s default said only *"override for better performance"*. It is two **awaited** `node_degree` calls per pair against the whole incident-edge set of the top entities — thousands of sequential requests per query, which is the defect this PR had to fix on two backends. It now says so and names the eight-line override.

**Incomplete after the boundary change:**
- `_reject_self_loop_relation` claimed manual creation was *"the only way such an edge can reach the graph"*. It now names every ingress — extraction, merge, custom-KG, the migration tool — and points at the store-level invariant.
- The migration tool's module docstring did not mention the self-loop drop: the one thing it deliberately does not copy verbatim.
- `detect_reciprocal_pairs`' own self-loop skip is now unreachable from the pipeline (the drop runs first); noted so it does not read as tolerance.

**Checked and left alone:** `docs/` states no self-loop or degree rule anywhere, so no design document could drift. Mongo's and OpenSearch's `node_degree` docstrings describing the scalar/batch divergence as unobservable are accurate under the new contract.

## Separate finding — not fixed here

**Mongo's edge collection has no index on `source_node_id` / `target_node_id`.** `MongoGraphStorage` creates only the `(edge_lo, edge_hi)` unique index, and `get_or_create_collection` creates none. So `node_degree`'s `$or` count and both `node_degrees_batch` aggregations are **collection scans**. A docstring introduced in this PR's first revision claimed *"the `source_node_id` index narrows it"* — that was false and is gone. Pre-existing and worth its own PR; the `edge_degrees_batch` override above reduces how often it is paid but does not fix it.

`MemgraphStorage` overrides neither `node_degrees_batch` nor `edge_degrees_batch` — the same N+1, untouched by this PR. Also follow-up.

## Changes Made

14 files, +774 / -39. The five backend implementations that only ever carried the reverted filter are absent from the diff.

**Core**
- `lightrag/base.py` — `node_degree`: a self-loop must not be in the store, so its degree is not contracted; records that query-time filtering was implemented and reverted, and what it cost. `get_node_edges`: the listing rule, reframed as the repair path rather than the opposite of a degree rule.
- `lightrag/kg/mongo_impl.py` — `edge_degrees_batch` override; `node_degrees_batch` seeds zeros; `node_degree` docstring no longer claims an endpoint index that does not exist.
- `lightrag/kg/opensearch_impl.py` — `edge_degrees_batch` override; `node_degrees_batch` seeds zeros.
- `lightrag/tools/migrate_graph_storage.py` — `drop_self_loop_edges` at the read boundary + `dropped_self_loop_edges` in the run report.
- `lightrag/kg/pgtable_impl.py`, `lightrag/utils_graph.py` — docstring/comment alignment only, no behavior change (see below).

**Tests**
- new: `test_mongo_edge_degrees_batch.py`, `test_opensearch_edge_degrees_batch.py` — one `node_degrees_batch` call over the distinct ids, no `node_degree` calls, and the zero-seeding.
- new: `test_mongo_self_loop_contract.py`, `test_opensearch_self_loop_contract.py` — the LISTING rule only (a self-loop appears once and is never hidden).
- `test_migrate_graph_storage.py` / `_networkx.py` — the drop helper, and the real round trip asserting the self-loop is dropped, counted, and verification still passes on the filtered set.
- `test_pgtable_graph_storage.py` / `test_pgtable_smoke.py` — the three pre-existing self-loop degree tests renamed and reworded (see Additional Notes).

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Additional Notes

**Verification.** Full suite at the final commit (`base.py` is cross-cutting): **8539 passed, 244 skipped**. `tests/kg` + `tests/tools`: 2526 passed. `ruff`: clean, formatted with the pinned pre-commit hooks.

Every new assertion was verified to fail when its subject is removed: 3 per backend for the `edge_degrees_batch` overrides and zero-seeding, 3 for the migration drop (plus the real round-trip test).

**Tests removed or reframed.** The degree-on-self-loop contract tests added in the earlier revisions are gone — they pinned a case the contract no longer defines, and re-adding a "counts twice" or "counts zero" assertion would re-establish exactly what this PR declines to promise. `pgtable`'s three pre-existing self-loop degree tests are kept with the opposite meaning: renamed and reworded to pin the **query shape**, so they now guard against someone re-adding the predicate that was measured and rejected.

**History.** Five commits, the first three pointing in different directions as the rule was worked out: adopt NetworkX's "a self-loop counts twice" → invert it to "exclude at the query" → move it to the boundary once the query-time cost was measured → reframe the pgtable tests that still asserted the abandoned contract → sweep the surrounding docstrings. **Squash at merge** — the intermediate states are not worth keeping in `main`'s history.


danielaskdd:
Superseded by #3929.

Same tree, byte-identical to this PR's head (`git diff` between the two branches is empty) — re-opened as a **single commit** so the diff is reviewable on its own.

This PR's five commits point in different directions, because the rule was worked out in the open: adopt NetworkX's "a self-loop counts twice" → invert it to "exclude at the query" → move it to the boundary once the query-time cost was measured (Neo4j loses its O(1) `GetDegree` plan; OpenSearch's `get_popular_labels` and `get_knowledge_graph('*')` become a script evaluation per document in the edge index) → reframe the tests that still asserted the abandoned contract → sweep the surrounding docstrings. Useful as a record, not as something to review commit by commit.

Closing in favour of #3929.

PR #3927, opened 2026-09-12
perf(opensearch): refresh the KV index only where a search reads it

### LINKED ISSUE #3923: OpenSearchKVStorage refreshes the index even when the flush wrote nothing
## Summary

`OpenSearchKVStorage.index_done_callback` refreshes the index **unconditionally**, including when the flush it just ran wrote nothing. Every commit of an idle KV namespace therefore costs one `indices.refresh` round trip that publishes none of this process's data.

```python
async def index_done_callback(self) -> None:
    await self._flush_pending_kv_ops()      # returns at once on an empty buffer
    if not self._index_ready:
        return
    await self.client.indices.refresh(index=self._index_name)   # runs regardless
```

## Why it is worth tracking

The two lines are unrelated mechanisms. The flush is about the **client-side buffer** — it is why this backend's commit does real work at all, where PG / Redis / Mongo write immediately (`index_done_callback` is `pass`) and `JsonKVStorage` writes its file only when dirty. The refresh is about **Lucene's near-real-time visibility** — a row an acknowledged bulk call has written is not yet in a searchable segment. Removing the buffer entirely would not remove the need for the refresh, and an empty buffer does not remove the refresh today.

The cost became visible on the query path. PR #3916 orders the LLM cache commit behind its `text_chunks` references at every commit site, including `_query_done`, because a cache commit publishes the whole namespace and would otherwise put extract rows on disk ahead of the chunk rows that name them. That ordering is not negotiable. But on OpenSearch it now adds a refresh of `text_chunks` — usually the largest index in the deployment — to **every query**, in every process, even though a query-only worker never has a chunk operation buffered.

## Why skipping the refresh looks safe, and what still has to be checked

With an empty buffer this process published nothing, so the refresh can only make **another** process's unrefreshed writes visible. LightRAG sets no `index.refresh_interval`, so the server's own periodic refresh converges on its own.

The readers it could affect are narrower than they look. Every runtime read of `OpenSearchKVStorage` goes through `mget` by `_id` — `get_by_id`, `get_by_id_strict`, `get_by_ids`, `filter_keys` — and a GET by id is real time, unaffected by refresh. Only two readers go through search, and neither is on the request path:

| Search-based reader | API | Who calls it |
|---|---|---|
| `is_empty` | `client.count` | `_migrate_chunk_tracking_storage`'s startup check for `entity_chunks` / `relation_chunks` |
| `_iter_raw_docs` | PIT + `search_after` | the offline tools only: `clean_llm_query_cache`, `migrate_llm_cache`, `rebuild_vdb` |

Where this codebase needs read-after-write through search it does **not** rely on this refresh: `OpenSearchDocStatusStorage` passes `refresh="wait_for"` on the write itself, in all four of its write paths (`upsert`, `update_doc_status_fields`, `repair_source_conflict`, `delete`), because the scheduling readers must see the row immediately. The strong guarantee is per-write and at the write site; the KV commit's refresh is the weak one.

Open questions a change must answer, both outside the request path:

- the startup `is_empty` could read a concurrently-written namespace as empty and run a migration that was not needed;
- an offline tool started immediately after an ingestion process could scan a view that is briefly short of rows.

## Proposed change

Have `_flush_pending_kv_ops` report whether it actually wrote anything, and refresh only then. Same question applies to `OpenSearchVectorDBStorage`, which should be checked rather than assumed to match.

Not included, and deliberately: no change to the commit ordering in PR #3916, and no change to `refresh="wait_for"` on doc-status writes.

## Before implementing

Measure. The saving is one round trip against an index that usually has nothing new to refresh, so the benefit has to be established at realistic scale and concurrency rather than argued from the code. A change here alters the visibility semantics of a storage every caller shares, which is why it is not folded into #3916.

## Background

The mechanism, the reader inventory and the accepted residue are written up in `docs/design/PurgeRecoveryContract.md`, section *What the OpenSearch KV refresh actually protects* (added in #3916).

danielaskdd:
## Description

`OpenSearchKVStorage.index_done_callback` refreshed the index **unconditionally**, including when the flush it had just run wrote nothing:

```python
await self._flush_pending_kv_ops()      # returns at once on an empty buffer
if not self._index_ready:
    return
await self.client.indices.refresh(index=self._index_name)   # runs regardless
```

The two lines are unrelated mechanisms. The flush is about the **client-side buffer**; the refresh is about **Lucene's near-real-time visibility**. An empty buffer means this process published nothing, and the refresh still cost a broadcast round trip across every shard.

The cost became visible on the query path. #3916 orders the LLM cache commit behind its `text_chunks` references at every commit site, `_query_done` included, because a cache commit publishes the whole namespace and would otherwise put extract rows on disk ahead of the chunk rows that name them. That ordering is not negotiable — but it put a refresh of `text_chunks`, usually the largest index in the deployment, on **every query, in every process**, even though a query-only worker never has a chunk operation buffered.

This PR moves the refresh to the two readers that actually need it, and makes the commit-side refresh something the storage **owes** rather than something it always does.

## Related Issues

Closes #3923.

Follows #3916, which introduced the per-query cost and recorded it as an accepted residue in the purge contract. That residue is closed here; the ordering itself is untouched, as #3923 requires.

## Changes Made

**`lightrag/kg/opensearch_impl.py`**

- **`_refresh_for_search()`** — new helper, called by the only two readers that go through `search` rather than `mget` by `_id`:

  | Search-based reader | API | Who calls it |
  |---|---|---|
  | `is_empty` | `client.count` | `_migrate_chunk_tracking_storage`'s startup check for `entity_chunks` / `relation_chunks` |
  | `_iter_raw_docs` | PIT + `search_after` | the offline tools only: `clean_llm_query_cache`, `migrate_llm_cache`, `rebuild_vdb` |

  Every runtime read — `get_by_id`, `get_by_id_strict`, `get_by_ids`, `filter_keys` — is an `mget` by `_id`, which consults the translog and is real time. This is the same shape `OpenSearchDocStatusStorage` already uses with `refresh="wait_for"` on all four of its write paths: the guarantee belongs at the site that needs it, not spread over every commit. `_iter_raw_docs` refreshes **before opening the PIT**, since the point-in-time freezes the view for the whole scan.

  Best effort by contract: a failure is logged and swallowed, leaving the caller the pre-refresh view — which is what every one of these readers got unconditionally before.

- **`index_done_callback` keeps a refresh debt.** A flush that issues writes bumps `_write_generation`; a refresh that returned records the generation it covered in `_refreshed_generation`. A commit of an idle namespace touches neither and is a full no-op.

- **A missing index defers the debt; it does not settle it.** `_mark_index_missing` only flips `_index_ready`, and `index_done_callback`'s missing-index branch returns with the counters apart. Most of that helper's dozen call sites are read paths, and a read cannot know whether a streaming bulk is still landing rows — `get_by_id` releases `_flush_lock` before its mget, so its `index_not_found` can be concurrent with a bulk that auto-creates the index and goes on writing. A read settling the debt on the commit path's behalf would strand exactly those rows. While the index is gone nothing is refreshed at all (the readiness check short-circuits ahead of the debt check, and an empty-buffer flush returns before `_ensure_index_ready`), so a dropped namespace stays free; the debt is paid once, by the first commit that finds the index back. `initialize()` likewise does not settle it — an extra refresh is harmless and a dropped one is not.

**`docs/design/PurgeRecoveryContract.md`** — the residue row for the per-query refresh is marked closed, the *What the OpenSearch KV refresh actually protects* section gains a *Where the refresh lives now* subsection with the two rules below, the missing-index deferral rule, and what the change trades, and the quarantine residue row records the narrowing described under *Additional Notes*.

### Why a debt, and not "did this flush write?"

The shape proposed in the issue — have the flush report whether it wrote anything, and refresh only then — drops the one thing the unconditional refresh was really providing.

`index_done_callback` **re-raises** a refresh failure that is not a missing-index error. At that point the bulk has landed but the rows are not in a searchable segment. Today the *next* commit repairs it, including an idle one, precisely because the refresh is unconditional. Gate on what the current call's flush wrote and the repair never happens: the buffer is empty on every later commit, so nothing retries it, and the rows stay outside every search-based reader until the server's own periodic refresh catches up.

"LightRAG sets no `index.refresh_interval`, so the server converges anyway" is an argument for skipping an **idle** refresh that had nothing local to publish. It does not extend to discarding a **failed** refresh over rows this process did write.

Two further rules follow, and neither is expressible with a single dirty bool:

- **Sample the generation before the refresh, record it after.** A flush landing mid-refresh may or may not be covered by it, so the debt has to stand.
- **Bump before the bulk, not from its success count.** `async_bulk` streams, so a transport error can raise with earlier chunks already written — and the aborting pipeline then discards the buffer that would have resent them. Over-counting costs one refresh, which is what this storage did unconditionally until now.

### Out of scope, deliberately

- **No change to the commit ordering from #3916**, and none to `refresh="wait_for"` on doc-status writes.
- **`OpenSearchVectorDBStorage` is untouched.** #3923 asks for it to be "checked rather than assumed to match". It does not match: its reads go through k-NN `search` and genuinely depend on the commit-side refresh. Same code shape, different reader inventory, different risk.

## How it works / what's broken

- **Nothing changes for any other backend.** PG / Redis / Mongo write on `upsert` and their `index_done_callback` is a no-op; `JsonKVStorage` writes its file only when dirty. This is an OpenSearch-only change.
- **One behavior change worth naming:** a commit no longer publishes *another* process's unrefreshed writes as a side effect. That incidental guarantee is not lost where it mattered — a reader-side refresh publishes the **index**, not this process's share of it, so both search-based readers now see every process's writes, which the commit-side refresh only did by luck of timing. What is given up is the sub-second convergence a busy namespace got for free everywhere else, which the server's own periodic refresh covers.
- **Cost:** one refresh per `is_empty` (twice at startup) and one per `_iter_raw_docs` scan (offline tools), against one saved per idle commit — which on the query path is one per query, per process.

### On measuring

#3923 asks for a measurement before changing the visibility semantics of a shared storage, and that is the right bar for a pure round-trip saving. This change is not only that: moving the refresh to the readers is a **correctness improvement independent of the saving**, since `_iter_raw_docs` previously opened its PIT against whatever the last commit happened to have refreshed, and `is_empty` decided a startup migration from the same. The saving on the query path rides along. I have not benchmarked the round trip at scale — what a benchmark would settle is how much the query path gains, not whether the change is right, and the two search-based readers are strictly better off either way.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (`PurgeRecoveryContract.md`: residue row closed, new *Where the refresh lives now* subsection, quarantine row narrowed)
- [x] Unit tests added

## Additional Notes

**A narrowed residue, for free.** The purge contract records a residue where a **transient refresh failure** quarantines extract cache rows even though every reference is durable — the flush has already popped its successful operations by the time the refresh runs, so the failure surfaces as a failed chunk commit. A commit that owes no refresh cannot reach the refresh at all, which takes the entire query path and every commit of an idle namespace out of that. Narrowed, not closed: a commit that did flush writes still refreshes, and can still fail this way.

**New tests** — 11 in `TestKVRefreshGating` (`tests/kg/opensearch_impl/test_opensearch_storage.py`): a commit that wrote still refreshes; an idle commit refreshes nothing and flushes nothing; a failed refresh is retried by the next **idle** commit and not paid twice; a write landing mid-refresh is refreshed again; a bulk that raised still owes a refresh after the aborting pipeline discards its buffer; a missing index defers the debt rather than settling it, and pays it exactly once after a recreate; a read observing `index_not_found` mid-bulk does not settle it either; `is_empty` refreshes before counting, survives a failed refresh, and reports empty when the refresh finds no index; `_iter_raw_docs` refreshes before opening the PIT.

**Fix-proof.** Each test was run against the shape it rules out, and the failures are behavioural rather than missing symbols:

| Wrong shape | Tests that go red |
|---|---|
| unpatched `main` (unconditional refresh) | 5 — idle commit, failed-refresh retry, both `is_empty` cases, `_iter_raw_docs` |
| gate on what **this call's** flush wrote (the issue's literal proposal) | 2 — failed-refresh retry, mid-refresh write |
| a single dirty bool cleared from live state | 1 — mid-refresh write |
| bump from the bulk's **reported success** | 1 — flush that raised |
| settle the debt whenever the index goes missing | 2 — missing-index deferral, read racing an in-flight bulk |

The remaining cases are regression guards that pass before and after, as intended. The first rewrite of the flush-that-raised test passed against all four shapes and was rewritten to route through `drop_pending_index_ops`, which is what makes it discriminate.

**Test subsets run** (mirroring the changed module, plus the callers of the two readers): `tests/kg/opensearch_impl` — **412 passed, 20 skipped**; `tests/pipeline` + `tests/tools` + `tests/test_storage_migrations_strict_read.py` — **1111 passed**; `tests/test_docstring_budget.py` — **16 passed**. `ruff check lightrag/ tests/kg/opensearch_impl/` and `ruff format --check` clean.

https://claude.ai/code/session_01Mc83Ph8mL5gFXGDJduQm4x

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mc83Ph8mL5gFXGDJduQm4x)_


danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3932, opened 2026-09-13
fix(storage): tell a lost reference from a durable one when a chunk commit raises

### LINKED ISSUE #3924: Typed flush exception: tell a lost reference from a durable one when a chunk commit raises
## Summary

When a `text_chunks` commit raises, the caller cannot tell whether the raise could have lost a reference. It therefore assumes the worst and quarantines **every** buffered `extract` LLM-cache row in the process — rows whose only reference is the chunk's `llm_cache_list`, and which would be unreachable if that reference never landed. On the two paths where the reference is in fact durable, that discard is pure waste: completed LLM calls are paid for again on the next run.

## The three raises, and why they are indistinguishable today

`OpenSearchKVStorage.index_done_callback` is `flush` then `refresh`, and all three failure modes reach the caller as a plain `Exception`:

| Raise | Reference state | Right response |
|---|---|---|
| `_flush_pending_kv_ops` → `OpenSearchException` from the bulk | buffers intact, replays on the next flush | **defer** |
| `_flush_pending_kv_ops` → `RuntimeError`, permanent non-retryable 4xx | operation already removed, reference gone for good | **quarantine** |
| `indices.refresh` → transient `OpenSearchException` | flush already succeeded and popped every successful op; **every reference is durable** | **do nothing** |

Inspecting the buffer afterwards does not separate them: a permanent bulk failure empties the buffer exactly as a successful flush does, so `has_pending_index_ops()` reads the same in rows 2 and 3. One flush can also carry permanent and retryable failures at once, so even a non-empty buffer does not prove no reference was lost.

Quarantine is the only answer that preserves the governing invariant (*a purge must never delete something that carries attribution and leave the objects behind* — see `AGENTS.md`), which is why it is what the code does. The cost is money and time, never data.

## Proposed change

Surface a typed exception through `lightrag/base.py` that answers exactly one question: **could this raise have lost a durable reference?** The caller then quarantines only when the answer is yes, and defers otherwise. One primitive closes all three rows.

Points to settle in the design:

- Where the type lives (`lightrag/exceptions.py` alongside `IndexFlushError`, or `base.py` next to the storage contract) and what every non-overriding backend implies by raising a plain exception — the safe default must stay "assume a reference may be lost".
- A single flush can mix permanent and retryable per-item failures. The type has to describe *that flush*, not one operation, so the honest answer for a mixed flush is still "yes, a reference may be lost".
- `OpenSearchVectorDBStorage` has the same flush/refresh shape and should be checked rather than assumed to match.

## Not in scope

No change to the commit ordering, to the fence, or to the quarantine's fail-safe direction. Until the typed exception exists, over-discarding is the correct behavior and must not be relaxed by guessing from buffer state.

## Background

The residue is recorded in `docs/design/PurgeRecoveryContract.md`, section *LLM extraction cache reachability* → *Accepted residue*, row "A **transient** chunk-commit failure quarantining the extract rows instead of deferring them". The unconditional refresh that makes row 3 reachable is tracked separately in #3923.

danielaskdd:
## Description

When a `text_chunks` commit raises, the caller cannot tell whether the raise could have lost a reference, so it assumes the worst and quarantines **every** buffered `extract` LLM-cache row in the process. On the paths where the reference is in fact intact, that discard is pure waste: completed LLM calls are paid for again on the next run.

This PR gives the caller the one fact it was missing, as a typed exception surfaced through the storage contract: **could this raise have lost a durable reference?** The caller quarantines only when the answer is yes.

### Why the buffer cannot answer it

`OpenSearchKVStorage._flush_pending_kv_ops` removes a permanently-failed operation *before* raising, so a permanent bulk failure empties the buffer exactly as a successful flush does — `has_pending_index_ops()` reads the same in both. A single flush can also carry permanent and retryable failures at once, so even a non-empty buffer does not prove no reference was lost. Quarantine was therefore the only answer that preserved the governing invariant.

### The three raises

| Raise | Reference state | Response now |
|---|---|---|
| `_flush_pending_kv_ops` → `OpenSearchException` from the bulk | buffers intact, replays on the next flush | defer |
| `_flush_pending_kv_ops` → `RuntimeError`, permanent non-retryable 4xx | operation already removed, reference gone for good | quarantine (unchanged) |
| `indices.refresh` → transient `OpenSearchException` | flush already popped every successful op; every reference is durable | defer |

The fail-safe direction is preserved by making the safe answer the *absence* of a claim: a backend that raises a plain exception keeps today's behavior with no code of its own.

## Related Issues

Closes #3924. Follows #3923 / #3927, which narrowed row 3 by taking the query path and idle-namespace commits off the refresh; this closes what remained of it.

## Changes Made

- **`lightrag/exceptions.py`** — `ReferencesIntactFlushError` and the `flush_may_have_lost_reference` predicate. The predicate unwraps the `IndexFlushError` transport wrapper and **nothing else**: an exception chained onto an intact one has made a claim of its own (`OpenSearchKVStorage.finalize` chains a "these writes have been lost" `RuntimeError` onto exactly such a flush).
- **`lightrag/base.py`** — the contract stated on `StorageNameSpace.index_done_callback`, including that the proof must hold for the whole flush rather than one operation.
- **`lightrag/kg/opensearch_impl.py`** — `OpenSearchReferencesIntactError` subclasses both the typed contract and `OpenSearchException`, so every existing `except OpenSearchException` keeps catching these. Raised on the KV and vector bulk-transport and refresh paths. The permanent-4xx `RuntimeError`, a mixed flush, and `_ensure_index_ready` deliberately keep the default — `_ensure_index_ready` can surface a permanent mapping rejection no later flush clears.
- **`lightrag/lightrag.py` / `lightrag/pipeline.py`** — `_flush_storages`, `_query_done` and the stage epilogue skip the quarantine when the raise dropped nothing. The epilogue also trusts its own retry again: what makes a retry untrustworthy is a buffer drained behind its back, and this is how the partial cache of a document that failed midway survives a transport blip.
- **`docs/design/PurgeRecoveryContract.md`** — new section *What a failed commit says about references*; the accepted-residue row rewritten to what remains.

### Not in scope

No change to the commit ordering, the fence, or the quarantine's fail-safe direction.

`_discard_pending_index_ops` and `_commit_cache_pair_before_finalize` needed no change: both already read their own retry or the buffer, and both become truthful once the flag is no longer set over a flush that dropped nothing. That is also where **"intact" is deliberately not collapsed into "durable"** — `_discard_pending_index_ops` drops the buffer next, so a merely-retained reference is as good as a lost one there, and it must not read an optimistic answer. A third state would be readable only at that one site.

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Additional Notes

**Tests.** 10 new tests across three files: the backend's answers (`tests/kg/opensearch_impl/test_opensearch_storage.py::TestFlushReferenceLossTyping` — bulk, refresh, permanent, mixed, `OpenSearchException` compatibility, missing index, vector), the predicate and the caller gating (`tests/pipeline/test_insert_done_error_handling.py`), and the query path (`tests/pipeline/test_cache_reference_flush_order.py`). Each was falsified: reverting the five gates one at a time turns the corresponding test red, and the gate was then restored to green.

- `tests/pipeline` + `tests/kg/opensearch_impl`: **1351 passed, 20 skipped**.
- Full suite: **8416 passed**, 143 failed — the same 143 fail on a clean checkout of `main` in this environment (missing spaCy models and tokenizer data), so none are from this change.

**On the vector storage.** `OpenSearchVectorDBStorage` was checked rather than assumed to match: it has the identical flush/refresh shape and answers the same way. No caller branches on a vector commit's answer today — the reference carrier is a KV namespace — but the contract belongs to the storage layer rather than to one namespace.

https://claude.ai/code/session_01Kse2aYctKM8ubQeC1rfQka


danielaskdd:
@codex review

PR #3936, opened 2026-09-13
fix(docx): reject out-of-range w:ilvl before it drives an unbounded loop

Shizoqua:
## Summary
Fix a denial-of-service in DOCX numbering resolution.

## Problem
w:ilvl (legally 0-8) is parsed with no bound check and can drive an unbounded loop, hanging the parser on one crafted paragraph.

## Changes
Reject any ilvl outside 0-8 before it reaches the loop.

## Testing
tests/parser/docx/test_numbering_resolver.py: 92 passed. Ruff and pre-commit passed.

## Compatibility and Risk
Low risk. Only out-of-range ilvl values are affected.

## Related Issue
No related issue.


danielaskdd:
@codex review

PR #3940, opened 2026-09-14
fix(docx): record Roman and ideographTraditional counts past their domain

### LINKED ISSUE #3934: numbering_resolver: lowerRoman/upperRoman degrade silently above 3999, ideographTraditional wraps at 10
## Summary

`NumberingResolver.FORMAT_CONVERTERS` has two converters whose domain limits are not registered in `LIMITED_DOMAIN_FORMATS`, so exceeding them produces a wrong label with no warning and no `numbering_out_of_range_formats` counter. `ideographTraditional` is the worse of the two: it does not degrade, it *wraps*, which is the exact failure mode #3886 / #3902 fixed for `lowerLetter` / `upperLetter`.

Found while reviewing #3886 against #3905 — this is pre-existing on `main` and unrelated to either PR, so it was deliberately left out of #3905's scope.

## Where

All in `lightrag/parser/docx/numbering_resolver.py`.

### 1. `lowerRoman` / `upperRoman` degrade silently above 3999

`_to_roman` (line 577) already does the right thing at the edge:

```python
if n <= 0 or n >= 4000:
    return str(n)
```

but neither format appears in `LIMITED_DOMAIN_FORMATS` (line 64), so `_note_out_of_range` never fires:

```
lowerRoman n=3999  -> 'mmmcmxcix'   out_of_range=-  warn=-
lowerRoman n=4000  -> '4000'        out_of_range=-  warn=-
```

The fallback itself is correct and legible; only the diagnostic is missing. This is the same contract the `LIMITED_DOMAIN_FORMATS` docstring already states for the counting families.

### 2. `ideographTraditional` wraps silently, and has no `n <= 0` guard

Line 48:

```python
"ideographTraditional": lambda n: "甲乙丙丁戊己庚辛壬癸"[(n - 1) % 10],
```

The Heavenly Stems are a 10-element sequence, and `% 10` makes item 11 restart at 甲. Python's modulo on negatives also keeps zero and negative counters inside the string, so they too yield a plausible-looking stem — every other converter (`_to_roman`, `_to_alpha`, `_to_chinese`, `_to_ideograph_digital`) returns `str(n)` for `n <= 0`:

```
ideographTraditional n= 0 -> '癸'      # expected a decimal fallback
ideographTraditional n=-1 -> '壬'      # expected a decimal fallback
ideographTraditional n=10 -> '癸'
ideographTraditional n=11 -> '甲'      # wraps; item 1 and item 11 are indistinguishable
ideographTraditional n=21 -> '甲'
```

This is structurally identical to the `chr(ord("a") + (n - 1) % 26)` wrap that #3886 reported and #3905 fixed, and it is the failure the `LIMITED_DOMAIN_FORMATS` docstring singles out as hardest to notice: a wrong-yet-plausible label rather than an obviously wrong one.

For reference, `ideographDigital` is fine — it is digit-by-digit and correct for any positive `n`, so it needs no entry.

## Expected behaviour

Both formats should follow the contract already written in the `LIMITED_DOMAIN_FORMATS` docstring: outside the converter's real domain, degrade to the decimal string **and** record it, so the event is findable in a real document instead of silently producing a wrong ordinal.

## Suggested work

These are independent; the first is mechanical and the second needs a spec decision, so they should not block each other.

**1. Roman (small, no open questions)**

- Add `"lowerRoman": 3999` and `"upperRoman": 3999` to `LIMITED_DOMAIN_FORMATS`.
- Extend the docstring to note that the Roman cutoff is the converter's own, not a rendering choice.
- Test: count 3999 renders `mmmcmxcix` with no warning; 4000 renders `4000`, sets `out_of_range_formats`, and increments `numbering_out_of_range_formats`.

**2. `ideographTraditional` (needs [MS-DOCX] verification first)**

- Decide what Word actually renders for item 11 and beyond. I did not verify this against [MS-DOCX] "numFmt Extensions" and it should not be guessed — the file's existing policy is that an unimplemented rendering is left unimplemented *and made findable*, rather than approximated.
- If the real sequence is implemented, drop the modulo. If it is not, the minimum correct behaviour is to stop wrapping: add an `n <= 0 or n > 10` guard returning `str(n)`, and register `"ideographTraditional": 10` so the fallback is reported. Either way item 11 must stop being indistinguishable from item 1.
- Test: items 1–10 render 甲…癸; item 11 no longer returns 甲; `0` and `-1` fall back to decimal.

Run `./scripts/test.sh tests/parser/docx`.

## Impact

Low frequency, but silent when it happens. A list with more than 10 天干 items or more than 3999 Roman items is rare; the concern is that when one does appear, the resolver emits a label that looks valid, `smart_heading` classifies it with the wrong ordinal, and nothing in the warnings surfaces it. Registering the domains is what makes such a document show up as evidence.

RizgarOzan:
## Description

`NumberingResolver` had two converters whose domain limits were not in `LIMITED_DOMAIN_FORMATS`, so exceeding them produced a label with no out-of-range diagnostic:

- `lowerRoman` / `upperRoman`: `_to_roman` already falls back to decimal at 4000, but the fallback was silent (no `numbering_out_of_range_formats` count, no warning).
- `ideographTraditional`: `(n - 1) % 10` wrapped, so item 11 rendered 甲 like item 1, and 0 / negative counts rendered 癸 / 壬 — the same failure mode fixed for the letter formats.

## Related Issues

Closes #3934

## Changes Made

- Register `lowerRoman` / `upperRoman` at 3999 and `ideographTraditional` at 10 in `LIMITED_DOMAIN_FORMATS` (comment updated to say why).
- Replace the wrapping lambda with `_to_heavenly_stem`, which renders 甲…癸 for 1–10 and falls back to `str(n)` outside that range, matching how `_to_alpha` handles letters. Word's rendering past 癸 is not guessed.
- Tests in `tests/parser/docx/test_numbering_resolver.py`: Roman 3999 vs 4000 (nothing recorded vs recorded), the ten stems, and -1 / 0 / 11 / 21 falling back instead of wrapping.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

How I tested:
- `./scripts/test.sh tests/parser/docx/test_numbering_resolver.py` → 99 passed. With only the source change reverted, the new tests fail (6 failed, 93 passed: both Roman cases and the four fallback cases).
- `./scripts/test.sh tests/parser/docx` → 903 passed, 45 skipped.
- `pre-commit run --files` on both files → all hooks passed.


danielaskdd:
@codex review

PR #3941, opened 2026-09-14
fix(chunker): correct custom-chunker drift diagnostics and selection ergonomics

danielaskdd:
## Description

Follow-up to the server-startup selection of installed third-party chunkers (#3869), from a post-merge review of that implementation. Four defects and rough edges, each with a regression test.

Nothing here changes what `C` means, how it is admitted, or how it falls back. The six-argument callback contract, `is_builtin_chunker`, the `chunk_method` strings and the fallback-warning cadence are all untouched.

## Related Issues

Follows up #3869 / #3868. No issue is closed by this PR.

## Changes Made

**1. The drift WARNING was manufactured out of nothing.** An explicit `F`/`R`/`V`/`P` attempt never consults `chunking_func`, yet it overwrote `doc_status.metadata.custom_chunker` with a null observation. A `C` → `F` → `C` sequence therefore warned

```
Custom chunker identity changed for doc_id ...: None@None -> 'acme'@'1'
```

on a deployment whose configuration had never changed. The record is now written only by attempts that actually route through the callback, so carry-over preserves the last real observation — the only baseline a comparison can be made against. Read the field together with `chunk_method`, which says what actually ran.

**2. Unselectable names were offered as choices.** A duplicated name stays in the registry (last wins, plus the ERROR line naming both origins), but `resolve_chunker` refuses it. Listing it under "registered names" sent the operator to a value guaranteed to abort startup. New `selectable_chunker_names()` backs both the unknown-name message and the custom-chunking 422; `registered_chunker_names()` keeps its full list for the startup summary, which legitimately shows what each provider registered.

**3. An invalid `CUSTOM_CHUNKER` printed a raw traceback**, while every other operator misconfiguration in `create_app` renders a boxed notice (the `smart_heading` block right below it). Boxed now, and still re-raised — presentation only; an unselectable chunker must never reach a running server.

**4. `doc_status_reset_metadata` carried `custom_chunker` through a bespoke branch** bolted next to the directive whitelist. Named as `_DOC_STATUS_METADATA_OBSERVATION_KEYS`, so the next field of this kind is one entry rather than a second special case. An observation is neither a directive nor a per-attempt result: nothing may be gated on it, but the next attempt compares against it, so it has to survive the reset.

**5. No-selector documents never reported chunker drift.** The drift WARNING was gated on `doc_process_opts.chunking == "C"`, inside a block only `C` and no-selector attempts reach — both consult `chunking_func`, only one compared. That left the *quieter* path silent. On `C`, drift has other witnesses: `chunk_method` separates `custom_chunking_func` from `custom_chunking_fallback_fixed_token`, the INFO line separates `C(custom)` from `C(fallback F)`, and a fallback WARNING fires when the callback is gone. On the no-selector path none of that exists — `chunk_method` is `legacy_chunking_func` whether the built-in or a plugin ran, the INFO line is `F(legacy)` either way, and there is no fallback warning because the built-in *is* the default. So `custom_chunker` is the only evidence such a document's chunking changed between attempts, and it was written silently.

No-selector is also where the accidental case lives: `C` is an explicit per-document opt-in, while no-selector is the default population, so a `CUSTOM_CHUNKER` change is a global flip and a FAILED retry or restart-recovery can re-run a handful of documents under a configuration their siblings never saw.

The gate is now `isinstance(previous_identity, dict)` under the existing `uses_custom_callback` block. FR-3.4 stays safe **by construction**: warning requires a recorded dict → requires a prior write → whose bootstrap requires `chunker_identity()` non-None → requires a `ChunkerBinding`, so a deployment that never selected a plugin has no key and cannot warn. FR-6.4's per-document/per-attempt/never-collapsed cadence is reused unchanged. Write-side gate, dispatch, `chunk_method` and the identity's non-authoritative status are all untouched — one log line, nothing else.

**Documentation** now states the consequence that is easiest to miss: selecting a chunker changes DEFAULT ingestion, not just `C`. No-selector inserts then report `chunk_method=legacy_chunking_func` and lose source-span sidecar backfill eligibility, exactly as a constructor-supplied callback does. The startup INFO line, `env.example` and both authoring guides say so. The guides also note that the `executor_safe` pool is single-worker and shared with the built-in strategies — the flag buys the event loop back, not parallelism.

## Validation

Python 3.11, `ruff check .` clean.

Affected subsets (`tests/chunker`, `tests/pipeline`, `tests/api`, plus `tests/test_docstring_budget.py`), compared against the same command on unmodified `main` at `61f1245`:

```
main  61f1245 : 99 failed, 2505 passed, 14 skipped
this branch   : 99 failed, 2508 passed, 14 skipped
```

The two failure sets are **identical** (diffed, zero entries on either side), so there are no regressions; the +3 are exactly the new tests. Those 99 are pre-existing failures of this sandbox, not of the tree: the `tiktoken` encoding download is proxy-blocked (403), and the API prefix / auth / CORS suites fail here for the same environmental reason. I have not diagnosed them; CI is the place that will.

Fix 5 landed after that comparison, so it was verified separately on macOS (network available, so the proxy-blocked failures above do not occur): `tests/pipeline tests/chunker tests/api/test_custom_chunker_startup.py tests/test_docstring_budget.py` — **1352 passed, 0 failed**, `ruff check .` clean. Its three new `"!"` parametrizations were observed failing on `c57a6e2` (the tree without fix 5). Its claim about FAILED retry / restart-recovery was checked against the real reset path, not only the test harness: both reset sites (`_run_exclusive_failed_reset` for `/reprocess_failed`, and the batch consistency validator) funnel through `_build_pending_reset_update` → `doc_status_reset_metadata`, which is the function the test asserts preserves `custom_chunker`.

Each new test was observed **failing against the unfixed tree** and passing after the fix:

- `tests/pipeline/test_custom_chunking_selector.py::test_explicit_builtin_attempt_does_not_manufacture_drift`
- `tests/chunker/test_registry.py::test_unselectable_duplicate_is_not_offered_as_a_choice`
- `tests/api/test_custom_chunker_startup.py::test_invalid_selection_prints_a_boxed_notice_and_still_aborts`
- `tests/api/test_custom_chunker_startup.py::test_ingress_error_names_config_and_selectable_choices` (renamed, extended)
- `tests/pipeline/test_custom_chunking_selector.py::test_registered_identity_survives_reset_and_warns_once_on_drift` — parametrized over `["C!", "!"]` rather than duplicated, keeping the `("acme", "1")` no-drift case on both paths so it pins the absence of false positives too. The three new `"!"` combinations were observed failing before fix 5.

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed
- [x] Documentation updated
- [x] Unit tests added

## Additional Notes

One review finding is deliberately **not** addressed here:

- **The custom-chunking 422 lists installed plugin names to any API caller.** FR-6.5 explicitly asks for that list, so narrowing it would contradict an accepted requirement. Flagging it only because it is deployment information leaving an authenticated-but-not-admin surface.

FR-6.3's wording says "a persisted `C` document", which is narrower than what fix 5 now implements. That is a wording fix to make on #3868, not a behavior disagreement — see the reasoning under fix 5.

AI-assisted implementation and local verification; no independent human review is claimed.

https://claude.ai/code/session_019JchtE3Rtsrb3tN3Et4o5U

---
_Generated by [Claude Code](https://claude.ai/code/session_019JchtE3Rtsrb3tN3Et4o5U)_


danielaskdd:
@codex review

PR #3943, opened 2026-09-14
fix(milvus): extend event-loop offloading to delete_entity_relation and id lookups

### LINKED ISSUE #3942: [Bug]: Milvus event-loop offloading gaps in `delete_entity_relation`, `get_by_ids`, and `get_by_id`

### Describe the bug

`MilvusVectorDBStorage.delete_entity_relation`, `_query_rows_by_ids` (the
helper shared by `get_by_ids` and `get_vectors_by_ids`), and `get_by_id`
still call the synchronous `pymilvus` SDK directly inside `async def`
methods — [`lightrag/kg/milvus_impl.py:2768`](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/kg/milvus_impl.py#L2768) (`delete_entity_relation`, calling `self._client.query`/`self._client.delete` at [L2817](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/kg/milvus_impl.py#L2817) and [L2833](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/kg/milvus_impl.py#L2833)), [`lightrag/kg/milvus_impl.py:2902`](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/kg/milvus_impl.py#L2902) (`_query_rows_by_ids`, calling `self._client.query` at [L2935](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/kg/milvus_impl.py#L2935)), and [`lightrag/kg/milvus_impl.py:2962`](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/kg/milvus_impl.py#L2962) (`get_by_id`, calling `self._client.query` at [L2980](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/kg/milvus_impl.py#L2980)). Each also calls `self._ensure_collection_loaded()` synchronously first, which itself blocks on a cold `load_collection` poll.

`pymilvus` is a synchronous SDK (blocking gRPC): calling it directly from an
`async def` freezes the *entire* event loop for the round trip — every
other in-flight request on that loop stalls until the call returns.

**This is exactly the failure mode #3862 already fixed — just not here.**
PR [#3862](https://github.com/HKUDS/LightRAG/pull/3862) added a
dedicated, bounded thread pool for Milvus calls —
[`get_milvus_executor()`](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/kg/milvus_impl.py#L82) /
[`run_in_milvus_executor()`](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/kg/milvus_impl.py#L95) —
and converted every blocking call in `query()` and
`_flush_pending_vector_ops()` to use it. The PR's own description
self-scoped explicitly: *"Scoped to the hot read/write path;
startup/migration/admin methods in this file share the same limitation and
are left for a follow-up."* 

However, hese three methods are not "startup/migration/admin" in the sense that
phrase describes (they don't run once at process start or only during a
schema migration) — but they were also never individually named or
reviewed against by #3862. They sit in a gap the PR's own scoping sentence
doesn't quite cover, and are the only methods on the class, besides the
genuinely-once-per-run admin/migration cluster, that still make a
synchronous `pymilvus` call. Two of the three carry real, ordinary
production traffic:

- **`delete_entity_relation`** is called from `adelete_by_entity()` in
  [`lightrag/utils_graph.py:669`](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/utils_graph.py#L669), which is wired directly to the
  entity-delete API route ([`lightrag/api/routers/graph_routes.py:895`](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/api/routers/graph_routes.py#L895)).
  Every entity deletion through the WebUI or API — an ordinary
  knowledge-graph-maintenance action — fires this method once,
  synchronously, blocking the loop for the round trip.
- **`_query_rows_by_ids`** is exercised through `get_vectors_by_ids`, its
  sibling public method (same helper, same blocking pattern), from
  `pick_by_vector_similarity()` at
  [`lightrag/utils.py:6484`](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/utils.py#L6484) — the
  vector-similarity chunk-selection step of the retrieval pipeline. This
  runs on every query using that selection strategy.
- **`get_by_id`**'s only callers are `get_entity_info`/`get_relation_info`
  (gated behind an `include_vector_data` flag whose docstring at
  [`lightrag/lightrag.py:8071`](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/lightrag.py#L8071) states *"No LightRAG code path sets
  this — it is kept for backward compatibility only"*) and the
  `aexport_data()` admin/CSV-export helper ([`lightrag/utils.py:4746`](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/utils.py#L4746)/[`4785`](https://github.com/HKUDS/LightRAG/blob/824f33d360a63168381f6555d6139008d7f56288/lightrag/utils.py#L4785)).
  Included here for class-wide consistency rather than because it carries
  hot traffic today.

**Proposed fix**: convert the same four blocking call sites (the two
`_ensure_collection_loaded()` calls plus `query()`/`delete()` in
`delete_entity_relation`, the `_ensure_collection_loaded()` + `query()` in
`_query_rows_by_ids`, and the `_ensure_collection_loaded()` + `query()` in
`get_by_id`) to `await run_in_milvus_executor(...)`, mirroring exactly what
#3862 already did for `query()` and `_flush_pending_vector_ops()` — no new
mechanism, just extending the established one to the call sites it hasn't
reached yet. A PR implementing this is ready and will reference this issue.

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### LightRAG Config Used

# Paste your config here

### Logs and screenshots

_No response_

### Additional Information

- LightRAG Version:  1.5.8 (evaluated against `upstream/main` @ `824f33d360a63168381f6555d6139008d7f56288`)
- Operating System:
- Python Version:
- Related Issues:
- A PR implementing the fix described above, with new regression tests, is ready and will reference this issue

AdalbertoDam:
## Description

Extends the bounded-executor offloading PR #3862 introduced for
`query()`/`_flush_pending_vector_ops()` to the three remaining
`MilvusVectorDBStorage` methods that still call the synchronous `pymilvus`
SDK directly inside `async def`: `delete_entity_relation`,
`_query_rows_by_ids` (shared by `get_by_ids` and `get_vectors_by_ids`), and
`get_by_id`. Each of these still froze the event loop for the duration of
its Milvus round trip, exactly the failure mode #3862 fixed elsewhere in
the same file — #3862's own description scoped itself to "the hot
read/write path" and deferred "startup/migration/admin methods" as a
follow-up, but these three are neither part of the fixed path nor
startup/migration/admin; they simply weren't individually named or
reviewed against in that PR. `delete_entity_relation` and
`get_vectors_by_ids` (via `_query_rows_by_ids`) carry ordinary production
traffic — every entity deletion and every vector-similarity chunk-selection
query goes through them respectively.
## Related Issues

Fixes #3942;

Extends #3862 ("fix(milvus): run search/upsert/delete off the event loop
thread") — this PR reuses `run_in_milvus_executor`/`get_milvus_executor`
exactly as introduced there; no new mechanism is added.
## Changes Made

- `lightrag/kg/milvus_impl.py`: wrap the remaining synchronous `pymilvus`
  calls (`_ensure_collection_loaded`, `self._client.query`,
  `self._client.delete`) in `delete_entity_relation`, `_query_rows_by_ids`,
  and `get_by_id` with `await run_in_milvus_executor(...)`, mirroring the
  call shape #3862 already established for `query()` and
  `_flush_pending_vector_ops()`. `get_by_ids` and `get_vectors_by_ids` are
  fixed as a side effect of fixing their shared `_query_rows_by_ids`
  helper.
- `tests/kg/milvus_impl/test_milvus_off_event_loop.py`: add coverage for
  the three newly-offloaded methods, mirroring the existing
  thread-identity/thread-name assertions used for `query()` and the flush
  path — plus two negative-case tests (`delete_entity_relation` with no
  server client; `get_by_id` served entirely from the pending-upsert
  buffer) confirming the buffer-only short-circuit paths, which never
  touch the client, are unaffected.

No behavior change beyond concurrency — these methods return the same
results as before, just off the event loop thread.
## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Additional Notes


danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3951, opened 2026-09-15
fix(parser): retain Markdown table headers in chunk content

### LINKED ISSUE #3950: Native Markdown parser omits pipe-table header rows from block content (paragraph-semantic chunks lose column headers)
## Summary

For Markdown pipe tables, the native parser writes the header row into the table sidecar
but omits it from the block content that chunking consumes. Paragraph-semantic chunks
therefore contain a table's data rows with no column names.

## Reproduce (parser only — no server, LLM, or embeddings)

```bash
printf '| component_header_contract | limit_header_contract |\n| --- | --- |\n| Widget | 10 |\n| Sensor | 20 |\n' > contract.md
NATIVE_MD_IMAGE_DOWNLOAD_ENABLED=false python -m lightrag.parser.cli contract.md --engine native --preview 0
```

Requires the parser-side dependencies from the `api` extra (`python-docx`, `defusedxml`,
`langchain-text-splitters`), since `lightrag.parser.cli` imports the DOCX parser unconditionally.

## Actual

`contract.md.parsed/contract.blocks.jsonl` — the content block:

```json
{"type": "content", "format": "plain_text",
 "content": "<table id=\"tb-...-0001\" format=\"json\">[[\"Widget\", \"10\"], [\"Sensor\", \"20\"]]</table>"}
```

`contract.md.parsed/contract.tables.json` — the same table, with the header kept out of band:

```json
{"dimension": [2, 2],
 "format": "json",
 "content": "[[\"Widget\", \"10\"], [\"Sensor\", \"20\"]]",
 "table_header": "[[\"component_header_contract\", \"limit_header_contract\"]]"}
```

The P chunker loads `table_header` from the sibling `.tables.json` sidecar, but its
JSON header-recovery path assumes those rows already prefix the inline table body.
Native Markdown emits body rows only. Unsplit tables pass through unchanged, and
split tables fail the `json_pinned` recovery precondition, leaving the original
column headers absent from the resulting chunks.

Source: [sidecar loading](https://github.com/HKUDS/LightRAG/blob/v1.5.7/lightrag/chunker/paragraph_semantic.py#L319-L348),
[JSON header precondition](https://github.com/HKUDS/LightRAG/blob/v1.5.7/lightrag/chunker/paragraph_semantic.py#L777-L789),
[header-injection gate](https://github.com/HKUDS/LightRAG/blob/v1.5.7/lightrag/chunker/paragraph_semantic.py#L932-L948),
and [unsplit-table path](https://github.com/HKUDS/LightRAG/blob/v1.5.7/lightrag/chunker/paragraph_semantic.py#L1003-L1009).

## Expected

The header cells should reach the text that chunking consumes — inline as a row, repeated
per row, or prepended once per table. The adjacent case already behaves this way: HTML
tables keep `<th>` cells inline in block content and survive chunking intact.

## Impact

Retrieved context and graph extraction see only `Widget 10 / Sensor 20` with no indication
of what the columns mean. Markdown pipe tables are the common table form in notes, so this
silently degrades table-heavy documents.

## Scope

- Affected: 1.5.4, 1.5.5, 1.5.6, 1.5.7, and current `main`.
- 1.5.0–1.5.3 are not affected because they have no native Markdown engine at all
  (`error: engine 'native' does not support .md files (supported: docx)`); 1.5.4 is the
  earliest affected release.
- HTML tables (`<table><tr><th>...`) keep their headers in both the sidecar and block
  content, so the loss is specific to the Markdown pipe-table path.

Lu1sDV:
## Description

Native Markdown extracts pipe-table headers correctly, but its IR builder currently emits only data rows in `IRTable.rows`. The header survives in `table_header` metadata, not in the inline table content.

The P chunker already reads that metadata. Unsplit tables pass through unchanged, while split JSON header recovery expects the original header to prefix the inline rows. Emitting the complete table grid at the native Markdown boundary satisfies both paths without changing the chunker.

## Related Issues

Fixes #3950

## Changes Made

- Include the extracted header rows once in the native Markdown table grid, retaining the existing `table_header` metadata.
- Calculate dimensions from the complete grid, including the header, as the sidecar format already specifies.
- Add offline parser → sidecar → P chunker regressions for header-only, unsplit, and split tables. Assert intact table markup, one header per table fragment, preserved data rows in order, token caps, and consistent sidecar content/dimensions.
- Update the existing Markdown IR regression to the corrected table contract.

Production changes are confined to `lightrag/parser/markdown/ir_builder.py`. No chunker, shared writer, DOCX, HTML, API, or configuration changes.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation checked: no changes needed; this restores the existing [sidecar dimension contract](https://github.com/HKUDS/LightRAG/blob/main/docs/LightRAGSidecarFormat.md#L254) and [header-recovery contract](https://github.com/HKUDS/LightRAG/blob/main/docs/ParagraphSemanticChunking.md#L161-L170).
- [x] Unit/regression tests added and updated

## Additional Notes

Validation used Python 3.12.13 and dependencies installed from `uv.lock` with `uv sync --frozen --extra api --extra pytest`.

- Failing-first check: all four targeted cases failed before the fix and passed afterward.
- Affected suites: **1,590 passed, 44 skipped**. The optional `en_core_web_sm` and `zh_core_web_sm` language models were not installed.
  ```sh
  python -m pytest tests/parser/markdown tests/parser/docx tests/parser/test_parse_native_lightrag_e2e.py tests/sidecar tests/chunker -q --tb=short
  ```
- Actual native parser CLI + real Tiktoken/P chunker smoke: six synthetic cases passed (Markdown header-only/unsplit/split, HTML, DOCX unsplit/split), with preserved headers and data rows, no duplicate JSON header rows, and token caps respected. No LLM calls or external services.
- Pinned pre-commit hooks passed on all three changed files.

Existing parsed/indexed documents must be reparsed and re-ingested to obtain corrected content; this change does not rewrite cached sidecars or stored chunks.


danielaskdd:
@codex review

kokhlo (commented):
Verified independently on `main@202be78` (Python 3.14, deps via `uv sync --extra api --extra test`, fully offline).

**Repro confirmed.** The CLI output matches the report: the content block carries a body-only grid (`[["Widget","10"],["Sensor","20"]]`) while `tables.json` holds the header out-of-band.

**Root cause agrees with the PR's diagnosis.** `extract.py:_consume_pipe_table` (L639-650) returns the header separately from body rows, and the pipe branch of `_build_table` emitted only the body into `IRTable.rows`. The P chunker's `json_pinned` precondition (`paragraph_semantic.py` L785, `rows[:len(header_rows)] == header_rows`) is designed for header-prefixed bodies — native DOCX (`parser/docx/ir_builder.py` L147-165) emits exactly that shape, so native Markdown was the outlier. Restoring the full grid at the parser boundary (rather than special-casing the chunker) is the right layer.

**Checks with this diff applied:**
- Failing-first: stashing only `ir_builder.py` fails all four new/updated cases (header-only / unsplit / split / IR contract); reapplying turns them green.
- `tests/parser/markdown/test_ir_builder.py + test_e2e_sidecar.py`: 14 passed.
- Wider offline baseline: `tests/parser` 1109 passed / 44 skipped; `tests/chunker` + `tests/sidecar` 423 passed — identical to clean `main`, no regressions.

**Two edge probes beyond the PR's matrix (char tokenizer, offline):**
- `chunk_size=60` forcing the whole-table recursive degrade (the `paragraph_semantic.py` L861 warning path): header text survives in the degraded chunks and all 10 data rows are preserved across the 10 fragments.
- A data row identical to the header row: the pinned path holds the header out of the split body, so injection adds exactly one header and the duplicate-looking data row survives intact — no spurious dedup, matching the design comment at L780-786.

Sidecar consistency also checked: `tables.json` `content` and `dimension` now include the header row, which matches the sidecar format doc's dimension contract, and no other `md_tables`/`table_header` consumer assumes a body-only grid.

PR #3952, opened 2026-09-15
fix(llm): report LaTeX escape damage followed by a non-ASCII character

### LINKED ISSUE #3949: LaTeX escape damage goes unreported when a word character follows (CJK, digits, underscore)
### Summary

`_WS_LATEX_SUSPECT_PATTERN` in `lightrag/utils.py` detects whitespace-class LaTeX escape damage (`"\tau"` emitted with a single backslash inside JSON decodes to a tab plus `au`). Outside dollar math the damage is deliberately not repaired — a tab there is legitimate whitespace — so this warning is the **only** trace that a LaTeX command was destroyed.

The pattern ends each residue with `\b`. Python's `re` uses Unicode word semantics, so a CJK ideograph **is** a word character and no boundary exists between `au` and `为`. Damage followed by a word character is therefore never reported.

```python
>>> from lightrag.utils import _WS_LATEX_SUSPECT_PATTERN as P
>>> bool(P.search("阈值\tau为0.5"))     # CJK follows  -> silent
False
>>> bool(P.search("阈值\tau。"))         # punctuation  -> reported
True
>>> bool(P.search("the \tau2 value"))   # digit        -> silent
False
>>> bool(P.search("col\text_id header"))  # underscore -> silent
False
```

### Why the `\b` is there

It suppresses the case where a residue happens to start an English word: `"col1\tauthor list"` must not be reported. That is a real constraint and the reason the in-math variant cannot simply be reused — `_WS_LATEX_MATH_PATTERN` relaxes the guard to `(?![A-Za-z])`, which is correct inside a confirmed math span but would report `col\text_id` in ordinary tab-separated prose.

### Impact

Warning-only; no data is rewritten and no behavior changes. But the information loss is real:

- A description is persisted as `阈值au为0.5` with nothing in the logs, so the occurrence cannot be located, counted, or attributed to a model or document batch after the fact.
- The exposure is skewed toward Chinese corpora: models frequently write inline formulas in Chinese prose without `$` delimiters (`阈值 \tau 为 0.5`), which is exactly the shape that is both unrepairable (no math span) and unreported (CJK follows).
- It masks prompt regressions. `entity_extraction_json_system_prompt` and the three multimodal prompts already require double-escaped backslashes; if a model stops honoring that, Chinese corpora would produce no signal at all.

### Proposed fix

Widen the prose guard to "word boundary **or** a following non-ASCII character", leaving digits and underscores alone:

```python
r"(?:\b|(?=[^\x00-\x7F]))"
```

- Covers the CJK case, which carries almost all of the real-world value.
- Keeps `col\text_id` and `\tau2` silent — in tab-separated data those can be genuine values, so the noise is not worth it.
- A Latin fragment pressed directly against a CJK character with no space is essentially only produced by this damage, so the false-positive risk is low.

Risk ceiling is log noise: this path only emits a WARNING and never rewrites text.

### Scope

Not part of #3946, which changed *what gets repaired* (the dollar-math pairing rules). This is the independent question of *what gets reported* outside math. Touches one regex plus tests for three shapes: CJK follows, punctuation/space follows (must still report), and legitimate tab-separated data (`col1\tauthor`, `col\text_id` — must stay silent).

Relevant code: `lightrag/utils.py` (`_WS_LATEX_RESIDUES` / `_WS_LATEX_SUSPECT_PATTERN`), tests in `tests/llm/test_vlm_json_escape_repair.py`.

danielaskdd:
## Description

`_WS_LATEX_SUSPECT_PATTERN` in `lightrag/utils.py` detects whitespace-class LaTeX escape damage — `"\tau"` emitted with a single backslash inside JSON decodes to a tab plus `au`, destroying the command. Outside dollar math the damage is deliberately **not** repaired (a tab there is legitimate whitespace), so this WARNING is the *only* trace that a LaTeX command was ever there.

The pattern ended each residue with `\b`. Python's `re` uses Unicode word semantics, so a CJK ideograph **is** a word character and no boundary exists between `au` and `为`. Damage followed by a word character was therefore never reported:

```python
>>> bool(P.search("阈值\tau为0.5"))    # CJK follows  -> silent
False
>>> bool(P.search("阈值\tau。"))       # punctuation  -> reported
True
```

This is skewed toward Chinese corpora, where models routinely write inline formulas into prose without `$` delimiters (`阈值 \tau 为 0.5`) — exactly the shape that is both unrepairable (no math span) and unreported. A description gets persisted as `阈值au为0.5` with nothing in the logs, so the occurrence cannot be located, counted, or attributed after the fact. It also masks prompt regressions: if a model stops honoring the double-escaped-backslash requirement in `entity_extraction_json_system_prompt` or the multimodal prompts, Chinese corpora would produce no signal at all.

The fix widens the prose guard to **word boundary OR a following non-ASCII character**:

```python
_WS_LATEX_RESIDUES.replace("__END__", r"(?:\b|(?=[^\x00-\x7F]))")
```

A Latin fragment pressed directly against a non-ASCII character with no space between them is essentially only produced by this damage, so the false-positive risk is low. ASCII word characters stay excluded deliberately: in tab-separated data `col<tab>ext_id` and `<tab>au2` are plausible values, and the log noise is not worth them.

This is warning-only — no text is rewritten and no behavior changes. The risk ceiling is one extra WARNING line.

## Related Issues

Closes #3949.

Independent of #3946, which changed *what gets repaired* (the dollar-math pairing rules). This is the separate question of *what gets reported* outside math.

## Changes Made

- **`lightrag/utils.py`** — one regex: `_WS_LATEX_SUSPECT_PATTERN`'s trailing guard becomes `(?:\b|(?=[^\x00-\x7F]))`. `_WS_LATEX_MATH_PATTERN` is untouched; the comment above the pair is updated so it still describes the guards accurately.
- **`docs/design/LatexEscapeRepairContract.md`** — *Residue whitelist and its two boundaries* now documents the second alternative and why it is admitted. The "Accepted gap" paragraph is narrowed from "any word character" to "any **ASCII** word character", which is what remains silent.
- **`tests/llm/test_vlm_json_escape_repair.py`** — four new tests covering the three shapes the issue names plus the invariant the widening rests on:
  - `test_damage_followed_by_a_non_ascii_character_is_reported` — CJK and a Greek letter following the residue; also asserts the text is *not* rewritten.
  - `test_word_boundary_shapes_are_still_reported` — ideographic full stop, full-width comma, space, end of string, ASCII punctuation.
  - `test_ascii_word_characters_after_the_residue_stay_silent` — `col1<tab>author`, `col<tab>ext_id`, `<tab>au2`, `row<CR>ightmost` must produce no log record.
  - `test_in_math_pattern_stays_a_superset_of_the_prose_pattern` — sweeps every residue in the whitelist against nine followers, pinning the contract's reason a repaired span can never re-trigger the prose warning. Widening the prose guard is only safe while this holds.

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed

## Additional Notes

**Verification.** `tests/llm/test_vlm_json_escape_repair.py` + `tests/test_docstring_budget.py` + `tests/utils/`: 97 and 460 passed. The new tests were proven to fail against the old `\b` guard before the fix was applied — 5 of the 6 non-ASCII cases go red, the sixth (full-width comma) was already caught by `\b` and has been moved into the "still reported" group so each parametrize stays honest.

The full suite was also run (`6823 passed`); the 274 failures and 73 collection errors in it are pre-existing in this sandbox (missing optional backends — `qdrant_client`, `asyncpg` — and undownloadable spaCy models), confirmed by reproducing the same failures on a stashed tree. None are in the touched modules.

https://claude.ai/code/session_015tLhc5LrAbsXcGMjsEsaUm

---
_Generated by [Claude Code](https://claude.ai/code/session_015tLhc5LrAbsXcGMjsEsaUm)_


danielaskdd:
@codex review

PR #3955, opened 2026-09-15
fix(opensearch): floor legacy edge weight to evidence count

Shizoqua:
## Summary
Fix a relation weight contract violation in the OpenSearch legacy edge dedup migration.

## Problem
A duplicate edge with a missing weight can raise the evidence count without raising the weight, leaving it too low or unset.

## Changes
Floor the merged weight to the evidence count and guard against sum overflow, mirroring the existing MongoDB fix.

## Testing
tests/kg/opensearch_impl: 398 passed. Ruff and pre-commit passed.

## Compatibility and Risk
Low risk. Only affects the one-time legacy edge dedup migration on OpenSearch-backed deployments.

## Related Issue
No related issue.


Shizoqua:
Already fixed in upstream, closing as duplicate.

PR #3958, opened 2026-09-15
feat(storage): add Hologres backend for KV, Vector, Graph, and Doc Status

TimothyDing:
## Description

Add Alibaba Cloud Hologres as a first-class storage backend for all four
LightRAG storage roles. The implementation ships two graph storage options,
capability probing, resumable schema management, and a comprehensive test
suite with a ≥90% per-module coverage gate.

### New backends (`lightrag/kg/hologres/`)

| LightRAG role | Implementation | Notes |
|---|---|---|
| KV storage | `HologresKVStorage` | JSONB documents, workspace-scoped |
| Vector storage | `HologresVectorStorage` | Similarity queries via pgvector-compatible semantics |
| Graph storage | `HologresGraphStorage` (default) | Two shared tables (`graph_nodes` / `graph_edges`), no AGE dependency |
| Graph storage (optional) | `HologresAGEGraphStorage` | Embedded Apache AGE engine; one AGE graph per workspace, falls back to two-table backend if probe fails |
| Doc status storage | `HologresDocStatusStorage` | Aligned with live Hologres SQL contracts |

### Key features

- **Resumable schema management** — migrations are tracked in a per-schema
  ledger so interrupted deployments recover on restart.
- **Capability probes** — startup probes detect server support (including the
  AGE contract) before any data is written; unsupported features fall back or
  fail fast with clear errors.
- **Stream COPY bulk upserts** — bulk writes route through Hologres' native
  COPY channel for throughput.
- **Columnar JSONB + row-column hybrid orientation** — table layouts tuned for
  Hologres' storage engine.
- **Workspace isolation** — all roles scope data by LightRAG's `WORKSPACE`;
  KV/Vector/DocStatus use schema-prefixed tables, graph uses workspace
  partitions or per-workspace AGE graphs.
- **Backend registry integration** — all implementations are registered in
  `lightrag/kg/__init__.py` and selectable via standard
  `LIGHTRAG_*_STORAGE` env vars.

### Configuration

All settings use the `HOLOGRES_` prefix (host, port, user, password, database,
schema). See `env.example` and the storage guide for the full reference.

```ini
HOLOGRES_HOST=example.hologres.aliyuncs.com
HOLOGRES_PORT=80
HOLOGRES_USER=your_username
HOLOGRES_PASSWORD=your_password
HOLOGRES_DATABASE=your_database
HOLOGRES_SCHEMA=lightrag

LIGHTRAG_KV_STORAGE=HologresKVStorage
LIGHTRAG_VECTOR_STORAGE=HologresVectorStorage
LIGHTRAG_GRAPH_STORAGE=HologresGraphStorage          # or HologresAGEGraphStorage
LIGHTRAG_DOC_STATUS_STORAGE=HologresDocStatusStorage
```

### Dependencies

- Requires Hologres **5.0+** and `asyncpg` (already included in the
  `offline-storage` extra).

## Related Docs

- `docs/HologresStorageGuide.md` — full operator/maintainer guide (English)
- `docs/HologresStorageGuide-zh.md` — Chinese translation
- `docs/ProgramingWithCore.md` / `env.example` / `README.md` — updated to
  reference the new backend

## Testing

- **Offline unit tests** (`tests/kg/hologres_impl/`, no Hologres instance
  required): ~20 test files covering client, capabilities, config, KV, vector,
  graph (both backends), doc status, schema lifecycle, retry policy, SQL
  guard, workspace selection, and backend registration.
- **Integration gating**: `tests/conftest.py` deliberately changes the global
  `pytest_collection_modifyitems` behavior so Hologres-specific live tests are
  selected by `--run-hologres-live`; existing `--run-integration` behavior
  remains equivalent for the rest of the suite. This backend-specific gate is
  intentionally separate from the general `--run-integration` flag.
- **Live integration tests**: gated behind `--run-hologres-live`; exercise real
  capability probes, schema recovery, CRUD, similarity queries, graph behavior,
  and an AGE cross-chunk hydration regression against a real instance.
- **Coverage gate**: every module under `lightrag/kg/hologres/` maintains
  **≥90% statement coverage**.

Run offline tests:

```bash
./scripts/test.sh tests/kg/hologres_impl
```

## Checklist

- [x] New storage backend registered in `lightrag/kg/__init__.py`
- [x] All four storage roles implemented and covered by tests
- [x] Offline tests pass; live-only tests skipped by default
- [x] Documentation added (EN + zh) and cross-referenced
- [x] No changes to existing backend behavior


TimothyDing:
@danielaskdd Hello, Could you help me to review it~Thanks

lightrag/kg/hologres/graph_age.py
```diff
@@ -0,0 +1,1034 @@
...
+                    return
+            except BaseException:
+                if owns_client:
+                    try:
+                        await actual_client.close()
+                    except Exception:
+                        pass
+                raise
+
+            # Probe failed: fall back to the two-table implementation on the
```
danielaskdd: **Silent fallback on any non-`PASSED` probe result.**

The fallback branch treats every non-`PASSED` outcome the same way — including a transient connection loss or a temporary permission problem — and delegates to the two-table backend without emitting a log line or raising.

The consequence is not just a slower path: an existing AGE graph becomes invisible, and subsequent writes land in a *different physical store* (`graph_nodes` / `graph_edges`). A deployment that has been running on AGE can silently split its graph across two backends after one restart that happened to hit a blip, and nothing in the process output says so.

`docs/HologresStorageGuide.md` tells operators to "monitor whether the startup probe passed", but there is nothing to monitor — no log, no metric, no status attribute.

Suggested minimum: log at WARNING with the probe's failure reason whenever the fallback is taken, and consider distinguishing a definitive "AGE not available on this server" (fall back quietly at INFO) from an indeterminate probe failure (WARNING, or fail fast when `HologresAGEGraphStorage` was explicitly requested).

---

lightrag/kg/hologres/kv.py
```diff
@@ -0,0 +1,520 @@
...
+                    not isinstance(identifier, str)
+                    or identifier not in requested
+                    or identifier in chunk_existing
+                ):
+                    raise HologresKVError("Hologres KV filter response is corrupt")
+                chunk_existing.add(identifier)
+            existing.update(chunk_existing)
+        return keys - existing
+
+    async def upsert(self, data: dict[str, dict[str, Any]]) -> None:
```
danielaskdd, line 379: **`upsert` never writes the storage-managed `create_time` / `update_time`.**

`BaseKVStorage.upsert` makes these two fields the storage's responsibility, not the caller's — the payload arrives without them and the backend is expected to stamp them (insert sets both, update refreshes `update_time` only). This implementation passes the payload straight through to JSONB, so neither field is ever written.

The visible breakage is in the LLM extraction cache: `operate.py` orders cached extraction results chronologically by `create_time`. Every Hologres-backed row reads back with `create_time` absent (defaulting to `0`), so the ordering collapses to whatever order the backend happens to return — the cached extraction results are no longer replayed in the order they were produced.

`PGKVStorage` stamps both fields server-side in the upsert statement; worth mirroring that here so the value is generated by the database rather than the client.

---

lightrag/kg/hologres/kv.py
```diff
@@ -0,0 +1,520 @@
...
+            "ON CONFLICT (workspace, namespace, id) DO UPDATE SET "
+            "payload = EXCLUDED.payload, "
+            "updated_at = CURRENT_TIMESTAMP"
+        )
+        use_stream_copy = bool(getattr(client, "stream_copy_available", False))
+        for payload_json in chunks:
+            payload_dict = json.loads(payload_json)
+            keys = list(payload_dict.keys())
+            if is_full_docs:
+                existing = await self.get_by_ids(keys)
```
danielaskdd: **Client-side read-modify-write on `full_docs` loses concurrent writes.**

The protected-field merge reads the existing rows here, merges in Python, then writes the full payload back. The base contract forbids exactly this: the merge has to happen server-side, in the same statement as the write, because there is no lock between the read and the write.

The data loss is wider than the protected keys alone. Because the whole JSONB payload is replaced rather than patched, two concurrent upserts of the same `doc_id` mean the later writer overwrites `content` with the version it read before the earlier writer committed — the earlier write disappears entirely. With `max_parallel_insert` > 1 this is reachable in normal ingestion, not just in a contrived race.

`PGKVStorage` does the equivalent merge in a single `INSERT ... ON CONFLICT DO UPDATE` with `COALESCE` over the incoming and stored values, so there is no window. The same shape should work here — Hologres supports `ON CONFLICT` upserts and JSONB concatenation (`||`), which would let the protected keys be preserved by the statement itself.

---

lightrag/kg/hologres/client.py
```diff
@@ -0,0 +1,1106 @@
...
+    if not is_connection_loss(error):
+        return False
+    return kind is OperationKind.READ or replay_safe
+
+
+def _error_category(error: BaseException) -> str:
+    return "connection_loss" if is_connection_loss(error) else "database_error"
+
+
+def _requires_invalidation(
```
danielaskdd, line 490: **`TimeoutError` only aliases `asyncio.TimeoutError` from Python 3.11.**

This function (and the timeout branch in `close()`) matches on the builtin `TimeoutError`. The two became the same class in 3.11; on 3.10 `asyncio.TimeoutError` is a distinct class inheriting from `Exception`, so the check does not catch what `asyncio.wait_for` actually raises.

`pyproject.toml` still declares `requires-python = ">=3.10"`, so this is a supported configuration. On 3.10 a write that times out and is not replay-safe never gets its connection invalidated: the connection goes back to the pool while the server may still be executing the statement, and the next borrower of that connection can observe the previous operation's result or an out-of-band error.

Matching `asyncio.TimeoutError` explicitly works on every supported version, since on 3.11+ it is the same object:

```python
if isinstance(error, asyncio.TimeoutError):
    ...
```

The offline tests presumably run on 3.11+, which is why this is green — a regression test that pins the 3.10 behavior would need to raise `asyncio.TimeoutError` directly rather than relying on the alias.

---

lightrag/kg/hologres/vector.py
```diff
@@ -0,0 +1,802 @@
...
+            )
+        else:
+            vector = _normalize_vector(
+                query_embedding,
+                self._dimension,
+                "Hologres vector is invalid",
+            )
+
+        client, table = self._ready()
+        # approx_cosine_distance returns cosine SIMILARITY (higher is
```
danielaskdd, line 692: **The similarity-vs-distance assumption is never enforced at runtime.**

The comment is right that everything downstream depends on `approx_cosine_distance` returning *similarity* (higher = closer) — the `ORDER BY ... DESC` and the `cosine_better_than_threshold` comparison both invert if it returns a distance. But the "contract frozen by the live HGraph probe" only holds where that probe actually runs, and it is gated behind `--run-hologres-live`.

On the startup path, `probe_production_capabilities` checks the server version and nothing else, and its `blocking_failures` result is never consulted by the caller. So on a server where the function carries distance semantics, nothing fails and nothing warns — queries return the *least* similar rows, and the threshold filter keeps the wrong side of the set. A silently-inverted retriever is hard to attribute back to the storage layer from the symptom (answers are simply poor).

Two things would close it: have the startup path actually consult `blocking_failures` and refuse to serve when a blocking probe failed, and add a cheap semantic assertion to the probe itself — one known pair where similarity and distance disagree in sign is enough to pin the orientation without a full live suite.

---

lightrag/kg/hologres/graph_age.py
```diff
@@ -0,0 +1,1034 @@
...
+            "RETURN properties(r)",
+            "props ag_catalog.agtype",
+            descriptor="age.edge.read",
+            params={"src": src, "tgt": tgt},
+        )
+        if not rows:
+            return None
+        return _decode_map(rows[0][0])
+
+    async def upsert_edge(
```
danielaskdd, line 559: **Non-atomic edge write leaves a property-less edge; the residue is undocumented.**

`upsert_edge` creates the `DIRECTED` edge in one autocommit statement and then sets its properties in a second. An interruption between the two — connection loss, timeout, process death — leaves an edge in the graph with no `weight`, `source_id` or `description`.

That residue is worse than a missing edge, because it is not self-healing in the way `AGENTS.md` *Consistency without transactions* requires: a subsequent `upsert_edge` for the same pair will find the edge present and may take the update path, and in the meantime a query traversing it gets an edge that carries no evidence and no attribution. It also interacts with the *Relation weight contract* — a weight-less edge is not a valid row under `weight >= len(distinct real source IDs)`.

The two-table backend writes the edge and its properties in a single statement, so the two graph backends do not have the same failure model, which is itself worth calling out for operators choosing between them.

Preferred fix is to make it one statement (an AGE `MERGE ... SET` in a single Cypher call, or both statements inside one explicit transaction rather than autocommit). If AGE genuinely cannot express it atomically, then per the repo's consistency rule the residue needs writing down next to the code with its recovery path — a documented residue is a decision, an undocumented one is a defect.

---

danielaskdd (commented):
Thanks for a thorough contribution — the module layout, the resumable schema ledger and the offline test suite are all in good shape, and `./scripts/test.sh tests/kg/hologres_impl` passes locally (836 passed, 13 skipped) with `ruff check` clean.

I read every production module (`client`, `config`, `capabilities`, `schema`, `kv`, `vector`, `doc_status`, `graph`, `graph_age`) plus the registry and packaging changes, cross-checked against the `lightrag/base.py` contracts and the PostgreSQL reference implementation. Six findings, left inline.

**Blocking:**

1. `graph_age.py` — any non-`PASSED` probe, transient failures included, silently falls back to the two-table backend, so an existing AGE graph goes invisible and writes land in a different physical store with nothing logged.
2. `kv.py` — `upsert` never stamps the storage-managed `create_time` / `update_time`, which breaks chronological ordering of LLM extraction-cache rows in `operate.py`.
3. `kv.py` — the `full_docs` protected-field merge is a client-side read-modify-write; concurrent upserts of the same doc lose `content`, not just the protected keys.
4. `client.py` — `TimeoutError` is only an alias of `asyncio.TimeoutError` from 3.11, so on 3.10 (still allowed by `requires-python`) a timed-out non-replay-safe write never invalidates its connection.

**Worth addressing, your call on how:**

5. `vector.py` — the similarity-vs-distance orientation of `approx_cosine_distance` is asserted only by a live-only probe, and `blocking_failures` is never consulted at startup; a server with distance semantics silently returns the least similar rows.
6. `graph_age.py` — `upsert_edge` is two autocommit statements, so an interruption leaves a property-less edge. Either make it atomic or document the residue with its recovery path, per *Consistency without transactions* in `AGENTS.md`.

Items 2, 3 and 6 are places where the existing `PGKVStorage` / two-table graph implementations already have the right shape, so they should mostly be a matter of porting rather than designing something new.

---

TimothyDing:
Thank you for the thorough review — all six findings are addressed in 8914231a2.

**Blocking**

1. **AGE silent fallback** (`graph_age.py`): a `FAILED` probe now splits by certainty. `age_extension_missing` is definitive and falls back at INFO; any other failure (transient connection loss, permissions, unexpected probe behavior) now **fails initialization** with the probe's detail code instead of falling back — falling back on an indeterminate outcome is what would make an existing AGE graph invisible and land writes in a different physical store. The class docstring and both storage guides describe the new semantics, and a regression test pins the fail-fast path.

2. **KV timestamps** (`kv.py`): `create_time`/`update_time` are now stamped inside the single upsert statement with `EXTRACT(EPOCH FROM CURRENT_TIMESTAMP)::bigint` — server-generated on insert, and the conflict branch restores the stored numeric `create_time` (anything else reads back as `0`, matching `normalize_kv_create_time`). Caller-supplied values are ignored, per the `BaseKVStorage.upsert` contract. `operate.py`'s cache ordering works again.

3. **KV full_docs merge** (`kv.py`): the read-modify-write is gone. The merge happens in the same statement as the write: `current.payload || EXCLUDED.payload || <per-key restore CASEs>`. The only client-side step drops protected keys whose incoming value is unusable (None/""/"""/{}), which inspects the incoming payload alone and cannot race a concurrent writer; the EXCLUDED row carries a protected key exactly when the caller supplied a usable value. Concurrent upserts of the same doc now lose nothing — the later writer's `content` lands atomically. KV upserts no longer use stream COPY: its `on_conflict update` replaces the whole row, which can neither stamp timestamps nor restore protected keys server-side (stream COPY remains for vector bulk writes).

4. **`TimeoutError` on 3.10** (`client.py`): both the non-replay-safe invalidation path and pool close now match `asyncio.TimeoutError` explicitly. A regression test raises `asyncio.TimeoutError` directly so it pins the 3.10 behavior rather than riding the 3.11+ alias.

**Worth addressing**

5. **Similarity orientation** (`vector.py`, `capabilities.py`): vector initialize now proves the orientation cheaply at startup — one scalar query on constant vectors asserting `approx_cosine_distance(same) > approx_cosine_distance(opposite)` — and refuses to serve on a mismatch, so a distance-semantics server fails loudly instead of silently returning the least similar rows. The full HGraph score-contract probe (which already pins expected scores 1.0/0.0/-1.0) stays in the isolated probe suite exercised by the live tests; the startup check was deliberately kept DDL-free so restricted deployments (no CREATE SCHEMA privilege) can still start.

6. **Atomic edge upsert** (`graph_age.py`): `MERGE (a)-[r:DIRECTED]->(b) SET r = {...}` is now a single cypher call, so an interruption can never expose a property-less edge. The endpoint-stub residue (crash between endpoint creation and edge write) is documented next to the code with its recovery path, matching the two-table backend's accepted residue.

Verification: `./scripts/test.sh tests/kg/hologres_impl` — 842 passed, 13 skipped; `ruff check .` clean.

TimothyDing:
@danielaskdd Thank you, I have fixed all~ Please review ~

danielaskdd:
## Review：Hologres 后端与现有存储实现的语义兼容性问题

感谢这个 PR，接口层面是完整的（五个类的 `__abstractmethods__` 都为空，方法签名与基类一致，`lightrag/kg/__init__.py` 注册也齐全）。

我把 `lightrag/kg/hologres/` 逐模块与现有实现做了对照 —— KV / Vector / DocStatus 对照 `lightrag/kg/postgres_impl.py`，两表图对照 PR 自己声明的参照实现 `lightrag/kg/pgtable_impl.py`。**语义层面有若干处与现有实现不兼容，其中 4 处我认为必须在合并前处理。** 以下按影响排序，麻烦逐条整改或说明理由。

---

### 一、必须整改

#### 1. 缺少 `HOLOGRES_WORKSPACE` 覆盖与 `"default"` 兜底

现有每个数据库后端都实现同一条 workspace 优先级链：`<BACKEND>_WORKSPACE 环境变量 > 实例 workspace > "default"`。参见 `lightrag/kg/postgres_impl.py:3140-3153`，Mongo / Redis / Milvus / Qdrant / Neo4j / Memgraph / OpenSearch 同构（`grep -o "[A-Z_]*_WORKSPACE" lightrag/kg/*.py` 可以看到完整清单）。

Hologres 五个类都只做 `self.workspace = validate_workspace(self.workspace)`：

- `lightrag/kg/hologres/kv.py:186`
- `lightrag/kg/hologres/vector.py:187`
- `lightrag/kg/hologres/doc_status.py:276`
- `lightrag/kg/hologres/graph.py:165`
- `lightrag/kg/hologres/graph_age.py:176`

`env.example` 里也没有 `HOLOGRES_WORKSPACE`。后果：

- 未设全局 `WORKSPACE` 时数据落在 `workspace = ''`，而 PG 落在 `'default'`；
- 运维没有其它后端都有的"事后用环境变量把已有数据认回来"的逃生口。

**请补上 `HOLOGRES_WORKSPACE` 与 `"default"` 兜底，对齐现有优先级链，并同步更新 `env.example` 和 `docs/HologresStorageGuide*.md`。**

#### 2. doc_status 的 `content_hash` 丢失 write-once 保护

PG 侧用的是：

```sql
content_hash = COALESCE(
    NULLIF(EXCLUDED.content_hash, ''),
    LIGHTRAG_DOC_STATUS.content_hash
)
```

并且在 `PGDocStatusStorage.upsert` 的注释里明确写了这是**必需**的：状态迁移（processing → processed）复用旧的 `DocProcessingStatus` 载荷、不重新提供 hash，而 `_persist_parsed_full_docs` 是用一次单独的 targeted update 补写 hash 的。注释还特别标注这是"对 `JsonDocStatusStorage` 的有意偏离"。

Hologres 是裸覆盖 —— `lightrag/kg/hologres/doc_status.py:656`：

```
"content_hash = EXCLUDED.content_hash, "
```

只要有一次状态迁移 upsert 携带的 `status_doc.content_hash` 是 `None`（例如 `status_doc` 在 parse 补写 hash 之前就已装载，见 `lightrag/pipeline.py:6127`），hash 就被清空，`get_doc_by_content_hash` 的跨文件名去重随之失效。

**请改回 `COALESCE(NULLIF(EXCLUDED.content_hash, ''), current.content_hash)`，并补一个回归测试。**

#### 3. 图存储依赖 `datcollate = 'C'`，但运行时从不校验

`lightrag/kg/hologres/graph.py:16` 与 `graph.py:949` 都声称"capability probe 已证明 `datcollate='C'`，所以不需要 `COLLATE \"C\"` 子句"。但实际在 `initialize()` 里跑的只有 `probe_production_capabilities()`，它**只执行 `SELECT hg_version()`**（`lightrag/kg/hologres/capabilities.py:279-288`）。唯一的 collation 检查在 `capabilities.py:720-726`，属于 HGraph 分区探针 —— 只在 `--run-hologres-live` 的隔离测试套件里跑，生产启动路径完全不经过。

对比参照实现 `lightrag/kg/pgtable_impl.py:1322` 是显式写死 `ORDER BY degree DESC, n.id COLLATE "C" ASC`，且模块 docstring 特别警告 Python `min`/`max` 与 SQL 排序混用在非 C collation 下会分叉。

后果：在非 C collation 的 Hologres 实例上，`get_all_labels` / `get_popular_labels` / `search_labels` / `get_knowledge_graph` 的排序与截断会与参照实现不一致，而且**没有任何启动期报错**。

值得注意的是同一个 PR 里向量侧做法恰恰相反：`prove_similarity_orientation()` 是在 `vector.initialize()` 里真跑的（`capabilities.py:240-276`），并在不匹配时拒绝服务。图侧这个缺口看起来是漏了而不是设计选择。

**两种修法二选一：把 `datcollate = 'C'` 检查挪进启动路径（做法照抄 `prove_similarity_orientation`），或者恢复 `COLLATE "C"` 子句。同时请修正 `graph.py:16` / `graph.py:949` 这两处与实际不符的注释。**

#### 4. `HologresAGEGraphStorage` 会在运行时静默降级成另一个物理存储

`lightrag/kg/hologres/graph_age.py:242-258`：AGE 扩展缺失时打一条 `logger.info` 就委托给 `HologresGraphStorage`。

现有实现没有这个先例 —— PG 的 `configure_age_extension`（`lightrag/kg/postgres_impl.py:929+`）在 AGE 缺失或版本不受支持时直接 `RuntimeError` 拒绝启动，宁可起不来也不换后端，只有显式的 `AGE_ALLOW_UNSUPPORTED` 环境变量能放行。

降级的后果是 `LIGHTRAG_GRAPH_STORAGE` 不再决定物理 schema，两套物理存储互不可见。PR 已经把"非 `age_extension_missing` 的探针失败"改成 fail loud 了（这点很好），但"一个存储类在运行时变成另一个存储类"本身就是引入了新语义。

**请改为 fail loud，或者要求显式 opt-in 环境变量放行（对齐 PG 的 `AGE_ALLOW_UNSUPPORTED` 模式）。**

---

### 二、需要明确结论

#### 5. 落库的 `status` 被 `DocProcessingStatus.__post_init__` 改写

`lightrag/kg/hologres/doc_status.py:497` + `512-514`：`_normalize_complete_input` 先调 `DocProcessingStatus.from_stored(raw)`，再取 `document.status.value` 写入 status 列。而 `__post_init__`（`lightrag/base.py:1385-1391`）会在 `multimodal_processed is False and status == PROCESSED` 时把 status 改成 `PREPROCESSED` —— **于是 Hologres 持久化的是转换后的值**。

三个后端对同一批文档的表现因此各不相同：

| 后端 | 存储值 | 读取值 |
|---|---|---|
| Hologres | `PREPROCESSED` | `PREPROCESSED` |
| PG | `PROCESSED` | `PROCESSED`（表里没有 `multimodal_processed` 列，字段被丢弃，转换永不触发） |
| Json | `PROCESSED` | `PREPROCESSED` |

`get_status_counts()` 是 `GROUP BY status` 的原始列聚合，所以三个后端给出的计数会不一样。

**建议：落库时用调用方传入的原始 status，`__post_init__` 的转换只保留在读路径。如果认为持久化转换后的值才是对的，请在 PR 里说明，并同步处理 PG / Json 的不一致。**

#### 6. 向量行的 `created_at` 实际是 `updated_at`

`lightrag/kg/hologres/vector.py:693`：

```
"EXTRACT(EPOCH FROM updated_at)::bigint AS created_at "
```

Hologres 的向量表压根没有 create_time 列。PG 侧是 `EXTRACT(EPOCH FROM create_time)::BIGINT as created_at`（`lightrag/kg/postgres_impl.py:5113`）。

这个值会进入 LLM 上下文的 `created_at` 列（`lightrag/operate.py:5560-5588`），所以重新抽取之后，实体/关系的"创建时间"会整体跳到最近一次写入时间。

**建议给向量表加 `create_time` 列并在 ON CONFLICT 时保留；如果不打算加，请至少不要把 `updated_at` 命名成 `created_at`。**

---

### 三、建议修正（非阻塞，但请逐条回应）

#### 7. doc_status 读回的 `status` 是枚举而非字符串

`_decode_raw_row` 把 `raw["status"]` 转成 `DocStatus` 实例（`doc_status.py:377-379`），而 PG 的 `get_by_id` 返回裸 `str`。`DocStatus` 是 `str` 子类，等值比较和 `json.dumps` 没问题，但 `str()` / f-string 在 Python < 3.11 会得到 `"DocStatus.PROCESSED"`。建议与其它后端保持一致返回字符串。

#### 8. KV `full_docs` 是 JSONB 整体 MERGE，PG 是固定列 REPLACE

六个受保护键（`sidecar_location` / `parse_format` / `content_hash` / `process_options` / `parse_engine` / `chunk_options`）的保留语义两边对上了（`kv.py:431-439` vs `SQL_TEMPLATES["upsert_doc_full"]`），这点没问题。差异在其余键：PG 是固定列，调用方写的未知键被直接丢弃；Hologres 的 `current.payload || EXCLUDED.payload` 会**永久保留**任何写过的键，而且调用方"删掉某个键"在 Hologres 上不生效。`_persist_parsed_full_docs` 做 `{**existing, **record}` 时，Hologres 侧会不断累积 PG 侧不存在的字段。请在 `docs/HologresStorageGuide.md` 里把这个差异写下来。

#### 9. 边 upsert 拆成两条语句

`pgtable_impl.py` 的 `upsert_edge` 用一条 CTE 语句在同一事务里完成"建端点 + 写边"。Hologres 拆成两条独立 autocommit（`graph.py:568-626`），崩溃窗口会留下没有边的孤立 stub 节点。

按 `AGENTS.md` 的"Consistency without transactions"，这属于可接受残留，代码注释也写了恢复路径 —— 但原子性与参照实现不同。按仓库约定，**每个被接受的残留都要连同理由和恢复路径写下来**，请确认这条在 `docs/HologresStorageGuide.md` 或 `docs/design/` 里有对应记录，而不只是一行代码注释。

#### 10. 其它次要差异

| 项 | 差异 |
|---|---|
| 批量粒度 | doc_status upsert 是 **per-record 一条语句**（`doc_status.py:664`），PG 用 `executemany` + 字节/条数预算分批。KV / vector / graph 侧 Hologres 都有分批，唯独 doc_status 没有 |
| 失败粒度 | PG 的 doc_status upsert 跳过非法记录并记日志继续；Hologres 整批抛 `HologresDocStatusError` |
| `chunks_count` | PG 缺省 `-1`（`v.get("chunks_count", -1)`），Hologres 缺省 `None`。`chunk_fields_from_status_doc` 两者都容忍，但在 SQL 过滤和 UI 上不等价 |
| 配置来源 | 不支持 `config.ini`（PG 走 `configparser.read("config.ini")`），也不走 `lightrag.utils.get_env_value` |
| 测试收集 | `tests/conftest.py` 改了全局 `pytest_collection_modifyitems`，`--run-integration` 不再"跑全部"。行为上等价，但影响的是全仓库的收集逻辑，请在 PR 描述里显式说明 |

---

### 小结

阻塞项是 **1 / 2 / 3 / 4**，**5 / 6** 需要一个明确结论（改掉或说明理由），**7–10** 请逐条回应（修掉或补文档）。

其中 2 和 3 是我最担心的两条：前者会静默破坏文档去重，后者会在非 C collation 实例上静默给出与参照实现不同的排序结果 —— 两者都没有任何报错，只能靠代码审查发现。

整改后麻烦 ping 一下，我再过一遍。

---

TimothyDing:
@danielaskdd 
Thanks for the detailed semantic pass. Pushed `b4af24117` with fixes and explicit decisions for all ten items.

**Blocking**

1. **Workspace selection** — all five roles now follow `HOLOGRES_WORKSPACE` > instance workspace > `"default"`, with fail-closed validation. `env.example` and both guides document the chain.
2. **`content_hash` write-once** — doc-status upsert now uses `COALESCE(NULLIF(EXCLUDED.content_hash, ''), current.content_hash)`, and a regression test pins the omitted/empty update case.
3. **Graph collation** — we restored the reference implementation's explicit `COLLATE "C"` on identifier tie-breakers instead of assuming server `datcollate`; the stale “probe proves C” comments were removed. All emitted identifier sort keys now pin the collation, and tests assert this.
4. **AGE fallback** — missing AGE now fails startup. `HOLOGRES_AGE_ALLOW_UNSUPPORTED=true` is the explicit opt-in to the different two-table physical schema; other probe failures continue to fail closed. Tests cover default refusal and explicit fallback.

**Explicit decisions**

5. **Doc-status status value** — persistence now stores the producer's raw status. `DocProcessingStatus.__post_init__` derivation remains a read/model-time behavior. This makes persisted `status` aggregation match PostgreSQL; JSON already performs the derivation on read.
6. **Vector `create_time`** — added a nullable `create_time timestamptz` through an additive ledger descriptor, populated it on insert/COPY, preserved it on conflict, and changed query output from `updated_at` to `COALESCE(create_time, 0)` for pre-migration rows.

**Non-blocking responses**

7. Public doc-status rows now return plain `status` strings, matching PostgreSQL; enum conversion remains local validation/model construction.
8. The intentional JSONB breadth difference is now documented in both guides (including explicit-null clearing for non-protected keys). Protected-key restore semantics remain aligned.
9. The two-table endpoint-stub residue and recovery path are documented in both storage guides under “Accepted write residues,” not just beside the code.
10. `chunks_count` now defaults to `-1` to match PG. The guides explicitly record the intentional doc-status fail-closed validation/single-record statement behavior, environment-only configuration, and the test-gating collection change. The latter is also being added to the PR description.

Verification: `./scripts/test.sh tests/kg/hologres_impl --test-workers 4` — **852 passed, 13 skipped**; `uv run ruff check .` is clean.

PR #3973, opened 2026-09-15
fix(anthropic): extract answer text from the correct content block

Shizoqua:
## Summary
Fixes a crash calling Claude through the native Anthropic binding when extended thinking is enabled.

## Problem
The handler read content from the first block unconditionally, which is a reasoning block with no text when thinking is on.

## Changes
Handler now picks the first content block that actually carries a text field.

## Testing
Added a regression test for a leading reasoning block followed by the answer. Ran the full Anthropic binding test suite, all passing.

## Compatibility and Risk
No API changes. Normal responses are unaffected.

## Related Issue
No related issue.


danielaskdd:
Thanks for the fix — the bug is real and reachable, and the direction is right. I verified against the SDK and ran the suite; one blocking issue before merge.

## Verification

Confirmed on `anthropic==1.6.0`, the response content block types are:

```
ThinkingBlock          ['signature', 'thinking', 'type']   <- no .text
RedactedThinkingBlock  ['data', 'type']                    <- no .text
TextBlock              ['citations', 'text', 'type']
```

So `response.content[0].text` does raise `AttributeError` once `thinking={"type": "enabled", ...}` is passed through `llm_model_kwargs` (kwargs are forwarded straight into `create_params`). Not a theoretical problem.

A point in favour of the change that the description does not mention: the **streaming branch already behaves this way**. `stream_response()` filters on `hasattr(event.delta, "text")`, and `ThinkingDelta` only carries `.thinking` while `SignatureDelta` only carries `.signature`, so streaming has always skipped thinking correctly. This PR aligns non-streaming with existing streaming semantics rather than inventing a new one.

`./scripts/test.sh tests/llm/anthropic_impl` — **19 passed**.

## Blocking: the new raise conflicts with the repo-wide retry policy

`InvalidResponseError` is in this binding's own tenacity predicate:

```python
retry=retry_if_exception_type(
    (RateLimitError, APIConnectionError, APITimeoutError, InvalidResponseError)
),
```

With extended thinking on, by far the most likely way to get zero text blocks is the model spending its entire `max_tokens` inside the thinking budget, i.e. `stop_reason == "max_tokens"`. That is precisely the case the codebase already routes away from retry, and the rationale is written down in `lightrag/exceptions.py`:

> Deliberately absent from every binding's retry predicate... hitting the token limit is a property of the prompt and the configured output budget, so re-running the same call re-buys the same full-budget generation just to fail identically.

`openai.py` (`finish_reason == "length"`) and `gemini.py` (`length_limited`) both follow it: token-limit exhaustion raises the non-retryable `EmptyTruncatedResponseError`, everything else raises `InvalidResponseError`. `utils.py::empty_length_truncated_hint` states the intent explicitly — "Shared by every binding so the four providers describe the same failure identically."

As written, this PR makes Anthropic the only binding that violates that: the failure mode the PR is specifically about would buy **three full-budget billed generations plus exponential backoff**, then surface a `RetryError` whose message does not tell the operator which knob to turn. Suggested shape:

```python
if content is None:
    stop_reason = getattr(response, "stop_reason", None)
    # ...diagnostics...
    if stop_reason == "max_tokens":
        raise EmptyTruncatedResponseError(message)   # not retried
    raise InvalidResponseError(message)              # retried
```

## Non-blocking

1. **Diagnostics are thin.** The other three bindings carry `finish_reason` / `completion_tokens` / `reasoning_tokens` / `reasoning_content_len` and route through `empty_length_truncated_hint(...)`. This one carries only `stop_reason=`. Since the PR is about thinking, the thinking-block length and `usage.output_tokens` are the most actionable numbers — they tell the operator where the budget went. Reusing `empty_length_truncated_hint("consider raising max_tokens or lowering thinking.budget_tokens", reasoning_consumed_budget=...)` keeps the four providers' messages aligned.

2. **`hasattr(block, "text")` vs `block.type == "text"`.** Duck-typing works against today's SDK (I checked all 27 block types; only `TextBlock` and `ParsedTextBlock` carry `text`), but `type` is the documented discriminator.

3. **`next(...)` keeps only the first text block.** Under interleaved thinking the content list can be `[thinking, text, thinking, text]`, and the later half of the answer would be dropped silently. Joining every text block is the same amount of code and strictly safer.

4. **Missing test for the new failure mode.** The added `[thinking, text]` case is good, and rewriting rather than deleting the "usage recorded before extraction" regression test correctly preserves that invariant. What is not covered is `content=[ThinkingBlock]` with `stop_reason="max_tokens"` — the case item 1 above turns into a non-retryable error.

## On discarding the thinking content

Dropping it is the right call here, not just an omission — it matches what the streaming path already does and what `enable_cot=False` means everywhere else in the tree. Worth noting that flattening Anthropic thinking into the answer string would be actively wrong: `ThinkingBlock` carries a `signature` that must be round-tripped verbatim for multi-turn/tool-use continuation, and a flattened `<think>` block would then be regex-stripped by `remove_think_tags` in `use_llm_func_with_cache` and persisted into the answer cache on the way.

A one-line docstring note next to the existing `enable_cot=True is not supported for the Anthropic API and will be ignored` would help — stating that thinking may be requested but its output is discarded, so a reader knows this is "requested and dropped" rather than "unsupported".

---

PR #3975, opened 2026-09-16
feat: support video inputs in chat completions

octo-patch:
Reason: MiniMax video input cannot currently be passed through the existing chat completion binding.

Add optional `video_inputs` that become `video_url` content parts alongside text and images. Accept URL strings or dictionaries carrying a URL and video options, including uploaded-file references. Invalid or empty URLs fail before a request is sent; existing text-only payloads remain unchanged.

The configured model and endpoint must support video input. This adds transport at the binding layer; video parsing and upload orchestration remain caller responsibilities.

Validation: 68 offline tests passed in `tests/llm/openai_impl`; Ruff formatting and lint checks passed for the changed files. Dependencies were installed with `uv sync --frozen --extra test`.


lightrag/llm/openai.py
```diff
@@ -456,20 +462,33 @@ async def openai_complete_if_cache(
...
+                        },
+                    }
+                )
+        for video in video_inputs or []:
+            if isinstance(video, str):
+                video_url = {"url": video}
+            elif isinstance(video, dict):
+                video_url = dict(video)
+            else:
+                raise TypeError("video_inputs elements must be URL strings or dicts")
```
danielaskdd, line 485: **Resource leak on invalid video input.** `create_openai_async_client()` has already created the client at line 450, but the new `TypeError`/`ValueError` paths here run before the lifecycle `try/finally` that begins at line 517. As a result, every rejected `video_inputs` call leaves `openai_async_client.close()` uncalled. The new invalid-input tests only assert that no request is sent, so they do not catch this leak. Please either normalize and validate `video_inputs` before constructing the client, or move message construction under a `try/finally` that always closes it, and add a regression assertion that the fake client is closed exactly once on these validation failures.

danielaskdd (changes requested):
The invalid-input path leaks the OpenAI client, so this needs to be fixed before merge.

PR #3982, opened 2026-09-16
fix(memgraph): dedupe get_all_edges and iter_edges by relationship id

Shizoqua:
## Summary
Fixes get_all_edges and iter_edges returning every relationship twice on Memgraph.

## Problem
Same defect as the Neo4j backend: the undirected match pattern yields one row per orientation of each relationship, and DISTINCT on the projected columns does not collapse them.

## Changes
Both queries now dedupe on the relationship's own id instead of the projected endpoint columns.

## Testing
Added regression tests covering single and multiple relationships. Ran the full memgraph test suite, all passing.

## Compatibility and Risk
No API changes. Only affects whole-graph edge reads; returned edge count is now accurate.

## Related Issue
No related issue.


danielaskdd:
@codex review

lightrag/kg/memgraph_impl.py
```diff
@@ -1298,16 +1310,23 @@ async def iter_edges(self, batch_size: int):
...
             result = await session.run(
                 f"""
                 MATCH (a:`{workspace_label}`)-[r]-(b:`{workspace_label}`)
-                RETURN DISTINCT a.entity_id AS source, b.entity_id AS target,
+                RETURN id(r) AS rel_id, a.entity_id AS source, b.entity_id AS target,
                        properties(r) AS properties
                 """
             )
             batch: list[dict] = []
+            seen_rel_ids = set()
```
danielaskdd: `iter_edges()` was introduced specifically so whole-graph maintenance such as `chunk_tracking_repair` keeps Python memory bounded by `batch_size`. This set survives every `yield` and grows to one entry per physical relationship, so a large Memgraph repair is back to O(E) client memory and can OOM even though each emitted batch is small. Please make Cypher return each relationship once—for example, enumerate the stored direction with `MATCH (a:`workspace`)-[r]->(b:`workspace`)` without filtering the relationship type—then remove `rel_id` and this global set. The test should pin the one-row query/bounded batching rather than simulate two rows and require a graph-wide client-side dedupe table.

danielaskdd (changes requested):
The duplicate-row diagnosis is correct, but the current client-side fix breaks the bounded-memory contract of `iter_edges()`. Please move the deduplication into the Cypher query so Memgraph emits one row per physical relationship, remove the graph-sized `seen_rel_ids`, and update the regression tests to cover the bounded iterator behavior. `get_all_edges()` should use the same one-row-per-relationship query for consistency.

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #3985, opened 2026-09-16
fix(api): require a word boundary after unsuffixed context mode keys

### LINKED ISSUE #3983: [Bug]: /context mode prefix also matches "/contextualize" and truncates the query on Ollama-compatible endpoints

### Describe the bug

The context mode keys in `parse_query_mode()` (`lightrag/api/routers/ollama_api.py`) are matched with `query.startswith(prefix)` and no word-boundary check. The `/context` key (unlike the space-suffixed keys such as `/local `) therefore also prefix-matches ordinary English words:

Any query beginning with "/contextualize ...", "/contexts ...", etc. is silently treated as `/context` mode, with two effects:

1. The matched fragment is cut out of the retrieval query text ("context" disappears from the user's question).
2. `only_need_context` is set to True, so no LLM answer is generated — the endpoint returns retrieved context only.

Affected paths: the RAG branches of the Ollama-compatible endpoints `/api/chat` and `/api/generate`, which call `parse_query_mode()` on the incoming prompt / last chat message.

### Steps to reproduce

No services or API keys needed — the repro is at the function level:

python -c "from lightrag.api.routers.ollama_api import parse_query_mode; print(parse_query_mode('/contextualize the following passage'))"

Output:

('ualize the following passage', <SearchMode.mix: 'mix'>, True, None)

For comparison, the legitimate form still works as documented:

python -c "from lightrag.api.routers.ollama_api import parse_query_mode; print(parse_query_mode('/context how does docker networking work'))"

('how does docker networking work', <SearchMode.mix: 'mix'>, True, None)

Against a running server: POST a chat request to /api/chat whose last user message is "/contextualize the following passage" — the reply contains retrieved context only (no LLM answer), built from the truncated query text "ualize the following passage".

### Expected Behavior

A mode key should only match at a word boundary (followed by whitespace or end of query), so a query like "/contextualize the following passage" passes through unchanged as an unrecognized prefix — exactly like every other unknown mode, e.g. "/nosuchmode tell me about X" — and gets normal mix retrieval plus an LLM answer.

### LightRAG Config Used

Not config-sensitive: any deployment that exposes the Ollama-compatible endpoints (the default server). The repro above exercises the parsing function directly, with no storage or LLM provider involved.

### Logs and screenshots

(Console output of the two repro commands above; no relevant server logs — the misrouting is
silent.)

### Additional Information

- LightRAG Version: v1.5.8
- Python Version: 3.12.10
- Related Issues: none found
- Root cause: the unsuffixed context keys ("/context", "/localcontext", "/mixcontext", ...) are matched without a boundary check; the space-suffixed keys ("/local ", "/bypass ", ...) carry the boundary implicitly and are unaffected.

hxxxi-malog:
## Description

The context mode keys in `parse_query_mode()` match without a trailing space, so `"/context"` also prefix-matches ordinary English words such as `"/contextualize"` and `"/contexts"` on the Ollama-compatible endpoints (`/api/chat`, `/api/generate` RAG branches): the matched fragment is cut out of the retrieval query text and the query silently flips to context-only retrieval (no LLM answer).

A mode key now matches only at a word boundary - end of the query or whitespace - so such queries pass through unchanged as unknown modes, exactly like every other unrecognized prefix. The space-suffixed keys (`"/local "`, `"/bypass "`, ...) carry the boundary implicitly and are untouched.

## Related Issues

Closes #3983

## Changes Made

- `lightrag/api/routers/ollama_api.py`: word-boundary check in the `mode_map` matching loop; the docstring now states the matching rule.
- `tests/api/test_ollama_query_mode.py`: `test_unsuffixed_context_keys_end_at_a_word_boundary` - 2 regression cases for the truncated queries, plus 3 cases pinning the legitimate unsuffixed forms (`/context ...`, bare `/context`, `/localcontext ...`) against regressions.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Verified red/green: the two new defect cases fail on `main` and pass with this fix; all 16 pre-existing cases in the file are unchanged. Ran the mirror subset `tests/api/` (1245 passed; the failures present on the author's Windows machine are byte-identical on the unmodified baseline - Unix-only `fcntl` imports, event-loop timing, symlink permissions - and CI on Linux is not subject to them). ruff (pinned v0.15.17) lint and format are clean.


danielaskdd:
@codex review

PR #3993, opened 2026-09-17
fix(kg): merge relation evidence instead of overwriting it

Shizoqua:
## Summary
Fixes lost relation evidence in ainsert_custom_kg.

## Problem
Re-declaring an existing relation overwrote its weight and source_id instead of combining them.

## Changes
Existing relations are now merged: source_id unioned, weight floored to combined evidence.

## Testing
Added regression tests. Full pipeline and extraction suites pass.

## Compatibility and Risk
No API changes. New relations unaffected.

## Related Issue
No related issue.


lightrag/lightrag.py
```diff
@@ -4687,8 +4710,24 @@ async def _do_graph_and_vdb_writes() -> None:
...
+                            )
+                            weight = apply_relation_weight_floor(
+                                max(weight, existing_edge.get("weight") or 0.0),
+                                source_id,
+                            )
+
                         edge_data = {
-                            "weight": relationship_data["weight"],
+                            "weight": weight,
                             "description": relationship_data["description"],
```
AmirF194: I ran this against d8833e826673eef98683eee2c53751b90d16808f in a clean container, using the fixtures from your own `test_custom_kg_relation_merge.py` with an existing edge that carries a distinct description and keywords.

Merging into an existing edge unions weight and source_id but still replaces description, keywords and file_path outright, the same overwrite this PR sets out to fix. `_merge_edges_then_upsert` in `operate.py` (the document-ingest merge path) collects `already_description`/`already_keywords`/`already_file_paths` from the existing edge before writing; the `ainsert_custom_kg` path here only reads `relationship_data` for those three fields.

```
$ python desc_check.py
FINAL edge_data: {'weight': 2.0, 'description': 'new description only', 'keywords': 'new-keyword', 'source_id': 'chunk-old<SEP>chunk-7c6b3e302d2557da6f3e570b6ea6e080', 'file_path': 'custom_kg', 'created_at': 1789644760}
CONFIRMED: old description and keywords are silently dropped, only source_id/weight merged.
```

(fixture seeded `existing_edge` with description "OLD DESCRIPTION THAT MUST SURVIVE" and keywords "old-keyword", called `ainsert_custom_kg` with new description/keywords for the same pair, and read the `edge_data` passed to `upsert_edges_batch`.)

A second `ainsert_custom_kg` call re-declaring the same pair loses the prior description and keywords entirely rather than combining them. Folding `existing_edge.get("description")`/`.get("keywords")`/`.get("file_path")` into `edge_data` the same way `source_id` is merged above would close it.

AmirF194: Read the new commit: routing through _merge_attributes with the same strategies _merge_entities_impl uses fixes the overwrite, description/keywords/file_path are combined now instead of replaced.

PR #3994, opened 2026-09-17
fix(kg): reap expired lease before scan job cancel

Shizoqua:
## Summary
Fixes inconsistent state when cancelling an expired scan job.

## Problem
Cancel skipped the lease-expiry reap every other method performs.

## Changes
Cancel now reaps an expired lease before checking status.

## Testing
Added a regression test. Full scan job store suite passes.

## Compatibility and Risk
No API changes. Only affects already-timed-out jobs.

## Related Issue
No related issue.


danielaskdd:
@codex review

PR #3995, opened 2026-09-17
fix(parser): make filename-hint validation and classification agree

RizgarOzan:
## Description

`_filename_hint_match` (the non-throwing classifier used for scan grouping and `canonicalize_parser_hinted_basename`) and `_validate_filename_hint_for_resolution` (the ingestion gate) disagreed on which bracket hints exist, in both directions:

1. **Classifier accepts, validator rejects.** The classifier strips the bracket inner; the validator did not. `report.[ -iet].docx` was grouped and canonicalized as an options-only hint, then failed ingestion with a misleading message:
   `uses unsupported parser engine ''; supported engines: docling, legacy, mineru, native`.
   (`[ native-iet]` already worked because `normalize_parser_engine` strips; only the `-OPTIONS` branch broke.)
2. **Classifier recognises what the spec does not.** A chunk-parameter block made an unknown engine count as a hint: `abc.[xyz-R(chunk_ts=800)].docx` canonicalized to `abc.docx`, while `abc.[xyz-R].docx` is kept — contradicting §2.5 / the duplicate rule ("an unrecognized hint is not stripped"). Validation correctly rejects it.

## Changes Made

- Validator strips the inner before and after `_extract_param_blocks`, mirroring the classifier.
- Classifier: without a supported engine, only an inner starting with `-` is a hint.

## Checklist

- [x] Changes tested locally
- [x] Unit tests added

## Additional Notes

How I checked the two functions now agree: generated 12,348 hints (engines, unknown tokens, `-`, options, `R(chunk_ts=800)`, `(do_ocr=true)`, with/without surrounding spaces) and counted "classifier accepts / validator rejects":

```
main                    276
strip only               64   (all case 2: xyz-/iet-/R-/!- + params)
this PR                   0   (reverse direction also 0)
```

Tests (Windows 11, Python 3.14):
- 7 new cases in `tests/pipeline/test_pipeline_release_closure.py` (`@pytest.mark.offline`): 4 fail on main, 7 pass.
- `pytest tests/ -m offline`: same failure set with and without the change (the remaining failures are Windows-only: `fcntl`, path/URI, symlinks). One FAISS timing test differed once and passes 3/3 on rerun.
- `pre-commit run --files` clean.


danielaskdd:
@codex review

danielaskdd:
@codex review

Issue #4003, opened 2026-09-17
[Feature Request]: per-call provider context — no supported way to attach request-scoped data to a query's LLM, embedding, or rerank calls


pennycoders:
### Do you need to file a feature request?

- [x] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

### Feature Request Description

There's no supported way to attach request-scoped data to the outbound provider calls a single query triggers.

A common need: forwarding a caller-scoped identity (e.g. a header value) onto those calls, so a downstream proxy/observability layer can attribute them to the right end user. A query can make three kinds of outbound provider call — an LLM completion, an embedding (to vector-search for the query text itself and, in local/global/hybrid/mix mode, for keywords), and, when enabled, a rerank call. The natural approach — set a `contextvars.ContextVar` right before calling `aquery`/`aquery_llm`, read it inside the custom `llm_model_func`/`embedding_func`/`rerank_model_func` — doesn't work with how LightRAG dispatches any of these.

Each is wrapped once by `priority_limit_async_func_call` (`lightrag/utils.py`) in a small, fixed pool of persistent `asyncio.Task` workers, created lazily on first use and never recreated per call. A call is handed to a worker as a plain tuple on an `asyncio.PriorityQueue`; the worker executes it inside *its own* task context, captured when the worker task was first created — long before any particular request existed. A `ContextVar.set()` made by the caller's own task is invisible inside that worker task. This is correct, standard `asyncio` behavior (contextvars propagate through a task/coroutine chain, or are copied at `asyncio.create_task()` time — never through an independent queue hand-off), not a bug in the queue itself, but it does mean there's currently no clean way to carry per-call data through to any of these calls without reimplementing the queue's own concurrency limiting and timeout handling on the caller's side.

Proposal: add fields to `QueryParam` that let a caller attach extra keyword arguments to that one query's provider calls — merged in as plain data at each call site, so they ride through the queue exactly like every other kwarg already does (`system_prompt`, `stream`, etc.), without bypassing the queue's concurrency limiting or timeout handling. One field per kind of call (LLM / embedding / rerank), since these are different callables with different signatures, often different providers entirely — forcing one caller-supplied dict onto all three risks a key valid for one colliding with, or being silently rejected by, another.

[...]

Issue #3847, opened 2026-09-06
RFC: cross-worker propagation for runtime configuration (`addon_params` and role LLM config)


danielaskdd:
## Summary

`LightRAG` has two runtime-mutable configuration surfaces — `addon_params` and the role LLM runtime driven by `update_llm_role_config()` / `aupdate_llm_role_config()`. Both are **process-local instance state**.

Under `lightrag-gunicorn --workers N` the app is preloaded (`gunicorn_config.py:32`, `preload_app = True`), so the `LightRAG` object is constructed once in the master and each worker inherits a copy-on-write copy at fork. A runtime change applied in one worker is invisible to the other N-1 workers, and no mechanism exists to propagate it.

This RFC proposes a versioned cross-worker runtime-configuration channel so that a change made in any worker becomes the effective configuration in every worker, with bounded staleness, atomic per-worker application, and without moving credentials into shared storage.

### `addon_params`

- `addon_params` is an `InitVar`; the live store is `_addon_params: ObservableAddonParams` (`lightrag/lightrag.py:939`), exposed through a property attached after class creation (`lightrag/lightrag.py:6984`).
- Any top-level mutation calls `_on_addon_params_changed` (`lightrag/lightrag.py:1024`) → sets `_addon_params_dirty` → the derived caches (`_resolved_summary_language`, `_entity_extraction_prompt_profile`) are recomputed by `_ensure_addon_params_cache()` on the next `_build_global_config()` (`lightrag/lightrag.py:1217`, `:1225`).
- Only top-level keys are observed. Nested in-place mutation is corrected opportunistically at the next enqueue by `resolve_chunk_options()` (`lightrag/parser/routing.py:679-693`).
- There is no REST write path. The server only seeds `{"language": args.summary_language}` at construction (`lightrag/api/lightrag_server.py:2444`).
- Per-document `chunk_options` are frozen into `full_docs` at enqueue, so document processing is already independent of which worker picks the document up. That invariant is what keeps today's divergence from corrupting ingestion.

### Role LLM config

- The `role_llm_configs` constructor field is read exactly once, in `__post_init__` (`lightrag/lightrag.py:1566`); the live state is `_role_llm_states` (`lightrag/lightrag.py:1581`). Mutating `rag.role_llm_configs` at runtime silently does nothing.
[...]

Issue #2133, opened 2025-09-22
[Question]:How to implement multi tenant solutions?


harish-trench:
### Do you need to ask a question?

- [x] I have searched the existing question and discussions and this question is not already answered.
- [ ] I believe this is a legitimate question, not just a bug or feature request.

### Your Question

I'm exploring using LightRAG + POSTGRES for storage in a multi-tenant application and am running into some challenges with managing data isolation between tenants.
What I've Considered
I noticed the workspace isolation option, which seems useful for defining different environments. However, it feels difficult to dynamically switch this configuration at runtime for each incoming API request. The persistence of the connection pools seems to prevent a clean switch from one tenant's database/schema to another's on a per-request basis.

My Question
Is there a recommended or existing way to manage a multi-tenant setup with LightRAG that I might be missing?

For example, is there a way to:

Dynamically pass tenant-specific database postgres credentials/configurations to the LightRAG or other components for each operation? OR use a single database with workspace isolations that can easily be modified during run time...

Any guidance or best practices you could share on this would be greatly appreciated. Thank you for the fantastic library!

BEEFF:
+1 on this. Would love to know the recommended approach here.

superfive666:
We've encountered the same challenge.
In the method `await ClientManager.get_client()`

```
    @classmethod
    async def get_client(cls) -> PostgreSQLDB:
        async with cls._lock:
            if cls._instances["db"] is None:
                config = ClientManager.get_config()
                db = PostgreSQLDB(config)
                await db.initdb()
                await db.check_tables()
                cls._instances["db"] = db
                cls._instances["ref_count"] = 0
            cls._instances["ref_count"] += 1
            return cls._instances["db"]
```
This is a global parameter, it will always return the same workspace each time selecting from the DB.

In the `postgres_impl.py`, suggest the workspace parameter to accept the programmatically passed-in function argument instead of taking the class attribute (which will never be changed once initialized).

This will help a lot if using the LightRAG core for multi-tenant solution.

I have no idea why the DB workspace takes higher priority:
```
    async def initialize(self):
        if self.db is None:
            self.db = await ClientManager.get_client()
            # Implement workspace priority: PostgreSQLDB.workspace > self.workspace > "default"
            if self.db.workspace:
                # Use PostgreSQLDB's workspace (highest priority)
                final_workspace = self.db.workspace
            elif hasattr(self, "workspace") and self.workspace:
                # Use storage class's workspace (medium priority)
[...]

PR #4004, opened 2026-09-17
feat(query): let QueryParam carry extra per-call kwargs for LLM, embedding, and rerank

pennycoders:
## Description

Adds three `QueryParam` fields that let a caller attach extra keyword arguments to the outbound provider calls a single query triggers — `extra_llm_kwargs`, `extra_embedding_kwargs`, and `extra_rerank_kwargs` — merged in at each relevant call site right before invoking the queue-wrapped provider func, so the data rides through the same concurrency-limited, priority-ordered queue every other query-time kwarg already goes through. Nothing about the queue or worker-pool internals changes.

The motivating case is forwarding request-scoped data — e.g. a caller identity header — to the calls a query makes, so a downstream proxy or observability layer can attribute them to the right end user. A query can trigger three kinds of outbound provider call: an LLM completion, an embedding (to vector-search for the query text itself and, in local/global/hybrid/mix mode, for keywords), and, when enabled, a rerank call. The natural-looking approach, setting a `contextvars.ContextVar` right before calling `aquery`/`aquery_llm` and reading it inside a custom `llm_model_func`/`embedding_func`/`rerank_model_func`, does not work with how LightRAG dispatches any of these: each is wrapped once by `priority_limit_async_func_call` in a small, fixed pool of persistent `asyncio.Task` workers, created lazily and never recreated per call. A call is handed to a worker as a plain tuple on an `asyncio.PriorityQueue` and executed inside *that worker's own* task context — captured when the worker task was first created, long before any particular request existed. A `ContextVar.set()` made by the caller's own task is therefore invisible there. This is correct, standard `asyncio` behavior, not a queue bug, but it leaves no supported way to carry per-call data through short of reimplementing the queue's own concurrency limiting and timeout handling on the caller's side.

Full design rationale — including why three fields rather than one, and the different failure/collision behavior of each — is in `docs/design/QueryTimeExtraProviderKwargs.md`.

## Related Issues

- #4003 (opened alongside this PR, describes the problem)
- Related, same general class of problem (request-scoped state not reaching a call made through LightRAG's shared, once-initialized machinery), though none overlaps this specific mechanism: #2904, #2133, #3847

## Changes Made

- `lightrag/base.py`: three new `QueryParam` fields — `extra_llm_kwargs`, `extra_embedding_kwargs`, `extra_rerank_kwargs` (each `dict[str, Any] | None = None`), added at the end per the existing "new fields go last" convention (positional-construction safety), each documenting what it's for and its collision/failure behavior.
- `lightrag/operate.py`:
  - `kg_query`, `naive_query`, `extract_keywords_only` spread `**(query_param.extra_llm_kwargs or {})` into their LLM call.
  - `_perform_kg_search`'s batched query/keyword embedding pre-compute (covers local/global/hybrid/mix mode) merges `extra_embedding_kwargs` in, and stops swallowing that call's failure when the field is set.
  - `naive_query` gains a new, parallel embedding pre-compute — only runs when `extra_embedding_kwargs` is set, so existing callers get zero added round-trips.
- `lightrag/lightrag.py`: `aquery_llm`'s `bypass` mode branch (the one LLM call outside `operate.py`) does the same `extra_llm_kwargs` merge.
- `lightrag/utils.py`: `apply_rerank_if_enabled` takes a new `extra_rerank_kwargs` parameter, merges it into the rerank call, and stops swallowing that call's failure when the field is set; `process_chunks_unified` threads `query_param.extra_rerank_kwargs` into it.
- `lightrag/llm/openai.py`: `openai_embed` gains a `**kwargs` passthrough to the embeddings API call, mirroring `openai_complete_if_cache`'s existing pattern — see "Found while verifying" below.
- `docs/design/QueryTimeExtraProviderKwargs.md` (renamed from `QueryTimeExtraLLMKwargs.md` to reflect the broader scope): the problem, why a queue can't carry contextvars, every call site each field covers, the two different failure behaviors, binding-support status, a rejected alternative, and a stated limitation.
- `tests/llm/test_query_extra_provider_kwargs.py` (renamed from `test_query_extra_llm_kwargs.py`): 19 regression tests covering all three fields — forwarding, no-op defaults, and both failure behaviors (`TypeError` for a duplicate LLM-call keyword; explicit opt-out of graceful degradation for embedding/rerank, confirmed both when the field is set and, as a baseline regression guard, when it is not).
- `tests/llm/openai_impl/test_openai_embed_extra_kwargs.py`: regression coverage for the `openai_embed` fix.

This is purely additive: every field defaults to `None`/no-op, nothing about `priority_limit_async_func_call`, the queue, or the worker pool changes, and no existing behavior is touched.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed
- [x] Documentation updated (`docs/design/QueryTimeExtraProviderKwargs.md`, `QueryParam` docstrings, `apply_rerank_if_enabled`/`openai_embed` docstrings)
- [x] Unit tests added

## Additional Notes

**Found while verifying, not just designed on paper**: I ran this end to end against a real local LLM (LM Studio) and a real embedding model, through a logging reverse proxy, to confirm the header actually survives the real queue/worker-pool hand-off rather than trusting mocks. That surfaced a genuine pre-existing gap: `openai_embed` had a closed parameter list and rejected `extra_embedding_kwargs` with `TypeError` even though the field itself reached it correctly — `openai_complete_if_cache` already forwarded unrecognized kwargs, `openai_embed` did not. Fixed alongside this PR since `extra_embedding_kwargs` is otherwise non-functional against the most common embedding binding. `azure_openai_embed` has its own separate, still-closed parameter list and is not covered — noted as a known gap in the design doc rather than silently left broken.

**Known limitation** (stated in the design doc too): this covers query-time calls only. The ingestion-time `extract`/`vlm` LLM roles and document-chunk embedding run per document chunk with no `QueryParam` in scope, so they're out of scope for this mechanism. `pick_by_vector_similarity`'s own embedding fallback (`lightrag/utils.py`) isn't separately wired — in the normal flow it only fires when `_perform_kg_search` didn't already supply an embedding, so `extra_embedding_kwargs` reaches it transitively whenever it would otherwise fire.

**Why three fields instead of one shared dict**: the LLM, embedding, and rerank functions are different callables, often different providers entirely. Forcing one caller-supplied dict onto all three risks a key valid for one silently colliding with, or being rejected by, another — which is exactly the class of bug the `openai_embed` finding above illustrates in practice.

**Failure behavior, briefly**: for the four LLM call sites, a key in `extra_llm_kwargs` repeating a kwarg the call site itself sets raises `TypeError` immediately — that's plain Python call syntax rejecting a duplicate keyword, not an added check. The embedding pre-compute and the rerank call already catch every exception and degrade gracefully by design (fall back to on-demand embedding, or unreranked order); silently keeping that behavior once a caller has asked for data to be forwarded would mean silently dropping it on any error, so setting `extra_embedding_kwargs`/`extra_rerank_kwargs` disables that particular call's fallback and lets the error propagate instead.

Ran locally: `ruff check` and `pre-commit run --all-files` clean on touched files; the mirroring test subset passes (960+ tests across `tests/llm`, `tests/extraction`, `tests/test_operate_*.py`, `tests/test_lightrag_query_validation.py`, `tests/test_docstring_budget.py`); and, since this change touches `lightrag/base.py` and `lightrag/utils.py` (both called out as cross-cutting in this repo's own contribution guide), the **full suite** as well — confirmed via a stash/diff against a clean checkout that the failures present in both runs are pre-existing environment gaps in my sandbox (missing optional provider SDKs and services — postgres/redis/ollama/bcrypt, etc.), identical before and after this change, i.e. zero new failures introduced.

Verified against a real local LLM and embedding model (LM Studio) end to end for every field, not just mocks: a small logging reverse proxy confirmed `extra_headers` set via each of `extra_llm_kwargs`/`extra_embedding_kwargs`/`extra_rerank_kwargs` actually reaches the corresponding outbound HTTP request after surviving the real queue/worker-pool hand-off, across a full ingest-then-query round trip, a `local`-mode KG query, and both failure-propagation paths (a broken rerank function correctly failed the query loudly instead of silently answering unreranked).


HyperionC21 (approved):

danielaskdd:
@codex review

pennycoders:
> @codex review

Hi! This did not seem to work. Can you please re-trigger it?

Thanks!

pennycoders:
> You have reached your Codex usage limits. You can see your limits in the [Codex usage dashboard](https://chatgpt.com/codex/cloud/settings/usage).

@danielaskdd it seem the usage limit has been reached. Any way in which I can help to get this over the fence?

Thanks,
Alex

danielaskdd:
@codex review

danielaskdd:
> > You have reached your Codex usage limits. You can see your limits in the [Codex usage dashboard](https://chatgpt.com/codex/cloud/settings/usage).
> 
> @danielaskdd it seem the usage limit has been reached. Any way in which I can help to get this over the fence?
> 
> Thanks, Alex

There appears to be an issue with my Codex Cloud environment: I am only able to trigger reviews on my own pull requests.

pennycoders:
> > > You have reached your Codex usage limits. You can see your limits in the [Codex usage dashboard](https://chatgpt.com/codex/cloud/settings/usage).
> > 
> > 
> > @danielaskdd it seem the usage limit has been reached. Any way in which I can help to get this over the fence?
> > Thanks, Alex
> 
> There appears to be an issue with my Codex Cloud environment: I am only able to trigger reviews on my own pull requests.

Hi @danielaskdd. Any way around this issue? Can I help in any way?

Thanks,
Alex

PR #4005, opened 2026-09-17
fix(parser): apply the forbidden-character rule on the engine-param dict path

RizgarOzan:
## What

Enforce the "one parameter value may never contain `,` `(` `)` `]`" rule inside
`_coerce_engine_value`, the helper both engine-param paths share, instead of only in
`parse_engine_params`.

## Why

No issue — found while reading `param_schema.py`. Two invariants the module states about
itself are false today, because the rule is coded in the text path only:

- `_coerce_engine_value`: *"Shared by the text path and the resolved-dict path so both apply
  identical rules"* — the forbidden-character rule sits outside it, so
  `normalize_engine_params` does not apply it.
- `encode_parse_engine`: *"raises ValueError on invalid input, so it can never emit a string
  that `decode_parse_engine` would later reject"*.

On `main`, `encode_parse_engine("mineru", {"language": "a)b"})` returns
`mineru(language=a)b)`. `take_paren_block` stops at the injected `)`, so with a second
parameter the directive is silently truncated — `mineru(language=a)b,page_range=1-3)` reads
back as `language=a` and drops `page_range=1-3`. The module comment on `_VALUE_FORBIDDEN`
says `]` is rejected explicitly "so the error is friendly rather than a silent truncation";
this closes the same hole for the dict path.

Reach: every in-tree caller feeds `normalize_engine_params` a dict that came from routing /
`decode_parse_engine`, which already applied the rule, so no upload, scan or `parse_engine`
string path can hit this. It is reachable through the seam that function documents itself
for — a direct SDK caller or a third-party parser building the dict. So: a public validator
not enforcing one of its own declared rules, not a user-visible regression.

`,` is the canonical joiner for the list-type `page_range`, so list values are checked per
segment and `page_range=1-3,5` still round-trips.

## How

- `_coerce_engine_value`: check the forbidden characters first, per segment for `is_list`
  specs, so the text path keeps its exact messages.
- `parse_engine_params`: drop its now-redundant copy — one encoding of the rule instead of
  two.

## How I tested

Windows, `uv sync --frozen --extra api --extra offline-storage --extra offline-llm --extra pytest`.

- 8 new `@pytest.mark.offline` cases (the file had no marker, so unmarked tests would not run
  in CI's `pytest tests/ -m offline`): on `main` 8 fail, with the fix `tests/parser/test_engine_params.py`
  is 54 passed.
- Full `pytest tests/ -m offline`: 7302 passed / 369 failed before and after, identical
  failure set (diff empty). The 369 are pre-existing and Windows-only (symlinks, POSIX
  permissions, `sh` scripts); `tests/api/test_gunicorn_worker_orphan_exit.py` is excluded
  because it imports `fcntl`, which aborts collection on Windows on `main` too.
- `pre-commit run --files lightrag/parser/param_schema.py tests/parser/test_engine_params.py`: clean.


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #4010, opened 2026-09-18
fix: pass document context to custom chunkers

### LINKED ISSUE #3603: Per-chunk text "breadcrumb" / metadata hook — `chunking_func` is silently bypassed on `/documents/scan` & `/documents/upload`

dafyy321-pixel:
## Description

Custom chunkers need the document context to make source-aware decisions, but the pipeline only passed the legacy chunking arguments.

## Changes

- Added the keyword-only `ChunkingContext` contract.
- Added explicit opt-in support for constructor callbacks and entry-point chunkers.
- Forwarded context through both custom chunking paths while keeping existing six-argument callbacks unchanged.
- Added regression coverage and updated the third-party chunker documentation.

## Validation

- Registry and pipeline regression tests pass.
- Compatibility and parity tests pass.

Closes #3603 and #3664


danielaskdd:
@codex review

danielaskdd:
@codex review

PR #4013, opened 2026-09-18
fix: correct four misspelled strings in exceptions and logs

simpleqt:
## Changes Made

Four user-visible/log strings carried misspellings:

- `lightrag/llm/lmdeploy.py:145`: "currently using **lmdeloy** ..." → "lmdeploy" (raised to users on the do_sample path; the same message spells it correctly one line above)
- `lightrag/api/lightrag_server.py:1692`: "**Unvicorn** Mode" → "Uvicorn" (the comment above and the sibling Gunicorn branch are correct)
- `lightrag/kg/json_kv_impl.py:175` and `lightrag/kg/json_doc_status_impl.py:278`: "**writting**" → "writing"

Each verified as the only occurrence repo-wide. String-only change, tested locally via grep after the change.


danielaskdd:
@codex review

Issue #4006, opened 2026-09-18
Configuration storage: one workspace-independent place for server and per-workspace settings


danielaskdd:
## Why

Two needs meet in the same place.

**1. Configuration must be reachable independently of the active workspace.** Every storage LightRAG opens today is bound to `self.workspace` — PostgreSQL by a `workspace` column, Redis and MongoDB by a name prefix, the JSON backends by a subdirectory. That is right for knowledge-base data and wrong for configuration: the server still has to read its own settings while switching between workspaces, and it has to read them before any workspace is open. Configuration is the server's property, not a tenant's.

**2. A container name is an assertion, not a record.** `BaseVectorStorage._generate_collection_suffix()` (`lightrag/base.py:357`) builds `{folded_model}_{dim}d` from the *current* configuration, so it can never disagree with the current configuration. Milvus, Qdrant and PostgreSQL name their vector containers that way, which means they cannot tell whether the data in an existing container was written by the model configured right now. See #3978 for what that misses in practice.

Both are answered by the same thing: a durable place to record what a deployment is configured with, separate from the data that configuration produced.

**The design is settled in [`docs/design/ConfigurationStorage.md`](https://github.com/HKUDS/LightRAG/blob/dev/docs/design/ConfigurationStorage.md)** (merged in #4008, on `dev`). That contract governs; this issue is the work item and must stay checkable against it. Where the two disagree, the contract wins.

## The facility

A configuration namespace in the KV storage, held in one fixed reserved workspace.

**Container.** KV namespace `config`, workspace `_lightrag_config`. PostgreSQL is the only backend needing **schema** work (`NAMESPACE_TABLE_MAP` entry, `TABLES` DDL, the per-namespace SQL templates `get_by_id_config` / `get_by_ids_config` / `upsert_config`). The other four need no new *physical container type* — a collection, a key prefix or a file named after the namespace — but that is not the same as "no work": every one of them owes the enumeration surface below. KV container names carry no model suffix, so a record kept here does not move when the embedding model changes, which is exactly the property the vector container name lacks.

**Keys.** `<workspace>/embedding/<target>` for a per-workspace setting, `_lightrag_server/<key>` for a server-global one.

[...]

PR #4017, opened 2026-09-18
fix(redis): acquire the shared connection pool in initialize(), not __post_init__

### LINKED ISSUE #4016: RedisKVStorage acquires its shared connection-pool reference at construction, so a failed LightRAG construction leaks it
## Problem

`RedisKVStorage.__post_init__` and `RedisDocStatusStorage.__post_init__` call `RedisConnectionManager.get_pool(url)` **at construction**. That manager is class-level state: one shared `ConnectionPool` per URL plus a reference count (`_pool_refs`), incremented by `get_pool` and decremented only by `release_pool_ref` / `release_pool`, which run from the storage's async `close()` / `finalize()`.

`LightRAG.__post_init__` is synchronous, constructs twelve storage objects in sequence, and then runs a series of validations (`llm_model_func` present, `role_llm_configs` well-formed, ...). It has no teardown path: if any constructor or validation raises, every Redis storage already constructed keeps its pool reference forever. Repeated failed constructions grow the count without bound, and a later valid instance can never bring it to zero, so the shared pool is never closed.

## Why only Redis

Every other backend defers connecting to `initialize()` and leaves the constructor side-effect free:

| backend | constructor | connection |
| --- | --- | --- |
| MongoDB (all four storages) | validate workspace, build collection name | `initialize()` → `await ClientManager.get_client()` |
| PostgreSQL (all four) | validate workspace | `initialize()` → `await ClientManager.get_client()`, pool via `asyncpg.create_pool` |
| OpenSearch (three) | build index name | `initialize()` → `await ClientManager.get_client()` |
| Milvus / Qdrant | validate embedding, compute suffix, `self._client = None` | `initialize()` creates the client |
| Neo4j / Memgraph | handle workspace override, `self._driver = None` | `initialize()` → `AsyncGraphDatabase.driver()` |
| JSON KV / DocStatus, NetworkX, NanoVectorDB, Faiss | record paths | `initialize()` binds shared dict / loads file |
| **Redis KV / DocStatus** | **`RedisConnectionManager.get_pool()`** | `initialize()` only pings |

## How it surfaced

Found by review on #4011 (configuration storage, slice 1). That PR adds one more KV storage object to `__post_init__` and a new refusal path (`validate_workspace_override`, raised in an ordinary storage's constructor), which made the leak reachable through the configuration storage's reference. The PR sidesteps it by constructing the configuration storage as the **last** statement of `__post_init__` that can raise, and records the leak between the business storages themselves as a pre-existing residue in `docs/design/ConfigurationStorage.md` (*Cleanup before `INITIALIZED` exists*, last paragraph). This issue tracks the root fix.

## Proposed fix

Align Redis with the other ten backends: move `RedisConnectionManager.get_pool()` and the `Redis(connection_pool=...)` client creation from `__post_init__` into `initialize()` (under `get_data_init_lock()`, guarded by `_initialized` as the ping already is). The constructor then records `_redis_url` only and sets `_pool = _redis = None`.

Things to check while doing it:

- The constructor's existing error path (`release_pool_ref` + `schedule_pool_close` on a failed `get_pool`) moves with the acquisition.
- `close()` must stay idempotent and tolerate a never-initialized instance (`_pool is None`).
- Tests that construct a Redis storage and read `_pool` / `_redis` before `initialize()`, and tests that mock `RedisConnectionManager.get_pool` around construction (e.g. `tests/workspace/test_reserved_workspace.py`, `tests/kg/redis_impl/`), need to move their expectations to after `initialize()`.
- `_get_redis_connection()` should raise the existing "Redis connection not initialized" error when called before `initialize()`; the strict-read path in `config_store` already relies on that message.

## Acceptance

- Constructing any Redis storage, alone or via `LightRAG(...)`, leaves `RedisConnectionManager._pools` / `_pool_refs` untouched.
- A `LightRAG` construction that fails after Redis storages were built leaves the reference count unchanged (the `tests/workspace/test_reserved_workspace.py` pool-reference test can then drop its "constructed last" premise and assert directly).
- After the fix lands, the construction-order paragraph in `ConfigurationStorage.md` can be reduced to a note that the ordering is no longer load-bearing.

Related: #4011, #4006.

danielaskdd:
## Description

`RedisKVStorage` and `RedisDocStatusStorage` took a reference on the shared, reference-counted `ConnectionPool` from their **constructors** (`RedisConnectionManager.get_pool()` in `__post_init__`). A constructor has no teardown path, so that reference could never be given back when construction failed afterwards.

`LightRAG.__post_init__` is synchronous, builds twelve storage objects in sequence, and then runs a series of validations (`llm_model_func` present, `role_llm_configs` well-formed, …). When one of those raises, every Redis storage already built is dropped on the floor — nothing calls `close()` on it, and there is no half-built `LightRAG` to call `finalize_storages()` on. Repeated failed constructions grew `_pool_refs` without bound, so no later valid instance could bring it to zero and the shared pool was never disconnected.

Redis was the only backend doing this. The other ten all leave the constructor side-effect free and connect in `initialize()`:

| backend | constructor | connection |
| --- | --- | --- |
| MongoDB / PostgreSQL / OpenSearch | validate workspace, build names | `initialize()` → `await ClientManager.get_client()` |
| Milvus / Qdrant / Neo4j / Memgraph | validate, `self._client = None` | `initialize()` creates the client |
| JSON KV / DocStatus, NetworkX, NanoVectorDB, Faiss | record paths | `initialize()` binds shared dict / loads file |
| **Redis KV / DocStatus** (before) | **`RedisConnectionManager.get_pool()`** | `initialize()` only pinged |

This fix aligns Redis with the rest.

## Related Issues

Fixes #4016. Related: #4011, #4006.

## Changes Made

**`lightrag/kg/redis_impl.py`**

- `__post_init__` (both storages) now only records `_redis_url` and leaves `_pool` / `_redis` as `None`. It can no longer fail, so its acquisition error path goes away with it.
- `initialize()` acquires the pool and builds the client under the existing `get_data_init_lock()`, guarded by `_initialized` as the ping already was. The pool is assigned before the client is built, so a failing `Redis(...)` still leaves the reference visible to the cleanup in the same `except` block.
- The acquisition's error path moves with it: `initialize()`'s existing `await self.close()` now releases the reference, and it can `await` the disconnect. That makes the synchronous-context helpers unreachable, so `schedule_pool_close`, `_close_pool_safely` and `_cleanup_tasks` are removed.
- `close()` gates the release on `_pool` instead of `_redis_url`. This is load-bearing: `_redis_url` is set at construction while the reference is not, so without the change, closing a never-initialized storage would release a reference it never took and disconnect the pool out from under a live sibling. `_pool` becomes the detach token that keeps `close()` idempotent.
- `close()` keeps `_redis_url` (configuration, not a resource) and resets `_initialized`, so a closed storage can be initialized again and a failed `initialize()` can be retried — previously a retry would have hit `get_pool(None)`.
- `RedisConnectionManager`'s docstring now states the acquire/release pairing and why it must not run from a constructor.

**`tests/kg/redis_impl/`**

- `test_redis_finalize.py`: the three `__post_init__` init-error tests are replaced by tests of the new contract, and five new tests are added — see below.
- Three fixtures (`test_redis_doc_status_lookup.py`, `test_redis_doc_status_track_id_hydration.py`, `test_redis_scheduling_pages.py`) attached the fake client via the constructor; they now attach it explicitly, standing in for `initialize()`.

## What's broken / behavior change

No public API change; `LightRAG` usage is unaffected because `initialize_storages()` is already required. Two internals changed for anyone reaching into them:

- `_pool` / `_redis` are `None` until `initialize()` runs. Reads before that get the existing `ConnectionError("Redis connection not initialized")` from `_get_redis_connection()` — the message the `config_store` strict-read path relies on is unchanged.
- `close()` no longer clears `_redis_url`. `chunk_tracking_repair.py`'s `connection_target()` reads that attribute on an initialized backend and is strictly better off.
- `RedisConnectionManager.schedule_pool_close` / `_close_pool_safely` / `_cleanup_tasks` are gone (internal, no callers outside the removed path).

## Checklist

- [x] Changes tested locally
- [ ] Code reviewed
- [x] Documentation updated (if necessary) — docstrings only; no `docs/design/` contract covers this file

## Additional Notes

**Tests run:** `tests/kg/redis_impl` — 88 passed, 24 skipped (skips are the live-Redis integration file). Also `tests/kg` as a whole: 2515 passed. The 15 failures in `tests/kg/test_batch_graph_operations.py` are pre-existing — they reproduce identically on a clean tree with this branch stashed, and concern `ainsert_custom_kg` / PG, not Redis.

**New regression tests**, each verified to fail against the pre-fix implementation and pass after:

- `test_construction_takes_no_pool_reference` — construction leaves `_pools` / `_pool_refs` untouched.
- `test_discarded_storages_leave_no_pool_reference` — the reported symptom: three storages built and discarded the way a failing `LightRAG.__post_init__` discards them, refcount unchanged.
- `test_close_before_initialize_releases_nothing` — the under-release risk the new gate closes: a sibling's reference survives.
- `test_initialize_client_failure_releases_the_pool_ref` — the moved error path; original exception propagates unmasked.
- `test_initialize_ping_failure_releases_the_pool_ref` — failure after the client exists still gives the reference back.

Once this lands, the "constructed last" ordering premise in #4011 is no longer load-bearing and the construction-order paragraph in `ConfigurationStorage.md` can be reduced to a note.

https://claude.ai/code/session_014keeHtKtYrRq2FSZNrbPur

---
_Generated by [Claude Code](https://claude.ai/code/session_014keeHtKtYrRq2FSZNrbPur)_


danielaskdd:
@codex review

danielaskdd:
@codex review

danielaskdd:
@codex review

PR #4024, opened 2026-09-20
test(pipeline): make multimodal fail-fast assertion independent of elapsed time

### LINKED ISSUE #4021: Flaky: `test_analyze_multimodal_fail_fast_preserves_successes` times fail-fast with a wall clock
## Symptom

`tests/pipeline/test_pipeline_cancellation.py::test_analyze_multimodal_fail_fast_preserves_successes` fails intermittently on CI:

```
AssertionError: fail-fast still waited 0.81s for slow task
assert 0.8142666139999903 < 0.8
```

Seen on `Offline Tests (3.12)` ([run 35453869085](https://github.com/HKUDS/LightRAG/actions/runs/35453869085)), which also cancelled the 3.14 job by fail-fast. 1 failed, 7990 passed. The next run on the next commit was green.

## Why retuning the threshold will not fix it

The test builds three VLM calls — 0.05 s success, 0.1 s failure, 1.2 s slow — and asserts the caller returned without waiting for the third:

```python
assert elapsed < 0.8, f"fail-fast still waited {elapsed:.2f}s for slow task"
```

Measured locally, three consecutive runs:

```
MEASURED elapsed=0.115s (budget 0.8, slow task 1.2)
MEASURED elapsed=0.115s
MEASURED elapsed=0.110s
```

So the real cost is **0.11 s against a 0.8 s budget — already a 7× margin** — and CI still measured 0.814 s. That is roughly 0.7 s of runner stall, **six times the entire operation being measured**.

Raising the budget does not help. The usable window is bounded above by the slow task itself: anything at or beyond 1.2 s stops distinguishing "returned early" from "waited". So the whole room available is 0.11 s → 1.2 s, about 10×, while the jitter already observed is 7×. **No absolute wall-clock threshold is safe in that window.**

## The assertion is not redundant, so it cannot just be deleted

The test already asserts the outcomes:

```python
assert statuses == ["failure", "failure", "success"]
assert len(cancelled_items) == 1
```

If the code genuinely waited for the 1.2 s call it would have **succeeded**, returning `"late"`, and both of those would fail. So they cover the plain "waited for it" regression.

What they do **not** cover is the one the timing assertion uniquely catches: **cancellation recorded but not actually preempting** — the caller stays blocked until the slow work finishes, and the item is still reported as cancelled. Deleting the assertion loses that.

## Proposed fix: park the slow task on an event, not a clock

Replace the timed sleep with an event the test controls, so completion is impossible rather than merely slow:

```python
release_slow = asyncio.Event()
slow_completed = asyncio.Event()

async def vlm_func(prompt, **kwargs):
    ...
    await release_slow.wait()          # never completes on its own
    slow_completed.set()
    return json.dumps({...})
```

and assert on state instead of duration:

```python
assert not slow_completed.is_set(), "fail-fast waited for the slow task"
```

- **Fail-fast working** → the call returns promptly and `slow_completed` is necessarily unset. Deterministic; no clock anywhere.
- **Fail-fast regressed** → the caller blocks, and the existing `asyncio.wait_for(..., timeout=15.0)` bounds it. The failure is explicit rather than a near-miss inequality.
- **Teardown gets faster.** `finally` sets `release_slow` before shutting the role workers down. Today's `asyncio.sleep(1.2)` keeps running to completion regardless — the source comment says so, and it is why the sleep was lowered from 5 s to 1.2 s in the first place. That constraint disappears.

Net: the clock goes away, the coverage the assertion uniquely provides stays, the failure mode is bounded, and the test gets quicker.

## Scope

Not part of #4011, which touches nothing in this area — its changes to the storage lifecycle sit outside the measured window (it opens after `initialize_storages()` and closes before `finalize_storages()`). Worth a small standalone PR.

## Definition of Done

- No wall-clock assertion remains in this test.
- The "cancelled in bookkeeping but still blocked" regression is still caught — verified by deliberately making the fail-fast path await the slow task and watching the test go red.
- The test's own runtime does not grow, and teardown no longer waits on a sleeping worker.

danielaskdd:
## Description

The multimodal fail-fast regression test can fail on a stalled CI runner because it requires completion within 0.8 seconds. Replace that wall-clock assertion with an event-gated slow VLM call: the call cannot finish before teardown, so waiting for it is bounded by the existing 15-second timeout.

## Related Issues

Fixes #4021.

Overlaps with #4022, which independently implements the same issue proposal. Only one of these PRs needs to merge.

## Changes Made

- Park the slow VLM call on an `asyncio.Event` and assert it has not completed when analysis raises.
- Remove the elapsed-time measurement and threshold while retaining success, failure, and cancellation assertions.
- Release the slow worker in `finally` before shutting down role workers.

## Checklist

- [x] Changes tested locally
- [x] Code reviewed

## Additional Notes

Only the existing regression test is changed; no production behavior or documentation changes are required.

Validation on main-based branch (base `50a0b5b64`):
- `./scripts/test.sh tests/pipeline -q`: 954 passed.
- Cancellation test file: 6 passed.
- Ruff and `git diff --check`: passed.

During implementation, temporarily wrapped the fail-fast cache-persistence hook to verify that the sidecar already recorded cancellation, then wait for the slow VLM completion event. The test failed with `TimeoutError` at the existing 15-second bound and cleaned up successfully. The temporary mutation was removed. This verifies coverage of cancellation recorded in bookkeeping while the caller remains blocked.

On the original development baseline, the target test's single-run call duration decreased from 2.06s to 0.48s; these are local observations, not test assertions.


danielaskdd:
@codex review

PR #4026, opened 2026-09-20
Make JSON mode optional for query keyword extraction

### LINKED ISSUE #4023: [Bug]: hardcoded response_format={"type": "json_object"} in keyword extraction breaks OpenAI-compatible servers that only accept json_schema/text (e.g. LM Studio) — hybrid/local/global queries always fail
# [Bug]: hardcoded response_format={"type": "json_object"} in keyword extraction breaks OpenAI-compatible servers that only accept json_schema/text (e.g. LM Studio) — hybrid/local/global queries always fail

## Do you need to file an issue?

## Describe the bug

`_extract_keywords_for_query()` in `lightrag/operate.py` unconditionally sends
`response_format={"type": "json_object"}` to the LLM when extracting keywords
for `hybrid` / `local` / `global` query modes:

```python
# lightrag/operate.py (around line 5047)
result = await use_model_func(kw_prompt, response_format={"type": "json_object"})
```

There is no configuration knob to change or disable this. Servers that
implement the OpenAI-compatible `/v1/chat/completions` endpoint but only
accept `response_format.type` of `json_schema` or `text` — LM Studio being a
prominent one — reject the request with HTTP 400:

```json
{"error": "'response_format.type' must be 'json_schema' or 'text'"}
```

LightRAG then surfaces this as a successful-but-empty query
(`"No relevant context found for the query."`), so **every** hybrid/local/
global query silently returns nothing while `naive` (which skips keyword
extraction) works fine. The failure is invisible unless you read the server
logs:

```
ERROR: keyword LLM func: Error in decorated function ...: Error code: 400 -
  {'error': "'response_format.type' must be 'json_schema' or 'text'"}
ERROR: Query failed: Error code: 400 - ...
```

This matters because `json_object`-without-schema is the legacy OpenAI JSON
mode; an increasing number of local-inference servers (LM Studio ≥ 0.4.x
documents only `json_schema` + `text` for `/v1/chat/completions`:
https://lmstudio.ai/docs/developer/openai-compat/structured-output) accept
only the newer subset. Hardcoding it makes LightRAG incompatible with them
with no workaround short of patching LightRAG itself.

## Steps to reproduce

1. Point LightRAG (v1.5.7, Docker `ghcr.io/hkuds/lightrag:latest`) at LM
   Studio's OpenAI-compatible endpoint (`http://host.docker.internal:1234/v1`)
   for the `keyword` role (default when using an OpenAI-compatible binding).
2. Index any document (indexing works — only query-time keyword extraction
   is affected).
3. `POST /query` with `{"query": "anything", "mode": "hybrid"}` (same for
   `local` / `global`).
4. Observe `{"response": "No relevant context found for the query."}` and the
   HTTP 400 in the LightRAG server logs shown above.

Minimal reproduction without LightRAG (proves the server-side rejection):

```bash
# 400 — what LightRAG sends today
curl -X POST http://127.0.0.1:1234/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"<any>","messages":[{"role":"user","content":"hi"}],
       "max_tokens":20,"response_format":{"type":"json_object"}}'
# → {"error":"'response_format.type' must be 'json_schema' or 'text'"}

# 200 — the same request without the legacy mode
curl -X POST http://127.0.0.1:1234/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"<any>","messages":[{"role":"user","content":"hi"}],
       "max_tokens":20,"response_format":{"type":"text"}}'
# → valid completion
```

## Expected Behavior

Keyword extraction should not hard-fail on servers that don't implement the
legacy `json_object` mode. Since `_parse_keywords_payload()` already
tolerates plain-text responses (strips markdown fences, `json_repair`
fallback) and the keyword prompt itself explicitly asks for JSON, sending the
request *without* `response_format` degrades gracefully: I verified locally
that hybrid queries return correct, cited answers through LM Studio once the
parameter is omitted.

## Proposed fix (contribution-ready)

Make the keyword-extraction `response_format` resilient with a
try-fallback inside `_extract_keywords_for_query()`
(`lightrag/operate.py`), keeping `json_object` as the first attempt so
default behavior against OpenAI is unchanged:

```python
# lightrag/operate.py, in _extract_keywords_for_query()
try:
    result = await use_model_func(kw_prompt, response_format={"type": "json_object"})
except Exception as json_mode_error:
    # Some OpenAI-compatible servers (e.g. LM Studio) only accept
    # 'json_schema' or 'text' for response_format.type. The keyword prompt
    # already requests JSON explicitly and _parse_keywords_payload()
    # tolerates plain-text responses, so retry without structured output.
    logger.warning(
        "Keyword extraction: JSON mode unsupported (%s), retrying as plain text",
        json_mode_error,
    )
    result = await use_model_func(kw_prompt)
```

Notes for reviewers (per `.github/CONTRIBUTING.md`):

- No new dependency, no config surface change, default path untouched.
- The fallback triggers only when the provider raises, so OpenAI/Azure/
  Ollama behavior is byte-identical to today.
- Happy to add this behind an explicit opt-in (e.g. `KEYWORD_JSON_MODE=auto|json_object|text`)
  instead if you prefer configuration over fallback — say the word and I'll
  shape the PR that way.
- Will run `pre-commit run --all-files` + `python -m pytest tests` and cover
  the fallback with a unit test (mocked `use_model_func` raising on the
  first call) before opening the PR on a `fix/...` branch from `main`.

## LightRAG Config Used

```env
LLM_BINDING=openai
LLM_BINDING_HOST=http://host.docker.internal:1234/v1
LLM_MODEL=<local instruct model via LM Studio>
EMBEDDING_BINDING=openai
EMBEDDING_BINDING_HOST=http://host.docker.internal:1234/v1
EMBEDDING_MODEL=text-embedding-nomic-embed-text-v1.5
```

## Logs and screenshots

```
ERROR: keyword LLM func: Error in decorated function for task ...: Error code: 400 -
  {'error': "'response_format.type' must be 'json_schema' or 'text'"}
ERROR: Query failed: Error code: 400 - {'error': "'response_format.type' must be 'json_schema' or 'text'"}
```

## Additional Information

- LightRAG Version: v1.5.7 (API 0344, `ghcr.io/hkuds/lightrag:latest`)
- Operating System: macOS 26 (Docker via OrbStack), LM Studio 0.4.25
- Python Version: 3.12 (server image)
- Related Issues: none found (searched "json_object", "LM Studio", "response_format")

chengwudi1:
## Description

Query keyword extraction asked for JSON structured output on every call. Servers that implement the OpenAI-compatible endpoint but only accept `response_format.type` of `json_schema` or `text` answer HTTP 400, and the query path turns that into `"No relevant context found for the query."` — so every `hybrid`, `local` and `global` query returns nothing, while `naive` (which skips keyword extraction) keeps working. The only trace is a line in the server's log.

JSON mode is not what makes this work. `_parse_keywords_payload` already handles a plain-text reply, think tags, markdown fences, and responses that need `json_repair`, so the field was costing compatibility and buying little.

## Related Issues

Fixes #4023

## Changes Made

- `extract_keywords_only` no longer sends `response_format={"type": "json_object"}` unless asked to.
- Added `keyword_extraction_use_json` (env `KEYWORD_EXTRACTION_USE_JSON`, default `false`), mirroring the existing `entity_extraction_use_json`: opt-in, env-backed, off by default.
- Documented it on the field, including why a server would reject the field.

## What's broken and how it works now

Before, on a server that rejects the field:

```
ERROR: keyword LLM func: Error code: 400 - {'error': "'response_format.type' must be 'json_schema' or 'text'"}
ERROR: Query failed: Error code: 400
```
and the caller saw an empty result.

Now the field is not sent, the model answers the prompt as it always did, and the tolerant parser reads the keywords out of the reply. Users who want JSON mode in their own deployment set `KEYWORD_EXTRACTION_USE_JSON=true`.

## Tests

Added to `tests/extraction/test_keyword_parsing.py`, which already drives `extract_keywords_only` through a fake keyword model:

- `test_keyword_extraction_does_not_request_json_mode_by_default` — asserts `response_format` is absent from the call, and that the plain-text reply still parses into keywords.
- `test_keyword_extraction_requests_json_mode_when_enabled` — asserts the field comes back when the flag is set.

The first fails against the current behaviour (`response_format` is present in the captured kwargs); I verified that by reverting only the `operate.py` change.

```
$ pytest tests/extraction/ tests/test_lightrag_query_validation.py tests/test_query_validation.py -q
415 passed, 1 failed
```

The failure is `test_entity_extraction_stability.py::test_prompt_file_rejects_absolute_paths`, which fails the same way on a clean `main` here (Windows path handling, unrelated). `tests/test_docstring_budget.py` reports the same two pre-existing failures on a clean tree as well.


danielaskdd (changes requested):
Changes requested before re-review:

1. Setting the default to `false` downgrades existing calls from an API-enforced JSON constraint to relying on the prompt alone. `_parse_keywords_payload()` can handle fences, think tags, and minor malformed JSON, but it cannot turn genuinely free-form text into the keyword structure. Once parsing produces empty lists, long queries return the fail response. This can therefore introduce silent retrieval failures for models that currently support `json_object`. Please do not use a Boolean switch to cover these two provider classes. Use three modes instead: `auto` (the default: send `json_object` first, then retry without it only for a clearly identified unsupported `response_format`/`json_object` error), `json_object`, and `none`. The fallback must be limited to that compatibility error; it must not retry every exception. Please add a regression test where the first request is rejected for this field and the retry without it succeeds, as well as coverage that a non-JSON response is not treated as valid keyword output.

2. The new `KEYWORD_EXTRACTION_USE_JSON` setting is mentioned only in the field docstring; it is absent from `env.example`, the README, and the API Server configuration documentation. Users following the project-standard .env setup cannot discover or choose this setting correctly. Please document the default, the compatibility rationale, and the appropriate use case for each policy in the public configuration entry points.

chengwudi1:
Addressed both points in `188aade` — the branch is now a single commit, since the boolean design you rejected was its first draft.

## 1. Three modes instead of a boolean

`keyword_extraction_use_json` is replaced by `keyword_extraction_json_mode`:

- **`auto` (default)**: sends `response_format={"type": "json_object"}` on the first attempt exactly as today — supported providers are **not** downgraded — then retries the same prompt once *without* the constraint, and only when `_is_unsupported_response_format_error` matches: the message must mention `response_format`/`json_object` **and** carry an explicit rejection marker (`must be`, `not supported`, `invalid`, `only accept`, ...). A connection reset, a 429, a generic 400 — anything without that shape — propagates untouched, after a single attempt.
- **`json_object`**: always sends the constraint.
- **`none`**: never sends it; keywords are parsed from plain text.

Tests added to `tests/extraction/test_keyword_parsing.py`:

- the compat regression you asked for: first call raises LM Studio's `'response_format.type' must be 'json_schema' or 'text'`, the retry without the field succeeds, and both attempts are asserted (2 calls, the second without `response_format`);
- a `ConnectionResetError` on the first attempt is **not** retried — exactly one call, exception propagates;
- the no-downgrade guard: `auto`'s first attempt still carries the constraint;
- `json_object` / `none` behaviors;
- a free-form prose reply yields no keywords — non-JSON output is not treated as valid keyword output;
- the rejection matcher itself, parametrized over 8 messages (4 match, 4 don't; the negatives include `"Request timed out while waiting for response_format"`, which has both tokens but no rejection intent);
- unknown mode values raise `ValueError` naming the accepted set.

Red check: restoring the previous behavior fails exactly the compat-retry and `none` tests.

## 2. Documentation at the public configuration entry points

- `env.example`, next to the `KEYWORD_*` role block — modes, default, and why the fallback exists;
- `README.md`, *Other Important Configurations for Document Querying* — the policy and when to pick each mode;
- `docs/LightRAG-API-Server.md` **and** `docs/LightRAG-API-Server-zh.md` — new *Query Keyword Extraction Configuration* section stating it is independent from `ENTITY_EXTRACTION_USE_JSON`, with the compatibility rationale and an example.

## Verification

Mirror-directory scope per the repo rules: `tests/extraction/` → **326 passed** (the one failure, `test_prompt_file_rejects_absolute_paths`, fails identically on a clean `main` here — Windows path handling, unrelated); `tests/test_docstring_budget.py` → the same two pre-existing failures as `main`, the new field docstring is within budget; `ruff check` reports no findings on any changed line.

PR #4027, opened 2026-09-20
fix(azure): forward max_token_size in azure_openai_embed

Shizoqua:
## Summary
Azure embedding calls now truncate oversized text instead of sending it untouched.

## Problem
Long text reached Azure's embeddings API untruncated, unlike the standard OpenAI path.

## Changes
Forwarded max_token_size into the underlying embed call.

## Testing
tests/llm/azure_openai_impl, tests/llm/openai_impl, asymmetric embedding: 78 passed.

## Compatibility and Risk
No API changes. Only oversized inputs are affected.

## Related Issue
No related issue.


danielaskdd:
Addressed the positional-argument compatibility issue in 7a277e607c1e508522c59920aaa969518625b857.

- Moved `max_token_size` to the end of `azure_openai_embed`'s signature so every pre-existing positional argument retains its original meaning. The named parameter remains discoverable by `EmbeddingFunc`, and forwarding to `openai_embed.func` is preserved.
- Added offline regression coverage for the complete legacy positional call, both without an explicit token limit and with `max_token_size=128` passed by keyword. The tests verify that all arguments reach the underlying implementation correctly.
- Replaced the truncation test's `EMBEDDING_DIM * 5` precondition with `MAX_TOKEN_SIZE = 8192`, so the input must actually exceed the truncation threshold.

Validation: the two new compatibility cases failed before the signature fix and passed afterward. `./scripts/test.sh tests/llm/azure_openai_impl tests/llm/openai_impl -q` completed with **60 passed**. Ruff and `git diff --check` also passed.

PR #4028, opened 2026-09-20
fix(jina): stop retrying permanent 4xx errors

Shizoqua:
## Summary
Jina API calls no longer retry on permanent client errors.

## Problem
Bad API key or malformed request errors were retried like transient failures.

## Changes
Retries now limited to connection errors and status 429 or 500+.

## Testing
tests/llm/jina_impl: 3 passed.

## Compatibility and Risk
No API changes. Transient errors still retry as before.

## Related Issue
No related issue.


danielaskdd:
@codex review

lightrag/llm/jina.py
```diff
@@ -14,11 +14,25 @@
...
+
+    ``fetch_data`` raises ``aiohttp.ClientResponseError`` for every non-200
+    Jina response alike, so retrying on the exception type alone retried a
+    permanent failure (bad API key, malformed request) exactly as eagerly as
+    a transient one (rate limit, server overload) -- three requests plus
+    exponential backoff before the caller ever saw the real error.
+    """
+    if isinstance(exc, aiohttp.ClientResponseError):
+        return exc.status == 429 or exc.status >= 500
+    return isinstance(exc, aiohttp.ClientError)
```
danielaskdd, line 33: **Please narrow this to actual connection failures.**

`aiohttp.ClientError` is the root class for more than transient connection failures. It also includes permanent client-side errors such as `InvalidURL`. Because `base_url` is caller-configurable, a malformed URL would still be retried three times with exponential backoff, contrary to this predicate's contract and the PR description.

Please use `aiohttp.ClientConnectionError` here, or explicitly enumerate any additional retryable subclasses such as `ClientPayloadError` if retrying them is intentional. Please also add coverage proving that `InvalidURL` is not retried and `ClientConnectionError` is retried.

tests/llm/jina_impl/test_jina_retry_classification.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+
+@pytest.mark.offline
+@pytest.mark.asyncio
+async def test_permanent_401_is_not_retried(monkeypatch):
+    """A 401 (bad API key) must fail on the first attempt, not be retried 3x."""
+    monkeypatch.setenv("JINA_API_KEY", "irrelevant")
+    # Keep the test fast regardless of outcome; only call count is asserted.
+    jina_embed.func.retry.wait = wait_none()
```
danielaskdd, line 33: **Please restore the shared retry configuration after each test.**

This assignment mutates the module-level Tenacity retry object and is not reverted when the test finishes. Consequently, every later Jina test in the same pytest process runs with `wait_none()` instead of the production exponential-backoff policy, making the suite order-dependent and potentially causing later integration tests to issue immediate repeated requests.

Please replace this with:

```python
monkeypatch.setattr(jina_embed.func.retry, "wait", wait_none())
```

and make the same change in `test_transient_503_is_retried` (line 58), so pytest restores the original value after each test.

danielaskdd (changes requested):
Please address the two inline findings below: narrow the retry predicate to actual transient failures and restore the shared Tenacity configuration after each test.

PR #4029, opened 2026-09-20
fix(gemini): match real google-genai exception types

Shizoqua:
## Summary
Gemini API calls now actually retry on transient errors.

## Problem
Retry predicate matched the wrong SDK's exceptions, so 503s and 429s failed after one try.

## Changes
Added matches for google-genai's real ServerError and 429 ClientError.

## Testing
tests/llm/gemini_impl, provider install guards: 28 passed.

## Compatibility and Risk
No API changes. Only widens what gets retried.

## Related Issue
No related issue.


danielaskdd:
@codex review

PR #4030, opened 2026-09-20
fix(chunker): reject overlap >= chunk size

Shizoqua:
## Summary
The recursive character chunker now rejects an invalid overlap instead of crashing.

## Problem
An overlap equal to or greater than the chunk size was silently accepted or crashed with a raw internal error.

## Changes
Added a clear validation check before the splitter is constructed.

## Testing
Full tests/chunker suite: 381 passed.

## Compatibility and Risk
No API changes. Only invalid overlap values are affected.

## Related Issue
No related issue.


danielaskdd:
@codex review

PR #4032, opened 2026-09-21
fix(tests): stop double-capturing lightrag log records on pytest 9.1

YHC66:
### Summary

Two tests fail on pytest 9.1+ because a record from the `lightrag` logger is captured twice.

### Motivation

pytest 9.1 attaches its capture handler to non-propagating loggers. `lightrag.utils` sets `propagate = False` on the `lightrag` logger, and these tests turned propagation back on so that caplog's root handler would see the records — the workaround pytest 9.1 makes counterproductive. With propagation on, each record now reaches caplog twice: once through the handler pytest attached to `lightrag`, once through root. Assertions on an exact count then fail:

```
tests/pipeline/test_strict_capability_gate.py::test_gaps_are_warned_with_their_operator_facing_consequence
    assert len(caplog.records) == 1
E   assert 2 == 1

tests/pipeline/test_chunk_options_regex_scrub.py::test_audit_line_bounds_the_untrusted_pattern
E   ValueError: too many values to unpack (expected 1)
```

`uv.lock` pins pytest 9.0.3, so CI does not see this; `pip install pytest` today gives 9.1.1, so a contributor does.

### What's changed

- `tests/conftest.py`: a `lightrag_log_records` fixture that collects from a handler installed on the `lightrag` logger itself, so it is independent of both pytest behaviours and pins how many records the code emitted rather than how many handlers observed them.
- The two affected tests use it instead of toggling `propagate`.

`test_a_first_party_backend_warns_about_nothing` moves over too: it asserted `caplog.records == []` without turning propagation on, so on pytest 9.0.3 it could not have observed a record either way.

The other 21 test files that toggle `propagate` still pass on both versions — they assert on message content rather than an exact count — so this leaves them alone. Happy to migrate them in a follow-up if you would prefer one pattern across the suite.

### How it was verified

| | `main` | this branch |
|---|---|---|
| pytest 9.1.1 | 2 failed, 13 passed | 15 passed |
| pytest 9.0.3 | 15 passed | 15 passed |

`tests/pipeline`: 954 passed on both pytest versions.

Full `pytest tests/ -m offline` run, same machine and dependency set: 316 failures on `main`, 314 on this branch — the difference is exactly the two tests above, and the remaining 314 are pre-existing macOS environment failures (the setup wizard needs bash 4+, the SVG tests need libcairo), identical on both sides.

macOS 26, Python 3.12.


danielaskdd:
@codex review

